PR #2, opened 2024-10-13
Fix matrix strategy for ARM and AMD Builds

vivekgsharma:


QuantumExplorer:
Latte used this. Thanks Vivek.

PR #27, opened 2024-10-30
fix: successful name registration was not being handled

pauldelucia:


src/platform/mod.rs
```diff
@@ -70,7 +70,7 @@ impl AppContext {
             BackendTask::IdentityTask(identity_task) => self
                 .run_identity_task(identity_task, &sdk, sender)
                 .await
-                .map(|_| BackendTaskSuccessResult::None),
+                .map(|_| BackendTaskSuccessResult::Message("Success".to_string())),
```
QuantumExplorer, line 73: I think this will have unintended side effects.

QuantumExplorer (changes requested):

src/backend_task/identity/register_dpns_name.rs
```diff
@@ -148,6 +149,58 @@ impl AppContext {
...
+
+        qualified_identity.dpns_names = owned_dpns_names;
+
+        // Insert qualified identity into the database
+        self.insert_local_qualified_identity(&qualified_identity)
+            .map_err(|e| format!("Database error: {}", e))?;
+
+        Ok(BackendTaskSuccessResult::Message(
+            "Successfully registered dpns name".to_string(),
+        ))
```
QuantumExplorer, line 204: Why do this, the put to platform and wait for response gives you the document that you just inserted as proof.

QuantumExplorer (changes requested):

src/backend_task/identity/register_dpns_name.rs
pauldelucia: (added a to-do because documents are returned for contested names as well)

QuantumExplorer (approved):

PR #79, opened 2024-11-20
fix: remove duplicate testnet nodes in env and remove unused env in exe dir

pauldelucia:
We had some duplicate nodes in the .env.example file (one of which is broken) and we weren't using .env in the exe dir.


.github/workflows/release.yml
```diff
@@ -68,7 +68,6 @@ jobs:
       - name: Setup prerequisites
         run: |
           mkdir -p dash-evo-tool/
-          cp .env.example dash-evo-tool/.env
```
ogabrielides, line 71: We need to keep this

src/app_dir.rs
```diff
@@ -45,16 +45,13 @@ pub fn copy_env_file_if_not_exists() {
...
     let env_file_in_app_dir = app_data_dir.join(".env".to_string());
-    if env_file_in_app_dir.exists() && env_file_in_app_dir.is_file() {
-    } else {
+    if !env_file_in_app_dir.exists() || !env_file_in_app_dir.is_file() {
         let env_example_file_in_exe_dir = PathBuf::from(".env.example");
         if env_example_file_in_exe_dir.exists() && env_example_file_in_exe_dir.is_file() {
             fs::copy(&env_example_file_in_exe_dir, env_file_in_app_dir)
                 .expect("Failed to copy main net env file");
         } else {
-            let env_file_in_exe_dir = PathBuf::from(".env");
```
ogabrielides, line 55: Same

ogabrielides (changes requested):

ogabrielides (approved):

PR #89, opened 2024-11-24
fix: network chooser screen statuses

pauldelucia:
- Fixed: network statuses would never switch from Online to Offline unless the app restarts
 - Fixed: only the currently active network status would update automatically

Now we check both networks every 3 seconds while the screen is active.

## Summary by CodeRabbit

- **New Features**
	- Enhanced network status handling with simultaneous retrieval of chain locks from both mainnet and testnet.
	- Introduced a new mechanism for rechecking network status every three seconds.

- **Bug Fixes**
	- Improved error handling and configuration loading for network tasks.

- **Refactor**
	- Simplified state management in the network chooser screen by removing unnecessary fields and methods.

### DIFF
--- a/src/backend_task/core/mod.rs
+++ b/src/backend_task/core/mod.rs
@@ -43,7 +43,7 @@ impl AppContext {
                         self.network,
                     ))
                 })
-                .map_err(|e| e.to_string()),
+                .map_err(|e| format!("Failed to get best chain lock: {}", e.to_string())),
             CoreTask::RefreshWalletInfo(wallet) => self.refresh_wallet_info(wallet),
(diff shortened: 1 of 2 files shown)


pauldelucia:
@coderabbitai review

QuantumExplorer (changes requested):
No, this is wrong, we don't want to get chain locks for both networks with that call, because if one network is offline you can't get the chain lock for the other network.

pauldelucia:
> No, this is wrong, we don't want to get chain locks for both networks with that call, because if one network is offline you can't get the chain lock for the other network.

Yes you can

QuantumExplorer (approved):

PR #115, opened 2024-12-03
fix: refresh owned dpns names

pauldelucia:
Refresh button wasn't actually refreshing owned DPNS names on that DPNS subscreen. It was just doing a contested resource query for the other subscreens.

Also made it so that self.refreshing is set to false if the screen changes. Otherwise, the refresh button can get stuck in a Refreshing state.


pauldelucia:
It's not perfect but it works. It seems that after refreshing My Usernames, you still need to navigate away from the screen and back again before the new name displays.

ogabrielides (approved):

PR #121, opened 2024-12-04
feat: show version in top panel

ogabrielides:


pauldelucia:
Why does mine say 0.3.1 - shouldn't it say 0.6.0?

Also, I suggest to change color to white and align the version text better.

pauldelucia (changes requested):
Why does mine say 0.3.1 - shouldn't it say 0.6.0?

Also, I suggest to change color to white and align the version text better.

pauldelucia:
Ah I see there is another PR for bumping version

pauldelucia (approved):

PR #238, opened 2025-05-05
feat: sams work

QuantumExplorer:
Includes various font support and various UI fixes and improvements.


pauldelucia (commented):
LGTM. The only thing is, I think normally backend tasks are for async tasks. Saving a token locally can probably just be an AppContext function

PR #252, opened 2025-05-07
fix: wallet not present when updating token config

lklimek:
When trying to update token config, `wallet not present` error is displayed.

The PR addresses an issue where wallet information was missing when updating token configurations, resulting in a "wallet not present" error. The root cause was in the get_identity_by_id function, which was retrieving identities from the database but not properly populating the associated_wallets field.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the process for locating identities, resulting in clearer error messages when identities cannot be listed or found.


lklimek:
@coderabbitai review all

QuantumExplorer:
This was fixed another way

PR #269, opened 2025-05-13
fix: get critical key for token action screens

pauldelucia:
On token action screens, we would select any authentication key High Medium or Critical, but only Critical can be used now on Platform.

## Summary by CodeRabbit

- **Refactor**
  - Updated key selection criteria for token-related actions to require only the highest security level (CRITICAL) authentication keys, enhancing the security standard across all token management screens.


lklimek:
@coderabbitai review

lklimek (approved):

PR #274, opened 2025-05-14
fix: display keyword search results even if optional description not present

thephez:
Previously, the UI would indicate "No tokens match your keyword" if results were returned but the token didn't define a description. Since token descriptions are optional, this PR removes the DET-only requirement to have a description for the search to work.

## Summary by CodeRabbit

- **Refactor**
  - Improved the handling of contract descriptions to ensure every contract ID is included in the results, even if the description is empty or missing. This change streamlines the display of contract information for users.


thephez:
@coderabbitai review

pauldelucia (approved):

PR #277, opened 2025-05-14
feat: deduplicate keywords automatically before trying to register token

### LINKED ISSUE #275: Unnecessary duplicate keyword error condition
If the Searchable box is checked, the token name is programatically inserted into the contract's keywords list. If the user also manually enters it in the _Contract Keywords_ field, it will result in this duplicate keyword error:

`Failed to register token contract: Failed to register DataContract: Protocol error: Data contract BEJi3E6raSJJFKEUqjywC9rYwc1CFqjUKNUrKJSsfaZP has duplicated keyword 'phez'.`

Image

thephez:
Uses a hash set to make sure the token keyword list is unique (automatically excludes duplicates).

Closes #275

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyword handling to remove duplicate entries from token search keywords, ensuring more accurate and efficient search results.

### DIFF
--- a/src/ui/tokens/tokens_screen.rs
+++ b/src/ui/tokens/tokens_screen.rs
@@ -4169,6 +4169,10 @@ Emits tokens in fixed amounts for specific intervals.
             }
         }
 
+        // ---- Deduplicate contract_keywords after adding token names ----
+        let mut seen = HashSet::new();
+        contract_keywords.retain(|kw| seen.insert(kw.clone()));
+
         let token_description = if self.token_description_input.len() > 0 {
             Some(self.token_description_input.clone())
         } else {


thephez:
@coderabbitai review

pauldelucia:
I think it would be better to throw an error in the UI rather than quietly removing the duplicate keyword @thephez . And also better than allowing the user to click Register and wait for Platform to return an error :)

thephez:
> I think it would be better to throw an error in the UI rather than quietly removing the duplicate keyword @thephez

If we throw an error it needs to be clear as to why it's present. The error currently reported is not clear when it's due to clicking Searchable AND putting the token name in the keyword list. This is because it's not intuitively obvious that the "Searchable" checkbox adds the token name to the keyword list - the UI doesn't show this. It took me about 15 minutes to figure out that was what was happening.

What do you think the error message should be? IMO there probably would need to be at least 2 error messages. One for the case where the error is happening because they've checked the Searchable box and one for everything else.

pauldelucia (changes requested):
See above

pauldelucia:
https://github.com/dashpay/dash-evo-tool/pull/279

PR #279, opened 2025-05-15
fix: show error if duplicate contract keywords in creator

pauldelucia:
The whole Searchable token name thing was very confusing. Gave the checkbox a new label and also checked for duplicate keywords and displayed error.

### DIFF
--- a/src/ui/tokens/tokens_screen.rs
+++ b/src/ui/tokens/tokens_screen.rs
@@ -2592,12 +2592,7 @@ impl TokensScreen {
                             let mut token_to_remove: Option<u8> = None;
                             for i in 0..self.token_names_input.len() {
                                 ui.label("Token Name (singular):");
-                                ui.horizontal(|ui| {
-                                    ui.text_edit_singleline(&mut self.token_names_input[i].0);
-                                    // Plural name
-                                    ui.label("(plural):");
-                                    ui.text_edit_singleline(&mut self.token_names_input[i].1);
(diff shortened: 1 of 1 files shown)


thephez:
This is definitely better. I still find the checkbox somewhat mysterious, but I can live with it. :slightly_smiling_face:

src/ui/tokens/tokens_screen.rs
```diff
@@ -2592,12 +2592,7 @@ impl TokensScreen {
                             let mut token_to_remove: Option<u8> = None;
                             for i in 0..self.token_names_input.len() {
                                 ui.label("Token Name (singular):");
-                                ui.horizontal(|ui| {
-                                    ui.text_edit_singleline(&mut self.token_names_input[i].0);
-                                    // Plural name
-                                    ui.label("(plural):");
-                                    ui.text_edit_singleline(&mut self.token_names_input[i].1);
-                                });
+                                ui.text_edit_singleline(&mut self.token_names_input[i].0);
```
lklimek, line 2595: As per description:

> The whole Searchable token name thing was very confusing. Gave the checkbox a new label and also checked for duplicate keywords and displayed error.

All I see is that the plural was removed, nothing about duplicate keywords.

thephez: See further down for that part (around comment "// If there are any duplicate keywords, show an error")

PR #280, opened 2025-05-19
feat: token direct purchase and set price

pauldelucia:
Add the screens for Token Direct Purchase and Set Price


QuantumExplorer (approved):
not really reviewed, but will see if it works now.

PR #295, opened 2025-05-29
fix: group info for token config update was not being set correctly

pauldelucia:
When initiating group actions we were always setting the group to manual mint control group! 💀

Identified the bug, told Claude to fix it. Just needed one follow-up prompt to fix a compile error.

```
I realized that in update_token_config.rs I am not properly setting self.group based on the ChangeItem that the user selects. I am just setting it to the Group authorized to change manual minting on screen construction, and always passing that group to the backend. Can you edit so that I properly set self.group dynamically based on the ChangeItem the user selects?
```

### DIFF
--- a/src/ui/tokens/update_token_config.rs
+++ b/src/ui/tokens/update_token_config.rs
@@ -24,7 +24,6 @@ use dash_sdk::dpp::data_contract::associated_token::token_configuration::accesso
 use dash_sdk::dpp::data_contract::associated_token::token_configuration_convention::TokenConfigurationConvention;
 use dash_sdk::dpp::data_contract::associated_token::token_configuration_item::TokenConfigurationChangeItem;
 use dash_sdk::dpp::data_contract::associated_token::token_distribution_rules::accessors::v0::TokenDistributionRulesV0Getters;
-use dash_sdk::dpp::data_contract::associated_token::token_marketplace_rules::v0::TokenTradeMode;
 use dash_sdk::dpp::data_contract::change_control_rules::authorized_action_takers::AuthorizedActionTakers;
 use dash_sdk::dpp::data_contract::group::accessors::v0::GroupV0Getters;
 use dash_sdk::dpp::data_contract::group::Group;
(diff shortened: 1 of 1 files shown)


src/ui/tokens/update_token_config.rs
```diff
@@ -87,29 +86,204 @@ impl UpdateTokenConfigScreen {
...
+    ) -> (Option<(GroupContractPosition, Group)>, Option<String>) {
+        let mut error_message = None;
+
+        let authorized_action_takers = match change_item {
+            TokenConfigurationChangeItem::Conventions(_) => identity_token_info
+                .token_config
+                .conventions_change_rules()
+                .authorized_to_make_change_action_takers(),
+            TokenConfigurationChangeItem::ConventionsControlGroup(_)
+            | TokenConfigurationChangeItem::ConventionsAdminGroup(_) => identity_token_info
```
lklimek, line 146: shouldn't this operator be `||` instead of `|`?

src/ui/tokens/update_token_config.rs
```diff
@@ -87,29 +86,204 @@ impl UpdateTokenConfigScreen {
...
+                .token_config
+                .conventions_change_rules()
```
lklimek, line 148: it's the same as above (Conventions, ConventionsControlGroup,ConventionsAdminGroup all return the same result), please double-check and consolidate into one match case if this is correct.

Also applies to matches below.

src/ui/tokens/update_token_config.rs
```diff
@@ -215,12 +364,11 @@ impl UpdateTokenConfigScreen {
...
 
-        let item = &mut self.change_item;
-
-        let default_token_configuration = &self.identity_token_info.token_config;
+        // Clone the token configuration to avoid borrowing issues
+        let default_token_configuration = self.identity_token_info.token_config.clone();
 
         ui.horizontal(|ui| {
-            let label = match item {
+            let label = match &self.change_item {
```
lklimek, line 371: cannot we just `impl Display for TokenConfigurationChangeItem` ?

PR #325, opened 2025-06-05
fix: searching for env and config files in wrong directories

pauldelucia:
## Summary by CodeRabbit

- **New Features**
	- Improved detection and copying of environment files, prioritizing the executable's directory before falling back to the current working directory.
	- Enhanced logic for locating the Dash configuration file, preferring the executable's directory for greater reliability.

- **Style**
	- Reformatted code for improved readability, including multi-line function signatures and argument lists.
	- Removed trailing whitespace in the user interface code.

### DIFF
--- a/src/app_dir.rs
+++ b/src/app_dir.rs
@@ -1,7 +1,7 @@
-use directories::{ProjectDirs, UserDirs};
+use dash_sdk::dpp::dashcore::Network;
+use directories::ProjectDirs;
 use std::fs;
 use std::path::PathBuf;
-use dash_sdk::dpp::dashcore::Network;
 
 const QUALIFIER: &str = ""; // Typically empty on macOS and Linux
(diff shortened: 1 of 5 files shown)


src/app_dir.rs
```diff
@@ -87,13 +92,38 @@ pub fn copy_env_file_if_not_exists() {
         app_user_data_dir_path().expect("Failed to determine application data directory");
     let env_file_in_app_dir = app_data_dir.join(".env".to_string());
     if !env_file_in_app_dir.exists() || !env_file_in_app_dir.is_file() {
+        // Get the directory where the executable is located
+        let exe_dir = std::env::current_exe()
```
lklimek, line 96: I think it would be better to embed the default .env file inside the binary.  Looks like `include_bytes!` can do the job, but I didn't test this.

PR #364, opened 2025-06-12
fix: error loading identities from local DB on token creation screen

lklimek:
Importing an identity from the network causes NULL wallet in database identity table. This causes errors when loading identities.


lklimek:
@coderabbitai review all

PR #366, opened 2025-06-12
feat: enable vertical scrolling on network chooser screen

thephez:
**Note:** I think there are other screens that have the same issue of sometimes displaying unreachable content. Probably all screens should be reviewed for that.

Previously, if the content did not fit on the screen there was no way to view it unless the application window could be made larger.

## Summary by CodeRabbit

- **New Features**
  - Added vertical scrolling to the network table in the central panel, allowing users to view more content within the Network Chooser screen.


thephez:
@coderabbitai review

Issue #427, opened 2025-08-10
Dash Evo Tool showing WIF private keys in Mainnet format when in Testnet mode 


qwizzie:
I created a few Platform Identities on Testnet through Dash Evo Tool and noticed the WIF private keys that DET shows us under Keys are not starting with c or 9 (testnet), they are starting with X (mainnet). 

We really do need the correct WIF private keys of Testnet Identities to show in Dash Evo Tool (set to Testnet mode), under Keys.
And frankly i would like their HEX as well, very useful in Testnet.

Meanwhile i will try creating an Identity on Platform TUI on Testnet, hopefully i get the correct Identity private key that way.

pshenmic:
I reported this issue before, and actually thought it has been fixed

https://github.com/dashpay/dash-evo-tool/issues/321

But PR with bugfix still not merged yet https://github.com/dashpay/dash-evo-tool/pull/327

qwizzie:
Maybe because pr #327 never left draft status ?

pauldelucia:
Resolved in #423

PR #433, opened 2025-08-14
feat: key selector widget

pauldelucia:


qwizzie:
I just hope this key selector widget will show the correct format of private keys, that it wont show mainnet format (X) when they need to be testnet format (c). See : https://github.com/dashpay/dash-evo-tool/issues/427

PR #444, opened 2025-10-07
feat: add existing identity by wallet + identity index

pauldelucia:
Just select a loaded wallet and identity index, we fetch by NonUniquePublicKeyHash, and derive the keys.

### DIFF
--- a/src/backend_task/identity/load_identity_from_wallet.rs
+++ b/src/backend_task/identity/load_identity_from_wallet.rs
@@ -355,10 +355,14 @@ impl AppContext {
                 .collect::<Vec<_>>()
                 .join(", ");
             format!(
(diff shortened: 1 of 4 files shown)


src/backend_task/identity/load_identity_from_wallet.rs
```diff
@@ -27,20 +28,84 @@ impl AppContext {
         sdk: &Sdk,
         wallet_arc_ref: WalletArcRef,
         identity_index: IdentityIndex,
```
lklimek, line 30: Feature idea:  

I think it's not trivial for the user to know identity index when the user recovers his wallet in new DET install.  Maybe we can also add iterative lookup for identity - I mean, instead of providing ID, try index 0, index 1, ... , up to some limit provided by the user?

pauldelucia: Implemented

PR #445, opened 2025-10-07
feat: remove wallet

pauldelucia:
Add ability to remove wallets. Associated identities are not removed with the wallets. The links are broken, but they can be re-established via the new Add Existing Identity by loaded wallet + identity index feature in #444

### DIFF
--- a/src/database/wallet.rs
+++ b/src/database/wallet.rs
@@ -68,6 +68,45 @@ impl Database {
         Ok(())
     }
 
+    /// Remove a wallet and all associated records from the database.
+    ///
+    /// This clears dependent records (addresses, utxos, asset locks, identity links)
+    /// to keep the database consistent before deleting the wallet itself.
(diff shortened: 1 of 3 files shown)


src/database/wallet.rs
```diff
@@ -68,6 +68,45 @@ impl Database {
...
+    ///
+    /// This clears dependent records (addresses, utxos, asset locks, identity links)
+    /// to keep the database consistent before deleting the wallet itself.
+    pub fn remove_wallet(&self, seed_hash: &[u8; 32], network: &Network) -> rusqlite::Result<()> {
+        let network_str = network.to_string();
+        let mut conn = self.conn.lock().unwrap();
+        let tx = conn.transaction()?;
+
+        let mut address_stmt =
+            tx.prepare("SELECT address FROM wallet_addresses WHERE seed_hash = ?")?;
```
lklimek, line 81: please review database schema and add `on delete cascade` statements to wallet foreign key in each table (if supported by the DB)

pauldelucia: This is already implemented

PR #447, opened 2025-10-08
fix: identity creation and top-up via QR code handling

pauldelucia:
Identity creation and top-up via QR code was getting stuck waiting for funds. The funds would go to the wallet balance and then you had to switch funding methods and finish. Now it works properly.


lklimek:
Tested and it works for me :+1: but didn't review the code.

PR #492, opened 2026-01-22
fix: top up with QR was validating against wallet balance

pauldelucia:
If we tried to top up via QR, it would give error if we input an amount higher than the selected wallet's balance, even though that's irrelevant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Max amount restrictions for identity top-ups now apply selectively based on the chosen funding method
  * Wallet balance is properly used to calculate the maximum available top-up amount when applicable
  * Amount input correctly synchronizes with available limits and button visibility

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


pauldelucia:
@coderabbitai review

PR #512, opened 2026-02-02
feat: use deterministic derivation path for platform address funding

pauldelucia:
Replace random one-time keys with deterministic keys derived from wallet seed for platform address funding asset locks.

Uses DIP-17 path structure with key_class=2 for funding: m/9'/coin_type'/17'/0'/2'/index

This is a local workaround until dash-sdk adds native support for platform_address_funding_path().

Changes:
- Add DerivationPathReference::PlatformAddressFunding variant
- Add PLATFORM_KEY_CLASS_FUNDING constant (key_class=2)
- Add platform_address_funding_ecdsa_private_key() method
- Add next_platform_address_funding_index() to track used indices
- Update generic_asset_lock_transaction() to use derived keys

## Summary by CodeRabbit

* **Security Improvements**
  * Wallet now uses deterministic key derivation from the seed for asset lock transactions, replacing random key generation for enhanced security and recoverability.

* **Infrastructure Updates**
  * Introduced platform address funding support with deterministic index tracking to improve wallet functionality and consistency.

### DIFF
--- a/src/model/wallet/mod.rs
+++ b/src/model/wallet/mod.rs
@@ -58,8 +58,10 @@ pub enum DerivationPathReference {
     BlockchainIdentityCreditInvitationFunding = 13,
     ProviderPlatformNodeKeys = 14,
     CoinJoin = 15,
-    /// DIP-17: Platform Payment Addresses
+    /// DIP-17: Platform Payment Addresses (key_class 0')
     PlatformPayment = 16,
+    /// DIP-17: Platform Address Funding (key_class 2')
+    PlatformAddressFunding = 17,
     Root = 255,
 }
 
@@ -85,6 +87,7 @@ impl TryFrom<u32> for DerivationPathReference {
             14 => Ok(DerivationPathReference::ProviderPlatformNodeKeys),
             15 => Ok(DerivationPathReference::CoinJoin),
             16 => Ok(DerivationPathReference::PlatformPayment),
+            17 => Ok(DerivationPathReference::PlatformAddressFunding),
             255 => Ok(DerivationPathReference::Root),
             value => Err(format!(
                 "value {} not convertable to a DerivationPathReference",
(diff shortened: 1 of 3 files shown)


src/model/wallet/mod.rs
```diff
@@ -1365,6 +1379,83 @@ impl Wallet {
...
+    /// This scans watched addresses for existing funding paths (key_class=2)
+    /// and returns the next unused index.
+    pub fn next_platform_address_funding_index(&self, network: Network) -> u32 {
+        let coin_type = match network {
+            Network::Dash => 5,
+            _ => 1,
+        };
+
+        let max_existing = self
+            .watched_addresses
```
lklimek, line 1424: When you have two computers with DET, there is no guarantee that watched_addresses will be the same on both. It means there is a potential index reuse weakness, leading to private key reuse.

I would rather derive the index from some identity metadata - I think we have identity index, don't we? So maybe we can use the identity index here instead?

PastaPastaPasta:
> This is a local workaround until dash-sdk adds native support for platform_address_funding_path().

can we just make this happen is dash-sdk?

thepastaclaw (changes requested):
## Code Review

Clean implementation of deterministic HD-derived keys for platform address funding. Derivation path follows DIP-17 conventions and concurrency is correctly handled within the wallet lock scope. However, the new funding path (key_class=2) is misclassified by `is_platform_payment()` which only checks the DIP-17 prefix without filtering on key_class.

*Reviewed commit: 2fa3a349*

🔴 1 blocking

### 1 additional finding

<details>
<summary>🔴 <b>blocking:</b> key_class=2 funding paths misclassified as Platform payment addresses</summary>

`src/model/wallet/mod.rs (line 174)`

`is_platform_payment()` only checks `m/9'/coin_type'/17'/...` and ignores the `key_class` component, so the new funding path `m/9'/coin_type'/17'/0'/2'/index` matches the payment predicate too. This leaks funding keys into payment-only flows:
- `platform_receive_address()` and new-address index selection
- `platform_addresses()` listing
- `get_platform_address_private_key()`
- Core refresh skips these Core P2PKH funding addresses as if they were Platform-only

The helper needs to require `key_class == 0` for payments, or callers need to filter on `DerivationPathReference::PlatformPayment` instead of the broad DIP-17 prefix.

</details>

<details>

```

In `src/model/wallet/mod.rs`:
- [BLOCKING] line 174: key_class=2 funding paths misclassified as Platform payment addresses
  `is_platform_payment()` only checks `m/9'/coin_type'/17'/...` and ignores the `key_class` component, so the new funding path `m/9'/coin_type'/17'/0'/2'/index` matches the payment predicate too. This leaks funding keys into payment-only flows:
- `platform_receive_address()` and new-address index selection
- `platform_addresses()` listing
- `get_platform_address_private_key()`
- Core refresh skips these Core P2PKH funding addresses as if they were Platform-only

The helper needs to require `key_class == 0` for payments, or callers need to filter on `DerivationPathReference::PlatformPayment` instead of the broad DIP-17 prefix.
```
</details>

lklimek:
Closing — superseded by #790, which takes a more complete approach to the same problem (deterministic HD keys for asset lock transactions). PR #790 uses `m/9'/coin_type'/15'/index` derivation with pre-registered funding addresses and full recovery support, replacing this earlier workaround.

PR #514, opened 2026-02-02
fix: incorrect fee calculation for platform-to-platform addresses transfer

lklimek:
Blocked by https://github.com/dashpay/platform/issues/3040#top


lklimek:
Replaced by #651

PR #521, opened 2026-02-03
refactor: use state transitions to estimate fees for identity create and top up

lklimek:
Fixes fee calculation for the following scnearios:

platform address -> existing identity
platform address -> new identity

Blocked by dashpay/platform#3040


lklimek:
Replaced by #651

PR #525, opened 2026-02-04
feat: working SPV mode, with identity creation support

PastaPastaPasta:
## Summary by CodeRabbit

* **New Features**
  * Added SPV finality event handling for asset lock proofs (instant and chain locks)
  * Integrated DAPI transaction queries for status verification
  * Display connected peer count in SPV sync status
  * Enhanced UI labels for keys, identities, and contracts

* **Bug Fixes**
  * Fixed transaction broadcast synchronization
  * Improved balance synchronization after asset locks
  * Enhanced address balance refresh in receive dialog

* **Improvements**
  * Optimized asset lock timeout based on backend mode (SPV vs RPC)
  * Better wallet state and UTXO management
  * Improved BIP32 account naming

## Summary by CodeRabbit

* **New Features**
  * SPV mode now handles finality events for improved transaction confirmation tracking.
  * Added unified transaction broadcasting across RPC and SPV backends.
  * Introduced UI helpers for clearer identity, contract, and key labeling.

* **Bug Fixes**
  * Improved wallet balance accuracy after spending UTXOs.
  * Fixed core address balance refresh in wallet receive dialogs.

* **Refactoring**
  * Streamlined transaction handling and asset lock proof waiting logic.
  * Centralized wallet balance recalculation across operations.

PR #528, opened 2026-02-04
feat: use DAPI for historical asset lock transaction info

PastaPastaPasta:
## Summary
- Add `get_transaction_info_via_dapi()` function that queries transaction status via DAPI gRPC instead of Core RPC
- Replace all `core_client.get_raw_transaction_info()` calls with the new DAPI function in identity registration, top-up, and platform address funding flows
- Removes the dependency on a local Core RPC node for checking whether asset lock transactions have been chainlocked

This is extracted from #525 as a standalone change that doesn't depend on the broader SPV improvements.

## Files changed
- `src/context.rs` — new `DapiTransactionInfo` struct and `get_transaction_info_via_dapi()` function
- `src/backend_task/identity/register_identity.rs` — 2 call sites replaced
- `src/backend_task/identity/top_up_identity.rs` — 2 call sites replaced
- `src/backend_task/wallet/fund_platform_address_from_asset_lock.rs` — 1 call site replaced

## Test plan
- [x] Verify `cargo build` and `cargo clippy` pass cleanly
- [ ] Test identity registration with an existing asset lock (UseAssetLock path)
- [ ] Test identity registration with wallet funding (FundWithWallet path — error fallback)
- [ ] Test identity top-up with an existing asset lock
- [ ] Test platform address funding from asset lock

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset lock proof verification accuracy for identity registration and top-up operations
  * Enhanced chain-lock status detection with better confirmation tracking
  * Refined error handling for expired lock proofs and fallback scenarios

* **Performance**
  * Improved stability during concurrent wallet operations
  * Streamlined transaction information retrieval and verification methods

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1708,6 +1708,39 @@ impl AppContext {
     }
 }
 
+pub(crate) struct DapiTransactionInfo {
+    pub is_chain_locked: bool,
+    pub height: u32,
+    pub confirmations: u32,
+}
+
+/// Query transaction info from DAPI. Works in both SPV and RPC modes
+/// since DAPI (platform gRPC) is always available via the SDK.
+pub(crate) async fn get_transaction_info_via_dapi(
+    sdk: &Sdk,
+    tx_id: &Txid,
+) -> Result<DapiTransactionInfo, String> {
+    use dash_sdk::dapi_client::{DapiRequestExecutor, IntoInner, RequestSettings};
+    use dash_sdk::dapi_grpc::core::v0::GetTransactionRequest;
+
+    let response = sdk
+        .execute(
+            GetTransactionRequest {
+                id: tx_id.to_string(),
+            },
+            RequestSettings::default(),
+        )
+        .await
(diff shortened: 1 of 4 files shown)


src/context.rs
```diff
@@ -1708,6 +1708,39 @@ impl AppContext {
...
 
+pub(crate) struct DapiTransactionInfo {
+    pub is_chain_locked: bool,
+    pub height: u32,
+    pub confirmations: u32,
+}
+
+/// Query transaction info from DAPI. Works in both SPV and RPC modes
+/// since DAPI (platform gRPC) is always available via the SDK.
+pub(crate) async fn get_transaction_info_via_dapi(
```
lklimek, line 1719: I would just name it `get_transaction_info`. Caller doesn't really want to know which method is used,to me it's just an implementation detail.

lklimek (approved):
just one really minor comment.

PR #529, opened 2026-02-05
fix: handle RegisteredTokenContract result to clear token creation spinner

pauldelucia:
The backend task returned RegisteredTokenContract on success but the tokens screen had no handler for it, causing the spinner to spin forever.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Token Creator UI state to properly reflect successful token contract registration, ensuring the status updates to completion when a token contract is registered.


PastaPastaPasta (approved):
utACK

PR #549, opened 2026-02-10
fix(db): add indexes, transaction atomicity, and N+1 query elimination

PastaPastaPasta:
## Summary
- Add missing network indexes to high-traffic database tables for faster queries
- Wrap insert_token() in a database transaction for atomicity
- Optimize identity and identity_order loading with LEFT JOIN to eliminate N+1 query patterns

## Test plan
- [ ] Verify app starts and loads identities/tokens without errors
- [ ] Check database query performance with existing data
- [ ] Confirm token insertion is atomic (all-or-nothing)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and safer parsing for data deserialization and ID fields.

* **Performance**
  * Faster identity and top-up loading via consolidated queries.
  * Added several network-specific database indexes (including for scheduled votes).

* **Chores**
  * Bumped database schema version and added migrations for new indexes.
  * Made token insertions transactional and introduced identity-token balance storage.

* **Tests**
  * Added comprehensive tests for token insertion, ordering, replacement, and deletion.


src/database/tokens.rs
```diff
@@ -128,11 +139,20 @@ impl Database {
...
+            let identity_id_bytes = identity.identity.id().to_vec();
+            tx.execute(
+                "INSERT INTO identity_token_balances
+                  (token_id, identity_id, balance, network)
+            VALUES (?1, ?2, ?3, ?4)
+            ON CONFLICT(token_id, identity_id, network) DO UPDATE SET
+                  balance = excluded.balance",
+                params![token_id_bytes, identity_id_bytes, 0u64, network],
+            )?;
         }
```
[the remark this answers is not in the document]
PastaPastaPasta: Won't this be resolved on refresh? Also is this new logic?

src/database/tokens.rs
```diff
@@ -107,7 +111,14 @@ impl Database {
         let token_id_bytes = token_id.to_vec();
         let data_contract_bytes = data_contract_id.to_vec();
 
-        self.execute(
+        // Collect identities before acquiring the connection lock for the transaction
+        let wallets = app_context.wallets.read().unwrap();
+        let identities = self.get_local_qualified_identities(app_context, &wallets)?;
+
+        let mut conn = self.conn.lock().unwrap();
```
lklimek, line 120: **MEDIUM — Latent deadlock risk: `wallets` read lock held across `conn` lock**

The `wallets.read()` lock (line 115) is held through the entire transaction (until line 153). This establishes a lock ordering of `wallets.read() -> conn.lock()`. If any future code path acquires these in reverse order (`conn.lock() -> wallets.write()`), it creates a deadlock.

Currently safe (no reverse ordering exists in the codebase), but fragile. The `identities` vector is already fully collected at line 116, so the `wallets` lock is no longer needed.

**Suggested fix:**
```rust
let identities = {
    let wallets = app_context.wallets.read().unwrap();
    self.get_local_qualified_identities(app_context, &wallets)?
};
// wallets lock released here, before conn.lock()
let mut conn = self.conn.lock().unwrap();
```

src/database/tokens.rs
```diff
@@ -128,11 +139,19 @@ impl Database {
...
-        let wallets = app_context.wallets.read().unwrap();
-        for identity in self.get_local_qualified_identities(app_context, &wallets)? {
-            self.insert_identity_token_balance(token_id, &identity.identity.id(), 0, app_context)?;
+        for identity in &identities {
+            let identity_id_bytes = identity.identity.id().to_vec();
+            tx.execute(
+                "INSERT INTO identity_token_balances
+                  (token_id, identity_id, balance, network)
+            VALUES (?1, ?2, ?3, ?4)
+            ON CONFLICT(token_id, identity_id, network) DO NOTHING",
```
lklimek, line 150: **INFO — `ON CONFLICT DO NOTHING` is correct but lacks test coverage**

This is the key behavioral fix from commit bcf4a30 — the old `ON CONFLICT DO UPDATE SET balance = excluded.balance` would overwrite existing non-zero balances back to 0. The new `DO NOTHING` correctly preserves them.

However, this critical fix has no direct regression test. Consider adding a test that:
1. Inserts a token (creates balance = 0)
2. Updates balance to a non-zero value
3. Calls `insert_token` again for the same token
4. Asserts balance is still non-zero

src/database/identities.rs
```diff
@@ -156,54 +156,54 @@ impl Database {
...
-            let mut top_ups = BTreeMap::new();
-            let mut rows = top_up_stmt.query(params![identity_id])?;
+            if current_id.as_ref() != Some(&id) {
+                // New identity row
+                let data: Vec<u8> = row.get(0)?;
+                let alias: Option<String> = row.get(1)?;
+                let wallet_index: Option<u32> = row.get(2)?;
+                let status: Option<u8> = row.get(3)?;
+
+                let mut identity = QualifiedIdentity::from_bytes(&data);
```
lklimek, line 184: **MEDIUM (pre-existing) — `from_bytes()` panics on corrupted data**

`QualifiedIdentity::from_bytes()` uses `.expect()` internally. If the `data` blob in the database is corrupted (partial write, model struct change, etc.), this will panic the entire application during identity loading.

This is not introduced by this PR (the old code had the same call), but the LEFT JOIN refactoring means a single corrupted identity now prevents loading ALL identities (previously only the corrupted one would fail via `query_map`).

Consider changing `from_bytes` to return `Result` and skipping corrupted entries with `tracing::warn!`.

lklimek (commented):
## Code Review Audit — Database Optimizations

Overall the PR is well-implemented. The N+1 → LEFT JOIN refactoring is correct, the transaction wrapping in `insert_token` adds proper atomicity, error handling improvements eliminate panics, and the new indexes are consistent between migration and fresh-install paths.

### Issues not addressable via inline comments (pre-existing, outside diff):

**CRITICAL — Guaranteed deadlock in `insert_remote_identity_if_not_exists` (`identities.rs:133-144`)**
This function acquires `self.conn.lock()` at line 133, then calls `self.execute()` at line 140 which tries to lock the same `Mutex` again. Rust's `std::sync::Mutex` is not reentrant — this is a guaranteed deadlock. Currently marked `#[allow(dead_code)]`, but if activated it will freeze the application. Fix: use the already-held `conn` reference directly (`conn.execute(...)`) instead of `self.execute(...)`.

**MEDIUM — `unreachable!()` on database data (`scheduled_votes.rs:162`)**
The `executed` column read as integer hits `unreachable!()` for any value other than 0 or 1. A corrupted or manually edited database would panic the app. Consider replacing with a default fallback or proper error.

**MEDIUM — `QualifiedIdentity::from_bytes().expect()` panic risk (`identities.rs:184, 243, 297, 333, 361`)**
All identity loading functions call `from_bytes()` which uses `.expect()` internally. Corrupted blob = app crash on startup. Consider returning `Result` and skipping corrupted entries with a warning log.

lklimek:
> **CRITICAL — Guaranteed deadlock in `insert_remote_identity_if_not_exists` (`identities.rs:133-144`)** 

Function not used, I removed it in #559 

> **MEDIUM — `unreachable!()` on database data (`scheduled_votes.rs:162`)** 
> **MEDIUM — `QualifiedIdentity::from_bytes().expect()` panic risk (`identities.rs:184, 243, 297, 333, 361`)**

I created a separate issue, #560, to track these

src/database/identities.rs
lklimek: to be fixed in #560

lklimek (approved):
Please address that src/database/tokens.rs:118 and we'll be good to merge.

PR #555, opened 2026-02-11
test: add 65 unit tests for wallet model and config

PastaPastaPasta:
## Summary
- Add 39 unit tests for wallet model core operations (balance calculations, UTXO selection, platform address info, WalletTransaction, WalletSeed lifecycle, derivation paths, address compatibility, BIP44 derivation, receive_address, utxos_by_address, WalletArcRef equality)
- Add 26 unit tests for config.rs parsing and validation (NetworkConfig validity, DAPI address list parsing, insight API URI, config_for_network across all network types, env var save format, envy parsing roundtrip with required/optional/missing/invalid fields)

Cherry-picked from `ralph/improvements` branch with fixes adapted to v1.0-dev signatures.

## Test plan
- [x] `cargo test --all-features -- wallet::mod::tests config::tests` — all 65 tests pass
- [x] `cargo clippy --all-features --all-targets -- -D warnings` — clean
- [x] `cargo test --doc --all-features --workspace` — passes

## Summary by CodeRabbit

* **Tests**
  * Expanded comprehensive test coverage for network configuration handling and wallet functionality, including validation logic, address derivation, balance management, transaction handling, and platform address operations across various scenarios.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -336,3 +336,546 @@ impl NetworkConfig {
         self
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Helper to create a minimal valid NetworkConfig for testing
+    fn make_network_config(
+        dapi_addresses: &str,
+        insight_api_url: &str,
+        port: u16,
+    ) -> NetworkConfig {
+        NetworkConfig {
+            dapi_addresses: dapi_addresses.to_string(),
+            core_host: "127.0.0.1".to_string(),
+            core_rpc_port: port,
+            core_rpc_user: "dashrpc".to_string(),
+            core_rpc_password: "password".to_string(),
+            insight_api_url: insight_api_url.to_string(),
+            core_zmq_endpoint: Some("tcp://127.0.0.1:23708".to_string()),
+            devnet_name: None,
+            wallet_private_key: None,
+            show_in_ui: true,
+        }
+    }
+
+    // ── NetworkConfig::is_valid ──────────────────────────────────────
+
+    #[test]
+    fn test_is_valid_with_all_fields() {
+        let config = make_network_config(
+            "https://127.0.0.1:443",
+            "https://insight.dash.org/insight-api",
+            9998,
+        );
+        assert!(config.is_valid());
+    }
+
+    #[test]
+    fn test_is_valid_empty_user() {
+        let mut config = make_network_config(
+            "https://127.0.0.1:443",
+            "https://insight.dash.org/insight-api",
+            9998,
+        );
+        config.core_rpc_user = String::new();
+        assert!(!config.is_valid());
+    }
+
+    #[test]
+    fn test_is_valid_empty_password() {
+        let mut config = make_network_config(
+            "https://127.0.0.1:443",
+            "https://insight.dash.org/insight-api",
+            9998,
+        );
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -336,3 +336,546 @@ impl NetworkConfig {
...
+    #[test]
+    fn test_dapi_address_list_multiple_addresses() {
+        let config = make_network_config(
+            "https://127.0.0.1:443,https://192.168.1.1:443,https://10.0.0.1:443",
+            "https://insight.dash.org/insight-api",
+            9998,
+        );
+        // Should not panic — valid addresses parse successfully
+        let _list = config.dapi_address_list();
+    }
```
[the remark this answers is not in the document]
PastaPastaPasta: Suggestion 2: dapi_address_list tests lack assertions

  Verdict: FALSE POSITIVE - not actionable

  AddressList is an opaque type from the external dash_sdk::dapi_client crate. It doesn't expose len() or iteration that we could assert on. The no-panic smoke tests are the best we can do for verifying that valid input parses without error, which is what
  they're meant to test. The companion test_dapi_address_list_empty_panics test with #[should_panic] already covers the failure case.

src/database/initialization.rs
```diff
@@ -118,7 +118,9 @@ impl Database {
                 self.initialize_token_table(tx)?;
                 self.drop_identity_token_balances_table(tx)?;
                 self.initialize_identity_token_balances_table(tx)?;
+                tx.execute("DROP TABLE IF EXISTS identity_order", [])?;
```
lklimek, line 121: why moved here from initialize_identity_order_table() and initialize_token_order_table()?

lklimek (approved):
I assume the tests don't break anything ;) so not checking them. 

One question posted, double-check and if it's ok, you can merge.

src/database/initialization.rs
PastaPastaPasta: The reviewer's question is valid and deserves an explanation, but the code change is correct.

  What happened

  Commit a5ae060b ("Fix race condition in parallel test DB initialization") deliberately moved the DROP TABLE IF EXISTS statements out of the initialize_identity_order_table() and initialize_token_order_table() functions and into the version 6 migration
  block.

  Why it was moved

  Before (the old pattern):
  // In initialize_identity_order_table():
  conn.execute("DROP TABLE IF EXISTS identity_order", [])?;
  conn.execute("CREATE TABLE identity_order (...)", [])?;

  This caused race conditions in parallel tests. When multiple tests initialize the DB concurrently:
  1. Test A drops the table
  2. Test B drops the table (no-op)
  3. Test A creates the table
  4. Test B tries to create → "table already exists" error

  After (the new pattern):
  // In initialize_identity_order_table():
  conn.execute("CREATE TABLE IF NOT EXISTS identity_order (...)", [])?;

  // In migration (version 6 only):
  tx.execute("DROP TABLE IF EXISTS identity_order", [])?;
  self.initialize_identity_order_table(tx)?;

  Why this is correct

  - Functions now use CREATE TABLE IF NOT EXISTS — safe for concurrent execution (tests, startup)
  - Migrations still explicitly drop + recreate — correct because migrations run exactly once per DB version upgrade, inside a transaction, and need to ensure a clean schema transition from v5→v6

PR #562, opened 2026-02-12
fix: replace expect()/unwrap() panics with error propagation

PastaPastaPasta:
## Summary

- Replace ~40 `.expect()` / `.unwrap()` / `panic!()` / `unreachable!()` / `unimplemented!()` calls across backend, database, context, and UI layers with proper error propagation or safe fallbacks
- **Backend tasks**: contested names, identity registration/top-up, contract loading, platform info, contacts — all now return `Result` instead of panicking
- **Database layer**: wallet loading, contested names, tokens, scheduled votes, contacts, asset lock DB, initialization/migration — errors propagate instead of crashing
- **AppContext::new()**: 13 `.expect()` calls replaced with match blocks that log errors and return `None`, so a misconfigured network context fails gracefully instead of taking down the whole app
- **Config/SDK**: `dapi_address_list()`, `insight_api_uri()`, and `initialize_sdk()` now return `Result` instead of panicking on bad input
- **UI**: `SystemTime::now().duration_since(UNIX_EPOCH).expect()` replaced with `.unwrap_or_default()` across 26 files; `unimplemented!()` in marketplace settings replaced with informational label

## Motivation

The application had numerous panic-prone code paths that could crash the entire GUI if any upstream data was malformed, a network was misconfigured, or a database query returned unexpected results. This PR systematically eliminates these crash vectors by:
1. Returning `Result`/`Option` from fallible functions
2. Using `match` blocks with `tracing::error!` logging before returning `None`
3. Using `.unwrap_or_default()` for non-critical timestamp operations
4. Replacing `unreachable!()` with safe default branches

## Test plan
- [x] `cargo build` passes
- [x] `cargo clippy --all-features --all-targets -- -D warnings` passes clean
- [x] `cargo +nightly fmt --all` applied
- [ ] Manual testing: verify app starts and functions normally on mainnet/testnet
- [ ] Verify graceful failure when a network config is intentionally broken (should log error, not crash)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Converted many crash paths into graceful errors, warnings, and safe fallbacks (app initialization, network/context selection, ZMQ listeners, background tasks, DB, wallets, identity flows).
  * Improved validation and automatic fallbacks for unavailable network contexts.

* **UI**
  * Made time/elapsed displays resilient to clock anomalies.
  * Replaced an unimplemented crash with a "marketplace not supported" label.
[...]

PR #563, opened 2026-02-12
fix: identity & token screen safety and validation

PastaPastaPasta:
## Summary

- Replace panic/unwrap() with proper error propagation for unsupported identity key types in IdentityKeys::to_public_keys_map()
- Add validation for missing signing key in 8 token screens (claim, destroy, freeze, mint, pause, resume, transfer, unfreeze) — shows user-friendly error instead of panicking
- Handle deleted identity gracefully in transfer/withdraw screen refresh
- Log silenced DB errors (.ok() / let _ =) in identity deletion, contact save, and contact list operations with tracing::warn!
- Enforce security level validation for ENCRYPTION/DECRYPTION key purposes in Identity Create screen (auto-sets MEDIUM, locks selection)
- Replace expect() on DocumentQuery creation in query_tokens.rs with map_err + ? error propagation
- Replace unwrap() on identity/key submission in token creator with validation check and user error message

## Test plan
- [ ] Verify identity creation with ECDSA keys still works
- [ ] Verify token creation flow with selected identity + key works
- [ ] Verify adding ENCRYPTION/DECRYPTION purpose keys auto-sets security level to MEDIUM
- [ ] Verify transfer/withdraw screens handle missing identity without crash
- [ ] Run cargo clippy --all-features --all-targets -- -D warnings (passes clean)

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened global error handling: many operations now return and surface errors instead of panicking.
  * Added precondition checks for missing signing keys across token flows to prevent crashes and show clear errors.
  * Database operations for contacts and identity deletes now log warnings on failure instead of being ignored.

* **Improvements**
  * Safer identity refresh/replace logic in transfer/withdraw flows to avoid unintended overwrites.
  * Automatic security-level adjustment when a key's purpose changes for clearer UX.
  * Query construction and parsing now fail gracefully with user-facing errors.

### DIFF
--- a/src/ui/dashpay/contacts_list.rs
+++ b/src/ui/dashpay/contacts_list.rs
@@ -1044,10 +1044,13 @@ impl ScreenLike for ContactsList {
 
                     // Clear all existing contacts for this identity from database first
                     // This prevents stale contacts from persisting
-                    let _ = self
+                    if let Err(e) = self
                         .app_context
                         .db
-                        .clear_dashpay_contacts(&owner_id, &network_str);
+                        .clear_dashpay_contacts(&owner_id, &network_str)
+                    {
+                        tracing::warn!("Failed to clear dashpay contacts from database: {}", e);
+                    }
 
                     // Convert ContactData to Contact structs and save to database
                     for contact_data in contacts_data {
@@ -1073,7 +1076,7 @@ impl ScreenLike for ContactsList {
                         self.contacts.insert(contact_data.identity_id, contact);
 
                         // Save to database
-                        let _ = self.app_context.db.save_dashpay_contact(
+                        if let Err(e) = self.app_context.db.save_dashpay_contact(
                             &owner_id,
                             &contact_data.identity_id,
                             &network_str,
@@ -1082,16 +1085,23 @@ impl ScreenLike for ContactsList {
                             contact_data.avatar_url.as_deref(),
                             None,       // public_message - not yet fetched
                             "accepted", // Only accepted contacts are returned from load_contacts
-                        );
+                        ) {
+                            tracing::warn!("Failed to save dashpay contact to database: {}", e);
+                        }
 
                         // Save private info if present
-                        if let Some(nickname) = &contact_data.nickname {
-                            let _ = self.app_context.db.save_contact_private_info(
+                        if let Some(nickname) = &contact_data.nickname
+                            && let Err(e) = self.app_context.db.save_contact_private_info(
                                 &owner_id,
                                 &contact_data.identity_id,
                                 nickname,
                                 &contact_data.note.unwrap_or_default(),
                                 contact_data.is_hidden,
+                            )
+                        {
+                            tracing::warn!(
+                                "Failed to save contact private info to database: {}",
+                                e
                             );
                         }
                     }
@@ -1163,7 +1173,7 @@ impl ScreenLike for ContactsList {
                     if let Some(identity) = &self.selected_identity {
                         let owner_id = identity.identity.id();
                         let network_str = self.app_context.network.to_string();
-                        let _ = self.app_context.db.save_dashpay_contact(
+                        if let Err(e) = self.app_context.db.save_dashpay_contact(
                             &owner_id,
                             &contact_id,
                             &network_str,
@@ -1172,7 +1182,12 @@ impl ScreenLike for ContactsList {
                             contact.avatar_url.as_deref(),
                             public_message.as_deref(),
                             "accepted",
-                        );
+                        ) {
+                            tracing::warn!(
+                                "Failed to save updated contact profile to database: {}",
+                                e
+                            );
+                        }
                     }
                 }
             }
--- a/src/ui/identities/add_new_identity_screen/mod.rs
+++ b/src/ui/identities/add_new_identity_screen/mod.rs
@@ -668,6 +668,7 @@ impl AddNewIdentityScreen {
                             });
                             row.col(|ui| {
                                 ui.vertical(|ui| {
+                                    let prev_purpose = *purpose;
                                     ComboBox::from_id_salt(format!("purpose_combo_{}", i))
                                         .selected_text(format!("{:?}", purpose))
                                         .show_ui(ui, |ui| {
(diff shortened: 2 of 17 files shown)


src/ui/dashpay/contacts_list.rs
```diff
@@ -1082,16 +1085,23 @@ impl ScreenLike for ContactsList {
                             contact_data.avatar_url.as_deref(),
                             None,       // public_message - not yet fetched
                             "accepted", // Only accepted contacts are returned from load_contacts
-                        );
+                        ) {
```
lklimek, line 1088: We need better error handling. See https://github.com/dashpay/dash-evo-tool/pull/562/changes#r2797870053

lklimek:
## Audit Summary

**Reviewed by:** Claude Code with a 3-agent team:
- `code-reviewer` — Correctness, edge cases, duplication, behavioral changes
- `security-engineer` — Panics, race conditions, key security, concurrency
- `rust-developer` — Rust idioms, error handling patterns, edition compatibility

### Overall Assessment

The PR is **well-structured and achieves its stated goals**. It systematically replaces 10+ panic/unwrap/expect paths with proper error handling across 17 files. No critical issues found. The changes significantly improve application resilience.

Cross-checked against PR #562 (`fix/replace-panics-with-error-propagation`) — no overlap. PR #562 addresses `SystemTime` expects and identity transition expects; this PR addresses signing key validation, identity refresh safety, DB error logging, and key type handling.

### Findings

| # | Severity | Location | Description |
|---|----------|----------|-------------|
| 1 | MEDIUM | `transfer_tokens_screen.rs` `refresh()` (~L273-291) | `refresh()` still uses double `.unwrap()` on identity lookup — can panic if identity is deleted while screen is open. Inconsistent with the graceful fix applied to `transfer_screen.rs` and `withdraw_screen.rs` in this same PR. |
| 2 | MEDIUM | 8 token screens (see inline) | Guard-then-unwrap anti-pattern: `is_none()` + early return, then later `.unwrap()`. Safe but fragile. `token_creator.rs` already uses the idiomatic `match` destructure — the 8 screens should follow that pattern. |
| 3 | MEDIUM | `transfer_tokens_screen.rs:231-233` | Pre-existing double `expect()` on `get_unqualified_contract_by_id()` — same code path as the `selected_key` fix but not addressed. Other token screens use `self.identity_token_info.data_contract.contract.clone()` which cannot panic. |
| 4 | MEDIUM | `add_new_identity_screen/mod.rs:696-715` | Security level enforcement for ENCRYPTION/DECRYPTION keys is UI-only. Backend `register_identity.rs` passes `IdentityKeys` through without validating purpose/security-level constraints. Low risk for desktop app (see inline). |
| 5 | LOW | `identities_screen.rs:934-943` | Identity removed from in-memory map *before* DB deletion. If DB fails, identity disappears from UI but persists in DB, reappearing on restart. |
| 6 | LOW | `pause_tokens_screen.rs`, `resume_tokens_screen.rs` | Key validation placed inside dialog match (after user confirms), unlike the other 6 screens which validate in `confirmation_ok()`. Minor UX inconsistency. |

### Pre-existing Issues (outside diff, for awareness)

- `transfer_tokens_screen.rs:231-233` — double `expect()` on data contract lookup (Finding #3 above)
- `register_identity.rs:297` — `asset_lock_proof.create_identifier().expect(...)` still present
- `identity/mod.rs:107-145` — sibling method `to_key_storage()` doesn't distinguish ECDSA_HASH160 from ECDSA_SECP256K1, inconsistent with the newly-fixed `to_public_keys_map()`

### Positive Observations

- `to_public_keys_map()` refactor from iterator+panic to for-loop+Result is clean and idiomatic
- Token creator validation uses proper `match` destructure — good reference pattern
- DB error logging with `tracing::warn!` is appropriate for the context
- Transfer/withdraw graceful refresh prevents real user-facing panics
- Let-chain syntax (`if let Some(...) && let Err(e)`) is valid for Rust Edition 2024
- No TOCTOU risk in egui single-threaded frame rendering
- SQL injection properly mitigated throughout (parameterized queries)

src/ui/tokens/mint_tokens_screen.rs
```diff
@@ -327,7 +333,7 @@ impl MintTokensScreen {
             sending_identity: self.identity_token_info.identity.clone(),
```
lklimek, line 336: **[MEDIUM] Guard-then-unwrap anti-pattern (affects all 8 token screens)**

This `.unwrap()` is safe because of the `.is_none()` guard + early return at line 280, but the pattern is fragile — if code between the guard and here is ever reordered, it panics.

`token_creator.rs:1487-1496` already uses the idiomatic approach in this same PR:

```rust
let (identity, signing_key) = match (&self.selected_identity, &self.selected_key) {
    (Some(id), Some(key)) => (id.clone(), key.clone()),
    _ => { /* error + return */ }
};
```

Consider applying the same pattern here and in the other 7 token screens (claim, destroy, freeze, pause, resume, transfer, unfreeze) to eliminate the `.unwrap()` entirely.

src/ui/identities/identities_screen.rs
```diff
@@ -934,22 +934,27 @@ impl IdentitiesScreen {
                             let mut lock = self.identities.lock().unwrap();
                             lock.shift_remove(&identity_id);
```
lklimek: **[LOW] Identity removed from UI before DB deletion succeeds**

`lock.shift_remove(&identity_id)` removes the identity from the in-memory map before the DB deletion at line 937. If the DB call fails (logged at line 940), the identity disappears from the UI but persists in the database — it will reappear on the next app restart.

Consider moving the `shift_remove` after a successful DB deletion, or showing a user warning if the DB operation fails.

src/ui/identities/add_new_identity_screen/mod.rs
```diff
@@ -681,7 +682,37 @@ impl AddNewIdentityScreen {
...
+                                                purpose,
+                                                Purpose::DECRYPTION,
+                                                "DECRYPTION",
+                                            );
                                         });
+                                    // Auto-set security level when purpose changes
+                                    if *purpose != prev_purpose {
+                                        match *purpose {
+                                            Purpose::TRANSFER => {
+                                                *security_level = SecurityLevel::CRITICAL;
```
lklimek, line 700: **[MEDIUM] Security level enforcement is UI-only**

The auto-set logic here correctly enforces MEDIUM for ENCRYPTION/DECRYPTION and CRITICAL for TRANSFER per Dash platform rules. However, the backend (`register_identity.rs`) passes the `IdentityKeys` struct through to `to_public_keys_map()` without validating that security levels match the key purpose.

For a desktop app this is low practical risk (UI is the only input path), but if the code is reused in a multi-user or API context, invalid combinations could slip through. The platform itself likely enforces these rules as defense-in-depth.

lklimek (changes requested):

lklimek (approved):

PR #564, opened 2026-02-12
feat: DashPay contacts improvements

PastaPastaPasta:
## Summary
- **Resolve usernames/profiles for contact request display** — incoming and outgoing contact requests now show resolved usernames and display names instead of raw identifiers
- **Connect ContactDetailsScreen to backend data flow** — contact detail view now fetches and displays live profile data from Platform, with local DB caching
- **Fix contact name loading and payment history DB queries** — payment history now loads from local database with proper contact name resolution instead of returning empty results
- **Replace misleading Cancel button on outgoing requests** — outgoing contact requests show an informational label instead of a non-functional Cancel button
- **Add timestamp data to contacts list** — enables Recent filter and DateAdded sort in the contacts list
- **Remove stale autoAcceptProof TODO** — clean up obsolete TODO comment and normalize tracing log levels in DashPay backend
- **Parallelize contact profile/username fetching** — use chunked join_all for concurrent profile resolution instead of sequential fetching

## Test plan
- [ ] Verify contact requests screen displays resolved usernames for both incoming and outgoing requests
- [ ] Open contact details and confirm profile data (display name, bio, avatar) loads from Platform
- [ ] Check payment history loads from local DB with correct contact names
- [ ] Verify outgoing contact requests show info label instead of Cancel button
- [ ] Confirm contacts list supports Recent filter and DateAdded sort ordering
- [ ] Verify no regressions in contact request accept/reject flows

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Payment history loads locally showing resolved contact names, direction (sent/received), amounts, memos, and recent records.
  * Contact details now persist local edits and trigger background fetches to refresh profile data.
  * Outgoing contact requests show recipient names/usernames when available.

* **Improvements**
  * Parallelized contact data loading for faster UI.
  * Better local name resolution with robust fallbacks and created_at-aware sorting.
  * Send-payment looks up contact names from local storage.

### DIFF
--- a/src/backend_task/dashpay.rs
+++ b/src/backend_task/dashpay.rs
@@ -22,6 +22,8 @@ pub mod validation;
 pub use contacts::ContactData;
 
 use crate::model::qualified_identity::QualifiedIdentity;
+use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
+use dash_sdk::dpp::platform_value::string_encoding::Encoding;
 use dash_sdk::platform::{Identifier, IdentityPublicKey};
 
 #[derive(Debug, Clone, PartialEq)]
@@ -165,21 +167,54 @@ impl AppContext {
                 identity,
                 request_id,
             } => contact_requests::reject_contact_request(self, sdk, identity, request_id).await,
-            DashPayTask::LoadPaymentHistory { identity: _ } => {
-                // TODO: Implement payment history loading according to DIP-0015
-                // This requires an SPV client to query the blockchain, which is not yet available.
-                // Once SPV support is added, the implementation would:
-                // 1. Get all established contacts (bidirectional contact requests)
-                // 2. For each contact, derive payment addresses from their encrypted extended public key
-                // 3. Query blockchain via SPV for transactions to/from those addresses
-                // 4. Build payment history records with amount, timestamp, memo, etc.
-                // 5. Store in local database for faster access
-                //
-                // The derivation path for DashPay addresses is:
-                // m/9'/5'/15'/account'/(our_identity_id)/(contact_identity_id)/index
-                //
-                // For now, return empty payment history until SPV client is available
-                Ok(BackendTaskSuccessResult::DashPayPaymentHistory(Vec::new()))
+            DashPayTask::LoadPaymentHistory { identity } => {
+                // Load locally stored payment records from database.
+                // Full blockchain-based history (scanning DIP-15 addresses via SPV)
+                // remains deferred until SPV support is available.
+                let identity_id = identity.identity.id();
+                let stored = self
+                    .db
+                    .load_payment_history(&identity_id, 100)
+                    .map_err(|e| format!("Failed to load payment history: {}", e))?;
+
+                let network_str = self.network.to_string();
+                let contacts = self
+                    .db
+                    .load_dashpay_contacts(&identity_id, &network_str)
+                    .unwrap_or_default();
+
+                let mut results = Vec::new();
+                for sp in stored {
+                    let is_incoming = sp.to_identity_id == identity_id.to_buffer().to_vec();
+                    let contact_bytes = if is_incoming {
+                        &sp.from_identity_id
+                    } else {
+                        &sp.to_identity_id
+                    };
+
+                    let contact_name = contacts
+                        .iter()
+                        .find(|c| c.contact_identity_id == *contact_bytes)
+                        .and_then(|c| c.username.clone().or(c.display_name.clone()))
+                        .unwrap_or_else(|| {
+                            if let Ok(cid) = Identifier::from_bytes(contact_bytes) {
+                                let s = cid.to_string(Encoding::Base58);
+                                format!("Unknown ({})", &s[..s.len().min(8)])
+                            } else {
+                                "Unknown".to_string()
+                            }
+                        });
+
+                    results.push((
+                        sp.tx_id,
+                        contact_name,
+                        sp.amount as u64,
+                        is_incoming,
+                        sp.memo.unwrap_or_default(),
+                    ));
+                }
+
+                Ok(BackendTaskSuccessResult::DashPayPaymentHistory(results))
             }
             DashPayTask::SendPaymentToContact {
                 identity,
--- a/src/backend_task/dashpay.rs
+++ b/src/backend_task/dashpay.rs
@@ -168,14 +168,9 @@ impl AppContext {
                 request_id,
             } => contact_requests::reject_contact_request(self, sdk, identity, request_id).await,
             DashPayTask::LoadPaymentHistory { identity } => {
-                // Load locally stored payment records from database.
-                // Full blockchain-based history (scanning DIP-15 addresses via SPV)
-                // remains deferred until SPV support is available.
+                // Reuse the shared payment history loader to avoid duplicating logic.
                 let identity_id = identity.identity.id();
-                let stored = self
-                    .db
-                    .load_payment_history(&identity_id, 100)
-                    .map_err(|e| format!("Failed to load payment history: {}", e))?;
+                let records = payments::load_payment_history(self, &identity_id, None).await?;
 
                 let network_str = self.network.to_string();
                 let contacts = self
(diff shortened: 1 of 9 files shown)


PastaPastaPasta:
@coderabbitai review

src/ui/dashpay/contact_requests.rs
```diff
@@ -220,6 +233,187 @@ impl ContactRequests {
...
+            return AppAction::None;
+        }
+
+        let identity = self.selected_identity.as_ref().unwrap().clone();
+        let mut action = AppAction::None;
+
+        for contact_id in unresolved_ids {
+            let task = BackendTask::DashPayTask(Box::new(DashPayTask::FetchContactProfile {
+                identity: identity.clone(),
+                contact_id,
```
lklimek: **CRITICAL: Only the last profile fetch task is dispatched; all others are silently dropped.**

`BitOrAssign` for `AppAction` (at `app.rs:156-165`) simply replaces `self` with `rhs` — it does not accumulate. So this loop only keeps the last iteration's task.

Fix: collect into a `Vec<BackendTask>` and return `AppAction::BackendTasks(tasks, BackendTasksExecutionMode::Concurrent)` which already exists in the codebase.

```suggestion
        let tasks: Vec<_> = unresolved_ids
            .into_iter()
            .map(|contact_id| {
                BackendTask::DashPayTask(Box::new(DashPayTask::FetchContactProfile {
                    identity: identity.clone(),
                    contact_id,
                }))
            })
            .collect();

        if tasks.is_empty() {
            return AppAction::None;
        }

        AppAction::BackendTasks(tasks, crate::app::BackendTasksExecutionMode::Concurrent)
```

src/backend_task/dashpay/payments.rs
```diff
@@ -349,20 +349,57 @@ pub async fn send_payment_to_contact_impl(
...
+            _ => PaymentStatus::Broadcast,
+        };
+
+        records.push(PaymentRecord {
+            id: sp.id.to_string(),
+            from_identity: from_id,
+            to_identity: to_id,
+            from_address: None,
+            to_address: Address::p2pkh(
+                &dash_sdk::dpp::dashcore::PublicKey::from_slice(&[0x02; 33])
```
lklimek: **HIGH: Dummy public key produces a fake but valid-looking address.**

`[0x02; 33]` is a technically valid compressed public key, so this generates a real-looking P2PKH address that doesn't correspond to the actual payment destination. If ever displayed to users or used in validation it will be wrong.

Consider making `to_address` an `Option<Address>` in `PaymentRecord`, or storing the actual destination address in the DB during `save_payment()`.

src/backend_task/dashpay.rs
```diff
@@ -165,21 +167,54 @@ impl AppContext {
...
+                                format!("Unknown ({})", &s[..s.len().min(8)])
+                            } else {
+                                "Unknown".to_string()
+                            }
+                        });
+
+                    results.push((
+                        sp.tx_id,
+                        contact_name,
+                        sp.amount as u64,
```
lklimek: **HIGH: `i64 as u64` silently wraps negative values.**

`StoredPayment.amount` is `i64` from SQLite. If a negative value exists (corruption, future refund logic), `as u64` wraps it to an astronomically large number (e.g. -1 → `u64::MAX`).

We should validate i return error when out of bounds.

src/backend_task/dashpay/payments.rs
```diff
@@ -349,20 +349,57 @@ pub async fn send_payment_to_contact_impl(
...
+                    .map_err(|e| format!("Key error: {}", e))?,
+                app_context.network,
+            ),
+            amount: sp.amount as u64,
```
lklimek: **HIGH: Same `i64 as u64` wrapping risk as in `dashpay.rs:211`.**

src/backend_task/dashpay/payments.rs
```diff
@@ -349,20 +349,57 @@ pub async fn send_payment_to_contact_impl(
...
+            tx_id: Some(sp.tx_id),
+            memo: sp.memo,
+            timestamp: sp.created_at as u64,
```
lklimek: **LOW: Same `i64 as u64` wrapping risk for timestamps.**

src/backend_task/dashpay.rs
```diff
@@ -165,21 +167,54 @@ impl AppContext {
...
-                // For now, return empty payment history until SPV client is available
-                Ok(BackendTaskSuccessResult::DashPayPaymentHistory(Vec::new()))
+            DashPayTask::LoadPaymentHistory { identity } => {
+                // Load locally stored payment records from database.
+                // Full blockchain-based history (scanning DIP-15 addresses via SPV)
+                // remains deferred until SPV support is available.
+                let identity_id = identity.identity.id();
+                let stored = self
+                    .db
+                    .load_payment_history(&identity_id, 100)
```
lklimek: **MEDIUM: Duplicate payment history logic.**

This inline implementation duplicates `payments::load_payment_history()` (at `payments.rs:351-403`) but with divergent return types (tuples vs `PaymentRecord`). The two implementations will drift apart over time. Consider delegating to the existing function and transforming its output.

src/ui/dashpay/contact_requests.rs
```diff
@@ -220,6 +233,187 @@ impl ContactRequests {
...
+            {
+                request.from_display_name = profile.display_name;
+                found = true;
+            }
+
+            // Try contacts cache (has username and display_name)
+            if let Some(selected_identity) = &self.selected_identity {
+                let owner_id = selected_identity.identity.id();
+                if let Ok(contacts) = self
+                    .app_context
```
lklimek: **MEDIUM: N+1 database query pattern.**

`load_dashpay_profile()` and `load_dashpay_contacts()` are called inside the loop for each unresolved request (both incoming and outgoing). Each call acquires the `Mutex<Connection>` and executes a full query. With N requests this means O(N) full DB reads of the same contacts data.

Load the contacts list once before the loops and reuse it.

src/ui/dashpay/contacts_list.rs
```diff
@@ -1069,6 +1079,7 @@ impl ScreenLike for ContactsList {
                             nickname: contact_data.nickname.clone(),
```
lklimek, line 1089: **MEDIUM: `created_at: None` breaks Recent filter and DateAdded sort for Platform-loaded contacts.**

Contacts arriving from `DashPayContactsWithInfo` backend results get `created_at: None`, so they're excluded from the Recent filter and sorted last by DateAdded. Consider using `Some(chrono::Utc::now().timestamp())` or carrying the timestamp from the DB save.

src/backend_task/dashpay/payments.rs
```diff
@@ -349,20 +349,57 @@ pub async fn send_payment_to_contact_impl(
...
+        }
 
-    tracing::debug!(
-        "Would load payment history for identity {} with contact filter: {:?}",
-        identity_id.to_string(Encoding::Base58),
-        contact_id.map(|id| id.to_string(Encoding::Base58))
-    );
+        let status = match sp.status.as_str() {
+            "confirmed" => PaymentStatus::Confirmed(1),
+            "failed" => PaymentStatus::Failed(sp.status.clone()),
```
lklimek: **LOW:** This stores the literal string `"failed"` as the failure reason (since `sp.status` is `"failed"`). If you want the actual error message, it would need to come from a separate DB column.

lklimek (changes requested):
There are a few important bugs in the code, please check and address them.

src/backend_task/dashpay/contacts.rs
```diff
@@ -409,81 +410,86 @@ pub async fn load_contacts(
...
+                    if let Ok(mut dpns_query) = DocumentQuery::new(dpns_contract, "domain") {
+                        dpns_query = dpns_query.with_where(WhereClause {
+                            field: "records.identity".to_string(),
+                            operator: WhereOperator::Equal,
+                            value: Value::Identifier(contact_id.to_buffer()),
+                        });
+                        dpns_query.limit = 1;
+
+                        if let Ok(results) = Document::fetch_many(sdk, dpns_query).await
+                            && let Some((_, Some(doc))) = results.into_iter().next()
```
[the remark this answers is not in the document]
lklimek: The SDK is already cloned out of the `RwLock<Sdk>` in `run_backend_task()` before `load_contacts()` is called, and the lock is dropped immediately. The `&Sdk` passed here is a plain borrow of that local clone — no lock is held across await points.

`Document::fetch_many()` takes `&Sdk`, so no ownership transfer occurs. The borrow is valid for the entire scope of the `join_all`, so there is no lifetime or lock-holding issue. Adding another clone would be harmless (it is cheap — all `Arc::clone()` internally) but unnecessary.

src/ui/dashpay/contact_requests.rs
```diff
@@ -220,6 +233,187 @@ impl ContactRequests {
...
+            .filter(|s| !s.is_empty());
+
+        if let Err(e) = self.app_context.db.save_dashpay_profile(
+            &contact_id,
+            &network_str,
+            display_name.as_deref(),
+            bio,
+            avatar_url,
+            None,
+        ) {
```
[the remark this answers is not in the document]
lklimek: `Document` implements a `get()` method directly (via `DocumentGetRawForDocumentType` or similar trait), so `doc.get("displayName")` compiles and works correctly. The `.properties().get()` pattern is an alternative accessor but is not required. The code compiles and passes all checks as-is.

lklimek (approved):

PR #616, opened 2026-02-22
fix(ui): prevent contract chooser from expanding multiple contracts on single click

thepastaclaw:
## Issue

Clicking any contract in the Contracts chooser panel would also expand another contract. For example, clicking only the GWR contract would expand both DPNS and GWR simultaneously (and vice versa).

## Root Cause

The GWR contract shown as `GWRSAVFMjXx8HpQFaNJMqBV7MBgMK4br5UESsB4S31Ec` **is** the DPNS contract — that's DPNS's Base58-encoded contract ID.

`get_contracts()` always injects system contracts (DPNS, DashPay, etc.) at fixed positions, but if the same contract was also added manually by the user (e.g. by fetching the DPNS contract by ID), it would exist in the database too. This resulted in two entries with the same contract ID but different display names — one showing "DPNS" (from the system alias) and one showing the raw Base58 ID (from the DB with no alias).

Since the expand/collapse state in `ContractChooserState.expanded_contracts` is keyed by contract ID, toggling either entry toggled the same key, causing both to expand or collapse simultaneously.

## Fix

1. **Deduplicate contracts**: After fetching from the DB, filter out any contracts whose ID matches a system contract before prepending the system contracts. This prevents duplicates.

2. **Scope egui widget IDs**: Wrap each contract's entire rendering block (header + content) in `ui.push_id(&contract_id)` for robustness — prevents potential widget ID collisions between different contracts' +/- buttons.

## Testing

- `cargo check` and `cargo fmt` pass
- Manual verification: expanding any single contract should only expand that contract
- The duplicate DPNS entry will no longer appear in the list

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the ordering of contract retrieval to display system contracts first in a consistent, predictable order.
  * Fixed an issue where duplicate contracts could appear in results through enhanced deduplication logic.
  * Enhanced pagination handling to ensure offset and limit parameters work correctly together.

## Validation

**What was tested:**
- `cargo check` — verified compilation after deduplication and egui `push_id` changes
- `cargo fmt` — formatting clean
- Manual UI verification: clicking a single contract in the chooser only expands that contract; duplicate DPNS entry no longer appears

**Results:**
- All local commands passed
- `Clippy` CI check — pass (5m25s)
- `Test Suite` CI check — pass (7m41s)

**Environment:** Local macOS arm64; GitHub Actions CI (ubuntu-latest)

### DIFF
--- a/src/context/contract_token_db.rs
+++ b/src/context/contract_token_db.rs
@@ -8,21 +8,28 @@ use dash_sdk::dpp::data_contract::accessors::v0::DataContractV0Getters;
 use dash_sdk::platform::{DataContract, Identifier};
 use dash_sdk::query_types::IndexMap;
 use rusqlite::Result;
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
 
 impl AppContext {
-    /// Retrieves all contracts from the database plus the system contracts from app context.
+    /// Retrieves system contracts plus contracts from the database.
+    ///
+    /// System contracts are always returned first in fixed order:
+    /// `dpns`, `token_history`, `withdrawals`, `keyword_search`, `dashpay`.
+    /// DB contracts are appended after and deduplicated by contract id against system contracts.
+    ///
+    /// `limit` and `offset` are applied to the DB query only. If `offset` is provided
+    /// without a `limit`, an unbounded limit is used so offset semantics remain valid.
     pub fn get_contracts(
         &self,
         limit: Option<u32>,
         offset: Option<u32>,
     ) -> Result<Vec<QualifiedContract>> {
-        // Get contracts from the database
-        let mut contracts = self.db.get_contracts(self, limit, offset)?;
+        let db_limit = limit.or(offset.map(|_| u32::MAX));
+        let db_contracts = self.db.get_contracts(self, db_limit, offset)?;
 
-        // Build the list of system contracts to prepend
-        let system_contracts = vec![
+        let system_contracts = [
             QualifiedContract {
                 contract: Arc::clone(&self.dpns_contract).as_ref().clone(),
                 alias: Some("dpns".to_string()),
@@ -45,20 +52,18 @@ impl AppContext {
             },
         ];
 
-        // Collect system contract IDs in a map to deduplicate DB contracts that match.
-        let system_contracts_by_id: std::collections::BTreeMap<_, _> = system_contracts
-            .iter()
-            .map(|c| (c.contract.id(), ()))
-            .collect();
-
-        // Remove any DB contracts that duplicate a system contract
-        contracts.retain(|c| !system_contracts_by_id.contains_key(&c.contract.id()));
-
-        // Prepend system contracts in order
-        let mut result = system_contracts;
-        result.append(&mut contracts);
-
-        Ok(result)
+        let mut seen_contract_ids = BTreeMap::new();
+        Ok(system_contracts
+            .into_iter()
+            .chain(db_contracts)
+            .filter_map(|qualified_contract| {
+                let contract_id = qualified_contract.contract.id();
+                seen_contract_ids
+                    .insert(contract_id, ())
+                    .is_none()
+                    .then_some(qualified_contract)
+            })
+            .collect())
     }
 
     pub fn get_contract_by_id(
--- a/src/context/contract_token_db.rs
+++ b/src/context/contract_token_db.rs
@@ -4,6 +4,7 @@ use crate::model::wallet::WalletSeedHash;
 use crate::ui::tokens::tokens_screen::{IdentityTokenBalance, IdentityTokenIdentifier};
 use bincode::config;
 use dash_sdk::dpp::data_contract::TokenConfiguration;
+use dash_sdk::dpp::data_contract::accessors::v0::DataContractV0Getters;
 use dash_sdk::platform::{DataContract, Identifier};
 use dash_sdk::query_types::IndexMap;
 use rusqlite::Result;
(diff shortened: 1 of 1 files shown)


PastaPastaPasta (approved):
utACK I think the title is a bit unclear; this fix is more "fix: deduplicate contract list when a system contract was manually added"

src/context/contract_token_db.rs
```diff
@@ -20,52 +21,42 @@ impl AppContext {
...
-            contract: Arc::clone(&self.dashpay_contract).as_ref().clone(),
-            alias: Some("dashpay".to_string()),
-        };
-
-        // Insert the DashPay contract at 4
-        contracts.insert(4, dashpay_contract);
-
-        Ok(contracts)
+        // Build the list of system contracts to prepend
+        let system_contracts = vec![
```
lklimek: I would just use BTreeMap or IndexMap keyed with contract id, but it's good enough.

lklimek (approved):

thepastaclaw:
Addressed Lukasz’s feedback on contract dedup in `get_contracts()`.

Updated to use a contract-id keyed map (`BTreeMap`) for deduplication while preserving ordering semantics (system contracts first, DB contracts appended only when not duplicates).

Commit: https://github.com/dashpay/dash-evo-tool/commit/c9db2ca6

lklimek:
@thepastaclaw claw you misunderstood my comment. Read whole function. See how you can benefit from BTreeMap or IndexMap. Like, get the result in lower number of lines.

thepastaclaw:
@lklimek Thanks — you were right, I had overcomplicated it. I re-read the whole `get_contracts()` flow and rewrote it as a single map-backed dedup pass.

What changed in `src/context/contract_token_db.rs`:
- switched to a single `BTreeMap`-based dedup pass over `system_contracts.into_iter().chain(db_contracts)`
- keeps system contracts first in fixed order, then appends DB contracts only when ID is new
- removed the separate `system_contracts_by_id + retain + append` sequence to reduce lines/complexity
- clarified doc comments on ordering/dedup behavior
- also made `offset`-without-`limit` safe for the DB query (`db_limit = limit.or(offset.map(|_| u32::MAX))`)

Commit pushed to this PR branch:
https://github.com/dashpay/dash-evo-tool/commit/ee9f4904

thepastaclaw:
@coderabbitai review

thepastaclaw (commented):
## Code Review

Clean UI fix that prevents contract chooser from expanding multiple contracts on a single click. The get_contracts dedup simplification is correct — preserves system-contract precedence and fixes the SQLite OFFSET-without-LIMIT edge case. No issues found.

*Reviewed commit: ee9f4904*

thepastaclaw:
✅ Review complete (commit ee9f4904)

thepastaclaw (commented):
## Code Review

The fix correctly addresses the duplicate-Identifier collision in `ContractChooserState.expanded_contracts`. The dedup-source approach is sound. Concerns raised by reviewers about hiding 'newer' DB revisions of system contracts don't hold in practice: `FetchContracts` uses `insert_contract_if_not_exists` (never overwrites), and `replace_contract` is only invoked from user-owned-contract update flows — so DB copies of system contracts can only be older or equal to the bundled snapshot. The remaining valid concerns are latent pagination semantics, the `u32::MAX` fallback masking an invalid-SQL bug at the DB layer, missing test coverage, and a small idiom cleanup.

*Reviewed commit: ee9f4904*

🟡 3 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/context/contract_token_db.rs`:
- [SUGGESTION] lines 24-67: Pagination is applied to raw DB rows, then dedup runs after — `limit`/`offset` will skew once duplicates exist
  After this change, `get_contracts(Some(N), Some(M))` returns the 5 system contracts unconditionally plus a paged-then-deduped slice of DB rows. If a duplicate of a system contract falls inside the requested DB window, it consumes one of the `limit` slots and is then dropped, so callers receive fewer unique results than requested and items can be skipped across pages. All current call sites pass `(None, None)`, so this is latent — but the function now silently means `system ∪ paged(db)` rather than `paged(system ∪ db)`. Either gate the system-contract prepend on `offset.unwrap_or(0) == 0` and adjust `limit` to account for system count, or split into `get_system_contracts()` / `get_user_contracts(limit, offset)` so the distinction is encoded in the API rather than relying on the doc comment.
- [SUGGESTION] lines 29-30: `u32::MAX` fallback papers over an invalid-SQL bug in `database::contracts::get_contracts`
  The `let db_limit = limit.or(offset.map(|_| u32::MAX));` exists because `database/contracts.rs::get_contracts` builds `... WHERE network = ? OFFSET ?` when `offset` is `Some` and `limit` is `None`, which SQLite rejects (LIMIT must precede OFFSET). Today every caller passes `(None, None)`, so this never fires, but the workaround entrenches a broken contract at the upper layer instead of fixing the actual bug at the DB layer (e.g. emit `LIMIT -1 OFFSET ?` or require both). Future direct callers of `db.get_contracts` will hit the same problem.
- [SUGGESTION] lines 24-67: No regression test for the dedup behavior
  The bug being fixed is exactly the kind of regression a small unit test would catch: insert a contract whose ID equals `dpns_contract.id()` into the DB, call `get_contracts(None, None)`, and assert the result has 5 entries with the system aliases preserved. Without a test, future refactors of this function can silently re-introduce the duplicate-Identifier collision that broke `ContractChooserState.expanded_contracts`. Per the project's testing conventions, new logic should have inline `#[test]` coverage, and this is a small pure function ideal for that.
```
</details>

Fresh dispatcher-requested review for exact queued SHA ee9f490425c5487705c783d178133f5687e64373. Existing/same-SHA review dedup was intentionally bypassed per coordinator instruction.

src/context/contract_token_db.rs
```diff
@@ -4,68 +4,66 @@ use crate::model::wallet::WalletSeedHash;
...
+            .chain(db_contracts)
+            .filter_map(|qualified_contract| {
+                let contract_id = qualified_contract.contract.id();
+                seen_contract_ids
+                    .insert(contract_id, ())
+                    .is_none()
+                    .then_some(qualified_contract)
+            })
+            .collect())
     }
```
thepastaclaw, line 67: **🟡 Suggestion: Pagination is applied to raw DB rows, then dedup runs after — `limit`/`offset` will skew once duplicates exist**

After this change, `get_contracts(Some(N), Some(M))` returns the 5 system contracts unconditionally plus a paged-then-deduped slice of DB rows. If a duplicate of a system contract falls inside the requested DB window, it consumes one of the `limit` slots and is then dropped, so callers receive fewer unique results than requested and items can be skipped across pages. All current call sites pass `(None, None)`, so this is latent — but the function now silently means `system ∪ paged(db)` rather than `paged(system ∪ db)`. Either gate the system-contract prepend on `offset.unwrap_or(0) == 0` and adjust `limit` to account for system count, or split into `get_system_contracts()` / `get_user_contracts(limit, offset)` so the distinction is encoded in the API rather than relying on the doc comment.

<details>

```

In `src/context/contract_token_db.rs`:
- [SUGGESTION] lines 24-67: Pagination is applied to raw DB rows, then dedup runs after — `limit`/`offset` will skew once duplicates exist
  After this change, `get_contracts(Some(N), Some(M))` returns the 5 system contracts unconditionally plus a paged-then-deduped slice of DB rows. If a duplicate of a system contract falls inside the requested DB window, it consumes one of the `limit` slots and is then dropped, so callers receive fewer unique results than requested and items can be skipped across pages. All current call sites pass `(None, None)`, so this is latent — but the function now silently means `system ∪ paged(db)` rather than `paged(system ∪ db)`. Either gate the system-contract prepend on `offset.unwrap_or(0) == 0` and adjust `limit` to account for system count, or split into `get_system_contracts()` / `get_user_contracts(limit, offset)` so the distinction is encoded in the API rather than relying on the doc comment.
```
</details>

src/context/contract_token_db.rs
```diff
@@ -4,68 +4,66 @@ use crate::model::wallet::WalletSeedHash;
...
-            contract: Arc::clone(&self.dashpay_contract).as_ref().clone(),
-            alias: Some("dashpay".to_string()),
-        };
-
-        // Insert the DashPay contract at 4
-        contracts.insert(4, dashpay_contract);
-
-        Ok(contracts)
+        let db_limit = limit.or(offset.map(|_| u32::MAX));
+        let db_contracts = self.db.get_contracts(self, db_limit, offset)?;
```
thepastaclaw, line 30: **🟡 Suggestion: `u32::MAX` fallback papers over an invalid-SQL bug in `database::contracts::get_contracts`**

The `let db_limit = limit.or(offset.map(|_| u32::MAX));` exists because `database/contracts.rs::get_contracts` builds `... WHERE network = ? OFFSET ?` when `offset` is `Some` and `limit` is `None`, which SQLite rejects (LIMIT must precede OFFSET). Today every caller passes `(None, None)`, so this never fires, but the workaround entrenches a broken contract at the upper layer instead of fixing the actual bug at the DB layer (e.g. emit `LIMIT -1 OFFSET ?` or require both). Future direct callers of `db.get_contracts` will hit the same problem.

<sub>source: ['claude']</sub>

<details>

```

In `src/context/contract_token_db.rs`:
- [SUGGESTION] lines 29-30: `u32::MAX` fallback papers over an invalid-SQL bug in `database::contracts::get_contracts`
  The `let db_limit = limit.or(offset.map(|_| u32::MAX));` exists because `database/contracts.rs::get_contracts` builds `... WHERE network = ? OFFSET ?` when `offset` is `Some` and `limit` is `None`, which SQLite rejects (LIMIT must precede OFFSET). Today every caller passes `(None, None)`, so this never fires, but the workaround entrenches a broken contract at the upper layer instead of fixing the actual bug at the DB layer (e.g. emit `LIMIT -1 OFFSET ?` or require both). Future direct callers of `db.get_contracts` will hit the same problem.
```
</details>

src/context/contract_token_db.rs
```diff
@@ -4,68 +4,66 @@ use crate::model::wallet::WalletSeedHash;
...
+            .into_iter()
+            .chain(db_contracts)
+            .filter_map(|qualified_contract| {
+                let contract_id = qualified_contract.contract.id();
+                seen_contract_ids
+                    .insert(contract_id, ())
+                    .is_none()
+                    .then_some(qualified_contract)
+            })
+            .collect())
```
thepastaclaw, line 66: **💬 Nitpick: Use `HashSet`/`BTreeSet` instead of `BTreeMap<Identifier, ()>` for the seen-id set**

`BTreeMap<Identifier, ()>` is being used as a set — only the `is_none()` of `insert` is consumed. `HashSet`/`BTreeSet::insert` returns `bool` (true if newly inserted), giving a more direct expression and removing the dummy `()`. `Identifier` is a fixed-length byte array so `HashSet` hashes cheaply, and ordering isn't relied on (the chain order determines output order). Pure cleanup, no behavioral change.

<details>

```suggestion
        let mut seen_contract_ids = std::collections::HashSet::new();
        Ok(system_contracts
            .into_iter()
            .chain(db_contracts)
            .filter(|qualified_contract| seen_contract_ids.insert(qualified_contract.contract.id()))
            .collect())
```

</details>

src/context/contract_token_db.rs
thepastaclaw, line 67: **🟡 Suggestion: No regression test for the dedup behavior**

The bug being fixed is exactly the kind of regression a small unit test would catch: insert a contract whose ID equals `dpns_contract.id()` into the DB, call `get_contracts(None, None)`, and assert the result has 5 entries with the system aliases preserved. Without a test, future refactors of this function can silently re-introduce the duplicate-Identifier collision that broke `ContractChooserState.expanded_contracts`. Per the project's testing conventions, new logic should have inline `#[test]` coverage, and this is a small pure function ideal for that.

<details>

```

In `src/context/contract_token_db.rs`:
- [SUGGESTION] lines 24-67: No regression test for the dedup behavior
  The bug being fixed is exactly the kind of regression a small unit test would catch: insert a contract whose ID equals `dpns_contract.id()` into the DB, call `get_contracts(None, None)`, and assert the result has 5 entries with the system aliases preserved. Without a test, future refactors of this function can silently re-introduce the duplicate-Identifier collision that broke `ContractChooserState.expanded_contracts`. Per the project's testing conventions, new logic should have inline `#[test]` coverage, and this is a small pure function ideal for that.
```
</details>

thepastaclaw:
Closing after cleanup review: this narrower contract chooser collision fix overlaps with #854’s broader duplicate-loaded-contracts path. Keeping both open would split/conflict the same area. If #854 is abandoned, the useful minimal behavior from this PR should be recreated with focused tests.

PR #619, opened 2026-02-22
fix(ui): replace f64 round-trip with integer arithmetic in max amount calculations

thepastaclaw:
## Issue

The "Max" button in transfer and withdraw screens uses a floating-point round-trip (`u64→f64→u64`) to subtract fee estimates from the available balance. This can cause precision loss with large credit amounts, potentially resulting in incorrect max values.

## Changes

Replace the floating-point arithmetic with integer `saturating_sub`:

```rust
// Before (precision loss possible)
let max_amount_minus_fee = (self.max_amount as f64 / 100_000_000_000.0 - 0.0002).max(0.0);
let max_amount_credits = (max_amount_minus_fee * 100_000_000_000.0) as u64;

// After (exact integer arithmetic)
let max_amount_credits = self.max_amount.saturating_sub(20_000_000);
```

- **transfer_screen.rs**: 0.0002 DASH = 20,000,000 credits
- **withdraw_screen.rs**: 0.005 DASH = 500,000,000 credits

`saturating_sub` also handles the edge case where balance < fee (returns 0 instead of underflowing).

## Validation

- `cargo check` ✅
- `cargo fmt` ✅

## Release Notes

* **Refactor**
  * Optimized internal calculation logic for transfer and withdrawal amount limits to improve performance and reliability.

### DIFF
--- a/src/ui/identities/transfer_screen.rs
+++ b/src/ui/identities/transfer_screen.rs
@@ -134,8 +134,8 @@ impl TransferScreen {
         ui.add_space(5.0);
 
         // Calculate max amount minus fee for the "Max" button
-        let max_amount_minus_fee = (self.max_amount as f64 / 100_000_000_000.0 - 0.0002).max(0.0);
-        let max_amount_credits = (max_amount_minus_fee * 100_000_000_000.0) as u64;
+        // 0.0002 DASH = 20_000_000 credits (1 DASH = 100_000_000_000 credits)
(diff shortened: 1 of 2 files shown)


thepastaclaw:
Duplicate of #622 — sub-agent created the same fix independently. Closing in favor of #622.

thepastaclaw:
@coderabbitai review

PR #621, opened 2026-02-22
fix: add 30-second timeout to SPV quorum public key lookup

thepastaclaw:
Wrap the SPV quorum public key lookup in a 30-second timeout to prevent indefinite hangs.

## Problem

`get_quorum_public_key` calls `get_quorum_by_height` on the SPV client interface with no timeout. If the SPV client is unresponsive or the network request stalls, this blocks the calling thread indefinitely (it runs inside `block_in_place`/`block_on`).

## Fix

Wrap the async lookup in `tokio::time::timeout(Duration::from_secs(30), ...)`. On timeout, a descriptive warning is logged and a clear error is returned with the quorum type, hash, and height for debugging.

30 seconds is generous for a quorum lookup — normal responses arrive in under a second.

Cherry-picked from `ralph/improvements` (commit 9c179886).

## Validation

**What was tested:**
- `cargo clippy --all-features --all-targets -- -D warnings` — lint check
- `cargo test --all-features --workspace` — full workspace test suite

**Results:**
- All local commands passed
- `Clippy` CI check — pass (5m49s)
- `Test Suite` CI check — pass (7m52s)

**Environment:** Local macOS arm64; GitHub Actions CI (ubuntu-latest)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added timeout handling for quorum lookup operations to prevent indefinite hangs. Operations exceeding 30 seconds now return a clear error message instead of waiting indefinitely, improving system reliability and responsiveness.

### DIFF
--- a/src/spv/manager.rs
+++ b/src/spv/manager.rs
@@ -649,14 +649,18 @@ impl SpvManager {
                 )
                 .await
                 .map_err(|_| {
-                    let msg = format!(
+                    tracing::warn!(
                         "Quorum lookup timed out after 30s at height {} for llmq_type={} hash=0x{}",
                         core_chain_locked_height,
                         quorum_type,
                         hex::encode(quorum_hash),
                     );
-                    tracing::warn!("{}", msg);
-                    msg
+                    format!(
+                        "Quorum lookup timed out after 30s at height {} for llmq_type={} hash=0x{}",
+                        core_chain_locked_height,
+                        quorum_type,
+                        hex::encode(quorum_hash),
+                    )
                 })?
                 .map(|q| {
                     tracing::debug!(
--- a/src/spv/manager.rs
+++ b/src/spv/manager.rs
@@ -643,28 +643,40 @@ impl SpvManager {
 
         tokio::task::block_in_place(|| {
             tokio::runtime::Handle::current().block_on(async {
-                interface
-                    .get_quorum_by_height(core_chain_locked_height, llmq_type, qh)
-                    .await
-                    .map(|q| {
-                        tracing::debug!(
-                            "Quorum public key found: type={}, hash={}, height={}",
-                            quorum_type,
-                            hex::encode(quorum_hash),
-                            core_chain_locked_height
-                        );
-                        *q.quorum_entry.quorum_public_key.as_ref()
-                    })
-                    .map_err(|e| {
-                        tracing::warn!(
-                            "Quorum lookup failed at height {} for llmq_type={} hash=0x{}: {}",
-                            core_chain_locked_height,
-                            quorum_type,
-                            hex::encode(quorum_hash),
-                            e
-                        );
-                        e.to_string()
-                    })
+                tokio::time::timeout(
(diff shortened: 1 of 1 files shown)


thepastaclaw:
@coderabbitai review

PastaPastaPasta (approved):
seems fine

src/spv/manager.rs
```diff
@@ -643,28 +643,40 @@ impl SpvManager {
...
+                )
+                .await
+                .map_err(|_| {
+                    let msg = format!(
+                        "Quorum lookup timed out after 30s at height {} for llmq_type={} hash=0x{}",
+                        core_chain_locked_height,
+                        quorum_type,
+                        hex::encode(quorum_hash),
+                    );
+                    tracing::warn!("{}", msg);
```
lklimek: Unnecessary use of format!. tracing::warn! will do the same work.

lklimek (changes requested):

src/spv/manager.rs
thepastaclaw: Good call — updated. I switched the warning to use `tracing::warn!` formatting directly and kept `format!` only for the returned error string. Change is in f5993b83.

thepastaclaw:
Addressed Lukasz's feedback by removing the unnecessary `format!` usage in the warning path (using `tracing::warn!` formatting directly) in commit f5993b83. I also resolved the review thread now that the change is in place.

Validation re-run on current branch:
- `cargo check -q` ✅

xdustinface (commented):
Has this been an issue? I wonder why this would be needed. There is no network request involved here its just a lookup in the masternode list.

xdustinface:
Also btw, you don't seem to be on a recent commit of rust-dashcore? The client interface doesn't exist anymore. The `DashSpvClient` is now clonable so you can probably simplify quite few things in DET in this regard.

thepastaclaw:
Good point — the current `v1.0-dev` branch pins rust-dashcore at `a05d256f` which is pretty old. Upgrading to pick up the clonable `DashSpvClient` and the removed client interface would be a nice follow-up, but it's a bigger change than this timeout fix and would need its own PR.

I'll open a tracker issue to bump the rust-dashcore dependency separately. Thanks for the heads up!

thepastaclaw:
You're right that the underlying operation is a local masternode list lookup — no network round-trip. But the call goes through `DashSpvClientInterface`'s command channel (`command_tx` → SPV client task → response), not a direct function call. If the SPV client task is busy, deadlocked, or its event loop is blocked, the channel wait hangs indefinitely inside `block_in_place`/`block_on`, which freezes the calling thread with no feedback.

The timeout is defensive against the command channel hanging, not against network latency. 30 seconds is generous enough to never trigger during normal operation, but prevents a silent hang from becoming a frozen UI.

That said, if you think the command channel pattern itself is robust enough that this can't happen in practice, happy to hear that — you know the SPV internals better than I do.

lklimek:
Downvote - I think we don't need this PR.

PR #658, opened 2026-02-25
refactor(spv): replace auto-stop with Connecting state and degraded warning

lklimek:
## Issue being fixed or feature implemented

When SPV starts and no peers connect, the old behavior was to auto-stop SPV after 10 seconds and show an error. This was counterproductive — SPV handles peer discovery and reconnection internally, so stopping it killed the very mechanism that would fix the problem.

### User Story

Imagine you're a user starting Dash Evo Tool in SPV mode on a slow or congested network. Previously, if peers took longer than 10 seconds to connect, SPV would abruptly stop and show a scary error banner, forcing you to manually restart. Now, the app shows a "Connecting..." state with a pulsing orange indicator while SPV keeps trying to find peers in the background. If it takes longer than 30 seconds, you get a gentle warning in the tooltip — but SPV stays running and will connect as soon as peers become available.

### Replaced auto-stop with graceful degraded state
- **Removed `stop_spv()` on peer timeout** from the main update loop in `app.rs`
- SPV now keeps running and retrying peer discovery when no peers are found

### Added four-state connection indicator
- `OverallConnectionState` now has `Disconnected`, `Connecting`, `Syncing`, `Synced`
- `Connecting` = SPV active but no peers yet (fast orange pulse at 2.5 Hz)
- `Syncing` = SPV active with peers (slow orange pulse at 1.2 Hz)
- Distinguished by peer count in `refresh_state()`

### Degraded-state warning
- After 30 seconds with zero peers (`SPV_PEER_DEGRADED_TIMEOUT`), a warning banner appears: "Having trouble finding peers. Check your network connection."
- SPV continues running — the warning is informational, not destructive

### Fixed MessageBanner::set_global idempotency
- `set_global` no longer resets timestamps when a banner with the same text already exists
- Safe to call from per-frame checks without side-effects

## How has this been tested?

- All 42 unit/integration tests pass
- Clippy clean with warnings-as-errors
- Manual test scenarios documented in `docs/ai-design/2025-02-25-spv-peer-rework/manual-tests.md`

## Breaking Changes

None

## Follow-up

- Event-driven peer status via `NetworkEvent` subscription (replaces polling): #659

## Summary by CodeRabbit

* **New Features**
  * SPV degraded-peer warning banner now appears when SPV has no peers for an extended period
  * Connection indicator expanded to four visible states (including a new "Connecting" state) with updated pulse behavior and tooltips

* **Bug Fixes**
  * Global banner handling made idempotent to avoid creating duplicate banner entries

* **Documentation**
  * Added manual test scenarios covering SPV connection states, degraded-warning behavior, and edge cases

### DIFF
--- a/docs/ai-design/2025-02-25-spv-peer-rework/manual-tests.md
+++ b/docs/ai-design/2025-02-25-spv-peer-rework/manual-tests.md
@@ -60,9 +60,10 @@ These scenarios verify the reworked SPV peer connection lifecycle:
 ### Expected Results
 
 - **Step 3:** Orange indicator with fast pulse. Tooltip: "Connecting...\nSPV: Starting\nDAPI: Available (...)".
-- **Step 5:** After ~30s, tooltip adds: "\nHaving trouble finding peers...".
-- **Step 6:** SPV **remains running** (does NOT stop). Indicator stays orange. No error banner appears. SPV continues trying DNS lookups and peer discovery in the background.
+- **Step 5:** After ~30s, tooltip adds: "\nHaving trouble finding peers. Check your connection." A **warning banner** also appears with the same text.
+- **Step 6:** SPV **remains running** (does NOT stop). Indicator stays orange. SPV continues trying DNS lookups and peer discovery in the background.
 - **Critical:** Verify NO "SPV disconnected" error banner. Verify `stop_spv()` is NOT called (check logs -- no "stopping SPV" message).
+- **Recovery:** If you restore connectivity and peers connect, the warning banner **automatically clears** and the indicator transitions to Syncing/Synced.
 
 ---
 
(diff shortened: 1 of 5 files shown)


xdustinface (commented):
Why would you want to stop SPV? It will still try to do DNS lookups and connect to other peers but if you just stop it it won't do anything anymore? Wouldnt it make more sense to distinguish in the UI between connecting and syncing instead? Also, it looks like DET is currently polling for peer status. There are network events that you can leverage for that instead which notify about any change in peers. See https://github.com/dashpay/rust-dashcore/blob/26e7bc4a89035018edc6117304c635b94841479c/dash-spv/src/client/events.rs#L34 and https://github.com/dashpay/rust-dashcore/blob/26e7bc4a89035018edc6117304c635b94841479c/dash-spv/src/network/event.rs#L14-L38

lklimek:
> Why would you want to stop SPV? It will still try to do DNS lookups and connect to other peers but if you just stop it it won't do anything anymore? Wouldnt it make more sense to distinguish in the UI between connecting and syncing instead? Also, it looks like DET is currently polling for peer status. There are network events that you can leverage for that instead which notify about any change in peers. See https://github.com/dashpay/rust-dashcore/blob/26e7bc4a89035018edc6117304c635b94841479c/dash-spv/src/client/events.rs#L34 and https://github.com/dashpay/rust-dashcore/blob/26e7bc4a89035018edc6117304c635b94841479c/dash-spv/src/network/event.rs#L14-L38

Thanks for feedback. For now, I will just add another state plus alert every 10 seconds if connection cannot be established. I'll create a issue followup issue to refactor this for full peer status pooling.

lklimek:
Follow-up issue: https://github.com/dashpay/dash-evo-tool/issues/659

PastaPastaPasta (approved):

PR #668, opened 2026-02-27
fix(spv): clear stale SPV state and banner on RPC mode switch

thepastaclaw:
## Issue Being Fixed

After switching from SPV to RPC backend mode, two pieces of stale SPV state could linger:

1. **`spv_connected_peers` and `spv_no_peers_since`** in `ConnectionStatus` were not cleared when entering the RPC branch of `trigger_refresh()`. If a user switched backend modes without a full `reset()`, the degraded peer timer could still be active, potentially causing `spv_peer_degraded()` to return `true` even though SPV was no longer in use.

2. **The SPV degraded banner** (`SPV_DEGRADED_BANNER`) was only cleared inside the `if backend_mode == CoreBackendMode::Spv` block. After switching to RPC, the banner would persist on screen since the clear path was never reached.

## Solution

- **`connection_status.rs`**: Clear `spv_connected_peers` (→ 0) and `spv_no_peers_since` (→ None) at the top of the `CoreBackendMode::Rpc` branch in `trigger_refresh()`.
- **`app.rs`**: Add an `else` branch to clear the SPV degraded banner when not in SPV mode.

## Validation

- Both fixes are additive — no existing behavior is altered for SPV mode
- RPC mode now always starts with clean SPV state, matching the behavior of `reset()`
- The banner clear is idempotent (no-op if already cleared)

Follow-up from #658 which introduced the degraded peer detection and banner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where stale SPV-related connection info could persist when switching between SPV and RPC modes, ensuring connection status updates accurately and preventing misleading peer/state displays.
* **Tests**
  * Added a unit test to verify SPV shutdown state is cleared and refresh timing behaves correctly after switching modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/dash-evo-tool/pull/668)

PR #669, opened 2026-02-27
fix(spv): clear stale SPV state and banner on backend mode switch

thepastaclaw:
## Issue

When switching from SPV to RPC backend mode, two related issues could leave stale SPV state visible to the user:

1. **Stale SPV state leak (Medium):** `refresh_zmq_and_spv()` enters the `Rpc` branch but never clears `spv_no_peers_since` or `spv_connected_peers`. If the mode switch happens without a full `reset()`, the degraded peer timer persists.

2. **Banner not cleared on mode switch (Low-Medium):** The SPV degraded banner is only managed inside `if core_backend_mode() == Spv`. Switching away while the banner is showing leaves it stuck.

Found in code review of #658.

## Changes

- **`connection_status.rs`:** Clear `spv_connected_peers` and `spv_no_peers_since` at the top of the `Rpc` branch in `refresh_zmq_and_spv()`.
- **`app.rs`:** Add an `else` branch to unconditionally clear the SPV degraded banner when not in SPV mode.

## Testing

`cargo check` passes.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1170,6 +1170,9 @@ impl App for AppState {
             } else {
                 MessageBanner::clear_global_message(ctx, SPV_DEGRADED_BANNER);
             }
+        } else {
+            // Ensure stale SPV banner is cleared after mode switch.
+            MessageBanner::clear_global_message(ctx, SPV_DEGRADED_BANNER);
         }
 
         for action in actions {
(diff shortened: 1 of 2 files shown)


thepastaclaw:
Duplicate of #668 — I opened this without checking that Codex had already created a PR. Closing.

PR #671, opened 2026-03-02
chore(deps): update dashpay/platform to rev 570e3af0

lklimek:
## What was done?

Update `dashpay/platform` dependency from rev `0fa82e66` to `570e3af0`.

Adapt to breaking changes in the transitive `rust-dashcore` dependency (`a05d256f` → `2824e52a`):

- **`FeeLevel` removed** → replaced with `FeeRate::normal()` direct calls
- **`WalletManager::create_unsigned_payment_transaction()` removed** → replaced with `TransactionBuilder` + `WalletManager::get_change_address()`
- **`DashSpvClientInterface` / `DashSpvClientCommand` removed** → store `Arc<SpvClient>` directly; use `client.get_quorum_at_height()` for quorum lookups
- **`DashSpvClient::start()` now private, `monitor_network()` removed** → use `client.run(CancellationToken)` which handles start + monitor + stop
- **`subscribe_sync_events()`, `subscribe_network_events()`, `subscribe_progress()` now async** → added `.await`
- **`SyncState::Initializing` removed** → replaced with `SyncState::WaitForEvents`

## How has this been tested?

- `cargo clippy --all-features --all-targets -- -D warnings` ✅
- `cargo +nightly fmt --all` ✅
- 274 unit tests passing ✅

## Test plan

No manual test scenarios needed — this is a pure dependency update with API adaptation, no user-facing behavior changes.

## Breaking Changes

None

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation if needed

## Summary by CodeRabbit

* **Refactor**
  * Enhanced transaction fee calculation system for improved fee estimation accuracy
  * Improved transaction building architecture to better manage payment transaction construction
  * Simplified SPV client management through optimized internal client reference handling
  * Refined network sync state display by streamlining progress state mappings

### DIFF
--- a/src/backend_task/core/mod.rs
+++ b/src/backend_task/core/mod.rs
@@ -23,7 +23,9 @@ use dash_sdk::dpp::dashcore::{
 use dash_sdk::dpp::fee::Credits;
 use dash_sdk::dpp::key_wallet::Network as WalletNetwork;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::ManagedWalletInfo;
-use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::fee::FeeLevel;
+use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::coin_selection::SelectionStrategy;
+use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::fee::FeeRate;
+use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::transaction_builder::TransactionBuilder;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::transaction_building::AccountTypePreference;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::wallet_info_interface::WalletInfoInterface;
 use dash_sdk::dpp::key_wallet_manager::wallet_manager::{WalletError, WalletId, WalletManager};
@@ -540,12 +542,12 @@ impl AppContext {
                 .map(|(addr, amt)| (addr.clone(), (*amt as f64 * scale_factor) as u64))
                 .collect();
 
-            match wm.create_unsigned_payment_transaction(
+            match Self::build_unsigned_payment_tx(
+                wm,
                 wallet_id,
                 DEFAULT_BIP44_ACCOUNT_INDEX,
-                Some(AccountTypePreference::BIP44),
+                AccountTypePreference::BIP44,
                 scaled_recipients,
-                FeeLevel::Normal,
                 current_height,
             ) {
                 Ok(tx) => return Ok(tx),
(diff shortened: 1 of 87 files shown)


lklimek:
## Summary for rust-dashcore developer

**Adapted to changes in rust-dashcore `a05d256f` → `2824e52a`:**

| Breaking change in rust-dashcore | How we adapted in dash-evo-tool |
|---|---|
| `FeeLevel` enum removed from `key_wallet::managed_wallet_info::fee` | Replaced all `FeeLevel::Normal.fee_rate().calculate_fee(size)` with `FeeRate::normal().calculate_fee(size)` |
| `WalletManager::create_unsigned_payment_transaction()` removed (`transaction_building.rs` gone from key-wallet-manager) | Built equivalent using `TransactionBuilder::new().set_fee_rate().set_change_address().add_output().select_inputs().build()` + `WalletManager::get_change_address()` for change address, `ManagedWalletInfo::accounts()` for UTXOs |
| `DashSpvClientInterface` / `DashSpvClientCommand` removed (client `interface` module gone) | Store `Arc<SpvClient>` directly; call `client.get_quorum_at_height()` for quorum lookups instead of going through the command channel |
| `DashSpvClient::start()` now `pub(super)`, `monitor_network()` removed | Use `client.run(CancellationToken)` which handles start + monitor loop + stop internally |
| `subscribe_sync_events()`, `subscribe_network_events()`, `subscribe_progress()` now `async` | Added `.await` to all subscribe calls |
| `SyncState::Initializing` variant removed | Replaced with `SyncState::WaitForEvents` (the new `#[default]`) in all match arms |

src/spv/manager.rs
```diff
@@ -641,10 +630,15 @@ impl SpvManager {
...
+            .load_full()
+            .ok_or_else(|| "SPV client not initialized".to_string())?;
+
         tokio::task::block_in_place(|| {
             tokio::runtime::Handle::current().block_on(async {
-                interface
-                    .get_quorum_by_height(core_chain_locked_height, llmq_type, qh)
+                client
+                    .get_quorum_at_height(core_chain_locked_height, llmq_type, qh)
                     .await
```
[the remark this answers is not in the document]
thepastaclaw: The `Handle::current()` call here is inside `tokio::task::block_in_place()`, which itself requires an active multi-threaded Tokio runtime — it will panic before `Handle::current()` ever gets a chance to if there's no runtime. Since DET always runs within a Tokio runtime (see `main.rs`), and `get_quorum_public_key` is only called via the `ContextProvider` trait from within the SDK's async context, this is not a real panic risk.

That said, using `Handle::try_current()` to convert a panic into an `Err` is a cheap defensive improvement. Worth considering in a follow-up, but not critical for this PR since the invariant (always inside tokio runtime) is well-established in this app.

thepastaclaw:
Reviewed the CodeRabbit nitpick comments:

**1. Change-address side effects across retry attempts (lines 539-552)**

`get_change_address(..., true)` is called inside `build_unsigned_payment_tx` which runs per retry iteration. Even if `true` advances the HD derivation index, this is harmless in practice — HD wallets handle index gaps gracefully (the gap limit scanning on recovery will find all used addresses). "Wasting" a few change indices during failed iterations has no real consequence for wallet functionality. Not worth adding complexity to cache/reuse the address.

**2. Fragile "Insufficient" substring matching (lines 672-675)**

The error comes from `TransactionBuilder::select_inputs()` in rust-dashcore, where the error message is a known, controlled string. The `contains("Insufficient")` check is pragmatic — it matches what the dependency actually emits. Making it case-insensitive (`.to_ascii_lowercase().contains("insufficient")`) would be marginally more defensive but doesn't add real value when the error source is pinned to a specific git rev. A better long-term fix would be for rust-dashcore to expose typed error variants instead of string matching, but that's upstream work, not something to address in this adaptation PR.

src/spv/manager.rs
```diff
@@ -551,9 +552,7 @@ impl SpvManager {
             *storage_guard = None;
         }
 
-        if let Ok(mut interface_guard) = self.client_interface.write() {
-            *interface_guard = None;
-        }
+        // spv_client is cleared asynchronously when the client stops; no action needed here.
```
xdustinface, line 555:
```suggestion
```

Adding comments like this is some imo weird claude behaviour, consider adding this to your `CLAUDE.md`:

- Never add comments explaining why removed code isn't needed, why something was removed, or what used to happen at a given location. If code is gone, it's gone and there is no explanation needed in its place.

src/backend_task/core/mod.rs
```diff
@@ -611,10 +613,72 @@ impl AppContext {
...
 
         let estimated_size = Self::estimate_p2pkh_tx_size(spendable_inputs, 1);
-        let fee = FeeLevel::Normal.fee_rate().calculate_fee(estimated_size);
+        let fee = FeeRate::normal().calculate_fee(estimated_size);
         Ok(spendable_total.saturating_sub(fee))
     }
 
+    /// Build an unsigned payment transaction using TransactionBuilder.
+    ///
+    /// Replaces the removed `WalletManager::create_unsigned_payment_transaction`.
```
xdustinface, line 610:
```suggestion
```

Same here.

src/spv/manager.rs
```diff
@@ -149,8 +149,9 @@ pub struct SpvManager {
     wallet: Arc<AsyncRwLock<WalletManager<ManagedWalletInfo>>>,
     // Storage manager for direct access to SPV data (shared component from client)
     storage: Arc<Mutex<Option<Arc<tokio::sync::Mutex<DiskStorageManager>>>>>,
-    // Interface for sending commands to the running SPV client (quorum lookups, etc.)
-    client_interface: Arc<RwLock<Option<DashSpvClientInterface>>>,
+    // Shared reference to the running SPV client (for quorum lookups, etc.)
+    // ArcSwapOption gives wait-free reads (quorum lookups) and atomic set/clear on start/stop.
```
xdustinface:
```suggestion
```

No point in documenting ArcSwap here imo.

src/spv/manager.rs
```diff
@@ -841,8 +831,11 @@ impl SpvManager {
             }
         }
 
+        // Store the client reference for quorum lookups (wait-free reads via ArcSwap)
```
xdustinface:
```suggestion
```

This and some other places in this PR are also some excessive claude commenting imo. Not an issue and you might want to just leave it but just FYI, i gave my claude bunch of rules about its weird commenting behaviours like here i think this would have avoided it most of the times: 

- Only add comments when they provide meaningful context that isn't obvious from the code itself. Don't comment self-explanatory code, simple one-liners, or anything a competent Rust developer would understand at a glance.

xdustinface (changes requested):
Looks good just few comment nits that you may or may not want to address but i think the CodeRabbit finding at least should be investigated. Didn't look that much into it but i think we could just create the change address outside the loop and pass it in there? The `true` marks the address as used and with 50 possible fee iterations this potentially might lead to lot of wasted addresses which even if its not an issue is not helpful or good practice either.

> `539-552`: **Verify change-address side effects across retry attempts.**
>
>Line 633 calls `get_change_address(..., true)` and this helper is invoked repeatedly in >the retry loop (Lines 539-552). If `true` reserves/advances change derivation, failed >attempts may consume multiple change addresses unnecessarily.
>
>
>```web
>In dash-sdk revision 570e3af0, what does the last boolean argument of >`WalletManager::get_change_address(...)` control? Does passing `true` advance/reserve >the change index?
>```
>
>Also applies to: 632-634

lklimek:
Follow-up PR: https://github.com/dashpay/dash-evo-tool/pull/676

lklimek:
Merged in as part of another PR.

PR #725, opened 2026-03-10
fix: fall back to ENCRYPTION key when recipient lacks DECRYPTION key for ECDH

### LINKED ISSUE #687: impossible to Add Contact by Username
on Add Contact screen we can add concat by Username or Identity ID, but when we set existing Username it couldn't be found

Image
Image

cc: @thepastaclaw  @lklimek

thepastaclaw:
## Issue

Fixes https://github.com/dashpay/dash-evo-tool/issues/687 (partial — contact request interoperability)

## Description

Contact requests fail when the recipient identity only has an `ENCRYPTION` key (no `DECRYPTION` key). Many identities created by the mobile wallet or other tools only carry an `ENCRYPTION` key, which is equally valid for ECDH.

### Root Cause

`send_contact_request_with_proof` only looked for a recipient key with `Purpose::DECRYPTION`. When the recipient had only `Purpose::ENCRYPTION`, the lookup returned `None` and the request failed — but the error message incorrectly implied the *sender* should add keys.

### Fix

1. **Key fallback**: Try `Purpose::DECRYPTION` first (preferred per DIP-15 convention), then fall back to `Purpose::ENCRYPTION` — both are valid ECDSA_SECP256K1 keys that work identically for ECDH
2. **New `RecipientMissingKey` error variant**: Distinguishes "recipient needs to add a key" from "sender needs to add a key", so the UI shows the correct guidance
3. **UI updates**: Both `AddContactScreen` and `ContactRequests` recognize the new error and display appropriate messaging

## Testing

- `cargo check` ✅
- `cargo clippy` ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Contact request key lookup now accepts an encryption key when a decryption key isn't available.
  * Error handling now prioritizes detecting a recipient who lacks a compatible key to avoid misleading key-type errors.

* **UI**
  * Added a clear, actionable message prompting recipients to add a compatible key when missing.

### DIFF
--- a/src/backend_task/dashpay/contact_requests.rs
+++ b/src/backend_task/dashpay/contact_requests.rs
@@ -254,8 +254,11 @@ pub async fn send_contact_request_with_proof(
             "Sender does not have a compatible ECDSA_SECP256K1 ENCRYPTION key for ECDH. Please add a DashPay-compatible encryption key to your identity.".to_string()
         })?;
 
-    // Find a recipient DECRYPTION key that supports ECDH (must be ECDSA_SECP256K1)
-    // Platform enforces MEDIUM security level for ENCRYPTION/DECRYPTION keys
+    // Find a recipient key that supports ECDH (must be ECDSA_SECP256K1).
+    // Try DECRYPTION first (preferred per DIP-15 convention), then fall back to
+    // ENCRYPTION — both are valid ECDSA_SECP256K1 keys and work identically for
+    // ECDH.  Many identities created by the mobile wallet or other tools only
+    // carry an ENCRYPTION key, so the fallback is essential for interoperability.
     let recipient_key = to_identity
         .get_first_public_key_matching(
             Purpose::DECRYPTION,
@@ -263,8 +266,16 @@ pub async fn send_contact_request_with_proof(
             HashSet::from([KeyType::ECDSA_SECP256K1]),
             false,
         )
+        .or_else(|| {
+            to_identity.get_first_public_key_matching(
+                Purpose::ENCRYPTION,
+                HashSet::from([SecurityLevel::MEDIUM]),
+                HashSet::from([KeyType::ECDSA_SECP256K1]),
+                false,
+            )
+        })
         .ok_or_else(|| {
-            "Recipient does not have a compatible ECDSA_SECP256K1 DECRYPTION key for ECDH. They need to add a DashPay-compatible decryption key to their identity.".to_string()
+            "Recipient does not have a compatible ECDSA_SECP256K1 key (ENCRYPTION or DECRYPTION) for ECDH. They need to add a DashPay-compatible key to their identity.".to_string()
         })?;
 
     // Step 4: Generate ECDH shared key and encrypt data
--- a/src/backend_task/dashpay/errors.rs
+++ b/src/backend_task/dashpay/errors.rs
@@ -36,6 +36,9 @@ pub enum DashPayError {
     #[error("Missing DECRYPTION key required for DashPay")]
     MissingDecryptionKey,
 
+    #[error("Recipient identity is missing a DashPay-compatible key")]
+    RecipientMissingKey,
+
     #[error("ECDH key generation failed: {reason}")]
     EcdhFailed { reason: String },
 
@@ -174,6 +177,9 @@ impl DashPayError {
             DashPayError::MissingDecryptionKey => {
                 "Your identity is missing a DECRYPTION key required for DashPay. Please add a DashPay-compatible decryption key.".to_string()
             }
+            DashPayError::RecipientMissingKey => {
+                "The recipient's identity does not have a DashPay-compatible key. They need to add an ENCRYPTION or DECRYPTION key to their identity before you can send them a contact request.".to_string()
+            }
             _ => "An error occurred. Please try again.".to_string(),
         }
     }
@@ -203,6 +209,7 @@ impl DashPayError {
                 | DashPayError::MissingField { .. }
                 | DashPayError::MissingEncryptionKey
                 | DashPayError::MissingDecryptionKey
+                | DashPayError::RecipientMissingKey
         )
     }
 }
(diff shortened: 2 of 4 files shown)


thepastaclaw:
@coderabbitai review

lklimek:
Claw this is a known Platform bug. File a bug report in dashpay/platform and close this pull request.

thepastaclaw:
Closing this PR per @lklimek's feedback — this is a known Platform-level bug that should be fixed upstream rather than worked around in dash-evo-tool.

Filed as dashpay/platform#3224 to track the fix in Platform.

thephez:
> Claw this is a known Platform bug. File a bug report in dashpay/platform and close this pull request.

@lklimek Why do contact requests (between identities lacking a decryption key) work okay in the Android wallet if it's a platform bug? Maybe you mean it's an rs-sdk bug?

thepastaclaw:
Reopening — confirmed this IS a DET client-side bug, not a Platform issue. thephez verified that Android wallet users can exchange contact requests successfully with only ENCRYPTION keys. Platform's data trigger doesn't validate key purposes for ECDH; it's our hardcoded `Purpose::DECRYPTION` lookup that fails. This PR's fallback approach is the correct fix.

lklimek:
@thephez I consulted in @QuantumExplorer , he said it's platform bug and should be fixed there. 

And it sounds suspicious from cryptography perspective - in some schemes, misuse of keys lead to potential attack scenarios. Not sure if here, but it raised yellow flag, so I'm not happy to just merge that without good understanding of impact.

thephez:
> @thephez I consulted in @QuantumExplorer , he said it's platform bug and should be fixed there.
> 
> And it sounds suspicious from cryptography perspective - in some schemes, misuse of keys lead to potential attack scenarios. Not sure if here, but it raised yellow flag, so I'm not happy to just merge that without good understanding of impact.

Got it. More a violation of best practice than a bug per se. Can DET support already established contacts but disallow creating new ones this way? For the case where someone wanted to import an Android created identity that already has contacts for use in DET. Probably not a very high priority, but maybe useful.

lklimek:
@thepastaclaw please check the original dashpay spec (dip) and let us know what the specs say

thepastaclaw:
## DIP Spec Review: Key Purposes for ECDH in Contact Requests

@lklimek I checked the relevant DIPs. Here's what the specs say:

### DIP-11 (Identities) — Key Purpose Distinction

DIP-11 explicitly distinguishes between ENCRYPTION and DECRYPTION key purposes and prescribes specific roles for each in ECDH:

> "When encrypting a field that is intended to be decrypted by a recipient, the owner of the document should use the private key of one of his **encryption keys** and a public **decryption key** valid on the recipient's Identity for the key exchange."

So per spec:
- **Sender** uses their **ENCRYPTION** key (private) for ECDH
- **Recipient's DECRYPTION** key (public) is used for the other half

DIP-11 also states that identities are required to maintain:
> "at least one encryption key at a high security level and one decryption at a medium security level"

This means a properly formed identity **should** have both key types. An identity with only an ENCRYPTION key and no DECRYPTION key is technically non-compliant with DIP-11.

### DIP-15 (DashPay) — Contact Request Key Indices

DIP-15 steps 3-4 for creating a contact request say:

> 3. `senderKeyIndex` should be set to a valid public key index from the sender's identity. This key must allow for a Diffie-Hellman key agreement protocol.
> 4. `recipientKeyIndex` should be set to a valid public key index from the recipient's identity. This key must be of the same type as the sender key and must allow for a Diffie-Hellman key agreement protocol.

DIP-15 itself doesn't explicitly say "must be DECRYPTION purpose" for the recipient key — it just says the key must support ECDH. But DIP-11 (the identity spec that DIP-15 depends on) does make this distinction.

### Platform Enforcement

I checked the Platform data trigger for contact requests (`create_contact_request_data_trigger_v0`). It only validates:
1. Owner can't send a contact request to themselves
2. Recipient identity must exist

**Platform does NOT validate key purposes** — it doesn't check whether `senderKeyIndex` points to an ENCRYPTION key or `recipientKeyIndex` points to a DECRYPTION key. This is purely a client-side concern.

### Cryptographic Reality

From a pure cryptography perspective, both ENCRYPTION and DECRYPTION keys are secp256k1 keys that work **identically** for ECDH. The purpose distinction is semantic/logical, not cryptographic. Using an ENCRYPTION key where a DECRYPTION key is expected doesn't weaken the ECDH exchange or introduce any attack vector — the key exchange math is the same regardless of the purpose label.

### Summary

| Question | Answer |
|---|---|
| Does the spec distinguish ENCRYPTION vs DECRYPTION for ECDH? | **Yes** — DIP-11 explicitly assigns sender→ENCRYPTION, recipient→DECRYPTION |
| Are identities required to have both key types? | **Yes** — DIP-11 requires at least one ENCRYPTION (high) and one DECRYPTION (medium) |
| Does Platform enforce key purposes for contact requests? | **No** — validation only checks identity existence |
| Is there a cryptographic risk to fallback? | **No** — both are secp256k1 keys, ECDH works identically |
| Should identities lacking DECRYPTION keys exist? | **Not per spec** — but they do in practice (Android wallet creates them) |

The root issue appears to be that some identity creation flows (e.g., Android wallet) don't register a DECRYPTION key, violating DIP-11's requirement. This could be fixed either:
1. **Upstream in Platform** — enforce DECRYPTION key requirement during identity creation/update, or
2. **Client-side with fallback** — like this PR does, accepting the reality that non-compliant identities exist on the network

@lklimek @thephez hope this helps clarify what the specs say. The "correct" fix depends on whether you want to enforce spec compliance (fix identity creation) or be tolerant of existing non-compliant identities (client fallback).

thepastaclaw:
@coderabbitai review

thepastaclaw (commented):
## Code Review

probe

*Reviewed commit: 2ccbf0d2*

lklimek:
Related: #810

lklimek:
Closed because this PR goes against the spec. This needs to be addressed in create flow.

PR #733, opened 2026-03-12
fix(ui): prevent 'No users found' flash during profile search

### LINKED ISSUE #684: Search Public Profiles always show "No users found" at the beginning

thepastaclaw:
## Issue

Closes https://github.com/dashpay/dash-evo-tool/issues/684

## Problem

When searching public profiles, "No users found" flashes briefly before actual results appear.

**Root cause:** `display_message()` unconditionally set `self.loading = false`. During a search, informational messages (banner updates) can arrive before the actual `DashPayProfileSearchResults` result. When `loading` is prematurely cleared while `has_searched` is true and `search_results` is still empty, the UI renders the "No users found" empty state — until the real results arrive a moment later via `display_task_result()`.

## Fix

Only clear the loading state in `display_message()` when the message is an error. The success path in `display_task_result()` already handles setting `loading = false` when results arrive, so this is safe.

## Validation

- `cargo check` — clean
- `cargo clippy -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- Verified that `display_task_result` still sets `loading = false` for all match arms, so no risk of a stuck spinner on success
- Verified that error messages still correctly stop the loading state

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the profile search would briefly display a "No users found" message while loading results, creating a confusing user experience. The loading state now properly persists until results are available or an actual error occurs.

### DIFF
--- a/src/ui/dashpay/profile_search.rs
+++ b/src/ui/dashpay/profile_search.rs
@@ -250,9 +250,14 @@ impl ProfileSearchScreen {
         action
     }
 
-    pub fn display_message(&mut self, _message: &str, _message_type: MessageType) {
+    pub fn display_message(&mut self, _message: &str, message_type: MessageType) {
         // Banner display is handled globally by AppState; this is only for side-effects.
-        self.loading = false;
+        // Only stop loading on errors — success results are handled by display_task_result.
+        // Setting loading = false unconditionally causes "No users found" to flash briefly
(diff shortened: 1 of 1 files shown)


thepastaclaw:
@coderabbitai review

thepastaclaw:
Superseded by #736 which addresses the same issue.

PR #736, opened 2026-03-12
fix(ui): prevent 'No users found' flash during profile search

### LINKED ISSUE #684: Search Public Profiles always show "No users found" at the beginning

thepastaclaw:
## Issue
Fixes #684 — Search Public Profiles always shows "No users found" briefly before results load.

## Root Cause
When a search starts, `has_searched` is set to `true` and results are cleared. The `display_message()` method (called for banner side-effects) sets `loading = false` before `display_task_result()` delivers actual results. This creates a frame where `has_searched && !loading && results.is_empty()` is true, causing the "No users found" message to flash.

## Fix
Added a `search_completed` flag that's only set in `display_task_result()`, and changed the empty-results condition to check `search_completed` instead of `has_searched`. This ensures the "No users found" message only appears after the backend has actually returned (empty) results.

## Testing
- Start a profile search → spinner shows, no flash of "No users found"
- Complete search with results → results display normally
- Complete search with no matches → "No users found" displays (after results arrive)
- Clear Results button → resets all state correctly

## Summary by CodeRabbit

* **New Features**
  * Real-time progress updates from long-running backend tasks are now shown in the UI and global refresh banner.
  * Search UI now tracks backend completion separately to drive loading and result visibility.

* **Bug Fixes**
  * Prevented premature "No users found" message by showing no-results only after backend search completes.
  * Improved loading/reset behavior on search and clear to avoid misleading states.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1112,6 +1112,10 @@ impl App for AppState {
                             self.visible_screen_mut()
                                 .display_task_result(unboxed_message);
                         }
+                        BackendTaskSuccessResult::Progress(_) => {
+                            self.visible_screen_mut()
+                                .display_task_result(unboxed_message);
+                        }
                         BackendTaskSuccessResult::UpdatedThemePreference(new_theme) => {
                             self.theme_preference = new_theme;
                             MessageBanner::set_global(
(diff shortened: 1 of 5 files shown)


thepastaclaw:
@coderabbitai review

thepastaclaw:
Closing as duplicate of #748, which has the more complete fix. Apologies for the noise — this was caused by a heartbeat loop that didn't check for existing PRs before re-investigating the same issue.

PR #748, opened 2026-03-13
fix(ui): prevent false 'No users found' flash in profile search

### LINKED ISSUE #684: Search Public Profiles always show "No users found" at the beginning

thepastaclaw:
## Issue

Closes #684

When searching public profiles, "No users found" would flash briefly before actual results appeared.

## Root Cause

`ProfileSearchScreen` manages a `loading` flag that controls whether to show a spinner or the "No users found" message. The problem: `loading` was being cleared in two places that didn't discriminate between search results and unrelated task completions:

1. **`display_message()`** — called by `AppState` for *any* task's success/error banner while this screen is visible. It unconditionally set `self.loading = false`.
2. **`display_task_result()`** — set `self.loading = false` at the *top* of the function, before matching on the result variant. Any `Message`, `Refresh`, or other variant from a concurrent task would clear loading.

When another backend task completed while a profile search was in flight, the loading flag was cleared prematurely → empty results + `has_searched = true` + `loading = false` → "No users found" shown before real results arrived.

## Fix

- **`display_message()`**: No longer touches `loading` state (it's for banner side-effects, not search lifecycle)
- **`display_task_result()`**: `self.loading = false` moved *inside* the `DashPayProfileSearchResults` match arm only
- **`display_task_error()`**: New override clears `loading` on errors so the spinner stops appropriately

## Validation

- `cargo check` — clean
- `cargo clippy -- -D warnings` — clean
- Traced the full `AppState` dispatch path (`app.rs:1100-1180`) to confirm `display_task_result` is called for `Message` variants from other screens' tasks, confirming the race condition

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for DashPay profile search operations with improved feedback when searches encounter issues or fail
  * Refined loading state management to accurately track and display operation progress, providing users with consistent status indicators throughout the profile search process


thepastaclaw:
@coderabbitai review

PR #750, opened 2026-03-13
fix(ui): improve auth key error message and suppress on startup auto-select

thepastaclaw:
## Issue
Closes #743 — On startup, DET shows a confusing "Identity doesn't have an authentication key for signing document transitions" error banner with no indication of which identity is affected.

## Changes
1. **Better error message:** Include the identity alias/DPNS name and ID in the error, so users with multiple identities can identify which one lacks a signing key.

   Before: `Identity doesn't have an authentication key for signing document transitions`
   After: `Identity 'my-evonode' (7xK3...) doesn't have an authentication key for signing document transitions`

2. **Suppress banner on auto-select:** The ContactRequests and QRCodeGenerator screens auto-select the first identity during construction. If that identity lacks a document signing key (e.g., evonode identities), the error was shown immediately before the user did anything. Now these screens silently set wallet to `None` on auto-select — the error only appears when the user explicitly selects an identity that lacks the key.

## Testing
- `cargo check` — clean
- `cargo clippy` — clean
- `cargo fmt` — clean

PR #739, opened 2026-03-12
refactor: migrate from Result<T, String> to typed TaskError (#660)

lklimek:
## Summary

- Promote i18n-ready strings guideline to general coding rule in CLAUDE.md
- Add 4 new `TaskError` variants: `LockPoisoned`, `Database`, `CoreRpc`, `P2P`
- Add new `P2PError` domain type for peer-to-peer network errors
- Extend `WalletError` with `KeyDerivation` and `Sighash` variants
- Extend `DashPayError` with `QueryCreation`, `CryptoKeyParsing`, `PrivateKeyResolution` variants; drop `Clone`/`PartialEq` derives (unused)
- Add `ConfigError::SaveError` for file write operations
- Replace `.map_err(|e| e.to_string())` with typed error propagation across `backend_task/`, `ui/`, `model/`, `database/`, `context/`, `spv/`, `config`, `components/`
- Add blanket `From<PoisonError<T>>` for `TaskError::LockPoisoned`
- Route `dashcore_rpc::Error` to `TaskError::CoreRpc` instead of `Generic`
- Clean up redundant `TaskError::from` conversions

~50 callsites migrated to typed errors in this PR. ~854 remaining callsites still compile via `From<String>` bridge for incremental follow-up.

Closes #660

## Test plan

- [x] `cargo build --all-features` compiles cleanly
- [x] `cargo clippy --all-features --all-targets -- -D warnings` passes
- [x] `cargo test --all-features --workspace` — 72 tests + 3 doc-tests pass
- [x] Spot-check error messages against everyday-user persona
- [x] Verify `MessageBanner` shows user-friendly text, not raw error strings

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * More consistent, user-friendly error messages and retry behavior across wallets, transactions, contracts, tokens, P2P and identity flows; improved handling of lock/timeouts and proof failures.

* **UI**
  * Contact success simplified; contact-request error dismissal improved.
  * Broadcast banners now surface "saved-after-proof" outcomes and update refresh/status.

* **New Features**
  * New UI validation error type.
  * Fee details shown for several token operations.

* **Documentation**
  * Added i18n-ready string rules and refined user-facing error guidance.

PR #761, opened 2026-03-15
fix(ui): improve auth key error message and suppress on startup auto-select

### LINKED ISSUE #743: Identity doesn't have an authentication key for signing document transitions
I just updated to the latest v1.0-dev (8598ca8e) and after opening the first time it shows this error. Unfortunately, there is no indication as to which of my 16 identities it refers to. I'm guessing it's the evonode I imported a long time ago, but 🤷‍♂️ 

Section of DET.log surrounding the error:
<details>
2026-03-12T20:58:39.673083Z DEBUG dash_evo_tool::ui::dashpay::contact_requests: Loading contact requests from database for identity 3LhV8ZBGxkZk33VzQEPe89SJVeHxacxdyNJWCYVeKSmf on network testnet
2026-03-12T20:58:39.673110Z DEBUG dash_evo_tool::ui::dashpay::contact_requests: Loaded 0 incoming requests from database
2026-03-12T20:58:39.673126Z DEBUG dash_evo_tool::ui::dashpay::contact_requests: Loaded 0 outgoing requests from database
2026-03-12T20:58:39.673141Z DEBUG dash_evo_tool::ui::dashpay::profile_screen: Loading profile from database for identity 3LhV8ZBGxkZk33VzQEPe89SJVeHxacxdyNJWCYVeKSmf on network testnet
2026-03-12T20:58:39.673155Z DEBUG dash_evo_tool::ui::dashpay::profile_screen: No profile found in database for identity 3LhV8ZBGxkZk33VzQEPe89SJVeHxacxdyNJWCYVeKSmf
2026-03-12T20:58:39.673818Z  INFO dash_evo_tool::ui::tools::grovestark_screen: Refreshed contracts: found 14 user contracts
2026-03-12T20:58:39.673826Z DEBUG dash_evo_tool::ui::dashpay::profile_screen: Loading profile from database for identity 3LhV8ZBGxkZk33VzQEPe89SJVeHxacxdyNJWCYVeKSmf on network testnet
2026-03-12T20:58:39.673892Z DEBUG dash_evo_tool::ui::dashpay::profile_screen: No profile found in database for identity 3LhV8ZBGxkZk33VzQEPe89SJVeHxacxdyNJWCYVeKSmf
2026-03-12T20:58:39.677745Z ERROR dash_evo_tool::ui::components::message_banner: Banner displayed banner="Identity doesn't have an authentication key for signing document transitions"
2026-03-12T20:58:39.679444Z  INFO dash_evo_tool::config: Successfully loaded .env file
2026-03-12T20:58:39.679500Z  INFO dash_evo_tool::config: Mainnet configuration loaded successfully
2026-03-12T20:58:39.679526Z  INFO dash_evo_tool::config: Testnet configuration loaded successfully
2026-03-12T20:58:39.679549Z  INFO dash_evo_tool::config: Devnet configuration loaded successfully
2026-03-12T20:58:39.679568Z  INFO dash_evo_tool::config: Local configuration loaded successfully
</details>

Image

thepastaclaw:
## Issue
Closes #743 — On startup, DET shows a confusing "Identity doesn't have an authentication key for signing document transitions" error banner with no indication of which identity is affected.

Supersedes #750 — recreated on top of v1.0-dev after the error handling refactor (#739) was merged, per lklimek's [request](https://github.com/dashpay/dash-evo-tool/pull/750#issuecomment-2726181571).

## Changes
1. **Better error message:** Include the identity alias/DPNS name and ID in the error, so users with multiple identities can identify which one lacks a signing key.
   - Before: `Identity doesn't have an authentication key for signing document transitions`
   - After: `Identity 'my-evonode' (4ZmR...) doesn't have an authentication key for signing document transitions`

2. **Suppress error banner on startup auto-select:** In `ContactRequests` and `QRCodeGenerator` constructors, the first identity is auto-selected. If it lacks a document signing key (e.g. evonodes), the error banner no longer fires. User-initiated selection still shows errors correctly.

## Testing
- `cargo check`, `cargo clippy`, `cargo fmt` — all clean
- Verified the three original issues from #743 are still present on v1.0-dev post-refactor, and all three are addressed by this branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Suppressed an unnecessary "missing document-signing key" error during automatic wallet/identity auto-selection in dashboard and QR code screens.
  * Improved error text to include identity label and ID when document-signing key retrieval fails, and added a helper to reliably detect that specific error case.

### DIFF
--- a/src/ui/dashpay/contact_requests.rs
+++ b/src/ui/dashpay/contact_requests.rs
@@ -15,7 +15,7 @@ use crate::ui::components::wallet_unlock_popup::{
 };
 use crate::ui::components::{MessageBanner, ResultBannerExt};
 use crate::ui::identities::keys::add_key_screen::AddKeyScreen;
-use crate::ui::identities::{get_selected_wallet, is_missing_document_signing_key_error};
+use crate::ui::identities::{auto_selected_wallet_or_banner, get_selected_wallet};
 use crate::ui::theme::DashColors;
 use crate::ui::{MessageType, Screen, ScreenLike, ScreenType};
 use dash_sdk::dpp::document::DocumentV0Getters;
@@ -115,19 +115,10 @@ impl ContactRequests {
                 .id()
                 .to_string(dash_sdk::dpp::platform_value::string_encoding::Encoding::Base58);
 
-            // Get wallet for the selected identity. Suppress the expected
-            // "missing document-signing key" error on auto-select (some
-            // identities, e.g. evonodes, legitimately lack one); surface
-            // anything else so we don't silently hide real failures.
-            new_self.selected_wallet =
-                match get_selected_wallet(&preferred, Some(&app_context), None) {
-                    Ok(wallet) => wallet,
-                    Err(e) if is_missing_document_signing_key_error(&e) => None,
-                    Err(e) => {
-                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
-                        None
-                    }
-                };
+            new_self.selected_wallet = auto_selected_wallet_or_banner(
+                app_context.egui_ctx(),
+                get_selected_wallet(&preferred, Some(&app_context), None),
+            );
         }
 
         new_self
--- a/src/ui/dashpay/contact_requests.rs
+++ b/src/ui/dashpay/contact_requests.rs
@@ -14,8 +14,8 @@ use crate::ui::components::wallet_unlock_popup::{
     WalletUnlockPopup, WalletUnlockResult, try_open_wallet_no_password, wallet_needs_unlock,
 };
 use crate::ui::components::{MessageBanner, ResultBannerExt};
-use crate::ui::identities::get_selected_wallet;
 use crate::ui::identities::keys::add_key_screen::AddKeyScreen;
+use crate::ui::identities::{get_selected_wallet, is_missing_document_signing_key_error};
 use crate::ui::theme::DashColors;
 use crate::ui::{MessageType, Screen, ScreenLike, ScreenType};
 use dash_sdk::dpp::document::DocumentV0Getters;
@@ -115,10 +115,19 @@ impl ContactRequests {
                 .id()
                 .to_string(dash_sdk::dpp::platform_value::string_encoding::Encoding::Base58);
 
-            // Get wallet for the selected identity (don't show error on auto-select;
-            // some identities like evonodes may lack document signing keys)
+            // Get wallet for the selected identity. Suppress the expected
+            // "missing document-signing key" error on auto-select (some
+            // identities, e.g. evonodes, legitimately lack one); surface
+            // anything else so we don't silently hide real failures.
             new_self.selected_wallet =
-                get_selected_wallet(&preferred, Some(&app_context), None).unwrap_or(None);
+                match get_selected_wallet(&preferred, Some(&app_context), None) {
+                    Ok(wallet) => wallet,
+                    Err(e) if is_missing_document_signing_key_error(&e) => None,
+                    Err(e) => {
+                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
+                        None
+                    }
+                };
         }
 
         new_self
--- a/src/ui/dashpay/contact_requests.rs
+++ b/src/ui/dashpay/contact_requests.rs
@@ -115,10 +115,10 @@ impl ContactRequests {
                 .id()
                 .to_string(dash_sdk::dpp::platform_value::string_encoding::Encoding::Base58);
 
-            // Get wallet for the selected identity
-            new_self.selected_wallet = get_selected_wallet(&preferred, Some(&app_context), None)
-                .or_show_error(app_context.egui_ctx())
-                .unwrap_or(None);
+            // Get wallet for the selected identity (don't show error on auto-select;
+            // some identities like evonodes may lack document signing keys)
+            new_self.selected_wallet =
+                get_selected_wallet(&preferred, Some(&app_context), None).unwrap_or(None);
         }
 
         new_self
--- a/src/ui/dashpay/qr_code_generator.rs
+++ b/src/ui/dashpay/qr_code_generator.rs
@@ -16,7 +16,7 @@ use crate::ui::components::wallet_unlock_popup::{
 use crate::ui::components::{MessageBanner, ResultBannerExt};
 use crate::ui::dashpay::dashpay_screen::DashPaySubscreen;
 use crate::ui::identities::funding_common::generate_qr_code_image;
-use crate::ui::identities::get_selected_wallet;
+use crate::ui::identities::{get_selected_wallet, is_missing_document_signing_key_error};
 use crate::ui::theme::DashColors;
 use crate::ui::{MessageType, RootScreenType, ScreenLike};
 use eframe::epaint::TextureHandle;
@@ -82,10 +82,19 @@ impl QRCodeGeneratorScreen {
             new_self.selected_identity = Some(preferred.clone());
             new_self.selected_identity_string = preferred.identity.id().to_string(Encoding::Base58);
 
-            // Get wallet for the selected identity (don't show error on auto-select;
-            // some identities like evonodes may lack document signing keys)
+            // Get wallet for the selected identity. Suppress the expected
+            // "missing document-signing key" error on auto-select (some
+            // identities, e.g. evonodes, legitimately lack one); surface
+            // anything else so we don't silently hide real failures.
             new_self.selected_wallet =
-                get_selected_wallet(&preferred, Some(&app_context), None).unwrap_or(None);
+                match get_selected_wallet(&preferred, Some(&app_context), None) {
+                    Ok(wallet) => wallet,
+                    Err(e) if is_missing_document_signing_key_error(&e) => None,
+                    Err(e) => {
+                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
+                        None
+                    }
+                };
         }
 
         new_self
--- a/src/ui/dashpay/qr_code_generator.rs
+++ b/src/ui/dashpay/qr_code_generator.rs
@@ -82,10 +82,10 @@ impl QRCodeGeneratorScreen {
             new_self.selected_identity = Some(preferred.clone());
             new_self.selected_identity_string = preferred.identity.id().to_string(Encoding::Base58);
 
-            // Get wallet for the selected identity
-            new_self.selected_wallet = get_selected_wallet(&preferred, Some(&app_context), None)
-                .or_show_error(app_context.egui_ctx())
-                .unwrap_or(None);
+            // Get wallet for the selected identity (don't show error on auto-select;
+            // some identities like evonodes may lack document signing keys)
+            new_self.selected_wallet =
+                get_selected_wallet(&preferred, Some(&app_context), None).unwrap_or(None);
         }
 
         new_self
--- a/src/ui/dashpay/qr_code_generator.rs
+++ b/src/ui/dashpay/qr_code_generator.rs
@@ -16,7 +16,7 @@ use crate::ui::components::wallet_unlock_popup::{
 use crate::ui::components::{MessageBanner, ResultBannerExt};
 use crate::ui::dashpay::dashpay_screen::DashPaySubscreen;
 use crate::ui::identities::funding_common::generate_qr_code_image;
-use crate::ui::identities::{get_selected_wallet, is_missing_document_signing_key_error};
+use crate::ui::identities::{auto_selected_wallet_or_banner, get_selected_wallet};
 use crate::ui::theme::DashColors;
 use crate::ui::{MessageType, RootScreenType, ScreenLike};
 use eframe::epaint::TextureHandle;
@@ -82,19 +82,10 @@ impl QRCodeGeneratorScreen {
             new_self.selected_identity = Some(preferred.clone());
             new_self.selected_identity_string = preferred.identity.id().to_string(Encoding::Base58);
 
-            // Get wallet for the selected identity. Suppress the expected
-            // "missing document-signing key" error on auto-select (some
-            // identities, e.g. evonodes, legitimately lack one); surface
-            // anything else so we don't silently hide real failures.
-            new_self.selected_wallet =
-                match get_selected_wallet(&preferred, Some(&app_context), None) {
-                    Ok(wallet) => wallet,
-                    Err(e) if is_missing_document_signing_key_error(&e) => None,
-                    Err(e) => {
-                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
-                        None
-                    }
-                };
+            new_self.selected_wallet = auto_selected_wallet_or_banner(
+                app_context.egui_ctx(),
+                get_selected_wallet(&preferred, Some(&app_context), None),
+            );
         }
 
         new_self
--- a/src/ui/identities/mod.rs
+++ b/src/ui/identities/mod.rs
@@ -51,16 +51,16 @@ impl fmt::Display for SelectedWalletError {
                 if let Some(identity_label) = identity_label {
                     write!(
                         f,
-                        "Identity {identity_label} ({identity_id}) cannot sign DashPay actions \
-                         because it has no signing key. Add a signing key with high or critical \
-                         security level to this identity, or choose a different identity."
+                        "Identity {identity_label} ({identity_id}) cannot sign this action \
+                         because it has no high-security signing key. Add a high-security \
+                         signing key to this identity, or choose a different identity."
                     )
                 } else {
                     write!(
                         f,
-                        "Identity {identity_id} cannot sign DashPay actions because it has no \
-                         signing key. Add a signing key with high or critical security level to \
-                         this identity, or choose a different identity."
+                        "Identity {identity_id} cannot sign this action because it has no \
+                         high-security signing key. Add a high-security signing key to this \
+                         identity, or choose a different identity."
                     )
                 }
             }
@@ -292,9 +292,9 @@ mod tests {
 
         assert_eq!(
             error.to_string(),
-            "Identity Test Identity (TestIdentityId) cannot sign DashPay actions because it has \
-             no signing key. Add a signing key with high or critical security level to this \
-             identity, or choose a different identity."
+            "Identity Test Identity (TestIdentityId) cannot sign this action because it has no \
+             high-security signing key. Add a high-security signing key to this identity, or \
+             choose a different identity."
         );
     }
 
@@ -304,9 +304,9 @@ mod tests {
 
         assert_eq!(
             error.to_string(),
-            "Identity TestIdentityId cannot sign DashPay actions because it has no signing key. \
-             Add a signing key with high or critical security level to this identity, or choose \
-             a different identity."
+            "Identity TestIdentityId cannot sign this action because it has no high-security \
+             signing key. Add a high-security signing key to this identity, or choose a \
+             different identity."
         );
     }
 
--- a/src/ui/identities/mod.rs
+++ b/src/ui/identities/mod.rs
@@ -1,12 +1,19 @@
+use std::error::Error as StdError;
+use std::fmt;
 use std::sync::{Arc, RwLock};
 
 use dash_sdk::{
-    dpp::data_contract::accessors::v0::DataContractV0Getters, platform::IdentityPublicKey,
+    dpp::{
+        data_contract::accessors::v0::DataContractV0Getters,
+        data_contract::errors::DataContractError,
+    },
+    platform::IdentityPublicKey,
 };
 
 use crate::{
     context::AppContext,
     model::{qualified_identity::QualifiedIdentity, wallet::Wallet},
+    ui::{MessageType, components::MessageBanner},
 };
 
 pub mod add_existing_identity_screen;
@@ -19,18 +26,79 @@ pub mod top_up_identity_screen;
 pub mod transfer_screen;
 pub mod withdraw_screen;
 
-/// Substring used to identify the "missing document-signing key" error returned
-/// by [`get_selected_wallet`] when an identity has no key suitable for signing
-/// document state transitions. Callsites that auto-select an identity (e.g.
-/// DashPay startup) use this to suppress the expected case while still
-/// surfacing unrelated errors.
-const MISSING_DOCUMENT_SIGNING_KEY_MARKER: &str =
-    "doesn't have an authentication key for signing document transitions";
+#[derive(Debug)]
+pub enum SelectedWalletError {
+    DpnsPreorderDocumentTypeNotFound {
+        source: DataContractError,
+    },
+    MissingDocumentSigningKey {
+        identity_label: Option<String>,
+        identity_id: String,
+    },
+    NoKeyProvided,
+}
 
-/// Returns `true` if the given error string is the expected
-/// "identity has no document-signing key" error from [`get_selected_wallet`].
-pub fn is_missing_document_signing_key_error(error: &str) -> bool {
-    error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)
+impl fmt::Display for SelectedWalletError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::DpnsPreorderDocumentTypeNotFound { .. } => {
+                write!(f, "DPNS preorder document type not found")
+            }
+            Self::MissingDocumentSigningKey {
+                identity_label,
+                identity_id,
+            } => {
+                if let Some(identity_label) = identity_label {
+                    write!(
+                        f,
+                        "Identity {identity_label} ({identity_id}) cannot sign DashPay actions \
+                         because it has no signing key. Add a signing key with high or critical \
+                         security level to this identity, or choose a different identity."
+                    )
+                } else {
+                    write!(
+                        f,
+                        "Identity {identity_id} cannot sign DashPay actions because it has no \
+                         signing key. Add a signing key with high or critical security level to \
+                         this identity, or choose a different identity."
+                    )
+                }
+            }
+            Self::NoKeyProvided => write!(f, "No key provided when getting selected wallet"),
+        }
+    }
+}
+
+impl StdError for SelectedWalletError {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        match self {
+            Self::DpnsPreorderDocumentTypeNotFound { source } => Some(source),
+            Self::MissingDocumentSigningKey { .. } | Self::NoKeyProvided => None,
+        }
+    }
+}
+
+/// Returns `true` if the given error is the expected "identity has no
+/// document-signing key" error from [`get_selected_wallet`].
+pub fn is_missing_document_signing_key_error(error: &SelectedWalletError) -> bool {
+    matches!(error, SelectedWalletError::MissingDocumentSigningKey { .. })
+}
+
+/// Applies the auto-selection policy for screen construction:
+/// suppress the expected missing-document-signing-key error for
+/// auto-selected identities, but surface every other failure.
+pub fn auto_selected_wallet_or_banner(
+    ctx: &egui::Context,
+    result: Result<Option<Arc<RwLock<Wallet>>>, SelectedWalletError>,
+) -> Option<Arc<RwLock<Wallet>>> {
+    match result {
+        Ok(wallet) => wallet,
+        Err(error) if is_missing_document_signing_key_error(&error) => None,
+        Err(error) => {
+            MessageBanner::set_global(ctx, &error, MessageType::Error);
+            None
+        }
+    }
 }
 
 /// Retrieves the appropriate wallet (if any) associated with the given identity.
@@ -57,8 +125,8 @@ pub fn is_missing_document_signing_key_error(error: &str) -> bool {
 /// # Returns
 ///
 /// Returns `Ok(Some(Arc<RwLock<Wallet>>))` if a matching wallet is found,
-/// `Ok(None)` if no wallet is associated with the key, or `Err(String)` if
-/// an error is encountered.
+/// `Ok(None)` if no wallet is associated with the key, or
+/// `Err(SelectedWalletError)` if an error is encountered.
 ///
 /// # Errors
 ///
@@ -69,7 +137,7 @@ pub fn get_selected_wallet(
     qualified_identity: &QualifiedIdentity,
     app_context: Option<&AppContext>,
     selected_key: Option<&IdentityPublicKey>,
-) -> Result<Option<Arc<RwLock<Wallet>>>, String> {
+) -> Result<Option<Arc<RwLock<Wallet>>>, SelectedWalletError> {
     // If `app_context` is provided, use the DPNS-based approach.
     let public_key = if let Some(context) = app_context {
         let dpns_contract = &context.dpns_contract;
@@ -77,7 +145,7 @@ pub fn get_selected_wallet(
         // Attempt to fetch the `preorder` document type from the DPNS contract.
         let preorder_document_type = dpns_contract
             .document_type_for_name("preorder")
-            .map_err(|e| format!("DPNS preorder document type not found: {}", e))?;
+            .map_err(|source| SelectedWalletError::DpnsPreorderDocumentTypeNotFound { source })?;
 
         // Attempt to retrieve the public key from the identity.
         qualified_identity
@@ -85,26 +153,20 @@ pub fn get_selected_wallet(
             .ok_or_else(|| {
                 use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
                 use dash_sdk::dpp::platform_value::string_encoding::Encoding;
-                let identity_label = qualified_identity
-                    .alias
-                    .as_deref()
-                    .or_else(|| {
-                        qualified_identity
-                            .dpns_names
-                            .first()
-                            .map(|n| n.name.as_str())
-                    })
-                    .unwrap_or("");
-                format!(
-                    "Identity {} ({}) {}",
-                    identity_label,
-                    qualified_identity.identity.id().to_string(Encoding::Base58),
-                    MISSING_DOCUMENT_SIGNING_KEY_MARKER,
-                )
+                let identity_label = qualified_identity.alias.as_deref().or_else(|| {
+                    qualified_identity
+                        .dpns_names
+                        .first()
+                        .map(|n| n.name.as_str())
+                });
+                SelectedWalletError::MissingDocumentSigningKey {
+                    identity_label: identity_label.map(str::to_owned),
+                    identity_id: qualified_identity.identity.id().to_string(Encoding::Base58),
+                }
             })?
     } else {
         // Fallback: directly use the provided selected key.
-        selected_key.ok_or_else(|| "No key provided when getting selected wallet".to_string())?
+        selected_key.ok_or(SelectedWalletError::NoKeyProvided)?
     };
 
     // Once we have the public key (either from DPNS or directly), ask which
@@ -214,6 +276,64 @@ mod tests {
         assert!(
             selected.is_some(),
             "the wallet deriving this key must be found whichever placement names it"
+        );
+    }
+
+    fn missing_document_signing_key_error(identity_label: Option<&str>) -> SelectedWalletError {
+        SelectedWalletError::MissingDocumentSigningKey {
+            identity_label: identity_label.map(str::to_owned),
+            identity_id: "TestIdentityId".to_string(),
+        }
+    }
+
+    #[test]
+    fn missing_document_signing_key_display_includes_label_when_present() {
+        let error = missing_document_signing_key_error(Some("Test Identity"));
+
+        assert_eq!(
+            error.to_string(),
+            "Identity Test Identity (TestIdentityId) cannot sign DashPay actions because it has \
+             no signing key. Add a signing key with high or critical security level to this \
+             identity, or choose a different identity."
+        );
+    }
+
+    #[test]
+    fn missing_document_signing_key_display_uses_identity_id_without_label() {
+        let error = missing_document_signing_key_error(None);
+
+        assert_eq!(
+            error.to_string(),
+            "Identity TestIdentityId cannot sign DashPay actions because it has no signing key. \
+             Add a signing key with high or critical security level to this identity, or choose \
+             a different identity."
+        );
+    }
+
+    #[test]
+    fn auto_selection_policy_suppresses_missing_document_signing_key() {
+        let ctx = egui::Context::default();
+
+        let wallet =
+            auto_selected_wallet_or_banner(&ctx, Err(missing_document_signing_key_error(None)));
+
+        assert!(wallet.is_none());
+        assert!(
+            !MessageBanner::has_global(&ctx),
+            "automatic selection must suppress the typed missing-document-signing-key banner"
+        );
+    }
+
+    #[test]
+    fn auto_selection_policy_surfaces_unrelated_selected_wallet_error() {
+        let ctx = egui::Context::default();
+
+        let wallet = auto_selected_wallet_or_banner(&ctx, Err(SelectedWalletError::NoKeyProvided));
+
+        assert!(wallet.is_none());
+        assert!(
+            MessageBanner::has_global(&ctx),
+            "automatic selection must only suppress MissingDocumentSigningKey"
         );
     }
 }
--- a/src/ui/identities/mod.rs
+++ b/src/ui/identities/mod.rs
@@ -69,8 +69,17 @@ pub fn get_selected_wallet(
         qualified_identity
             .document_signing_key(&preorder_document_type)
             .ok_or_else(|| {
-                "Identity doesn't have an authentication key for signing document transitions"
-                    .to_string()
+                use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
+                let identity_label = qualified_identity
+                    .alias
+                    .as_deref()
+                    .or_else(|| qualified_identity.dpns_names.first().map(|n| n.name.as_str()))
+                    .unwrap_or("unknown");
+                format!(
+                    "Identity '{}' ({}) doesn't have an authentication key for signing document transitions",
+                    identity_label,
+                    qualified_identity.identity.id()
+                )
             })?
     } else {
         // Fallback: directly use the provided selected key.
--- a/src/ui/identities/mod.rs
+++ b/src/ui/identities/mod.rs
@@ -19,6 +19,20 @@ pub mod top_up_identity_screen;
 pub mod transfer_screen;
 pub mod withdraw_screen;
 
+/// Substring used to identify the "missing document-signing key" error returned
+/// by [`get_selected_wallet`] when an identity has no key suitable for signing
+/// document state transitions. Callsites that auto-select an identity (e.g.
+/// DashPay startup) use this to suppress the expected case while still
+/// surfacing unrelated errors.
+const MISSING_DOCUMENT_SIGNING_KEY_MARKER: &str =
+    "doesn't have an authentication key for signing document transitions";
+
+/// Returns `true` if the given error string is the expected
+/// "identity has no document-signing key" error from [`get_selected_wallet`].
+pub fn is_missing_document_signing_key_error(error: &str) -> bool {
+    error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)
+}
+
 /// Retrieves the appropriate wallet (if any) associated with the given identity.
 ///
 /// # Description
@@ -70,15 +84,22 @@ pub fn get_selected_wallet(
             .document_signing_key(&preorder_document_type)
             .ok_or_else(|| {
                 use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
+                use dash_sdk::dpp::platform_value::string_encoding::Encoding;
                 let identity_label = qualified_identity
                     .alias
                     .as_deref()
-                    .or_else(|| qualified_identity.dpns_names.first().map(|n| n.name.as_str()))
-                    .unwrap_or("unknown");
+                    .or_else(|| {
+                        qualified_identity
+                            .dpns_names
+                            .first()
+                            .map(|n| n.name.as_str())
+                    })
+                    .unwrap_or("");
                 format!(
-                    "Identity '{}' ({}) doesn't have an authentication key for signing document transitions",
+                    "Identity {} ({}) {}",
                     identity_label,
-                    qualified_identity.identity.id()
+                    qualified_identity.identity.id().to_string(Encoding::Base58),
+                    MISSING_DOCUMENT_SIGNING_KEY_MARKER,
                 )
             })?
     } else {
--- a/src/ui/dashpay/profile_screen.rs
+++ b/src/ui/dashpay/profile_screen.rs
@@ -17,7 +17,7 @@ use crate::ui::components::wallet_unlock_popup::{
 };
 use crate::ui::components::{MessageBanner, ResultBannerExt};
 use crate::ui::helpers::{ModalOpeningGuard, clicked_outside_window_after_open};
-use crate::ui::identities::get_selected_wallet;
+use crate::ui::identities::{auto_selected_wallet_or_banner, get_selected_wallet};
 use crate::ui::state::AvatarCache;
 use crate::ui::theme::{ComponentStyles, DashColors, ResponseExt};
 use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
@@ -134,9 +134,10 @@ impl ProfileScreen {
                 new_self.selected_identity_string
             );
 
-            new_self.selected_wallet = get_selected_wallet(&preferred, Some(&app_context), None)
-                .or_show_error(app_context.egui_ctx())
-                .unwrap_or(None);
+            new_self.selected_wallet = auto_selected_wallet_or_banner(
+                app_context.egui_ctx(),
+                get_selected_wallet(&preferred, Some(&app_context), None),
+            );
         }
 
         new_self
--- a/src/ui/tokens/set_token_price_screen.rs
+++ b/src/ui/tokens/set_token_price_screen.rs
@@ -264,7 +264,7 @@ impl SetTokenPriceScreen {
         let selected_wallet =
             get_selected_wallet(&identity_token_info.identity, None, possible_key).unwrap_or_else(
                 |e| {
-                    set_error_banner(&e);
+                    set_error_banner(&e.to_string());
                     None
                 },
             );
--- a/src/ui/tokens/token_action_screen.rs
+++ b/src/ui/tokens/token_action_screen.rs
@@ -192,7 +192,7 @@ impl<A: TokenAction> TokenActionScreen<A> {
         let selected_wallet =
             get_selected_wallet(&identity_token_info.identity, None, possible_key.as_ref())
                 .unwrap_or_else(|e| {
-                    super::set_error_banner(app_context, &e);
+                    super::set_error_banner(app_context, &e.to_string());
                     None
                 });
 
--- a/tests/kittest/dashpay_screen.rs
+++ b/tests/kittest/dashpay_screen.rs
@@ -13,6 +13,7 @@ use dash_evo_tool::context::AppContext;
 use dash_evo_tool::model::qualified_identity::encrypted_key_storage::KeyStorage;
 use dash_evo_tool::model::qualified_identity::{IdentityStatus, IdentityType, QualifiedIdentity};
 use dash_evo_tool::ui::ScreenLike;
+use dash_evo_tool::ui::components::MessageBanner;
 use dash_evo_tool::ui::dashpay::add_contact_screen::AddContactScreen;
 use dash_evo_tool::ui::dashpay::contact_requests::ContactRequests;
 use dash_evo_tool::ui::dashpay::contacts_list::ContactsList;
@@ -27,7 +28,10 @@ use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
 use dash_sdk::dpp::version::PlatformVersion;
 use dash_sdk::platform::Identifier;
 use egui_kittest::Harness;
+use egui_kittest::kittest::Queryable;
+use std::cell::RefCell;
 use std::collections::BTreeMap;
+use std::rc::Rc;
 use std::sync::Arc;
 
 /// On the Contacts subscreen, a `MissingEncryptionKey` error must route to
@@ -166,6 +170,34 @@ fn contact_requests_defaults_to_app_scoped_identity() {
     });
 }
 
+/// `ContactRequests::new()` must suppress only the typed startup
+/// `MissingDocumentSigningKey` selection banner for wallet-less identities.
+#[test]
+fn contact_requests_startup_suppresses_missing_document_signing_key_banner() {
+    with_isolated_data_dir(|| {
+        let rt = tokio::runtime::Runtime::new().expect("tokio runtime");
+        let _guard = rt.enter();
+
+        let (_h, ctx) = build_ctx();
+
+        let selected = seed_dp_identity(&ctx, 0xD5, "CR Startup Missing Key");
+        ctx.set_selected_identity(Some(selected));
+        MessageBanner::clear_all_global(ctx.egui_ctx());
+
+        let screen = ContactRequests::new(ctx.clone());
+
+        assert_eq!(
+            screen.selected_identity.as_ref().map(|qi| qi.identity.id()),
+            Some(selected),
+            "ContactRequests must still select the app-scoped identity at startup"
+        );
+        assert!(
+            !MessageBanner::has_global(ctx.egui_ctx()),
+            "ContactRequests startup must suppress the typed missing-document-signing-key banner"
+        );
+    });
+}
+
 /// `PaymentHistory::new()` must default to the app-scoped identity.
 #[test]
 fn payment_history_defaults_to_app_scoped_identity() {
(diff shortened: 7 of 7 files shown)


src/ui/identities/mod.rs
```diff
@@ -78,8 +78,17 @@ pub fn get_selected_wallet(
...
-                "Identity doesn't have an authentication key for signing document transitions"
-                    .to_string()
+                use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
+                let identity_label = qualified_identity
+                    .alias
+                    .as_deref()
+                    .or_else(|| qualified_identity.dpns_names.first().map(|n| n.name.as_str()))
+                    .unwrap_or("unknown");
+                format!(
+                    "Identity '{}' ({}) doesn't have an authentication key for signing document transitions",
```
lklimek: claw remove quotes, and when no name found, put empty string. We will just have a few spaces when identity label is not availble.

thepastaclaw: Done — removed quotes and switched the fallback to empty string. 83f602cc

thepastaclaw: Done in 83f602cc — removed quotes and switched to empty string when no name is found.

thepastaclaw:
@coderabbitai review

src/ui/identities/mod.rs
```diff
@@ -78,8 +78,17 @@ pub fn get_selected_wallet(
...
+                    identity_label,
+                    qualified_identity.identity.id()
+                )
```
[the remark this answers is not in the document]
thepastaclaw: The Display impl for Identifier outputs Base58 — same encoding used elsewhere in the UI. No mismatch here.

src/ui/dashpay/qr_code_generator.rs
```diff
@@ -77,11 +77,10 @@ impl QRCodeGeneratorScreen {
...
                 identities[0].identity.id().to_string(Encoding::Base58);
 
-            // Get wallet for the selected identity
+            // Get wallet for the selected identity (don't show error on auto-select;
+            // some identities like evonodes may lack document signing keys)
             new_self.selected_wallet =
-                get_selected_wallet(&identities[0], Some(&app_context), None)
-                    .or_show_error(app_context.egui_ctx())
-                    .unwrap_or(None);
+                get_selected_wallet(&identities[0], Some(&app_context), None).unwrap_or(None);
```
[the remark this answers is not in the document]
thepastaclaw: Fair point in theory, but in practice the only other error path here is DPNS preorder document type not found — which means the DPNS contract is broken/missing, and the app would fail in many other places too. Selectively filtering error types here adds complexity for a scenario that's already catastrophic. The current approach keeps the auto-select path clean.

src/ui/dashpay/contact_requests.rs
```diff
@@ -103,11 +103,10 @@ impl ContactRequests {
...
                 .to_string(dash_sdk::dpp::platform_value::string_encoding::Encoding::Base58);
 
-            // Get wallet for the selected identity
+            // Get wallet for the selected identity (don't show error on auto-select;
+            // some identities like evonodes may lack document signing keys)
             new_self.selected_wallet =
-                get_selected_wallet(&identities[0], Some(&app_context), None)
-                    .or_show_error(app_context.egui_ctx())
-                    .unwrap_or(None);
+                get_selected_wallet(&identities[0], Some(&app_context), None).unwrap_or(None);
```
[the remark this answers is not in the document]
thepastaclaw: Same rationale as the QRCodeGenerator comment above — the only other error is a missing DPNS contract, which is a global failure, not something worth special-casing here.

thepastaclaw:
🕓 Review not started yet because the new head is waiting for the 30-minute push debounce.

- [ ] **Request normal review** — click when the PR is ready for review.
- [ ] **Request priority review** — click to move this review to the front of the queue.

_Commit d22ee86c. Normal review starts when eligible; priority review starts as soon as a slot is available._

src/ui/identities/mod.rs
thepastaclaw: Done in a6bd8f37 — preserved the no-quotes formatting and empty-string fallback after the amended update.

src/ui/identities/mod.rs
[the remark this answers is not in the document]
thepastaclaw: Fixed in a6bd8f37 — the identity id is now formatted explicitly with Encoding::Base58.

src/ui/dashpay/qr_code_generator.rs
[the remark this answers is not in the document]
thepastaclaw: Fixed in a6bd8f37 — startup auto-select now suppresses only the expected missing document-signing-key error and surfaces other get_selected_wallet errors with a banner.

src/ui/dashpay/contact_requests.rs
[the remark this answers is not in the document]
thepastaclaw: Fixed in a6bd8f37 — same selective handling as QRCodeGenerator: only the missing signing-key case is suppressed, other errors are shown.

src/ui/identities/mod.rs
```diff
@@ -28,6 +28,20 @@ pub mod top_up_identity_screen;
...
+/// DashPay startup) use this to suppress the expected case while still
+/// surfacing unrelated errors.
+const MISSING_DOCUMENT_SIGNING_KEY_MARKER: &str =
+    "doesn't have an authentication key for signing document transitions";
+
+/// Returns `true` if the given error string is the expected
+/// "identity has no document-signing key" error from [`get_selected_wallet`].
+pub fn is_missing_document_signing_key_error(error: &str) -> bool {
+    error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)
+}
```
thepastaclaw: **🟡 Suggestion: Substring-matching a `String` error is brittle — model the case as a typed variant**

`is_missing_document_signing_key_error` decides control flow by `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)` against the formatted error string returned by `get_selected_wallet` (which still returns `Result<_, String>`). This is exactly the stringly-typed pattern the recent #739 `TaskError` migration moved away from, and it has three concrete failure modes:

1. Any future tweak to the wording (capitalisation, punctuation, i18n) silently breaks the suppression — and there is no compile-time signal.
2. The marker is a sentence fragment embedded mid-message (`format!("Identity {} ({}) {}", ...)`); callers are matching on a substring of a constructed string, which is easy to invalidate by accident.
3. Other `get_selected_wallet` callers (e.g. `register_contract_screen.rs:69`, `update_contract_screen.rs:75`) use `.unwrap_or(None)` and silently swallow *all* errors — they could benefit from the same structural distinction to drop only this case while still surfacing real failures.

A dedicated error type (e.g. `enum WalletLookupError { MissingDocumentSigningKey { identity_id, label }, DpnsContractMissing(...), NoKeyProvided }` with `Display` for the user-facing rendering) would let call sites match `Err(WalletLookupError::MissingDocumentSigningKey { .. }) => None` instead of grepping a sentence, and would also keep `Display` text out of a const so i18n becomes possible later.

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] lines 31-43: Substring-matching a `String` error is brittle — model the case as a typed variant
  `is_missing_document_signing_key_error` decides control flow by `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)` against the formatted error string returned by `get_selected_wallet` (which still returns `Result<_, String>`). This is exactly the stringly-typed pattern the recent #739 `TaskError` migration moved away from, and it has three concrete failure modes:

1. Any future tweak to the wording (capitalisation, punctuation, i18n) silently breaks the suppression — and there is no compile-time signal.
2. The marker is a sentence fragment embedded mid-message (`format!("Identity {} ({}) {}", ...)`); callers are matching on a substring of a constructed string, which is easy to invalidate by accident.
3. Other `get_selected_wallet` callers (e.g. `register_contract_screen.rs:69`, `update_contract_screen.rs:75`) use `.unwrap_or(None)` and silently swallow *all* errors — they could benefit from the same structural distinction to drop only this case while still surfacing real failures.

A dedicated error type (e.g. `enum WalletLookupError { MissingDocumentSigningKey { identity_id, label }, DpnsContractMissing(...), NoKeyProvided }` with `Display` for the user-facing rendering) would let call sites match `Err(WalletLookupError::MissingDocumentSigningKey { .. }) => None` instead of grepping a sentence, and would also keep `Display` text out of a const so i18n becomes possible later.
```
</details>

src/ui/identities/mod.rs
```diff
@@ -78,8 +92,24 @@ pub fn get_selected_wallet(
...
+                            .map(|n| n.name.as_str())
+                    })
+                    .unwrap_or("");
+                format!(
+                    "Identity {} ({}) {}",
+                    identity_label,
+                    qualified_identity.identity.id().to_string(Encoding::Base58),
+                    MISSING_DOCUMENT_SIGNING_KEY_MARKER,
+                )
             })?
```
thepastaclaw: **💬 Nitpick: Empty identity label produces a double space in the rendered message**

When `alias` is `None` and `dpns_names` is empty, `identity_label` becomes `""` (line 106 `unwrap_or("")`), so the formatted message reads `"Identity  (4ZmR…) doesn't have an authentication key for signing document transitions"` — note the double space and dangling "Identity ". The recent `a6bd8f37` commit intentionally chose `""` over a placeholder like `"Unknown"`, so the format must branch on whether a label exists.

<details>

```suggestion
                let identity_label = qualified_identity.alias.as_deref().or_else(|| {
                    qualified_identity
                        .dpns_names
                        .first()
                        .map(|n| n.name.as_str())
                });
                let id_b58 = qualified_identity.identity.id().to_string(Encoding::Base58);
                match identity_label {
                    Some(label) => format!(
                        "Identity {} ({}) {}",
                        label, id_b58, MISSING_DOCUMENT_SIGNING_KEY_MARKER,
                    ),
                    None => format!(
                        "Identity ({}) {}",
                        id_b58, MISSING_DOCUMENT_SIGNING_KEY_MARKER,
                    ),
                }
```

</details>

<sub>source: ['claude']</sub>

src/ui/identities/mod.rs
```diff
@@ -28,6 +28,20 @@ pub mod top_up_identity_screen;
 pub mod transfer_screen;
 pub mod withdraw_screen;
 
+/// Substring used to identify the "missing document-signing key" error returned
+/// by [`get_selected_wallet`] when an identity has no key suitable for signing
+/// document state transitions. Callsites that auto-select an identity (e.g.
+/// DashPay startup) use this to suppress the expected case while still
+/// surfacing unrelated errors.
+const MISSING_DOCUMENT_SIGNING_KEY_MARKER: &str =
+    "doesn't have an authentication key for signing document transitions";
```
thepastaclaw: **🟡 Suggestion: Error message uses jargon and lacks an actionable next step**

Per the Dash Evo Tool review skill user-facing message rules: user-facing messages must avoid jargon and must include a concrete self-service action ("what happened + what to do"). The marker text — `"doesn't have an authentication key for signing document transitions"` — is the original pre-PR wording and fails on both counts: it uses platform terminology aimed at developers ("authentication key", "document transitions") and tells the user nothing about what to do (e.g., "Add a signing key in this identity's Keys screen"). Since this PR already reformats the message, it's a reasonable touchpoint to bring the wording into compliance — though doing so requires updating the marker constant, which underlines the brittleness of the substring-matching design.

<sub>source: ['claude']</sub>

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] lines 36-37: Error message uses jargon and lacks an actionable next step
  Per the Dash Evo Tool review skill user-facing message rules: user-facing messages must avoid jargon and must include a concrete self-service action ("what happened + what to do"). The marker text — `"doesn't have an authentication key for signing document transitions"` — is the original pre-PR wording and fails on both counts: it uses platform terminology aimed at developers ("authentication key", "document transitions") and tells the user nothing about what to do (e.g., "Add a signing key in this identity's Keys screen"). Since this PR already reformats the message, it's a reasonable touchpoint to bring the wording into compliance — though doing so requires updating the marker constant, which underlines the brittleness of the substring-matching design.
```
</details>

src/ui/identities/mod.rs
thepastaclaw: **🟡 Suggestion: No test pins the marker/message contract together**

Suppression in `ContactRequests::new` / `QRCodeGeneratorScreen::new` (and the proposed `ProfileScreen::new` fix) depends on the invariant that `get_selected_wallet` always embeds `MISSING_DOCUMENT_SIGNING_KEY_MARKER` verbatim in its returned error string. No test asserts this round-trip — if anyone reformats the message, the helper silently starts returning `false` and the regression returns. Add a small `#[test]` that constructs an identity without a document-signing key, calls `get_selected_wallet`, and asserts `is_missing_document_signing_key_error(&err)` returns true. This is a low-cost guard until the call is migrated to a typed error.

<sub>source: ['claude']</sub>

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] lines 41-43: No test pins the marker/message contract together
  Suppression in `ContactRequests::new` / `QRCodeGeneratorScreen::new` (and the proposed `ProfileScreen::new` fix) depends on the invariant that `get_selected_wallet` always embeds `MISSING_DOCUMENT_SIGNING_KEY_MARKER` verbatim in its returned error string. No test asserts this round-trip — if anyone reformats the message, the helper silently starts returning `false` and the regression returns. Add a small `#[test]` that constructs an identity without a document-signing key, calls `get_selected_wallet`, and asserts `is_missing_document_signing_key_error(&err)` returns true. This is a low-cost guard until the call is migrated to a typed error.
```
</details>

thepastaclaw (commented):
## Code Review

The fix is partial and built on a fragile substring marker. Codex correctly identifies that ProfileScreen::new — which is eagerly constructed at startup as part of every DashPayScreen — still calls `.or_show_error()` and will reproduce the exact bug this PR aims to fix. The substring-based detection is also at odds with the recent #739 typed-error migration, and the rendered message has both an empty-label formatting bug and a jargon problem.

*Reviewed commit: a6bd8f37*

🔴 1 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

### 1 additional finding

<details>
<summary>🔴 <b>blocking:</b> Startup banner still fires from `ProfileScreen::new` — fix is incomplete</summary>

`src/ui/dashpay/profile_screen.rs (lines 206-210)`

This PR suppresses the auto-select banner in `ContactRequests::new` and `QRCodeGeneratorScreen::new`, but `ProfileScreen::new` still calls `get_selected_wallet(...).or_show_error(app_context.egui_ctx()).unwrap_or(None)` at line 207–210. Confirmed in `src/app.rs:288-297`: AppState eagerly constructs `contracts_dashpay_screen`, `dashpay_contacts_screen`, `dashpay_profile_screen`, and `dashpay_payments_screen`, and `DashPayScreen::new` (`src/ui/dashpay/dashpay_screen.rs:33-41`) unconditionally embeds a `ProfileScreen` regardless of which subscreen is requested. So for an evonode-only user, the same "missing document-signing key" banner described in #743 will still appear at app startup — the regression is not actually eliminated.

<details>

```suggestion
            // Get wallet for the selected identity. Suppress the expected
            // "missing document-signing key" error on auto-select (some
            // identities, e.g. evonodes, legitimately lack one); surface
            // anything else so we don't silently hide real failures.
            new_self.selected_wallet =
                match get_selected_wallet(&identities[selected_idx], Some(&app_context), None) {
                    Ok(wallet) => wallet,
                    Err(e) if crate::ui::identities::is_missing_document_signing_key_error(&e) => None,
                    Err(e) => {
                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
                        None
                    }
                };
```

</details>

</details>

<details>

```

In `src/ui/dashpay/profile_screen.rs`:
- [BLOCKING] lines 206-210: Startup banner still fires from `ProfileScreen::new` — fix is incomplete
  This PR suppresses the auto-select banner in `ContactRequests::new` and `QRCodeGeneratorScreen::new`, but `ProfileScreen::new` still calls `get_selected_wallet(...).or_show_error(app_context.egui_ctx()).unwrap_or(None)` at line 207–210. Confirmed in `src/app.rs:288-297`: AppState eagerly constructs `contracts_dashpay_screen`, `dashpay_contacts_screen`, `dashpay_profile_screen`, and `dashpay_payments_screen`, and `DashPayScreen::new` (`src/ui/dashpay/dashpay_screen.rs:33-41`) unconditionally embeds a `ProfileScreen` regardless of which subscreen is requested. So for an evonode-only user, the same "missing document-signing key" banner described in #743 will still appear at app startup — the regression is not actually eliminated.

In `src/ui/identities/mod.rs`:
- [SUGGESTION] lines 31-43: Substring-matching a `String` error is brittle — model the case as a typed variant
  `is_missing_document_signing_key_error` decides control flow by `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)` against the formatted error string returned by `get_selected_wallet` (which still returns `Result<_, String>`). This is exactly the stringly-typed pattern the recent #739 `TaskError` migration moved away from, and it has three concrete failure modes:

1. Any future tweak to the wording (capitalisation, punctuation, i18n) silently breaks the suppression — and there is no compile-time signal.
2. The marker is a sentence fragment embedded mid-message (`format!("Identity {} ({}) {}", ...)`); callers are matching on a substring of a constructed string, which is easy to invalidate by accident.
3. Other `get_selected_wallet` callers (e.g. `register_contract_screen.rs:69`, `update_contract_screen.rs:75`) use `.unwrap_or(None)` and silently swallow *all* errors — they could benefit from the same structural distinction to drop only this case while still surfacing real failures.

A dedicated error type (e.g. `enum WalletLookupError { MissingDocumentSigningKey { identity_id, label }, DpnsContractMissing(...), NoKeyProvided }` with `Display` for the user-facing rendering) would let call sites match `Err(WalletLookupError::MissingDocumentSigningKey { .. }) => None` instead of grepping a sentence, and would also keep `Display` text out of a const so i18n becomes possible later.
- [SUGGESTION] lines 36-37: Error message uses jargon and lacks an actionable next step
  Per the Dash Evo Tool review skill user-facing message rules: user-facing messages must avoid jargon and must include a concrete self-service action ("what happened + what to do"). The marker text — `"doesn't have an authentication key for signing document transitions"` — is the original pre-PR wording and fails on both counts: it uses platform terminology aimed at developers ("authentication key", "document transitions") and tells the user nothing about what to do (e.g., "Add a signing key in this identity's Keys screen"). Since this PR already reformats the message, it's a reasonable touchpoint to bring the wording into compliance — though doing so requires updating the marker constant, which underlines the brittleness of the substring-matching design.
- [SUGGESTION] lines 41-43: No test pins the marker/message contract together
  Suppression in `ContactRequests::new` / `QRCodeGeneratorScreen::new` (and the proposed `ProfileScreen::new` fix) depends on the invariant that `get_selected_wallet` always embeds `MISSING_DOCUMENT_SIGNING_KEY_MARKER` verbatim in its returned error string. No test asserts this round-trip — if anyone reformats the message, the helper silently starts returning `false` and the regression returns. Add a small `#[test]` that constructs an identity without a document-signing key, calls `get_selected_wallet`, and asserts `is_missing_document_signing_key_error(&err)` returns true. This is a low-cost guard until the call is migrated to a typed error.
```
</details>

thepastaclaw (commented):
## Code Review

The PR's intent (suppress the noisy startup banner and add identity context to the missing-signing-key error) is correct, but the implementation has two real defects that reach exactly the case the PR targets: ProfileScreen is built eagerly by DashPayScreen at startup yet was not patched, so the banner still appears, and identities without an alias or DPNS name (the evonode case in the PR description) render as 'Identity  (id) ...' with awkward double whitespace and a dangling word. Architecturally, the suppression mechanism couples control flow to the substring of a translatable user-facing string, which is fragile and conflicts with project i18n rules.

*Reviewed commit: a6bd8f37*

Fresh current-run review completed for this dispatcher item.

🔴 2 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

<details>
<summary>🔴 <b>blocking:</b> ProfileScreen auto-select still surfaces the banner the PR claims to suppress</summary>

`src/ui/dashpay/profile_screen.rs:206-210`

`DashPayScreen::new` (src/ui/dashpay/dashpay_screen.rs:38) eagerly constructs `ProfileScreen` alongside `ContactsList` and `PaymentHistory` at startup, and `AppState` builds multiple `DashPayScreen` instances. `ContactRequests::new` and `QRCodeGeneratorScreen::new` were updated to drop the missing-signing-key error via `is_missing_document_signing_key_error`, but `ProfileScreen::new` still calls `get_selected_wallet(&identities[selected_idx], Some(&app_context), None).or_show_error(app_context.egui_ctx())` on the same auto-select path. For an evonode identity without a document-signing key — the exact scenario in #743 — the global banner still fires at startup, so the user-visible bug is only partially fixed. Apply the same `Err(e) if is_missing_document_signing_key_error(&e) => None` arm here, or push the suppression down into `get_selected_wallet`/`or_show_error` so all auto-select callsites get it consistently.

<details>

```suggestion
            // Get wallet for the selected identity
            new_self.selected_wallet =
                match get_selected_wallet(&identities[selected_idx], Some(&app_context), None) {
                    Ok(wallet) => wallet,
                    Err(e) if crate::ui::identities::is_missing_document_signing_key_error(&e) => {
                        None
                    }
                    Err(e) => {
                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
                        None
                    }
                };
```

</details>

</details>

<details>
<summary>🔴 <b>blocking:</b> Empty identity label produces malformed 'Identity  (id) ...' message</summary>

`src/ui/identities/mod.rs:97-112`

When an identity has neither an `alias` nor any `dpns_names`, `identity_label` falls back to `""` (line 106), and the format string `"Identity {} ({}) {}"` produces `"Identity  (4ZmR…) doesn't have an authentication key for signing document transitions"` — note the double space and the dangling word `Identity`. This is the most common shape for non-user identities (evonodes), which is exactly the example called out in the PR description, so the message the PR is trying to clarify becomes garbled for the primary affected user. Branch on whether a label exists rather than emitting an empty placeholder.

<details>

```suggestion
                let identity_id = qualified_identity
                    .identity
                    .id()
                    .to_string(Encoding::Base58);
                let identity_label = qualified_identity
                    .alias
                    .as_deref()
                    .or_else(|| {
                        qualified_identity
                            .dpns_names
                            .first()
                            .map(|n| n.name.as_str())
                    });
                match identity_label {
                    Some(label) => format!(
                        "Identity {} ({}) {}",
                        label, identity_id, MISSING_DOCUMENT_SIGNING_KEY_MARKER,
                    ),
                    None => format!(
                        "Identity ({}) {}",
                        identity_id, MISSING_DOCUMENT_SIGNING_KEY_MARKER,
                    ),
                }
```

</details>

</details>

<details>
<summary>🟡 <b>suggestion:</b> Substring-matching a user-facing error string is fragile and conflicts with i18n rules</summary>

`src/ui/identities/mod.rs:31-43`

`is_missing_document_signing_key_error` decides program behaviour (suppress vs. surface a banner) by `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)` against text that is also rendered verbatim to the user. Per CLAUDE.md, user-facing strings must be 'simple, complete sentences' and 'single extractable units with named placeholders' — but the marker is spliced into the middle of a larger sentence, so any future copy edit, rephrasing, or i18n extraction will silently break the suppression and the startup banner returns. It also concatenates a static fragment, which won't survive translation.

A typed signal removes the substring contract and matches the project's stated direction. Two options:

1. Change `get_selected_wallet`'s return type to a small enum (e.g. `Result<Option<Arc<RwLock<Wallet>>>, GetSelectedWalletError>` with variants like `MissingDocumentSigningKey { identity_label: Option<String>, identity_id: String }`, `DpnsPreorderTypeUnavailable(...)`, `NoKeyProvided`). Callers pattern-match the variant; the UI boundary formats `Display`.
2. If keeping the `String` error: return `Ok(None)` for the missing-signing-key case (it is legitimately a 'no wallet for this identity' outcome) and let the screens render a hint when they actually need a signer.

Either approach erases the `MISSING_DOCUMENT_SIGNING_KEY_MARKER` constant and the `is_missing_document_signing_key_error` helper.

</details>

<details>
<summary>🟡 <b>suggestion:</b> Error message lacks a self-service action and uses everyday-user jargon</summary>

`src/ui/identities/mod.rs:107-112`

CLAUDE.md's error-message rules require *what happened + what to do* with a concrete self-service action, and explicitly prohibit jargon — `state transition` is cited as an example. The current message — `"... doesn't have an authentication key for signing document transitions"` — describes mechanism (`authentication key`, `document transitions`) without telling the user how to resolve it. Since this banner now appears only on user-initiated identity selection (auto-select suppresses it), actionability matters more than before. Reword along the lines of *"Identity <label> (<id>) cannot sign DashPay actions because it has no signing key. Add a signing key with high or critical security level to this identity, or pick a different identity."*

</details>

<details>
<summary>🟡 <b>suggestion:</b> No tests pin the marker contract or the format-string output</summary>

`src/ui/identities/mod.rs:31-43`

Two new behaviours are now load-bearing without coverage: (1) the formatted error must contain `MISSING_DOCUMENT_SIGNING_KEY_MARKER` for `is_missing_document_signing_key_error` to suppress the banner, and (2) the label/id formatting (including the empty-label edge case). A future innocent edit to the format string will silently break suppression. A small unit test that builds a `QualifiedIdentity` with no signing key and asserts the substring is present and the rendered message is well-formed across the three label states (alias=Some, alias=None+dpns_name, alias=None+no dpns_name) would lock the contract until a typed-error refactor lands.

</details>

<details>
<summary>💬 <b>nitpick:</b> `use` statements buried inside closure body</summary>

`src/ui/identities/mod.rs:95-96`

`use dash_sdk::dpp::identity::accessors::IdentityGettersV0;` and `use dash_sdk::dpp::platform_value::string_encoding::Encoding;` are placed inside the `ok_or_else` closure. Hoisting them to the file's top-level `use` block (or at minimum to the function scope) makes the trait dependencies discoverable and matches the surrounding style — `IdentityGettersV0` is already imported via the same path elsewhere in this codebase.

</details>

<details>

```

In `src/ui/dashpay/profile_screen.rs`:
- [BLOCKING] lines 206-210: ProfileScreen auto-select still surfaces the banner the PR claims to suppress
  `DashPayScreen::new` (src/ui/dashpay/dashpay_screen.rs:38) eagerly constructs `ProfileScreen` alongside `ContactsList` and `PaymentHistory` at startup, and `AppState` builds multiple `DashPayScreen` instances. `ContactRequests::new` and `QRCodeGeneratorScreen::new` were updated to drop the missing-signing-key error via `is_missing_document_signing_key_error`, but `ProfileScreen::new` still calls `get_selected_wallet(&identities[selected_idx], Some(&app_context), None).or_show_error(app_context.egui_ctx())` on the same auto-select path. For an evonode identity without a document-signing key — the exact scenario in #743 — the global banner still fires at startup, so the user-visible bug is only partially fixed. Apply the same `Err(e) if is_missing_document_signing_key_error(&e) => None` arm here, or push the suppression down into `get_selected_wallet`/`or_show_error` so all auto-select callsites get it consistently.

In `src/ui/identities/mod.rs`:
- [BLOCKING] lines 97-112: Empty identity label produces malformed 'Identity  (id) ...' message
  When an identity has neither an `alias` nor any `dpns_names`, `identity_label` falls back to `""` (line 106), and the format string `"Identity {} ({}) {}"` produces `"Identity  (4ZmR…) doesn't have an authentication key for signing document transitions"` — note the double space and the dangling word `Identity`. This is the most common shape for non-user identities (evonodes), which is exactly the example called out in the PR description, so the message the PR is trying to clarify becomes garbled for the primary affected user. Branch on whether a label exists rather than emitting an empty placeholder.
- [SUGGESTION] lines 31-43: Substring-matching a user-facing error string is fragile and conflicts with i18n rules
  `is_missing_document_signing_key_error` decides program behaviour (suppress vs. surface a banner) by `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)` against text that is also rendered verbatim to the user. Per CLAUDE.md, user-facing strings must be 'simple, complete sentences' and 'single extractable units with named placeholders' — but the marker is spliced into the middle of a larger sentence, so any future copy edit, rephrasing, or i18n extraction will silently break the suppression and the startup banner returns. It also concatenates a static fragment, which won't survive translation.

A typed signal removes the substring contract and matches the project's stated direction. Two options:

1. Change `get_selected_wallet`'s return type to a small enum (e.g. `Result<Option<Arc<RwLock<Wallet>>>, GetSelectedWalletError>` with variants like `MissingDocumentSigningKey { identity_label: Option<String>, identity_id: String }`, `DpnsPreorderTypeUnavailable(...)`, `NoKeyProvided`). Callers pattern-match the variant; the UI boundary formats `Display`.
2. If keeping the `String` error: return `Ok(None)` for the missing-signing-key case (it is legitimately a 'no wallet for this identity' outcome) and let the screens render a hint when they actually need a signer.

Either approach erases the `MISSING_DOCUMENT_SIGNING_KEY_MARKER` constant and the `is_missing_document_signing_key_error` helper.
- [SUGGESTION] lines 107-112: Error message lacks a self-service action and uses everyday-user jargon
  CLAUDE.md's error-message rules require *what happened + what to do* with a concrete self-service action, and explicitly prohibit jargon — `state transition` is cited as an example. The current message — `"... doesn't have an authentication key for signing document transitions"` — describes mechanism (`authentication key`, `document transitions`) without telling the user how to resolve it. Since this banner now appears only on user-initiated identity selection (auto-select suppresses it), actionability matters more than before. Reword along the lines of *"Identity <label> (<id>) cannot sign DashPay actions because it has no signing key. Add a signing key with high or critical security level to this identity, or pick a different identity."*
- [SUGGESTION] lines 31-43: No tests pin the marker contract or the format-string output
  Two new behaviours are now load-bearing without coverage: (1) the formatted error must contain `MISSING_DOCUMENT_SIGNING_KEY_MARKER` for `is_missing_document_signing_key_error` to suppress the banner, and (2) the label/id formatting (including the empty-label edge case). A future innocent edit to the format string will silently break suppression. A small unit test that builds a `QualifiedIdentity` with no signing key and asserts the substring is present and the rendered message is well-formed across the three label states (alias=Some, alias=None+dpns_name, alias=None+no dpns_name) would lock the contract until a typed-error refactor lands.
```
</details>

thepastaclaw (commented):
## Code Review

Small, focused PR that suppresses a noisy startup banner and enriches an error with identity context. Two real concerns: an empty-label code path produces a malformed user-facing message ('Identity  (id) ...' with double space), and the suppression discriminator is a substring match against user-facing wording — fragile post the #739 typed-error refactor and at odds with the i18n-ready string rule. A secondary UX concern is that suppressing the banner defers the failure to the Accept/Reject click in ContactRequests.

*Reviewed commit: a6bd8f37*

Fresh dispatcher gate review for this claimed queue item. `review_poster.py` detected an existing same-SHA review, so this records the fresh verification as a top-level review without duplicating inline threads.

🟡 4 suggestion(s) | 💬 2 nitpick(s)

<details>
<summary>🟡 <b>suggestion:</b> Empty label renders malformed message with double space</summary>

`src/ui/identities/mod.rs:94-113`

When the identity has no `alias` and no `dpns_names`, `identity_label` becomes `""` via `unwrap_or("")`, so the format string `"Identity {} ({}) {}"` renders as `"Identity  (4ZmR…) doesn't have an authentication key for signing document transitions"` — note the double space. Commit `a6bd8f37` ("remove quotes from identity label and use empty string for unknown") created exactly this path. Branch on `Some/None` so the label segment (including its trailing space) is omitted when no label is known, e.g. `"Identity ({id}) {marker}"` vs `"Identity {label} ({id}) {marker}"`. The existing `Display for QualifiedIdentity` (`src/model/qualified_identity/mod.rs:295`) is not directly substitutable because its no-label fallback is the identity id itself, which would produce `"Identity {id} ({id}) ..."`.

</details>

<details>
<summary>🟡 <b>suggestion:</b> Suppression discriminator is a substring match on user-facing wording</summary>

`src/ui/identities/mod.rs:31-43`

`is_missing_document_signing_key_error` distinguishes the suppressible case via `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)`. This couples error identity to display text, exactly what #739 (`Result<T, String>` → `TaskError`) was meant to eliminate, and it conflicts with the CLAUDE.md i18n rule that messages be a single extractable translation unit (the marker is a sentence-fragment tail). If the wording is ever reworded, localized, or wrapped, suppression silently breaks and the original startup-banner regression returns with no compile-time signal. Convert `get_selected_wallet` to return a small typed error (e.g. `enum GetSelectedWalletError { MissingDocumentSigningKey { label, id }, DpnsPreorderType(#[source] …), NoKeyProvided }`); auto-select callsites then `match Err(GetSelectedWalletError::MissingDocumentSigningKey { .. }) => None`. That removes the marker constant, the helper, and the fragility — and aligns this helper with the rest of the codebase post-#739. The 30+ existing callsites mostly stringify or display the error, so the swap is largely mechanical.

</details>

<details>
<summary>🟡 <b>suggestion:</b> Suppressing the startup banner defers the failure to the Accept/Reject click</summary>

`src/ui/dashpay/contact_requests.rs:110-118`

When the missing-key error is suppressed, `selected_wallet` is set to `None`. The list view treats `None` as `wallet_locked = false` (`contact_requests.rs:834-836`) and renders active Accept/Reject buttons (line 843+). Both action paths still require an authentication key in the backend: `accept_contact_request()` fails with `DashPayError::MissingAuthenticationKey` (`backend_task/dashpay/contact_requests.rs:643`) and `reject_contact_request()` reaches `create_or_update_contact_info()` with the same requirement (`backend_task/dashpay/contact_info.rs:403`). Net effect: an auto-selected identity that lacks the key is now silent on arrival, but the first click still fails after the confirmation dialog instead of guiding the user to switch identities or add a key. Consider disabling the buttons (or showing an inline note on the row) when `selected_wallet` is `None` due to this specific case, so the limitation is visible before the user commits to an action.

</details>

<details>
<summary>🟡 <b>suggestion:</b> No test coverage for the suppression helper or the new error format</summary>

`src/ui/identities/mod.rs:92-113`

Both the new `is_missing_document_signing_key_error` helper and the wording produced at line 107-112 are now load-bearing for the suppression behaviour at startup, with no unit or UI test covering either. The most likely regression — someone editing the wording without updating the marker constant, or a future i18n pass — would silently reintroduce the original noisy-banner bug from #743 with no test failure. A small inline `#[test]` asserting `is_missing_document_signing_key_error(&format!("Identity {} ({}) {}", "label", "id", MISSING_DOCUMENT_SIGNING_KEY_MARKER))` is `true` (and obviously-unrelated strings are `false`) is the cheapest insurance while the substring coupling exists; switching to a typed error (other finding) makes this redundant.

</details>

<details>
<summary>💬 <b>nitpick:</b> Concatenated format violates the i18n-ready string guideline</summary>

`src/ui/identities/mod.rs:107-112`

CLAUDE.md: "All user-facing strings must be simple, complete sentences. Avoid concatenating fragments… Each string should be extractable as a single translation unit with named placeholders." The current `format!("Identity {} ({}) {}", label, id, MARKER)` splices a sentence-fragment tail onto a positional prefix — a translator would receive two disjoint pieces with locked English word order. Inline the full sentence as a single template with named placeholders (`"Identity {label} ({id}) doesn't have an authentication key for signing document transitions."`). Combined with the typed-error suggestion, this lives naturally on the variant's `#[error("...")]` attribute as one extractable unit.

</details>

<details>
<summary>💬 <b>nitpick:</b> Suppression block is duplicated across two callsites</summary>

`src/ui/dashpay/contact_requests.rs:110-118`

The exact match arm is repeated in `contact_requests.rs:110-118` and `qr_code_generator.rs:84-92`. At two callsites this is borderline, but if a third auto-select path needs the same policy (or the typed-error refactor lands), a small helper — e.g. `auto_select_wallet(&identity, &app_context, ctx) -> Option<Arc<RwLock<Wallet>>>` — keeps the suppression policy in one place. Take it or leave it.

</details>

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] lines 94-113: Empty label renders malformed message with double space
  When the identity has no `alias` and no `dpns_names`, `identity_label` becomes `""` via `unwrap_or("")`, so the format string `"Identity {} ({}) {}"` renders as `"Identity  (4ZmR…) doesn't have an authentication key for signing document transitions"` — note the double space. Commit `a6bd8f37` ("remove quotes from identity label and use empty string for unknown") created exactly this path. Branch on `Some/None` so the label segment (including its trailing space) is omitted when no label is known, e.g. `"Identity ({id}) {marker}"` vs `"Identity {label} ({id}) {marker}"`. The existing `Display for QualifiedIdentity` (`src/model/qualified_identity/mod.rs:295`) is not directly substitutable because its no-label fallback is the identity id itself, which would produce `"Identity {id} ({id}) ..."`.
- [SUGGESTION] lines 31-43: Suppression discriminator is a substring match on user-facing wording
  `is_missing_document_signing_key_error` distinguishes the suppressible case via `error.contains(MISSING_DOCUMENT_SIGNING_KEY_MARKER)`. This couples error identity to display text, exactly what #739 (`Result<T, String>` → `TaskError`) was meant to eliminate, and it conflicts with the CLAUDE.md i18n rule that messages be a single extractable translation unit (the marker is a sentence-fragment tail). If the wording is ever reworded, localized, or wrapped, suppression silently breaks and the original startup-banner regression returns with no compile-time signal. Convert `get_selected_wallet` to return a small typed error (e.g. `enum GetSelectedWalletError { MissingDocumentSigningKey { label, id }, DpnsPreorderType(#[source] …), NoKeyProvided }`); auto-select callsites then `match Err(GetSelectedWalletError::MissingDocumentSigningKey { .. }) => None`. That removes the marker constant, the helper, and the fragility — and aligns this helper with the rest of the codebase post-#739. The 30+ existing callsites mostly stringify or display the error, so the swap is largely mechanical.
- [SUGGESTION] lines 92-113: No test coverage for the suppression helper or the new error format
  Both the new `is_missing_document_signing_key_error` helper and the wording produced at line 107-112 are now load-bearing for the suppression behaviour at startup, with no unit or UI test covering either. The most likely regression — someone editing the wording without updating the marker constant, or a future i18n pass — would silently reintroduce the original noisy-banner bug from #743 with no test failure. A small inline `#[test]` asserting `is_missing_document_signing_key_error(&format!("Identity {} ({}) {}", "label", "id", MISSING_DOCUMENT_SIGNING_KEY_MARKER))` is `true` (and obviously-unrelated strings are `false`) is the cheapest insurance while the substring coupling exists; switching to a typed error (other finding) makes this redundant.

In `src/ui/dashpay/contact_requests.rs`:
- [SUGGESTION] lines 110-118: Suppressing the startup banner defers the failure to the Accept/Reject click
  When the missing-key error is suppressed, `selected_wallet` is set to `None`. The list view treats `None` as `wallet_locked = false` (`contact_requests.rs:834-836`) and renders active Accept/Reject buttons (line 843+). Both action paths still require an authentication key in the backend: `accept_contact_request()` fails with `DashPayError::MissingAuthenticationKey` (`backend_task/dashpay/contact_requests.rs:643`) and `reject_contact_request()` reaches `create_or_update_contact_info()` with the same requirement (`backend_task/dashpay/contact_info.rs:403`). Net effect: an auto-selected identity that lacks the key is now silent on arrival, but the first click still fails after the confirmation dialog instead of guiding the user to switch identities or add a key. Consider disabling the buttons (or showing an inline note on the row) when `selected_wallet` is `None` due to this specific case, so the limitation is visible before the user commits to an action.
```
</details>

thepastaclaw (commented):

The latest delta fixes prior-2 by replacing substring parsing with typed error matching, and the targeted ContactRequests constructor test passes. A newly identified current-head blocker remains: eager ProfileScreen construction still publishes the missing-key banner, defeating the PR's startup suppression. Carried-forward concerns remain around the malformed and unactionable message, enabled unsupported actions, and missing regression coverage.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 1 blocking | 🟡 3 suggestion(s)

<details>

```

In `src/ui/dashpay/profile_screen.rs`:
- [BLOCKING] src/ui/dashpay/profile_screen.rs:137-139: ProfileScreen still displays the banner during startup
  AppState eagerly constructs four DashPayScreen instances at src/app.rs:1110-1116, and every DashPayScreen::new constructs a ProfileScreen at src/ui/dashpay/dashpay_screen.rs:38-40. ProfileScreen::new calls get_selected_wallet for the same preferred identity and passes the error to or_show_error. Whenever the ContactRequests and QR generator suppression branches are relevant, this constructor therefore adds the missing-key error to the global banner queue during the same startup sequence. MessageBanner only deduplicates identical text; it does not remove the banner. Apply the same typed automatic-selection policy here while preserving the user-initiated selection behavior at lines 393-398. The duplicated policy in the other constructors has already drifted by omitting this eager callsite.

In `src/ui/identities/mod.rs`:
- [SUGGESTION] src/ui/identities/mod.rs:55-61: Missing-key message is malformed and offers no usable recovery
  When an identity has neither an alias nor a DPNS name, identity_label is empty and this renders "Identity  (<id>) ..." with two spaces. The message also exposes implementation terminology about an "authentication key for signing document transitions" and provides no self-service action, contrary to the user-facing error rules. Finally, interpolating MISSING_DOCUMENT_SIGNING_KEY_MESSAGE splits the sentence across translation units. Use complete labeled and unlabeled templates that briefly explain the limitation and tell the user to add a signing key or choose another identity.
- [SUGGESTION] src/ui/identities/mod.rs:76-80: Startup suppression and error formatting lack regression coverage
  No test references SelectedWalletError or is_missing_document_signing_key_error, and the constructor tests in tests/kittest/dashpay_screen.rs only assert which identity was selected. They do not inspect global banners, distinguish automatic from user-initiated selection, or cover labeled and unlabeled Display output. Add focused tests proving that every eager automatic-selection path suppresses only MissingDocumentSigningKey, user selection surfaces it, and both message formats remain well-formed.

In `src/ui/dashpay/contact_requests.rs`:
- [SUGGESTION] src/ui/dashpay/contact_requests.rs:122-130: Suppressed signing failure remains actionable in the request list
  This branch maps MissingDocumentSigningKey to selected_wallet = None. The request list treats None as unlocked at lines 707-718 and enables Reject and Accept. Accept then requires a critical or high authentication key at src/backend_task/dashpay/contact_requests.rs:748-758, while Reject reaches src/backend_task/dashpay/contact_info.rs:652-666 and also requires an authentication key. Track this specific unavailable-signing state and disable the actions or provide an inline recovery choice before the confirmation dialog.
```
</details>

_Note: GitHub does not allow me to request changes on my own PR, so this preliminary blocker review is posted as a COMMENT while the public gate remains blocked._

thepastaclaw:
The failing **Test Suite** job is unrelated to this PR's changes.

The sole failure was `context::wallet_lifecycle::tests::reconnect_restart_in_place_reuses_backend`: during its own stop/reconnect sequence, the SPV backend reported that its temporary data directory was still locked. PR #761 does not modify `src/context/wallet_lifecycle/` or the SPV backend, and that test/path is identical between base `3b32bed2` and head `59afcc14`. The preceding PR head `cd9c0d8f` also passed the full Test Suite with the same lifecycle code; the current head only changes `ProfileScreen` startup handling and its UI regression coverage, and Clippy passed.

The PR is proper as-is. I am not pushing a no-op change or rerunning CI for this unrelated test-isolation failure; maintainers can rerun the job if desired.

Failed job: https://github.com/dashpay/dash-evo-tool/actions/runs/29589455775/job/87914598070

src/ui/identities/mod.rs
```diff
@@ -28,6 +31,54 @@ pub mod top_up_identity_screen;
...
+            Self::DpnsPreorderDocumentTypeNotFound { .. } => {
+                write!(f, "DPNS preorder document type not found")
+            }
+            Self::MissingDocumentSigningKey {
+                identity_label,
+                identity_id,
+            } => write!(
+                f,
+                "Identity {identity_label} ({identity_id}) {MISSING_DOCUMENT_SIGNING_KEY_MESSAGE}"
+            ),
```
thepastaclaw: **🟡 Suggestion: Empty identity label still produces malformed error text**

get_selected_wallet (lines 134-147) falls back to an empty string for identity_label when the identity has neither an alias nor a resolved DPNS name. The Display impl for MissingDocumentSigningKey unconditionally interpolates this label between "Identity" and "({identity_id})", so an unlabeled identity produces "Identity  (<id>) doesn't have an authentication key..." — a visible double space and no way to tell which identity is affected. Unlabeled User identities are a normal, reachable state (no alias set, no DPNS name registered yet), so this isn't hypothetical. Omit the label segment entirely when it's empty.

```suggestion
            Self::MissingDocumentSigningKey {
                identity_label,
                identity_id,
            } => {
                if identity_label.is_empty() {
                    write!(
                        f,
                        "Identity ({identity_id}) {MISSING_DOCUMENT_SIGNING_KEY_MESSAGE}"
                    )
                } else {
                    write!(
                        f,
                        "Identity {identity_label} ({identity_id}) {MISSING_DOCUMENT_SIGNING_KEY_MESSAGE}"
                    )
                }
            }
```

src/ui/identities/mod.rs
```diff
@@ -64,26 +115,40 @@ pub fn get_selected_wallet(
...
+                            .dpns_names
+                            .first()
+                            .map(|n| n.name.as_str())
+                    })
+                    .unwrap_or("");
+                SelectedWalletError::MissingDocumentSigningKey {
+                    identity_label: identity_label.to_string(),
+                    identity_id: qualified_identity.identity.id().to_string(Encoding::Base58),
+                }
             })?
```
thepastaclaw, line 166: **🟡 Suggestion: Error formatting and the original suppression callsites remain untested**

src/ui/identities/mod.rs has no #[cfg(test)] module at all, so SelectedWalletError::Display is untested for the alias, DPNS-name, and unlabeled cases — the double-space regression above would slip through CI. Separately, only ProfileScreen gained suppression-policy tests in this push (profile_screen.rs:1176-1201, plus two kittests in tests/kittest/dashpay_screen.rs). ContactRequests::new (contact_requests.rs:118-130) and QRCodeGeneratorScreen::new (qr_code_generator.rs:85-97) implement the identical suppress-or-banner branch but have no test asserting the banner is suppressed on auto-select or surfaced for other errors. Given that this exact duplication already caused ProfileScreen's suppression to be missed once, the two original, still-untested copies are a real regression risk.

src/ui/identities/mod.rs
```diff
@@ -28,6 +31,54 @@ pub mod top_up_identity_screen;
...
+        match self {
+            Self::DpnsPreorderDocumentTypeNotFound { .. } => {
+                write!(f, "DPNS preorder document type not found")
+            }
+            Self::MissingDocumentSigningKey {
+                identity_label,
+                identity_id,
+            } => write!(
+                f,
+                "Identity {identity_label} ({identity_id}) {MISSING_DOCUMENT_SIGNING_KEY_MESSAGE}"
```
thepastaclaw: **💬 Nitpick: Missing-key message still built from a concatenated sentence fragment**

MISSING_DOCUMENT_SIGNING_KEY_MESSAGE ("doesn't have an authentication key for signing document transitions") is a grammatical fragment, not a complete sentence, and gets spliced into a second template at the write! call. Per this repo's i18n-ready string rule, every user-facing message should be one complete, directly extractable translation unit with named placeholders — concatenation bakes in English word order and prevents a translator working from the fragment alone from reordering label/ID/explanation correctly. Write the complete labeled and unlabeled sentences directly in the Display arms instead of composing from a shared fragment constant.

thepastaclaw (commented):

The current head fixes the ProfileScreen startup-banner regression (prior-7) by routing automatic selection through auto_selected_wallet_or_banner, and replaces the brittle string-matching discriminator with a typed SelectedWalletError variant (prior-2), both independently verified against source and covered by passing focused tests. Four in-scope concerns remain unresolved: an unlabeled identity still renders a malformed double-space message, the Display formatter and the two original suppression callsites (ContactRequests, QRCodeGenerator) have no test coverage, the user-facing message still concatenates a raw English sentence fragment, and the auto-selection suppression policy is now duplicated across three independent call sites instead of two.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)
- Verifier: `claude-sonnet-5` — verifier
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — general (failed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — general (failed), `claude-sonnet-5` — general (completed)

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] src/ui/identities/mod.rs:55-61: Empty identity label still produces malformed error text
  get_selected_wallet (lines 134-147) falls back to an empty string for identity_label when the identity has neither an alias nor a resolved DPNS name. The Display impl for MissingDocumentSigningKey unconditionally interpolates this label between "Identity" and "({identity_id})", so an unlabeled identity produces "Identity  (<id>) doesn't have an authentication key..." — a visible double space and no way to tell which identity is affected. Unlabeled User identities are a normal, reachable state (no alias set, no DPNS name registered yet), so this isn't hypothetical. Omit the label segment entirely when it's empty.
- [SUGGESTION] src/ui/identities/mod.rs:49-148: Error formatting and the original suppression callsites remain untested
  src/ui/identities/mod.rs has no #[cfg(test)] module at all, so SelectedWalletError::Display is untested for the alias, DPNS-name, and unlabeled cases — the double-space regression above would slip through CI. Separately, only ProfileScreen gained suppression-policy tests in this push (profile_screen.rs:1176-1201, plus two kittests in tests/kittest/dashpay_screen.rs). ContactRequests::new (contact_requests.rs:118-130) and QRCodeGeneratorScreen::new (qr_code_generator.rs:85-97) implement the identical suppress-or-banner branch but have no test asserting the banner is suppressed on auto-select or surfaced for other errors. Given that this exact duplication already caused ProfileScreen's suppression to be missed once, the two original, still-untested copies are a real regression risk.
```
</details>

src/ui/dashpay/contact_requests.rs
```diff
@@ -115,10 +115,19 @@ impl ContactRequests {
...
+            // anything else so we don't silently hide real failures.
+            new_self.selected_wallet =
+                match get_selected_wallet(&preferred, Some(&app_context), None) {
+                    Ok(wallet) => wallet,
+                    Err(e) if is_missing_document_signing_key_error(&e) => None,
+                    Err(e) => {
+                        MessageBanner::set_global(app_context.egui_ctx(), &e, MessageType::Error);
+                        None
+                    }
+                };
```
thepastaclaw: **💬 Nitpick: Automatic-selection suppression policy now duplicated three times**

ContactRequests::new (lines 122-130) and QRCodeGeneratorScreen::new (qr_code_generator.rs:92) each independently match on is_missing_document_signing_key_error to decide whether to suppress the banner on auto-select. This push fixes ProfileScreen's missing suppression by adding a third, file-local implementation of the same policy (auto_selected_wallet_or_banner, profile_screen.rs:1151-1163) rather than extracting a shared helper. ProfileScreen being missed in the first place is direct evidence of the maintenance risk this duplication creates — a fourth auto-selecting screen could just as easily omit or diverge from the policy. Extract auto_selected_wallet_or_banner (or equivalent) into src/ui/identities/mod.rs and have all three constructors call it.

src/ui/identities/mod.rs
thepastaclaw: Fixed in aba65b2c. MissingDocumentSigningKey now stores an optional label and renders complete labeled/unlabeled messages; unlabeled identities omit the label segment and remain identifiable by Base58 ID. The prior no-quotes maintainer wording is preserved.

src/ui/identities/mod.rs
thepastaclaw: Fixed in aba65b2c. Added labeled/unlabeled Display tests, centralized the constructor auto-selection policy, added missing-vs-unrelated error policy tests, and added direct startup-banner regression coverage for ContactRequests and QRCodeGeneratorScreen. Focused unit/kittests and full Clippy pass.

thepastaclaw:
Addressed the two exact-head review findings in aba65b2c: corrected labeled/unlabeled signing-key error formatting, centralized constructor-time suppression across Profile, Contact Requests, and QR Generator, and added focused unit plus kittest coverage. Local formatting, focused tests, and full all-feature/all-target Clippy pass. CI and automated CodeRabbit follow-through are now pending for the new head.

src/ui/identities/mod.rs
```diff
@@ -28,6 +32,81 @@ pub mod top_up_identity_screen;
...
+                         security level to this identity, or choose a different identity."
+                    )
+                } else {
+                    write!(
+                        f,
+                        "Identity {identity_id} cannot sign DashPay actions because it has no \
+                         signing key. Add a signing key with high or critical security level to \
+                         this identity, or choose a different identity."
+                    )
+                }
```
thepastaclaw, line 65: **🔴 Blocking: Generic signing-key error recommends a key the lookup will reject**

`MissingDocumentSigningKey` comes from `get_selected_wallet` whenever an `AppContext` is supplied, and that helper is used not only by DashPay screens but also by DPNS registration, contract registration/update, and generic document-action screens. The new text therefore misidentifies non-DashPay failures as “DashPay actions.” More importantly, `QualifiedIdentity::document_signing_key` passes a singleton set containing the DPNS `preorder` document type requirement; that schema has no override, so the requirement defaults to `HIGH`. `get_first_public_key_matching` performs an exact `HashSet::contains` check, so a newly added `CRITICAL` key does not satisfy this helper and the same banner remains. Describe the current action generically and recommend the high-security key that actually clears this code path.

```suggestion
                if let Some(identity_label) = identity_label {
                    write!(
                        f,
                        "Identity {identity_label} ({identity_id}) cannot sign this action because \
                         it has no high-security signing key. Add a high-security signing key to \
                         this identity, or choose a different identity."
                    )
                } else {
                    write!(
                        f,
                        "Identity {identity_id} cannot sign this action because it has no \
                         high-security signing key. Add a high-security signing key to this \
                         identity, or choose a different identity."
                    )
                }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The latest push fixes all actionable prior review issues: it uses a typed error, handles unlabeled identities, centralizes constructor-time suppression across the three affected DashPay screens, and adds passing unit and UI regression tests. One blocking user-facing defect remains in the newly expanded recovery message: the generic helper labels failures as DashPay-specific and recommends a critical-security key even though its exact DPNS preorder lookup accepts only a high-security key.

### Review provenance

🔴 1 blocking

<details>

```

In `src/ui/identities/mod.rs`:
- [BLOCKING] src/ui/identities/mod.rs:57-71: Generic signing-key error recommends a key the lookup will reject
  `MissingDocumentSigningKey` comes from `get_selected_wallet` whenever an `AppContext` is supplied, and that helper is used not only by DashPay screens but also by DPNS registration, contract registration/update, and generic document-action screens. The new text therefore misidentifies non-DashPay failures as “DashPay actions.” More importantly, `QualifiedIdentity::document_signing_key` passes a singleton set containing the DPNS `preorder` document type requirement; that schema has no override, so the requirement defaults to `HIGH`. `get_first_public_key_matching` performs an exact `HashSet::contains` check, so a newly added `CRITICAL` key does not satisfy this helper and the same banner remains. Describe the current action generically and recommend the high-security key that actually clears this code path.
```
</details>

Source: reviewers — gpt-5.6-sol (general), gpt-5.6-sol (rust-quality); verifier — gpt-5.6-sol.

_Note: GitHub does not allow me to request changes on my own PR, so this preliminary blocker review is posted as a COMMENT while the public gate remains blocked._

src/ui/identities/mod.rs
thepastaclaw: Resolved in this update — _Error formatting and the original suppression callsites remain untested_ no longer present.

src/ui/identities/mod.rs
thepastaclaw: Resolved in this update — _Missing-key message still built from a concatenated sentence fragment_ no longer present.

src/ui/dashpay/contact_requests.rs
thepastaclaw: Resolved in this update — _Automatic-selection suppression policy now duplicated three times_ no longer present.

src/ui/identities/mod.rs
thepastaclaw: Resolved in this update — _Empty identity label still produces malformed error text_ no longer present.

src/ui/identities/mod.rs
thepastaclaw: Fixed in cadc51f5.

`MissingDocumentSigningKey` Display now uses generic “this action” wording (covers DashPay, DPNS, contract, and document callers) and recommends adding a high-security signing key — the exact `SecurityLevel::HIGH` requirement this DPNS-preorder lookup accepts via exact `HashSet` membership.

Validation:
- `cargo test --lib ui::identities::tests --all-features` — 4/4 passed (labeled/unlabeled Display pins + auto-select policy)
- `cargo +nightly fmt --all` — clean
- Pre-commit hooks passed on commit
- Independent code-review gate on the wording-only diff: no accepted actionable findings

Pushed to `origin/fix/startup-auth-key-error-v1.0-dev`.

thepastaclaw (commented):

At exact head cadc51f522987a571bcb499be20152234e4d8a38, independently re-verified against the live source and test suite: SelectedWalletError::MissingDocumentSigningKey stores identity_label as Option<String> with two well-formed Display branches (src/ui/identities/mod.rs:53-72), is_missing_document_signing_key_error/auto_selected_wallet_or_banner do structural enum matching rather than string parsing (mod.rs:89-108), and all three constructors (ContactRequests, ProfileScreen, QRCodeGeneratorScreen) reuse the single centralized helper. Traced contact_requests.rs directly: selected_wallet only gates the optional wallet-unlock UI (line 698), while Accept/Reject dispatch (lines 331-396) sends only identity+request_id and resolves signing keys independently in the backend, confirming prior-3 is genuinely outdated. Ran the exact commands myself: 4/4 targeted unit tests pass, 14/14 dashpay_screen kittest tests pass, `cargo +nightly fmt --check` is clean, and `cargo clippy --all-features --all-targets -- -D warnings` is clean. The 'high-security signing key' wording matches document_signing_key's exact-set lookup against document_type.security_level_requirement() (model/qualified_identity/mod.rs:847-852). No in-scope findings remain.

### Review provenance
- Verifier: `claude-sonnet-5` — final-verifier
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — general (completed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — rust-quality (failed)

src/ui/identities/mod.rs
thepastaclaw: Resolved in `cadc51f` — _Generic signing-key error recommends a key the lookup will reject_ no longer present.

thepastaclaw:
@lklimek This is ready for maintainer review: head cadc51f5 is mergeable, all current checks are green, and exact-head Codex + Sonnet validation found no remaining in-scope issues after the startup auto-select and signing-key guidance fixes. Could you take a look when you have a chance?

src/ui/identities/mod.rs
```diff
@@ -19,6 +26,81 @@ pub mod top_up_identity_screen;
...
+/// auto-selected identities, but surface every other failure.
+pub fn auto_selected_wallet_or_banner(
+    ctx: &egui::Context,
+    result: Result<Option<Arc<RwLock<Wallet>>>, SelectedWalletError>,
+) -> Option<Arc<RwLock<Wallet>>> {
+    match result {
+        Ok(wallet) => wallet,
+        Err(error) if is_missing_document_signing_key_error(&error) => None,
+        Err(error) => {
+            MessageBanner::set_global(ctx, &error, MessageType::Error);
```
thepastaclaw, line 98: **🟡 Suggestion: Preserve error details when reporting unexpected selection failures**

SelectedWalletError::DpnsPreorderDocumentTypeNotFound preserves the underlying DataContractError, but its Display implementation now omits that source. MessageBanner::set_global stores only Display text, so this branch discards diagnostics that the previous error string included. Use set_global_with_error to retain the concise message while attaching Debug details for the details panel and logs. Explicit-selection callers using or_show_error also discard the newly separated diagnostics; preserve details at those SelectedWalletError reporting sites as well, without changing the expected startup suppression.

```suggestion
            MessageBanner::set_global_with_error(ctx, &error);
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Final validation — Phase 2 only (queue backlog)

Verified both supplied findings against exact head 8cbed5c888887f9e24673b277723f16412eb28e7 and consolidated them into one non-blocking diagnostic-preservation suggestion. The typed startup suppression is selective, but unexpected contract errors lose their underlying diagnostic when displayed. Validation was source-based; tests were not rerun.

### Review provenance
- Triage: `normal` by `gpt-6-astra` (effort low) — The change introduces typed wallet-selection errors and selective startup banner suppression across several screens, requiring checks for unintended error hiding and preserved user-initiated feedback, but leaves signing, key selection, and fund-handling logic unchanged.
- Phase 1 reviewers: **not run (skipped for throughput: 63 PRs queued, above the 10 limit)**
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — rust-quality (completed, effort high); agent `phase2-reviewer`

🟡 1 suggestion(s)

<details>

```

In `src/ui/identities/mod.rs`:
- [SUGGESTION] src/ui/identities/mod.rs:98: Preserve error details when reporting unexpected selection failures
  SelectedWalletError::DpnsPreorderDocumentTypeNotFound preserves the underlying DataContractError, but its Display implementation now omits that source. MessageBanner::set_global stores only Display text, so this branch discards diagnostics that the previous error string included. Use set_global_with_error to retain the concise message while attaching Debug details for the details panel and logs. Explicit-selection callers using or_show_error also discard the newly separated diagnostics; preserve details at those SelectedWalletError reporting sites as well, without changing the expected startup suppression.
```
</details>

lklimek:
closing as outdated - base has moved way too much to correctly apply. If this issue is still present, please reproduce and recreate new PR on top of current v1.0-dev.

PR #767, opened 2026-03-17
feat: manual DAPI node discovery, SDK update, and network config improvements

lklimek:
## Summary

Imagine you're a user whose testnet node addresses went stale after a network migration. Previously, you'd need to manually edit a `.env` file with new IP addresses. Now you open Network Settings, see "DAPI: 0/18 reachable", click **"Fetch Node List"**, and the app fetches fresh addresses from the Dash network, saves them, and reconnects — all in one click.

### Key changes

- **Manual DAPI discovery**: "Fetch Node List" button in Network Settings fetches fresh node addresses from a DCG-operated HTTPS service on demand. Available for Mainnet and Testnet only.
- **No auto-discovery**: The app always boots from configured addresses in the `.env` file. No magic, no silent network calls at startup.
- **Health ratio display**: "DAPI: {available}/{total} reachable" shown in the Node Addresses section, using the existing connection status tracking.
- **Confirmation dialog**: When addresses already exist, confirms before replacing.
- **Trust disclosure**: Info tooltip explains the DCG service trust model (convenience trust, not security trust).
- **SDK update**: `dash-sdk` + `rs-sdk-trusted-context-provider` track `branch = "feat/mempool-support"`. `initialize_sdk` accepts `AddressList` directly.
- **Network-aware RPC defaults**: All 7 `unwrap_or(9998)` callsites replaced with `NetworkConfig::rpc_port(network)` — testnet defaults to 19998, devnet to 29998, regtest to 19898.
- **Testnet seed nodes**: Updated to new infrastructure at `68.67.122.1-29` (29 nodes).
- **Network chooser**: Mainnet and Testnet always visible; Devnet and Local only shown in developer mode.
- **macOS accessibility**: Eager AccessKit activation and labeled sidebar buttons.

### Architecture

```
Boot:
  Read dapi_addresses from config → parse with AddressList::from_str()
  If empty → log error, context creation fails (user must configure addresses)

Refresh (user-triggered):
  Click "Fetch Node List" → BackendTask::DiscoverDapiNodes
  → try_discover_nodes() via HTTPS to DCG endpoint
  → Save addresses_csv to config
  → ReinitCoreClientAndSdk with new addresses
  → "Updated to {count} node addresses."
```

### What was moved to PR #803

- Lazy network context creation (only active network at startup)
- `BackendTask::SwitchNetwork` for async network switching
- `BackendTask::ReinitCoreClientAndSdk` for async SDK reinit
- Devnet dynamic discovery support

## Test plan

- [ ] `cargo +nightly fmt --all` passes
- [ ] App starts with configured `.env` DAPI addresses — uses them directly
- [ ] App starts with empty/missing DAPI addresses — graceful error, no crash
- [ ] Network Settings shows "DAPI: X/Y reachable" health ratio
- [ ] Click "Fetch Node List" on Mainnet/Testnet — fetches, saves, reconnects
- [ ] Click "Fetch Node List" with existing addresses — confirmation dialog appears
- [ ] Fetch failure — error banner, existing addresses untouched
- [ ] "Fetch Node List" button hidden for Devnet/Regtest
- [ ] Info tooltip explains DCG service trust model
- [ ] Testnet connects via new seed nodes (68.67.122.1-29 in `.env.example`)
- [ ] Network chooser: Mainnet + Testnet always visible, Devnet + Local in dev mode only
- [ ] Network-aware RPC port defaults work (testnet → 19998 when port omitted)


src/dapi_discovery.rs
```diff
@@ -0,0 +1,98 @@
...
...250:443,https://213.199.34.251:443,https://213.199.35.15:443,https://213.199.35.18:443,https://213.199.35.6:443,https://213.199.44.112:443,https://2.58.82.231:443,https://31.220.84.93:443,https://31.220.85.180:443,https://31.220.88.116:443,https://37.27.83.17:443,https://37.60.236.151:443,https://37.60.236.161:443,https://37.60.236.201:443,https://37.60.236.212:443,https://37.60.236.247:443,https://37.60.236.249:443,https://37.60.243.119:443,https://37.60.243.59:443,https://37.60.244.220:443,https://44.240.99.214:443,https://49.12.102.105:443,https://49.13.154.121:443,https://49.13.193.251:443,https://49.13.237.193:443,https://49.13.28.255:443,https://51.195.118.43:443,https://51.83.191.208:443,https://5.189.186.78:443,https://52.10.213.198:443,https://52.33.9.172:443,https://54.69.95.118:443,https://5.75.133.148:443,https://64.23.134.67:443,https://65.108.246.145:443,https://65.109.65.126:443,https://65.21.145.147:443,https://79.137.71.84:443,https://81.17.101.141:443,https://91.107.204.136:443,https://91.107.226.241:443,https://93.190.140.101:443,https://93.190.140.111:443,https://93.190.140.112:443,https://93.190.140.114:443,https://93.190.140.162:443,https://95.216.146.18:443";
```
lklimek: We don't want fallback. We wanted a migration - where we delete old mainnet and testnet nodes from `.env` in user home directory. But no fallback.

lklimek: Addressed — the migration approach is now fully implemented in `migrate_env_file_if_needed()`. It comments out known old mainnet/testnet hardcoded addresses in the user's `.env` file. No fallback mechanism exists; if `.env` has no explicit addresses, dynamic discovery is used. File permissions are restored to `0o600` after migration.

src/context/mod.rs
```diff
@@ -147,8 +147,21 @@ impl AppContext {
...
+            Err(e) => {
+                tracing::error!(?network, error = %e, "Failed to resolve DAPI addresses");
+                return None;
+            }
+        };
+
         // Default to SPV provider initially; UI can switch backend after
-        let sdk = match initialize_sdk(&network_config, network, spv_provider.clone()) {
+        let sdk = match initialize_sdk(address_list, network, spv_provider.clone()) {
             Ok(sdk) => sdk,
```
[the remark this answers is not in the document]
lklimek: Fixed in `eb903895` — only the active network's `AppContext` is now created at startup. Other networks use `LazyContextParams` and are initialized on-demand when the user switches to them. Worst-case startup time drops from ~40s to ~10s.

thepastaclaw:
✅ Review complete (commit aa2f0bb7)

src/context/mod.rs
```diff
@@ -547,29 +560,32 @@ impl AppContext {
...
                 let rpc_provider = RpcProvider::new(self.db.clone(), self.network, &cfg)
                     .map_err(|e| TaskError::RpcProviderCreationFailed { detail: e })?;
-                // Swap in the updated RPC provider for future switches
                 {
                     let mut guard = self.rpc_context_provider.write()?;
                     *guard = rpc_provider.clone();
                 }
-                initialize_sdk(&cfg, self.network, rpc_provider)
+                initialize_sdk(address_list, self.network, rpc_provider)
                     .map_err(|e| TaskError::SdkInitializationFailed { detail: e })?
```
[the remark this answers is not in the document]
lklimek: to be addressed in #803

src/app.rs
```diff
@@ -218,6 +219,7 @@ impl AppState {
         let data_dir = app_user_data_dir_path()?;
         ensure_data_dir_exists(&data_dir)?;
         ensure_env_file(&data_dir);
+        crate::config::Config::migrate_env_file_if_needed();
```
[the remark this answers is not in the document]
lklimek: The testing constructor intentionally runs migration too — tests use ephemeral data dirs, so migrating the `.env` there is harmless and ensures test coverage of the migration path. No change needed.

src/config.rs
```diff
@@ -296,17 +414,21 @@ impl NetworkConfig {
...
+        let addrs = match self.dapi_addresses.as_deref() {
+            Some(a) => a.trim(),
+            None => return Ok(None),
+        };
+        if addrs.is_empty() {
+            return Ok(None);
+        }
+        AddressList::from_str(addrs)
+            .map(Some)
+            .map_err(|e| format!("Could not parse DAPI addresses '{addrs}': {e}"))
```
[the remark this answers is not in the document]
lklimek: Acknowledged — `dapi_address_list()` returns `Result<_, String>` which erases the typed `AddressListError`. This was improved from the original `Option<Result<...>>` pattern per Copilot's earlier review. Full typed error migration for config parsing is tracked but deferred — the current signature is adequate for the single callsite in `dapi_discovery.rs`.

src/app.rs
```diff
@@ -238,43 +254,54 @@ impl AppState {
...
+        let devnet_app_context = if saved_network == Network::Devnet {
+            make_context(Network::Devnet)
+        } else {
+            None
+        };
+        let local_app_context = if saved_network == Network::Regtest {
+            make_context(Network::Regtest)
+        } else {
+            None
+        };
```
thepastaclaw: **🔴 Blocking: Deferred networks are hidden from the network chooser, so the lazy initialization path is unreachable from the UI**

The PR's intended lazy-on-switch behavior cannot be exercised for deferred networks because the user cannot select them in the first place.

<sub>source: ['codex']</sub>

<details>

```

In `src/app.rs`:
- [BLOCKING] lines 277-294: Deferred networks are hidden from the network chooser, so the lazy initialization path is unreachable from the UI
  The PR's intended lazy-on-switch behavior cannot be exercised for deferred networks because the user cannot select them in the first place.
```
</details>

src/config.rs
```diff
@@ -103,12 +120,12 @@ impl Config {
...
-                "{}dapi_addresses={}",
-                prefix, config.dapi_addresses
-            )
-            .map_err(|e| ConfigError::SaveError { source: e })?;
+            if let Some(ref addrs) = config.dapi_addresses
+                && !addrs.is_empty()
+            {
+                writeln!(env_file, "{}dapi_addresses={}", prefix, addrs)
+                    .map_err(|e| ConfigError::SaveError { source: e })?;
+            }
```
thepastaclaw, line 144: **🔴 Blocking: Omitting `*_dapi_addresses` during save does not reliably clear stale values for later loads in the same process**

If an old `MAINNET_dapi_addresses`/`TESTNET_dapi_addresses` value was previously loaded into the process environment, removing the key from the file may not remove it from subsequent in-process loads, defeating the intended switch back to dynamic discovery.

<sub>source: ['codex']</sub>

<details>

```

In `src/config.rs`:
- [BLOCKING] lines 123-128: Omitting `*_dapi_addresses` during save does not reliably clear stale values for later loads in the same process
  If an old `MAINNET_dapi_addresses`/`TESTNET_dapi_addresses` value was previously loaded into the process environment, removing the key from the file may not remove it from subsequent in-process loads, defeating the intended switch back to dynamic discovery.
```
</details>

thepastaclaw (changes requested):
## Code Review

Verified 5 findings across the lazy network initialization, DAPI discovery, and config persistence paths. The convergent finding (deferred networks hidden from chooser) is confirmed. Additionally, synchronous DAPI discovery on the UI thread is a real blocking concern.

*Reviewed commit: 8eac7ff0*

🔴 3 blocking | 🟡 2 suggestion(s)

### 3 additional findings

<details>
<summary>🔴 <b>blocking:</b> Switching to a deferred network performs synchronous DAPI discovery on the egui update path and can stall the UI for up to 10 seconds</summary>

`src/app.rs (lines 1646-1647)`

Selecting a deferred network can freeze the UI while network I/O runs on the main egui execution path.

</details>
<details>
<summary>🟡 <b>suggestion:</b> Lazy-created network contexts never get a Core ZMQ listener</summary>

`src/app.rs (lines 523-653)`

A lazily created network will lack ZMQ-backed status/events even if its config enables ZMQ, causing inconsistent behavior versus eagerly created networks.

</details>
<details>
<summary>🟡 <b>suggestion:</b> Saving RPC password can synchronously reinitialize the SDK on the UI thread, including blocking DAPI discovery</summary>

`src/ui/network_chooser_screen.rs (lines 485-532)`

A simple settings save can stall the interface for the same reason as lazy context creation.

</details>

<details>

```

In `src/app.rs`:
- [BLOCKING] lines 277-294: Deferred networks are hidden from the network chooser, so the lazy initialization path is unreachable from the UI
  The PR's intended lazy-on-switch behavior cannot be exercised for deferred networks because the user cannot select them in the first place.
- [SUGGESTION] lines 523-653: Lazy-created network contexts never get a Core ZMQ listener
  A lazily created network will lack ZMQ-backed status/events even if its config enables ZMQ, causing inconsistent behavior versus eagerly created networks.
- [BLOCKING] lines 1646-1647: Switching to a deferred network performs synchronous DAPI discovery on the egui update path and can stall the UI for up to 10 seconds
  Selecting a deferred network can freeze the UI while network I/O runs on the main egui execution path.

In `src/ui/network_chooser_screen.rs`:
- [SUGGESTION] lines 485-532: Saving RPC password can synchronously reinitialize the SDK on the UI thread, including blocking DAPI discovery
  A simple settings save can stall the interface for the same reason as lazy context creation.

In `src/config.rs`:
- [BLOCKING] lines 123-128: Omitting `*_dapi_addresses` during save does not reliably clear stale values for later loads in the same process
  If an old `MAINNET_dapi_addresses`/`TESTNET_dapi_addresses` value was previously loaded into the process environment, removing the key from the file may not remove it from subsequent in-process loads, defeating the intended switch back to dynamic discovery.
```
</details>

src/app.rs
lklimek: This is addressed in PR #803 (`refactor/lazy-active-context`), which replaces per-network fields with a `BTreeMap<Network, Arc<AppContext>>` and makes all networks visible in the chooser regardless of whether their context is initialized. The lazy init path in `change_network()` creates the context on demand when the user selects a deferred network.

This PR (#767) is the base — #803 stacks on top of it.

src/config.rs
lklimek: The `save()` method atomically replaces the entire `.env` file via `NamedTempFile::persist()`, so old keys *are* removed from the file. The concern about stale process-level env vars is valid in theory — `dotenv` loads vars into `std::env`, and `envy::prefixed()` reads from process env, not the file. However, in practice this edge case doesn't trigger because config saves that affect DAPI addresses prompt a network reinit or app restart.

If this becomes an issue, the fix would be `std::env::remove_var()` for the old key during save. LOW priority.

src/dapi_discovery.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    match network {
+        Network::Mainnet | Network::Testnet => {}
+        _ => return Err(DapiDiscoveryError::AddressesRequired { network }),
+    }
+
+    let discovery_result = match tokio::runtime::Handle::try_current() {
+        Ok(handle) => tokio::task::block_in_place(|| {
+            handle.block_on(try_discover_nodes(network, devnet_name))
+        }),
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #803 (`46e094e6`). Added runtime flavor guard before `block_in_place`:

```rust
if handle.runtime_flavor() == tokio::runtime::RuntimeFlavor::CurrentThread {
    Err(DapiDiscoveryError::NoRuntime)
} else {
    tokio::task::block_in_place(|| { ... })
}
```

src/dapi_discovery.rs
```diff
@@ -0,0 +1,212 @@
...
+        if !trimmed.is_empty() {
+            return AddressList::from_str(trimmed)
+                .map_err(|source| DapiDiscoveryError::InvalidAddresses { source });
+        }
+    }
+
+    match network {
+        Network::Mainnet | Network::Testnet => {}
+        _ => return Err(DapiDiscoveryError::AddressesRequired { network }),
+    }
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #803 (`46e094e6`). Devnet now attempts dynamic discovery when `devnet_name` is configured. No seed-node fallback for Devnet (no hardcoded seeds exist) — discovery failure is final.

src/context/mod.rs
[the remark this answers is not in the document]
lklimek: Fixed in PR #803 (`46e094e6`). `reinit_core_client_and_sdk` is now routed through `BackendTask::ReinitCoreClientAndSdk` instead of running synchronously on the UI thread. The NetworkChooserScreen shows a "Reconnecting to Dash Core..." progress banner during reinit and displays success/failure when complete.

src/spv/manager.rs
```diff
@@ -1463,7 +1463,7 @@ impl SpvManager {
     fn primary_peer_socket(&self) -> Option<std::net::SocketAddr> {
         let config = self.config.read().ok()?;
 
-        let host = config.core_host.as_str();
+        let host = config.core_host.as_deref().unwrap_or("127.0.0.1");
```
lklimek: We should fall back to configured (or discovered) DAPI nodes, not hardcode "127.0.0.1".

lklimek:
@coderabbitai review all

src/spv/manager.rs
lklimek: Still open — `src/spv/manager.rs:1466` still has `unwrap_or("127.0.0.1")`. Will address in a follow-up: the fallback should use configured/discovered DAPI nodes rather than hardcoded localhost.

src/config.rs
```diff
@@ -290,29 +327,25 @@ impl Config {
...
+            Some(a) => a.trim(),
+            None => return Ok(None),
+        };
+        if addrs.is_empty() {
+            return Ok(None);
+        }
+        AddressList::from_str(addrs)
+            .map(Some)
+            .map_err(|e| format!("Could not parse DAPI addresses '{addrs}': {e}"))
     }
```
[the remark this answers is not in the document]
lklimek: Acknowledged — the docstring at `src/config.rs:331` still says "dynamic discovery should be used" but discovery is user-triggered, not automatic. Will update wording to clarify `None` means "not configured; the user may trigger discovery from Network Settings."

docs/ai-design/2026-03-30-dapi-discovery-button/ux-spec.md
```diff
@@ -0,0 +1,396 @@
...
+### 3.6 Success State (No Existing Addresses)
+
+When the field was empty before the fetch:
+
+1. Text field is populated with the fetched addresses (comma-separated)
+2. Button returns to default state
+3. A success banner appears: "Found {count} node addresses. Review them below and save your settings."
+4. The field is **not** auto-saved -- the user must click the existing Save mechanism (or the addresses are saved when the network config is saved)
+
+**Important**: The fetched addresses populate the field but do not persist until the user explicitly saves. This gives all personas a chance to review and edit.
```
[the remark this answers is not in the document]
lklimek: Acknowledged — the UX spec (`docs/ai-design/2026-03-30-dapi-discovery-button/ux-spec.md`) still describes "Fetch Node List" with explicit save, while the implementation uses "Refresh DAPI endpoints" with auto-save. Will align the spec to match the implementation.

src/ui/network_chooser_screen.rs
```diff
@@ -832,12 +797,83 @@ impl NetworkChooserScreen {
...
+                    };
+                    let button =
+                        egui::Button::new(button_text).corner_radius(Shape::RADIUS_MD);
+
+                    let response = ui.add_enabled(!self.discovery_in_progress, button);
+                    let clicked = response.clicked();
+                    response.on_hover_text(
+                        "Updates list of DAPI nodes using a centralized server managed by Dash Core Group.",
+                    );
+                    if clicked {
```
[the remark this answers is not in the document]
lklimek: The trust model is disclosed in the hover tooltip: *"Updates list of DAPI nodes using a centralized server managed by Dash Core Group."* The label "Refresh DAPI endpoints" (vs. "Fetch Node List") was an intentional UX refinement. Considering this adequately addressed — the substantive concern (trust disclosure) is present.

src/backend_task/dapi_discovery.rs
```diff
@@ -0,0 +1,135 @@
...
+    Ok(valid_urls)
+}
+
+/// Discover DAPI nodes and return them as a comma-separated string
+/// suitable for storing in `NetworkConfig::dapi_addresses`.
+///
+/// Returns `(count, csv)` on success.
+pub async fn discover_and_format(
+    network: Network,
+    devnet_name: Option<&str>,
```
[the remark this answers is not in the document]
lklimek: Acknowledged — `src/backend_task/dapi_discovery.rs` currently has no unit tests. The network-rejection path (`AddressesRequired` error on unsupported networks) and URL filtering logic are testable without network access. Will consider adding in this or a follow-up PR.

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2226,67 @@ impl ScreenLike for NetworkChooserScreen {
...
+                            "Failed to reinit SDK after DAPI discovery for {:?}: {}",
+                            network,
+                            e
+                        );
+                    }
+                }
+
+                MessageBanner::set_global(
+                    self.mainnet_app_context.egui_ctx(),
+                    format!("Updated to {count} node addresses."),
```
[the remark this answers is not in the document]
lklimek: Valid point — `display_task_result` in `network_chooser_screen.rs` unconditionally shows a success banner after discovery, even if `config.save()` or SDK reinit failed (those errors are only logged). Will address: show a warning banner when save fails, and an info banner about restart when network context is missing.

thepastaclaw (changes requested):
## Code Review

DAPI discovery PR has two regressions: removing Testnet guard crashes app, saving settings blocks UI.

*Reviewed commit: dd5fe9f4*

### 2 additional findings

<details>
<summary>🔴 <b>blocking:</b> Selecting Testnet crashes when no context exists</summary>

`src/ui/network_chooser_screen.rs (line 386)`

Removed is_some() guard. AppState panics on missing context.

</details>
<details>
<summary>🟡 <b>suggestion:</b> Save blocks UI during DAPI discovery</summary>

`src/ui/network_chooser_screen.rs (line 522)`

Synchronous block_on(try_discover_nodes) freezes egui thread.

</details>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] line 386: Selecting Testnet crashes when no context exists
  Removed is_some() guard. AppState panics on missing context.
- [SUGGESTION] line 522: Save blocks UI during DAPI discovery
  Synchronous block_on(try_discover_nodes) freezes egui thread.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2226,84 @@ impl ScreenLike for NetworkChooserScreen {
...
+                    .with_details(e);
+                    return;
+                }
+
+                MessageBanner::set_global(
+                    self.mainnet_app_context.egui_ctx(),
+                    format!("Updated to {count} node addresses."),
+                    MessageType::Success,
+                );
+            }
```
thepastaclaw: **🔴 Blocking: Discovered DAPI addresses silently discarded when target network has no config block**

The `if let` chain at lines 2248-2249 requires `config_for_network(network)` to return `Some(...)`. For Testnet — which is now always selectable in the network dropdown — this returns `None` when no `TESTNET_*` vars exist in `.env`. The entire discovery result is silently dropped: `discovery_in_progress` resets to false on line 2245, but no banner is shown and no addresses are saved.

This is the primary use case for the feature: a user selects Testnet, clicks "Refresh DAPI endpoints", discovery succeeds against the DCG endpoint, but the result vanishes with no feedback.

The fix should create a default `NetworkConfig` when the existing one is `None`, then populate it with the discovered addresses. An `else` branch should also handle the `Config::load_from` failure case with an error banner.

<details>

```suggestion
            if let Ok(mut config) = Config::load_from(&self.mainnet_app_context.data_dir) {
                let mut network_cfg = config
                    .config_for_network(network)
                    .clone()
                    .unwrap_or_default();
                network_cfg.dapi_addresses = Some(addresses_csv);
                config.update_config_for_network(network, network_cfg.clone());
```

</details>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] lines 2248-2306: Discovered DAPI addresses silently discarded when target network has no config block
  The `if let` chain at lines 2248-2249 requires `config_for_network(network)` to return `Some(...)`. For Testnet — which is now always selectable in the network dropdown — this returns `None` when no `TESTNET_*` vars exist in `.env`. The entire discovery result is silently dropped: `discovery_in_progress` resets to false on line 2245, but no banner is shown and no addresses are saved.

This is the primary use case for the feature: a user selects Testnet, clicks "Refresh DAPI endpoints", discovery succeeds against the DCG endpoint, but the result vanishes with no feedback.

The fix should create a default `NetworkConfig` when the existing one is `None`, then populate it with the discovered addresses. An `else` branch should also handle the `Config::load_from` failure case with an error banner.
```
</details>

src/spv/manager.rs
```diff
@@ -1470,6 +1493,19 @@ impl SpvManager {
...
+    /// Returns `None` if the string is empty or the first URL cannot be parsed.
+    fn first_dapi_host(dapi_addresses: &str) -> Option<&str> {
+        let first_url = dapi_addresses.split(',').next()?.trim();
+        // DAPI addresses are URLs like "https://host:port" — strip the scheme and port.
+        let after_scheme = first_url
+            .strip_prefix("https://")
+            .or_else(|| first_url.strip_prefix("http://"))?;
+        // Take everything before the port separator
+        Some(after_scheme.split(':').next()?.trim()).filter(|h| !h.is_empty())
+    }
```
thepastaclaw: **🟡 Suggestion: first_dapi_host does not strip trailing path/slash from DAPI URLs**

For URLs like `https://45.135.180.114/` (the exact format in `.env.example` mainnet addresses), `first_dapi_host` returns `45.135.180.114/` because it splits on `:` but never strips the path component. This would produce `45.135.180.114/:9999` in `primary_peer_socket()`, which `to_socket_addrs()` cannot resolve.

**Currently unreachable in practice**: `primary_peer_socket()` is only called for Devnet/Regtest (which use `:port` format) or when `use_local_node()` is true (which always sets `core_host` to `127.0.0.1` first). So `first_dapi_host` is never reached with trailing-slash URLs today. However, the function's contract promises to extract a host from any DAPI URL, and the latent bug will surface if any new code path calls it without the `core_host` guard.

<details>

```suggestion
    fn first_dapi_host(dapi_addresses: &str) -> Option<&str> {
        let first_url = dapi_addresses.split(',').next()?.trim();
        let after_scheme = first_url
            .strip_prefix("https://")
            .or_else(|| first_url.strip_prefix("http://"))?;
        let host_and_port = after_scheme.split('/').next()?;
        Some(host_and_port.split(':').next()?.trim()).filter(|h| !h.is_empty())
    }
```

</details>

<sub>source: ['claude']</sub>

<details>

```

In `src/spv/manager.rs`:
- [SUGGESTION] lines 1500-1508: first_dapi_host does not strip trailing path/slash from DAPI URLs
  For URLs like `https://45.135.180.114/` (the exact format in `.env.example` mainnet addresses), `first_dapi_host` returns `45.135.180.114/` because it splits on `:` but never strips the path component. This would produce `45.135.180.114/:9999` in `primary_peer_socket()`, which `to_socket_addrs()` cannot resolve.

**Currently unreachable in practice**: `primary_peer_socket()` is only called for Devnet/Regtest (which use `:port` format) or when `use_local_node()` is true (which always sets `core_host` to `127.0.0.1` first). So `first_dapi_host` is never reached with trailing-slash URLs today. However, the function's contract promises to extract a host from any DAPI URL, and the latent bug will surface if any new code path calls it without the `core_host` guard.
```
</details>

src/config.rs
```diff
@@ -58,6 +61,36 @@ pub struct NetworkConfig {
...
+    /// - Testnet: 19998
+    /// - Devnet: 29998
+    /// - Regtest: 19898
+    pub fn default_rpc_port(network: Network) -> u16 {
+        match network {
+            Network::Mainnet => 9998,
+            Network::Testnet => 19998,
+            Network::Devnet => 29998,
+            Network::Regtest => 19898,
+            _ => 9998,
```
thepastaclaw, line 78: **🟡 Suggestion: Regtest default RPC port (19898) doesn't match the app's docs and .env.example (20302)**

`default_rpc_port(Network::Regtest)` returns `19898` (standard Dash Core regtest port), but `.env.example`, `docs/local-network.md`, and all test fixtures use `20302` (dashmate's port). Since this app's setup guide assumes dashmate for local networks, the default is inconsistent. The mismatch only manifests when a user configures regtest without explicitly setting `core_rpc_port`, which the docs/example always include — so practical impact is low. But the default should match the app's own conventions to avoid confusion.

<sub>source: ['codex']</sub>

<details>

```

In `src/config.rs`:
- [SUGGESTION] lines 72-78: Regtest default RPC port (19898) doesn't match the app's docs and .env.example (20302)
  `default_rpc_port(Network::Regtest)` returns `19898` (standard Dash Core regtest port), but `.env.example`, `docs/local-network.md`, and all test fixtures use `20302` (dashmate's port). Since this app's setup guide assumes dashmate for local networks, the default is inconsistent. The mismatch only manifests when a user configures regtest without explicitly setting `core_rpc_port`, which the docs/example always include — so practical impact is low. But the default should match the app's own conventions to avoid confusion.
```
</details>

src/context/mod.rs
```diff
@@ -159,8 +161,30 @@ impl AppContext {
...
+                }
+            },
+            _ => {
+                tracing::error!(
+                    ?network,
+                    "No DAPI addresses configured. Use Refresh DAPI endpoints in Network Settings or add addresses to .env."
+                );
+                return None;
+            }
+        };
```
thepastaclaw, line 184: **🟡 Suggestion: Missing mainnet DAPI addresses prevent app launch — Refresh button is unreachable**

`AppContext::new()` returns `None` when DAPI addresses are absent (line 177-183), and `AppState::new()` treats a missing mainnet context as fatal (line 257-258). If a user's mainnet addresses are cleared or their `.env` is corrupted, the app won't start and the "Refresh DAPI endpoints" recovery button is unreachable. This is a pre-existing constraint (not introduced by this PR), but worth considering as the discovery feature matures — e.g., allowing the app to start in a degraded state with just the network settings screen.

<sub>source: ['codex']</sub>

<details>

```

In `src/context/mod.rs`:
- [SUGGESTION] lines 164-184: Missing mainnet DAPI addresses prevent app launch — Refresh button is unreachable
  `AppContext::new()` returns `None` when DAPI addresses are absent (line 177-183), and `AppState::new()` treats a missing mainnet context as fatal (line 257-258). If a user's mainnet addresses are cleared or their `.env` is corrupted, the app won't start and the "Refresh DAPI endpoints" recovery button is unreachable. This is a pre-existing constraint (not introduced by this PR), but worth considering as the discovery feature matures — e.g., allowing the app to start in a degraded state with just the network settings screen.
```
</details>

src/spv/manager.rs
thepastaclaw: **🟡 Suggestion: No unit tests for first_dapi_host URL parsing**

`first_dapi_host` manually parses URL strings with multiple edge cases (trailing slashes, missing schemes, ports, empty strings). A few focused unit tests would catch the trailing-slash bug and prevent regressions.

<sub>source: ['claude']</sub>

<details>

```

In `src/spv/manager.rs`:
- [SUGGESTION] lines 1500-1508: No unit tests for first_dapi_host URL parsing
  `first_dapi_host` manually parses URL strings with multiple edge cases (trailing slashes, missing schemes, ports, empty strings). A few focused unit tests would catch the trailing-slash bug and prevent regressions.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2226,84 @@ impl ScreenLike for NetworkChooserScreen {
...
+                }
+
+                let app_context = self.context_for_network(network);
+                {
+                    if let Ok(mut cfg_lock) = app_context.config.write() {
+                        *cfg_lock = network_cfg;
+                    }
+                }
+
+                if let Err(e) = Arc::clone(app_context).reinit_core_client_and_sdk() {
```
thepastaclaw: **🟡 Suggestion: reinit_core_client_and_sdk called synchronously in display_task_result (UI thread)**

`reinit_core_client_and_sdk()` rebuilds the RPC client and SDK from the UI thread's `display_task_result` path. While `Client::new()` and SDK init are likely fast (no network I/O), they do hold and replace write locks. If SDK initialization ever adds blocking work, the UI would freeze. Consider dispatching as a `BackendTask` for consistency with the app's async architecture.

<sub>source: ['claude']</sub>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [SUGGESTION] line 2291: reinit_core_client_and_sdk called synchronously in display_task_result (UI thread)
  `reinit_core_client_and_sdk()` rebuilds the RPC client and SDK from the UI thread's `display_task_result` path. While `Client::new()` and SDK init are likely fast (no network I/O), they do hold and replace write locks. If SDK initialization ever adds blocking work, the UI would freeze. Consider dispatching as a `BackendTask` for consistency with the app's async architecture.
```
</details>

src/backend_task/dapi_discovery.rs
```diff
@@ -0,0 +1,135 @@
...
+) -> Result<(usize, String), DapiDiscoveryError> {
+    let urls = try_discover_nodes(network, devnet_name).await?;
+    let count = urls.len();
+    let csv = urls.join(",");
+    Ok((count, csv))
+}
```
thepastaclaw, line 135: **🟡 Suggestion: No unit tests for dapi_discovery module**

The module has no tests. While `try_discover_nodes` requires network access, the error construction paths and `discover_and_format` CSV formatting are testable locally with mocked inputs.

<sub>source: ['claude']</sub>

<details>

```

In `src/backend_task/dapi_discovery.rs`:
- [SUGGESTION] lines 1-135: No unit tests for dapi_discovery module
  The module has no tests. While `try_discover_nodes` requires network access, the error construction paths and `discover_and_format` CSV formatting are testable locally with mocked inputs.
```
</details>

thepastaclaw (changes requested):
## Code Review

Well-structured PR with clean error types and proper trust model documentation. The main real bug is in the discovery result handler: discovered DAPI addresses are silently discarded when the target network has no existing config block, which is the exact scenario the feature is meant to address. The first_dapi_host trailing-slash issue is real but currently unreachable due to calling context guards. Several architectural suggestions are valid.

*Reviewed commit: d6d90e5a*

🔴 1 blocking | 🟡 7 suggestion(s)

### 1 additional finding

<details>
<summary>🟡 <b>suggestion:</b> context_for_network silently falls back to mainnet — mutating actions may target wrong network</summary>

`src/ui/network_chooser_screen.rs (lines 219-233)`

`context_for_network()` returns `&mainnet_app_context` for any network without a context (the `_ =>` catch-all on line 231). With Testnet now always selectable, code that calls `context_for_network(Network::Testnet)` without checking context existence will silently operate on mainnet. The discovery handler at line 2265 does check `network_context_exists` before using the context, which is good — but this guard pattern is opt-in and easy to miss in future code. Consider returning `Option<&Arc<AppContext>>` or adding a separate `try_context_for_network` so the compiler enforces handling.

</details>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] lines 2248-2306: Discovered DAPI addresses silently discarded when target network has no config block
  The `if let` chain at lines 2248-2249 requires `config_for_network(network)` to return `Some(...)`. For Testnet — which is now always selectable in the network dropdown — this returns `None` when no `TESTNET_*` vars exist in `.env`. The entire discovery result is silently dropped: `discovery_in_progress` resets to false on line 2245, but no banner is shown and no addresses are saved.

This is the primary use case for the feature: a user selects Testnet, clicks "Refresh DAPI endpoints", discovery succeeds against the DCG endpoint, but the result vanishes with no feedback.

The fix should create a default `NetworkConfig` when the existing one is `None`, then populate it with the discovered addresses. An `else` branch should also handle the `Config::load_from` failure case with an error banner.
- [SUGGESTION] lines 219-233: context_for_network silently falls back to mainnet — mutating actions may target wrong network
  `context_for_network()` returns `&mainnet_app_context` for any network without a context (the `_ =>` catch-all on line 231). With Testnet now always selectable, code that calls `context_for_network(Network::Testnet)` without checking context existence will silently operate on mainnet. The discovery handler at line 2265 does check `network_context_exists` before using the context, which is good — but this guard pattern is opt-in and easy to miss in future code. Consider returning `Option<&Arc<AppContext>>` or adding a separate `try_context_for_network` so the compiler enforces handling.
- [SUGGESTION] line 2291: reinit_core_client_and_sdk called synchronously in display_task_result (UI thread)
  `reinit_core_client_and_sdk()` rebuilds the RPC client and SDK from the UI thread's `display_task_result` path. While `Client::new()` and SDK init are likely fast (no network I/O), they do hold and replace write locks. If SDK initialization ever adds blocking work, the UI would freeze. Consider dispatching as a `BackendTask` for consistency with the app's async architecture.

In `src/spv/manager.rs`:
- [SUGGESTION] lines 1500-1508: first_dapi_host does not strip trailing path/slash from DAPI URLs
  For URLs like `https://45.135.180.114/` (the exact format in `.env.example` mainnet addresses), `first_dapi_host` returns `45.135.180.114/` because it splits on `:` but never strips the path component. This would produce `45.135.180.114/:9999` in `primary_peer_socket()`, which `to_socket_addrs()` cannot resolve.

**Currently unreachable in practice**: `primary_peer_socket()` is only called for Devnet/Regtest (which use `:port` format) or when `use_local_node()` is true (which always sets `core_host` to `127.0.0.1` first). So `first_dapi_host` is never reached with trailing-slash URLs today. However, the function's contract promises to extract a host from any DAPI URL, and the latent bug will surface if any new code path calls it without the `core_host` guard.
- [SUGGESTION] lines 1500-1508: No unit tests for first_dapi_host URL parsing
  `first_dapi_host` manually parses URL strings with multiple edge cases (trailing slashes, missing schemes, ports, empty strings). A few focused unit tests would catch the trailing-slash bug and prevent regressions.

In `src/config.rs`:
- [SUGGESTION] lines 72-78: Regtest default RPC port (19898) doesn't match the app's docs and .env.example (20302)
  `default_rpc_port(Network::Regtest)` returns `19898` (standard Dash Core regtest port), but `.env.example`, `docs/local-network.md`, and all test fixtures use `20302` (dashmate's port). Since this app's setup guide assumes dashmate for local networks, the default is inconsistent. The mismatch only manifests when a user configures regtest without explicitly setting `core_rpc_port`, which the docs/example always include — so practical impact is low. But the default should match the app's own conventions to avoid confusion.

In `src/context/mod.rs`:
- [SUGGESTION] lines 164-184: Missing mainnet DAPI addresses prevent app launch — Refresh button is unreachable
  `AppContext::new()` returns `None` when DAPI addresses are absent (line 177-183), and `AppState::new()` treats a missing mainnet context as fatal (line 257-258). If a user's mainnet addresses are cleared or their `.env` is corrupted, the app won't start and the "Refresh DAPI endpoints" recovery button is unreachable. This is a pre-existing constraint (not introduced by this PR), but worth considering as the discovery feature matures — e.g., allowing the app to start in a degraded state with just the network settings screen.

In `src/backend_task/dapi_discovery.rs`:
- [SUGGESTION] lines 1-135: No unit tests for dapi_discovery module
  The module has no tests. While `try_discover_nodes` requires network access, the error construction paths and `discover_and_format` CSV formatting are testable locally with mocked inputs.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2226,97 @@ impl ScreenLike for NetworkChooserScreen {
...
+        } = backend_task_success_result
+        {
+            self.discovery_in_progress = false;
+
+            // Update config with new addresses
+            let mut config = match Config::load_from(&self.mainnet_app_context.data_dir) {
+                Ok(c) => c,
+                Err(e) => {
+                    MessageBanner::set_global(
+                        self.mainnet_app_context.egui_ctx(),
```
lklimek: why mainnet app context, not current context?

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2226,97 @@ impl ScreenLike for NetworkChooserScreen {
...
+            let mut network_cfg = config
+                .config_for_network(network)
+                .clone()
+                .unwrap_or_default();
+            network_cfg.dapi_addresses = Some(addresses_csv);
+            config.update_config_for_network(network, network_cfg.clone());
+
+            if let Err(e) = config.save(&self.mainnet_app_context.data_dir) {
+                MessageBanner::set_global(
+                    self.mainnet_app_context.egui_ctx(),
```
lklimek: why mainnet app context, not current context?

src/app.rs
```diff
@@ -972,14 +956,29 @@ impl AppState {
...
+                Network::Devnet => "Devnet",
+                Network::Regtest => "Local",
+                _ => "Unknown",
+            };
             tracing::error!(
                 "Cannot switch to {:?}: network context not available. Staying on current network.",
                 network
             );
+            MessageBanner::set_global(
+                self.mainnet_app_context.egui_ctx(),
```
lklimek: why mainnet app context, not current context?

src/ui/network_chooser_screen.rs
lklimek: Good catch — no reason to hardcode `mainnet_app_context` here. `data_dir` and `egui_ctx()` are shared across all network contexts. Replaced with `self.current_app_context()` in `display_task_result()` (commit b4561b0b).

src/ui/network_chooser_screen.rs
lklimek: Fixed — same change, now uses `current_app_context()` for both `data_dir` and `egui_ctx()` throughout `display_task_result()` (commit b4561b0b).

src/app.rs
lklimek: Fixed — `change_network()` now uses `self.current_app_context()` (the still-active network) instead of `mainnet_app_context` for the error banner (commit b4561b0b).

src/context/mod.rs
```diff
@@ -538,38 +563,55 @@ impl AppContext {
...
             &addr,
-            Auth::UserPass(cfg.core_rpc_user.clone(), cfg.core_rpc_password.clone()),
+            Auth::UserPass(
+                cfg.core_rpc_user.clone().unwrap_or_default(),
+                cfg.core_rpc_password.clone().unwrap_or_default(),
+            ),
         )
         .map_err(|e| TaskError::RpcProviderCreationFailed {
             detail: e.to_string(),
         })?;
```
thepastaclaw: **🔴 Blocking: reinit_core_client_and_sdk() drops cookie-based Core RPC auth**

Startup uses `create_core_rpc_client()` (line 640-663), which tries cookie auth first via `Auth::CookieFile` and falls back to `Auth::UserPass` only when the cookie path is unavailable.

`reinit_core_client_and_sdk()` bypasses this and unconditionally builds `Client::new(&addr, Auth::UserPass(...))`. When `core_rpc_user` and `core_rpc_password` are both `None` (the norm for cookie-auth setups), this sends `Auth::UserPass("", "")` — which Core will reject.

Triggered after DAPI discovery completes (line 2302 in network_chooser_screen.rs) or after saving an RPC password. Users on cookie auth will see "reconnection failed" after every DAPI refresh.

<details>

```suggestion
        // 2. Rebuild the RPC client with the new password
        let addr = format!("http://{}:{}", cfg.rpc_host(), cfg.rpc_port(self.network));
        let new_client = Self::create_core_rpc_client(
            &addr,
            self.network,
            &cfg.devnet_name,
            &cfg,
        )?;
```

</details>

<sub>source: ['codex']</sub>

<details>

```

In `src/context/mod.rs`:
- [BLOCKING] lines 565-576: reinit_core_client_and_sdk() drops cookie-based Core RPC auth
  Startup uses `create_core_rpc_client()` (line 640-663), which tries cookie auth first via `Auth::CookieFile` and falls back to `Auth::UserPass` only when the cookie path is unavailable.

`reinit_core_client_and_sdk()` bypasses this and unconditionally builds `Client::new(&addr, Auth::UserPass(...))`. When `core_rpc_user` and `core_rpc_password` are both `None` (the norm for cookie-auth setups), this sends `Auth::UserPass("", "")` — which Core will reject.

Triggered after DAPI discovery completes (line 2302 in network_chooser_screen.rs) or after saving an RPC password. Users on cookie auth will see "reconnection failed" after every DAPI refresh.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2223,97 @@ impl ScreenLike for NetworkChooserScreen {
...
+                );
+                return;
+            }
+
+            let app_context = self.context_for_network(network);
+            {
+                if let Ok(mut cfg_lock) = app_context.config.write() {
+                    *cfg_lock = network_cfg;
+                }
+            }
```
thepastaclaw, line 2304: **🟡 Suggestion: Poisoned config lock silently ignored — success banner shows with stale config**

`if let Ok(mut cfg_lock) = app_context.config.write()` silently discards a poisoned-lock error. If the write fails, `network_cfg` is not applied, but `reinit_core_client_and_sdk()` still runs and reads the old config. The success banner at line 2312 still shows, telling the user addresses were updated when they were not.

Note: the Save-password path at line 529 uses `.unwrap()` on the same lock — inconsistent handling.

A poisoned lock is rare (requires a prior panic), but when it happens the silent success is misleading.

<details>

```suggestion
            match app_context.config.write() {
                Ok(mut cfg_lock) => *cfg_lock = network_cfg,
                Err(e) => {
                    MessageBanner::set_global(
                        self.mainnet_app_context.egui_ctx(),
                        format!("Discovered {count} node addresses but could not apply them. Restart the app."),
                        MessageType::Error,
                    )
                    .with_details(e);
                    return;
                }
            }
```

</details>

<sub>source: ['claude']</sub>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [SUGGESTION] lines 2296-2300: Poisoned config lock silently ignored — success banner shows with stale config
  `if let Ok(mut cfg_lock) = app_context.config.write()` silently discards a poisoned-lock error. If the write fails, `network_cfg` is not applied, but `reinit_core_client_and_sdk()` still runs and reads the old config. The success banner at line 2312 still shows, telling the user addresses were updated when they were not.

Note: the Save-password path at line 529 uses `.unwrap()` on the same lock — inconsistent handling.

A poisoned lock is rare (requires a prior panic), but when it happens the silent success is misleading.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -2190,4 +2223,97 @@ impl ScreenLike for NetworkChooserScreen {
 
         action
     }
+
+    fn display_message(&mut self, _message: &str, message_type: MessageType) {
+        // Only reset discovery state on errors — other message types (success,
+        // info) may be unrelated global banners (theme change, scheduled votes, etc.)
+        if matches!(message_type, MessageType::Error) && self.discovery_in_progress {
+            self.discovery_in_progress = false;
+        }
```
thepastaclaw, line 2232: **🟡 Suggestion: Any unrelated error banner resets discovery in-progress flag**

`display_message()` clears `discovery_in_progress` on every `MessageType::Error`, not just DAPI discovery errors. If an unrelated backend task (RPC timeout, SPV error) completes with an error while DAPI discovery is in flight, the Refresh button re-enables prematurely, allowing the user to spawn a duplicate discovery task.

The comment on line 2228-2229 acknowledges that non-error messages can be unrelated, but doesn't account for unrelated errors. Consider gating the reset on a task-specific flag or only resetting in `display_task_result`.

<sub>source: ['codex']</sub>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [SUGGESTION] lines 2227-2232: Any unrelated error banner resets discovery in-progress flag
  `display_message()` clears `discovery_in_progress` on every `MessageType::Error`, not just DAPI discovery errors. If an unrelated backend task (RPC timeout, SPV error) completes with an error while DAPI discovery is in flight, the Refresh button re-enables prematurely, allowing the user to spawn a duplicate discovery task.

The comment on line 2228-2229 acknowledges that non-error messages can be unrelated, but doesn't account for unrelated errors. Consider gating the reset on a task-specific flag or only resetting in `display_task_result`.
```
</details>

src/backend_task/core/mod.rs
```diff
@@ -492,7 +496,7 @@ impl AppContext {
         if is_rpc_connection_error(e) {
             let url = config
                 .as_ref()
-                .map(|c| format!("{}:{} ({})", c.core_host, c.core_rpc_port, e))
+                .map(|c| format!("{}:{}", c.rpc_host(), c.rpc_port(network)))
                 .unwrap_or_else(|| "unknown".to_string());
             return Some(TaskError::CoreRpcConnectionFailed { url, source: None });
```
thepastaclaw, line 501: **🟡 Suggestion: RPC connection error details lost — source: None discards root cause**

`TaskError::CoreRpcConnectionFailed { url, source: None }` at line 501 discards the original `dashcore_rpc::Error`. The error variant has a `source: Option<Box<dashcore_rpc::Error>>` field specifically for preserving diagnostics (connection refused vs. timeout vs. DNS failure), but it's passed as `None`. The user sees "Could not connect to Dash Core at host:port" with no technical details available in the collapsible panel.

If `dashcore_rpc::Error` implements `Clone`, pass `Some(Box::new(e.clone()))`. Otherwise, consider restructuring to take ownership of the error.

<sub>source: ['claude']</sub>

<details>

```

- [SUGGESTION] lines 496-501: RPC connection error details lost — source: None discards root cause
  `TaskError::CoreRpcConnectionFailed { url, source: None }` at line 501 discards the original `dashcore_rpc::Error`. The error variant has a `source: Option<Box<dashcore_rpc::Error>>` field specifically for preserving diagnostics (connection refused vs. timeout vs. DNS failure), but it's passed as `None`. The user sees "Could not connect to Dash Core at host:port" with no technical details available in the collapsible panel.

If `dashcore_rpc::Error` implements `Clone`, pass `Some(Box::new(e.clone()))`. Otherwise, consider restructuring to take ownership of the error.
```
</details>

src/app.rs
```diff
@@ -238,43 +240,25 @@ impl AppState {
...
+                db.clone(),
+                password_info.clone(),
+                subtasks.clone(),
+                connection_status.clone(),
+                ctx.clone(),
+            )
+        };
+
+        let mainnet_app_context = make_context(Network::Mainnet)
+            .ok_or("Failed to create AppContext for mainnet. Check your Dash configuration.")?;
```
thepastaclaw, line 258: **🟡 Suggestion: Recovery flow unreachable when mainnet DAPI addresses are missing or malformed**

This PR adds DAPI node discovery to recover from stale addresses, but the recovery button is only reachable after successful startup. `AppContext::new()` returns `None` when `dapi_addresses` is `None`, empty, or unparseable (lines 165-183 in context/mod.rs). Since mainnet context is required for startup (line 257-258), a completely broken mainnet DAPI config prevents the app from launching — the user can never reach Refresh DAPI Endpoints.

This is pre-existing behavior, not a regression from this PR. The PR still helps for the common case (stale but parseable addresses). Consider a follow-up that allows startup with a degraded mainnet context when DAPI addresses are missing, enabling the recovery flow for this edge case.

<sub>source: ['codex']</sub>

<details>

```

In `src/app.rs`:
- [SUGGESTION] lines 257-258: Recovery flow unreachable when mainnet DAPI addresses are missing or malformed
  This PR adds DAPI node discovery to recover from stale addresses, but the recovery button is only reachable after successful startup. `AppContext::new()` returns `None` when `dapi_addresses` is `None`, empty, or unparseable (lines 165-183 in context/mod.rs). Since mainnet context is required for startup (line 257-258), a completely broken mainnet DAPI config prevents the app from launching — the user can never reach Refresh DAPI Endpoints.

This is pre-existing behavior, not a regression from this PR. The PR still helps for the common case (stale but parseable addresses). Consider a follow-up that allows startup with a degraded mainnet context when DAPI addresses are missing, enabling the recovery flow for this edge case.
```
</details>

src/backend_task/dapi_discovery.rs
```diff
@@ -0,0 +1,135 @@
...
+        provider.fetch_masternode_addresses(),
+    )
+    .await
+    .map_err(|_| DapiDiscoveryError::Timeout)?
+    .map_err(|source| DapiDiscoveryError::RequestFailed { source })?;
+
+    let mut valid_urls = Vec::new();
+    for url in urls {
+        match DapiAddress::from_str(url.as_str()) {
+            Ok(_) => valid_urls.push(url.to_string()),
```
thepastaclaw, line 105: **💬 Nitpick: Redundant url.to_string() clones an owned String**

`urls` is consumed by the `for url in urls` loop, so `url` is an owned `String`. `url.to_string()` allocates a needless copy. Use `url` directly since ownership is available in the `Ok` branch.

<details>

```suggestion
        Ok(_) => valid_urls.push(url),
```

</details>

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Solid PR adding DAPI node discovery and making NetworkConfig fields optional. Two blocking issues: (1) the network combo box desyncs from AppState when the switch is rejected, silently falling back to mainnet context for screen-local operations, and (2) reinit_core_client_and_sdk() bypasses cookie-based auth used at startup, breaking RPC for cookie-auth setups. Several lower-severity issues around silent error handling and missing user feedback.

*Reviewed commit: 0002edef*

🔴 2 blocking | 🟡 5 suggestion(s) | 💬 1 nitpick(s)

### 2 additional findings

<details>
<summary>🔴 <b>blocking:</b> Network selector desyncs from AppState when switch is rejected, enabling cross-network config corruption</summary>

`src/ui/network_chooser_screen.rs (lines 379-435)`

`selectable_value(&mut self.current_network, Network::Testnet, ...)` mutates `self.current_network` immediately on click. `AppAction::SwitchNetwork` is then dispatched to `AppState::change_network()`, which checks `context_available_for_network()` and returns early with an error banner when the context doesn't exist — **without** updating `AppState::chosen_network`.

The screen now believes it's on Testnet while the app is still on Mainnet. Since `context_for_network()` (line 219-232) silently falls back to `mainnet_app_context` when the requested context is missing, all screen-local operations (Save RPC password at line 527, DAPI result handler at line 2295) that call `context_for_network(self.current_network)` will silently operate on mainnet.

Concrete scenario: User selects Testnet (no context) → screen shows Testnet UI → user enters an RPC password → Save writes it to mainnet's in-memory config and triggers mainnet SDK reinit.

Fix: Don't let `selectable_value` mutate `self.current_network` directly. Use a local variable and only commit the network change after verifying the context exists, or restore `prev_network` (already captured at line 379) when the context is unavailable.

</details>
<details>
<summary>🟡 <b>suggestion:</b> Saving RPC password is a silent no-op when the network has no config block</summary>

`src/ui/network_chooser_screen.rs (lines 489-500)`

The `let Some(network_cfg) = config.config_for_network(self.current_network).clone()` guard (line 492-493) means that if the selected network has no config section in `.env` (e.g. Testnet with no `TESTNET_*` vars), the entire save block is skipped. No file is written, no banner is shown — the user clicks Save and nothing happens.

This is reachable: the combo box now shows Testnet unconditionally, and newly defaulted networks won't have a persisted config block. Add an `else` branch with a user-facing banner explaining that the network needs to be configured first.

</details>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] lines 379-435: Network selector desyncs from AppState when switch is rejected, enabling cross-network config corruption
  `selectable_value(&mut self.current_network, Network::Testnet, ...)` mutates `self.current_network` immediately on click. `AppAction::SwitchNetwork` is then dispatched to `AppState::change_network()`, which checks `context_available_for_network()` and returns early with an error banner when the context doesn't exist — **without** updating `AppState::chosen_network`.

The screen now believes it's on Testnet while the app is still on Mainnet. Since `context_for_network()` (line 219-232) silently falls back to `mainnet_app_context` when the requested context is missing, all screen-local operations (Save RPC password at line 527, DAPI result handler at line 2295) that call `context_for_network(self.current_network)` will silently operate on mainnet.

Concrete scenario: User selects Testnet (no context) → screen shows Testnet UI → user enters an RPC password → Save writes it to mainnet's in-memory config and triggers mainnet SDK reinit.

Fix: Don't let `selectable_value` mutate `self.current_network` directly. Use a local variable and only commit the network change after verifying the context exists, or restore `prev_network` (already captured at line 379) when the context is unavailable.
- [SUGGESTION] lines 2296-2300: Poisoned config lock silently ignored — success banner shows with stale config
  `if let Ok(mut cfg_lock) = app_context.config.write()` silently discards a poisoned-lock error. If the write fails, `network_cfg` is not applied, but `reinit_core_client_and_sdk()` still runs and reads the old config. The success banner at line 2312 still shows, telling the user addresses were updated when they were not.

Note: the Save-password path at line 529 uses `.unwrap()` on the same lock — inconsistent handling.

A poisoned lock is rare (requires a prior panic), but when it happens the silent success is misleading.
- [SUGGESTION] lines 489-500: Saving RPC password is a silent no-op when the network has no config block
  The `let Some(network_cfg) = config.config_for_network(self.current_network).clone()` guard (line 492-493) means that if the selected network has no config section in `.env` (e.g. Testnet with no `TESTNET_*` vars), the entire save block is skipped. No file is written, no banner is shown — the user clicks Save and nothing happens.

This is reachable: the combo box now shows Testnet unconditionally, and newly defaulted networks won't have a persisted config block. Add an `else` branch with a user-facing banner explaining that the network needs to be configured first.
- [SUGGESTION] lines 2227-2232: Any unrelated error banner resets discovery in-progress flag
  `display_message()` clears `discovery_in_progress` on every `MessageType::Error`, not just DAPI discovery errors. If an unrelated backend task (RPC timeout, SPV error) completes with an error while DAPI discovery is in flight, the Refresh button re-enables prematurely, allowing the user to spawn a duplicate discovery task.

The comment on line 2228-2229 acknowledges that non-error messages can be unrelated, but doesn't account for unrelated errors. Consider gating the reset on a task-specific flag or only resetting in `display_task_result`.

In `src/context/mod.rs`:
- [BLOCKING] lines 565-576: reinit_core_client_and_sdk() drops cookie-based Core RPC auth
  Startup uses `create_core_rpc_client()` (line 640-663), which tries cookie auth first via `Auth::CookieFile` and falls back to `Auth::UserPass` only when the cookie path is unavailable.

`reinit_core_client_and_sdk()` bypasses this and unconditionally builds `Client::new(&addr, Auth::UserPass(...))`. When `core_rpc_user` and `core_rpc_password` are both `None` (the norm for cookie-auth setups), this sends `Auth::UserPass("", "")` — which Core will reject.

Triggered after DAPI discovery completes (line 2302 in network_chooser_screen.rs) or after saving an RPC password. Users on cookie auth will see "reconnection failed" after every DAPI refresh.

- [SUGGESTION] lines 496-501: RPC connection error details lost — source: None discards root cause
  `TaskError::CoreRpcConnectionFailed { url, source: None }` at line 501 discards the original `dashcore_rpc::Error`. The error variant has a `source: Option<Box<dashcore_rpc::Error>>` field specifically for preserving diagnostics (connection refused vs. timeout vs. DNS failure), but it's passed as `None`. The user sees "Could not connect to Dash Core at host:port" with no technical details available in the collapsible panel.

If `dashcore_rpc::Error` implements `Clone`, pass `Some(Box::new(e.clone()))`. Otherwise, consider restructuring to take ownership of the error.

In `src/app.rs`:
- [SUGGESTION] lines 257-258: Recovery flow unreachable when mainnet DAPI addresses are missing or malformed
  This PR adds DAPI node discovery to recover from stale addresses, but the recovery button is only reachable after successful startup. `AppContext::new()` returns `None` when `dapi_addresses` is `None`, empty, or unparseable (lines 165-183 in context/mod.rs). Since mainnet context is required for startup (line 257-258), a completely broken mainnet DAPI config prevents the app from launching — the user can never reach Refresh DAPI Endpoints.

This is pre-existing behavior, not a regression from this PR. The PR still helps for the common case (stale but parseable addresses). Consider a follow-up that allows startup with a degraded mainnet context when DAPI addresses are missing, enabling the recovery flow for this edge case.
```
</details>

src/context/mod.rs
lklimek: Fixed — `reinit_core_client_and_sdk()` now reuses `create_core_rpc_client()` which tries cookie auth first, falling back to user/pass (commit aa2f0bb7).

thepastaclaw (commented):
## Code Review

Clean incremental addressing two prior review findings. The cookie auth regression is fully resolved by delegating to `create_core_rpc_client()`. The hardcoded `mainnet_app_context` references in the DAPI discovery handler are replaced with `current_app_context()`. The network selector desync (prior blocking finding about `selectable_value` mutating `current_network` before `AppState` confirms the switch) is not yet addressed in this push.

*Reviewed commit: aa2f0bb7*

🟡 1 suggestion(s)

### 1 additional finding

<details>
<summary>🟡 <b>suggestion:</b> Network selector desync still present (prior blocking finding)</summary>

`src/ui/network_chooser_screen.rs (lines 379-420)`

The prior review flagged that `selectable_value(&mut self.current_network, ...)` mutates `self.current_network` immediately on click, before `AppState::change_network()` validates whether the target context exists. If the switch is rejected, the screen believes it's on the new network while the app stays on the old one.

This push fixes the hardcoded `mainnet_app_context` references (good — reduces impact of the desync), but the root cause remains: `self.current_network` is still mutated by the combo box before validation. The `context_for_network()` fallback-to-mainnet still applies to all screen-local operations when the selected network has no context.

The fix from the prior review still applies: use a local variable for the combo box, and only commit the change to `self.current_network` after confirming the context exists (or restore `prev_network` on rejection).

</details>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [SUGGESTION] lines 379-420: Network selector desync still present (prior blocking finding)
  The prior review flagged that `selectable_value(&mut self.current_network, ...)` mutates `self.current_network` immediately on click, before `AppState::change_network()` validates whether the target context exists. If the switch is rejected, the screen believes it's on the new network while the app stays on the old one.

This push fixes the hardcoded `mainnet_app_context` references (good — reduces impact of the desync), but the root cause remains: `self.current_network` is still mutated by the combo box before validation. The `context_for_network()` fallback-to-mainnet still applies to all screen-local operations when the selected network has no context.

The fix from the prior review still applies: use a local variable for the combo box, and only commit the change to `self.current_network` after confirming the context exists (or restore `prev_network` on rejection).
```
</details>

PR #781, opened 2026-03-20
fix(ui): throttle system theme detection to prevent white flash

lklimek:
## Summary

- Throttle `dark_light::detect()` to once every 2 seconds when using System theme — prevents transient OS detection glitches from causing brief Light/Dark flips (visible as white flashes during high-frequency repaints like SPV sync)
- Cache resolved theme in `AppState` and only reapply when it actually changes (skip per-frame `apply_theme` when theme is unchanged)
- Use cached `resolved_theme` consistently — including the shutdown branch, which previously bypassed the throttle
- Replace process-global `FONTS_INITIALIZED` static with a per-`egui::Context` guard so fonts are correctly initialized for each Context instance (tests, multiple viewports)
- Use non-persisted (`temp`) storage for the font-init flag — persisted storage caused fonts to be skipped on app restart because the flag survived in `app.ron`

## Root Cause

`apply_theme()` was called on every frame. For System theme mode, each call queried the OS via `dark_light::detect()`. During rapid repaints (e.g. SPV sync progress), the detection could transiently return the wrong value → momentary theme flip → white flash.

## Test plan

- [ ] Start app with System color theme in a dark OS environment
- [ ] Trigger SPV sync and observe — no white flashes
- [ ] Switch theme preference (Light → Dark → System) — changes apply immediately
- [ ] Verify Light and Dark explicit modes still work correctly (no throttling needed)
- [ ] Verify shutdown banner renders with correct theme (no flash)
- [ ] Kill and restart app — verify custom Dash fonts load correctly (not default egui fonts)

## Summary by CodeRabbit

* **Refactor**
  * Reduce redundant theme reapplications so the UI updates only when the active theme changes
  * Throttle system-theme detection to avoid rapid toggling and unnecessary checks
  * Ensure the shutdown screen preserves the selected theme and applies preference changes immediately

* **Behavior Change**
  * Stop altering font definitions during theme updates to avoid resetting fonts unexpectedly

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1104,9 +1104,10 @@ impl App for AppState {
             let now = Instant::now();
             if now.duration_since(self.theme_last_checked) >= Duration::from_secs(2) {
                 self.theme_last_checked = now;
-                let new_resolved = crate::ui::theme::resolve_theme_mode(self.theme_preference);
-                if new_resolved != self.resolved_theme {
-                    self.resolved_theme = new_resolved;
+                if let Some(detected) = crate::ui::theme::try_detect_system_theme()
+                    && detected != self.resolved_theme
+                {
+                    self.resolved_theme = detected;
                 }
             }
         }
--- a/src/app.rs
+++ b/src/app.rs
@@ -87,6 +87,8 @@ pub struct AppState {
     pub task_result_sender: egui_mpsc::SenderAsync<TaskResult>, // Channel sender for sending task results
     pub task_result_receiver: tokiompsc::Receiver<TaskResult>, // Channel receiver for receiving task results
     pub theme_preference: ThemeMode,                           // Current theme preference
+    resolved_theme: ThemeMode, // Cached resolved theme (Light/Dark, never System)
+    theme_last_checked: Instant, // Last time we polled the OS for system theme
     last_scheduled_vote_check: Instant, // Last time we checked if there are scheduled masternode votes to cast
     last_repaint_request: Instant,      // Throttle periodic repaint scheduling to once per second
     pub subtasks: Arc<TaskManager>,     // Subtasks manager for graceful shutdown
@@ -720,6 +722,8 @@ impl AppState {
             core_message_receiver,
             task_result_sender,
             task_result_receiver,
+            resolved_theme: crate::ui::theme::resolve_theme_mode(theme_preference),
+            theme_last_checked: Instant::now(),
             theme_preference,
             last_scheduled_vote_check: Instant::now(),
             last_repaint_request: Instant::now(),
@@ -1089,8 +1093,19 @@ impl App for AppState {
             return;
         }
 
-        // Apply Dash theme with user preference
-        crate::ui::theme::apply_theme(ctx, self.theme_preference);
+        // Throttle OS theme detection to every 2 s to prevent white flash from
+        // transient dark_light::detect() glitches during high-frequency repaints.
+        if self.theme_preference == ThemeMode::System {
+            let now = Instant::now();
+            if now.duration_since(self.theme_last_checked) > Duration::from_secs(2) {
+                self.theme_last_checked = now;
+                let new_resolved = crate::ui::theme::resolve_theme_mode(self.theme_preference);
+                if new_resolved != self.resolved_theme {
+                    self.resolved_theme = new_resolved;
+                }
+            }
+        }
+        crate::ui::theme::apply_theme(ctx, self.resolved_theme);
 
         self.enforce_network_context_invariant();
         let active_context = self.current_app_context().clone();
@@ -1132,6 +1147,9 @@ impl App for AppState {
                         }
                         BackendTaskSuccessResult::UpdatedThemePreference(new_theme) => {
                             self.theme_preference = new_theme;
+                            self.resolved_theme = crate::ui::theme::resolve_theme_mode(new_theme);
+                            self.theme_last_checked = Instant::now();
+                            crate::ui::theme::apply_theme(ctx, self.resolved_theme);
                             MessageBanner::set_global(
                                 ctx,
                                 "Theme preference updated successfully",
--- a/src/app.rs
+++ b/src/app.rs
@@ -1102,7 +1102,7 @@ impl App for AppState {
         // transient dark_light::detect() glitches during high-frequency repaints.
         if self.theme_preference == ThemeMode::System {
             let now = Instant::now();
-            if now.duration_since(self.theme_last_checked) > Duration::from_secs(2) {
+            if now.duration_since(self.theme_last_checked) >= Duration::from_secs(2) {
                 self.theme_last_checked = now;
                 let new_resolved = crate::ui::theme::resolve_theme_mode(self.theme_preference);
                 if new_resolved != self.resolved_theme {
@@ -1157,8 +1157,6 @@ impl App for AppState {
                             self.theme_preference = new_theme;
                             self.resolved_theme = crate::ui::theme::resolve_theme_mode(new_theme);
                             self.theme_last_checked = Instant::now();
-                            // Force immediate re-apply even if resolved colour didn't change.
-                            self.last_applied_theme = None;
                             crate::ui::theme::apply_theme(ctx, self.resolved_theme);
                             self.last_applied_theme = Some(self.resolved_theme);
                             MessageBanner::set_global(
--- a/src/app.rs
+++ b/src/app.rs
@@ -1157,20 +1157,34 @@ impl App for AppState {
                         }
                         BackendTaskSuccessResult::UpdatedThemePreference(new_theme) => {
                             self.theme_preference = new_theme;
+                            let mut detection_failed = false;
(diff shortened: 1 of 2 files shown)


thepastaclaw (commented):
## Code Review

Clean, well-targeted throttle fix. All callers correctly pass pre-resolved themes so the system works as intended. Findings are valid but none are blocking — the most actionable is the `>` vs `>=` off-by-one that stretches the effective throttle from 2s to ~3s at 1s repaint intervals.

*Reviewed commit: 3e84e77c*

🟡 4 suggestion(s) | 💬 1 nitpick(s)

### 2 additional findings

<details>
<summary>🟡 <b>suggestion:</b> `apply_theme` internally re-resolves via `resolve_theme_mode` — redundant for current callers and latent throttle bypass</summary>

`src/ui/theme.rs (lines 1035-1038)`

All callers now pass a pre-resolved `Light` or `Dark` value (never `System`), so `resolve_theme_mode` at line 1037 is a no-op. However, `apply_theme` is `pub` and still accepts `ThemeMode::System`, meaning a future caller could bypass the throttle by calling `apply_theme(ctx, ThemeMode::System)` directly. Consider either:
1. Adding a debug assertion that `theme_mode != ThemeMode::System`, or
2. Changing the parameter type to only accept resolved themes.

Not urgent since all current callers are correct, but the throttling invariant is enforced by caller discipline rather than the type system.

</details>
<details>
<summary>🟡 <b>suggestion:</b> `resolve_theme_mode` silently swallows detection errors</summary>

`src/ui/theme.rs (lines 25-29)`

Line 27 uses `.unwrap_or(ThemeMode::Light)` which discards the error from `detect_system_theme()` with no logging. If theme detection is failing persistently (e.g., dbus unavailable on Linux), the user would be stuck on Light with no diagnostic trail. A `tracing::warn!` on the error path would help.

<details>

```suggestion
pub fn resolve_theme_mode(preference: ThemeMode) -> ThemeMode {
    match preference {
        ThemeMode::System => detect_system_theme().unwrap_or_else(|e| {
            tracing::warn!("System theme detection failed, defaulting to Light: {e}");
            ThemeMode::Light
        }),
        other => other,
    }
}
```

</details>

</details>

<details>

```

In `src/app.rs`:
- [SUGGESTION] line 1105: Strict `>` stretches the 2-second throttle to ~3 seconds at idle
  With `request_repaint_after(Duration::from_secs(1))` at line 1233, the frame ticks are roughly 1s apart. Using strict `>` means:
- t=0: `theme_last_checked` set
- t=1: `1s > 2s` → false
- t=2: `2s > 2s` → false
- t=3: `3s > 2s` → true

So the effective interval is ~3s, not the documented 2s. Using `>=` aligns behavior with the comment.
- [SUGGESTION] lines 1101-1116: Throttle and cache logic has no automated tests
  The throttle/cache state machine (three coordinated fields: `resolved_theme`, `last_applied_theme`, `theme_last_checked`) has no unit or integration test coverage. All four review agents flagged this independently. Given this is fixing a user-visible regression (white flash), a regression test would be valuable — though the tight coupling to egui's `Context` makes this harder to test in isolation.

In `src/ui/theme.rs`:
- [SUGGESTION] lines 1035-1038: `apply_theme` internally re-resolves via `resolve_theme_mode` — redundant for current callers and latent throttle bypass
  All callers now pass a pre-resolved `Light` or `Dark` value (never `System`), so `resolve_theme_mode` at line 1037 is a no-op. However, `apply_theme` is `pub` and still accepts `ThemeMode::System`, meaning a future caller could bypass the throttle by calling `apply_theme(ctx, ThemeMode::System)` directly. Consider either:
1. Adding a debug assertion that `theme_mode != ThemeMode::System`, or
2. Changing the parameter type to only accept resolved themes.

Not urgent since all current callers are correct, but the throttling invariant is enforced by caller discipline rather than the type system.
- [SUGGESTION] lines 25-29: `resolve_theme_mode` silently swallows detection errors
  Line 27 uses `.unwrap_or(ThemeMode::Light)` which discards the error from `detect_system_theme()` with no logging. If theme detection is failing persistently (e.g., dbus unavailable on Linux), the user would be stuck on Light with no diagnostic trail. A `tracing::warn!` on the error path would help.
```
</details>

src/app.rs
```diff
@@ -1089,8 +1098,22 @@ impl App for AppState {
             return;
         }
 
-        // Apply Dash theme with user preference
-        crate::ui::theme::apply_theme(ctx, self.theme_preference);
+        // Throttle OS theme detection to every 2 s to prevent white flash from
+        // transient dark_light::detect() glitches during high-frequency repaints.
+        if self.theme_preference == ThemeMode::System {
+            let now = Instant::now();
+            if now.duration_since(self.theme_last_checked) > Duration::from_secs(2) {
```
thepastaclaw: **🟡 Suggestion: Strict `>` stretches the 2-second throttle to ~3 seconds at idle**

With `request_repaint_after(Duration::from_secs(1))` at line 1233, the frame ticks are roughly 1s apart. Using strict `>` means:
- t=0: `theme_last_checked` set
- t=1: `1s > 2s` → false
- t=2: `2s > 2s` → false
- t=3: `3s > 2s` → true

So the effective interval is ~3s, not the documented 2s. Using `>=` aligns behavior with the comment.

<details>

```suggestion
            if now.duration_since(self.theme_last_checked) >= Duration::from_secs(2) {
```

</details>

<sub>source: ['codex']</sub>

<details>

```

In `src/app.rs`:
- [SUGGESTION] line 1105: Strict `>` stretches the 2-second throttle to ~3 seconds at idle
  With `request_repaint_after(Duration::from_secs(1))` at line 1233, the frame ticks are roughly 1s apart. Using strict `>` means:
- t=0: `theme_last_checked` set
- t=1: `1s > 2s` → false
- t=2: `2s > 2s` → false
- t=3: `3s > 2s` → true

So the effective interval is ~3s, not the documented 2s. Using `>=` aligns behavior with the comment.
```
</details>

src/app.rs
```diff
@@ -1132,6 +1155,12 @@ impl App for AppState {
                         }
                         BackendTaskSuccessResult::UpdatedThemePreference(new_theme) => {
                             self.theme_preference = new_theme;
+                            self.resolved_theme = crate::ui::theme::resolve_theme_mode(new_theme);
+                            self.theme_last_checked = Instant::now();
+                            // Force immediate re-apply even if resolved colour didn't change.
+                            self.last_applied_theme = None;
+                            crate::ui::theme::apply_theme(ctx, self.resolved_theme);
+                            self.last_applied_theme = Some(self.resolved_theme);
```
thepastaclaw: **💬 Nitpick: Dead assignment: `last_applied_theme = None` immediately overwritten**

Line 1161 sets `last_applied_theme` to `None`, then line 1162 calls `apply_theme`, and line 1163 immediately sets it to `Some(...)`. Nothing reads the field between lines 1161 and 1163, and `apply_theme` does not access `self.last_applied_theme`. The `None` assignment serves no purpose — the explicit `apply_theme` call on 1162 already forces re-application regardless.

<details>

```suggestion
                    // Force immediate re-apply even if resolved colour didn't change.
                    crate::ui::theme::apply_theme(ctx, self.resolved_theme);
                    self.last_applied_theme = Some(self.resolved_theme);
```

</details>

<sub>source: ['claude']</sub>

src/app.rs
```diff
@@ -1089,8 +1098,22 @@ impl App for AppState {
...
+                let new_resolved = crate::ui::theme::resolve_theme_mode(self.theme_preference);
+                if new_resolved != self.resolved_theme {
+                    self.resolved_theme = new_resolved;
+                }
+            }
+        }
+        if self.last_applied_theme != Some(self.resolved_theme) {
+            crate::ui::theme::apply_theme(ctx, self.resolved_theme);
+            self.last_applied_theme = Some(self.resolved_theme);
+        }
```
thepastaclaw, line 1118: **🟡 Suggestion: Throttle and cache logic has no automated tests**

The throttle/cache state machine (three coordinated fields: `resolved_theme`, `last_applied_theme`, `theme_last_checked`) has no unit or integration test coverage. All four review agents flagged this independently. Given this is fixing a user-visible regression (white flash), a regression test would be valuable — though the tight coupling to egui's `Context` makes this harder to test in isolation.

<details>

```

In `src/app.rs`:
- [SUGGESTION] lines 1101-1116: Throttle and cache logic has no automated tests
  The throttle/cache state machine (three coordinated fields: `resolved_theme`, `last_applied_theme`, `theme_last_checked`) has no unit or integration test coverage. All four review agents flagged this independently. Given this is fixing a user-visible regression (white flash), a regression test would be valuable — though the tight coupling to egui's `Context` makes this harder to test in isolation.
```
</details>

Claudius-Maginificent: Acknowledged — the throttle/cache state machine is tightly coupled to `egui::Context`, making isolated unit tests impractical without significant refactoring. The logic is straightforward (three fields, one timer check) and the fix is verified by manual testing. Deferring automated test coverage for now.

PR #786, opened 2026-03-23
feat: AddressInput component, RPC error handling, shielded retry/sync improvements

lklimek:
## Summary

Adds a reusable `AddressInput` component, improves RPC error handling, and refines shielded pool operations with retry logic and post-operation syncing.

### New: AddressInput Component
- **Unified address input** (`src/ui/components/address_input.rs`) with autocomplete, type detection (Core/Platform/Shielded/Identity), DPNS resolution, network validation, and balance display
- **Domain types** (`src/model/address.rs`) — `AddressKind` enum and `ValidatedAddress` type for address classification and validation
- **UI Components README** (`src/ui/components/README.md`) — catalog of all reusable components
- Migrated `WalletSendScreen`, `UnshieldCreditsScreen`, and mine dialog to use `AddressInput`

### New: AmountInput Migration
- `ShieldCreditsScreen`, `ShieldFromAssetLockScreen`, `ShieldedSendScreen`, and `UnshieldCreditsScreen` replaced hand-rolled amount parsing with the `AmountInput` component

### RPC Error Handling
- **Connection errors include host:port** — dedicated `CoreRpcConnectionFailed` variant with `is_rpc_connection_error()` detector; `rpc_error_with_url()` enriches connection failures across all RPC call sites
- **RPC error visibility** — chain lock errors surfaced on Networks tab as "Error: {message}" instead of bare "Disconnected"; `rpc_last_error` field added to `ConnectionStatus`
- **Network password save** — RPC password input works for all networks (was hardcoded to Regtest); stale error banners cleared on reconnect; warning banner shown if config save fails; ZMQ not disconnected on save
- **Config file permissions** — `chmod 0600` applied to `.env` on Unix after save (contains RPC credentials)

### Shielded Pool Improvements
- **Anchor retry helper** — `with_anchor_retry()` extracts shared state management (take/run/mark-spent/put-back) from transfer, unshield, and withdrawal tasks, with automatic retry on `ShieldedAnchorMismatch`
- **Post-operation note sync** — all shielded screens queue a `SyncNotes` task after successful operations so balances update without manual refresh
- **Safety-net resync** — queries all notes (including spent) to avoid false-positive resync triggers; propagates DB errors instead of silently swallowing
- **Commitment tree clearing** — refactored to check table existence before DELETE; moved out of network-switch transaction

### Error Message Improvements
- `IdentityInsufficientBalance` shows amounts in DASH instead of raw credit integers
- `AssetLockOutPointInsufficientBalance` shows amounts in DASH with actionable guidance
- `ShieldedFeeExceedsBalance` shows fee in DASH and suggests reducing amount
- `ShieldedInsufficientPoolNotes` uses plain language ("network needs more activity")
- `ShieldedAnchorMismatch` says "wallet data is slightly outdated"
- `ShieldedAddressInsufficientFunds` shows available/required in DASH
- New dedicated variants: `WalletUtxoReloadFailed`, `WalletBalanceRecalculationFailed`
- `WalletKeyDerivationFailed` now preserves error chain via `#[source]`

### Other Changes
- **Consolidated `format_credits_as_dash`** — uses `Amount::dash_from_credits()` with trimmed trailing zeros
- **`networks_address_compatible()`** — now `pub(crate)` in `model/wallet/mod.rs`, used in wallet dialogs so Testnet/Devnet/Regtest addresses are accepted interchangeably
- **SPV balance tracking** — `spv_balance_known` flag distinguishes "not yet synced" (None) from "synced with zero balance" (Some(0))
- **Funding method dropdown** — ComboBox height set to 200px so all options display without scrollbar
- **Wallet switch refresh** — switching wallets in RPC mode triggers a Core refresh
- **Transaction filtering** — transactions list filtered to only show entries involving the selected wallet's addresses, with defensive stale-Arc check
- **Zero-balance auto-show** — new wallets with <5 addresses automatically show all addresses
- **Database** — `shared_connection()` narrowed to `pub(crate)`; `TransactionStatus::from_u8` logs unknown values
- **DB migration tests** — regression tests for v27->v33 migration and fresh v33 install
- **Tracing** — detailed logging added to all shielded bundle operations; log levels corrected per coding best practices

## Test plan

- [ ] Switch between Mainnet/Testnet/Regtest, change RPC password on each — verify save works and ZMQ stays connected
- [ ] Stop Dash Core, trigger an RPC call — verify banner shows `host:port` in error; details visible only in developer mode
- [ ] Enter wrong RPC password — verify "check your RPC password" message appears; fix password and verify error clears
- [ ] Save RPC password when config file is read-only — verify warning banner ("changes apply for this session only")
- [ ] Top Up Identity > "Select funding method" dropdown shows all options without scrollbar
- [ ] Attempt state transition with insufficient identity balance — verify message shows available/required in DASH (e.g., "0.5 DASH") not raw integers
- [ ] Attempt unshield with amount that leaves no room for fee — verify message shows fee in DASH and suggests reducing amount
- [ ] Attempt unshield when pool has fewer notes than minimum — verify plain-language message (no "notes" or "pool" jargon)
- [ ] On Regtest, paste a `tdash`-prefixed address in wallet send dialog — verify it is accepted
- [ ] Open Wallets screen — verify Balances and Shielded tabs render correctly
- [ ] Set DB version to 27 manually, restart app — verify migrations run cleanly to v33 (idempotent)
- [ ] Fresh install — verify DB created at v33 with all tables present
- [ ] On a regtest node with no chain locks, check Networks tab — verify RPC shows "Error: ..." instead of bare "Disconnected"
- [ ] Connect to a node with chain locks — verify RPC shows "Connected" and error clears
- [ ] **Send screen**: type a partial address — verify autocomplete dropdown shows matching wallet addresses with balances and type badges
- [ ] **Send screen**: switch source (Core/Platform/Shielded) — verify allowed address types change and address input resets
- [ ] **Send screen**: paste wrong-network address — verify validation error appears
- [ ] **Unshield screen**: enter a platform address — verify "Platform address" indicator; enter a Core address — verify "Core address" indicator
- [ ] **Mine dialog**: verify address selector shows wallet addresses with balances
- [ ] **Shielded transfer**: after successful send, verify balance updates without manual refresh (post-op sync)
- [ ] **Wallet switch**: select a different wallet — verify transactions refresh automatically in RPC mode
- [ ] **New wallet**: create a wallet with no funds — verify addresses are shown (not blank)

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added `AddressInput` component with address validation, autocomplete, and support for Core, Platform, Shielded, and Identity address types.
  * Improved RPC error detection and display—connection failures now surface meaningful error messages to users.

* **Bug Fixes**
  * Fixed balance state tracking with explicit `spv_balance_known` flag to prevent premature balance display.
  * Improved transaction filtering for selected wallets to prevent display of unrelated transactions.

* **Improvements**
  * Enhanced error messages for wallet and shielded operations.
  * Consolidated address and amount input handling across send screens using structured components instead of string parsing.
  * Improved database schema migrations and commitment tree management.
  * Better balance formatting using standardized DASH display.

* **Documentation**
  * Added UI Components Catalog documenting available UI components and usage patterns.


src/backend_task/error.rs
```diff
@@ -698,6 +761,92 @@ pub fn is_instant_lock_proof_invalid(error: &SdkError) -> bool {
     )
 }
 
+/// Format a credit amount as Dash with 4 decimal places.
+///
+/// Credits use 10^11 as the conversion factor (not satoshis).
+fn format_credits_as_dash(credits: u64) -> String {
```
lklimek: Check if the Amount struct doesn't implement this already; if it does, use it. If not, extend Amount type.

src/context/shielded.rs
```diff
@@ -216,23 +216,50 @@ impl AppContext {
...
+                            block_height: row.block_height,
+                            is_spent: false,
+                            value: row.value,
+                        });
+                    }
                 }
+                state.recalculate_balance();
+            }
+            Err(e) => {
+                tracing::warn!("Failed to load shielded notes from DB (table may be missing): {e}");
```
lklimek: why not return error to the caller?

src/context/shielded.rs
```diff
@@ -371,6 +425,33 @@ impl AppContext {
         )
         .await;
 
+        // On anchor mismatch, sync notes once and retry
```
lklimek: Code duplication. Extract helper.

src/app.rs
```diff
@@ -1244,11 +1244,20 @@ impl App for AppState {
                     if !handled {
                         let msg = err.to_string();
                         let handle = MessageBanner::set_global(ctx, &msg, MessageType::Error);
-                        if self.current_app_context().is_developer_mode() {
+                        // Always show details for RPC connection errors (users
```
lklimek: Variants should display all required information in Display (#[error]), not in details.

lklimek:
@thepastaclaw claw please check if some comments from PR 789 are already fixed in this PR. If not, comment on PR 786. Mark these comments on 789 as resolved.

thepastaclaw:
## Cross-check: PR #789 review comments vs this PR

lklimek asked me to check which review comments from PR #789 are already addressed here. Summary:

### ✅ Already fixed in this PR

1. **🔴 Duplicate method definitions** (`create_shielded_tables`, `create_shielded_wallet_meta_table`, `add_nullifier_sync_timestamp_column`) — Fixed in commit 08495bf by removing the duplicates from `initialization.rs` (they live in `shielded.rs` on this branch).
2. **💬 Doc comment misplacement** (migration 29 doc on migration 30 function) — Moot since the methods were removed.

### ❌ Still present in this PR (suggestions/nitpicks, non-blocking)

3. **🟡 Schema DEFAULT mismatch** — `add_wallet_transaction_status_column` uses `DEFAULT 2` (Confirmed) for the migration, but the fresh `CREATE TABLE` in `wallet.rs` uses `DEFAULT 0` (Unconfirmed). Harmless if all INSERT paths specify `status` explicitly, but a latent inconsistency.

4. **💬 `rename_network_dash_to_mainnet` error wrapping** — Uses `rusqlite::Error::InvalidParameterName` to wrap migration UPDATE failures, which is semantically unrelated. Minor code smell.

5. **🟡 No test for v33 migration path** — The consolidated v33 migration (7 idempotent sub-steps) has no regression test starting from a v27/v28/v31 database shape.

### N/A — Only in PR #789 (not shared code)

- `spv_confirmed_balance` zero-balance ambiguity
- `WalletKeyDerivationFailed` using `String` instead of typed `#[source]`
- `TransactionStatus::from_u8` silent fallback

Items 3–5 above are suggestions/nitpicks — not blocking for either PR.

src/context/shielded.rs
```diff
@@ -317,113 +345,134 @@ impl AppContext {
...
+            .await
+            .map_err(|e| {
+                tracing::warn!("Note sync after anchor mismatch failed: {e}");
+                e
+            })?;
+            state.last_notes_synced_at = Some(std::time::Instant::now());
+            operation(&state).await
+        } else {
+            result
+        };
```
thepastaclaw, line 482: **🔴 Blocking: with_anchor_retry permanently leaks shielded state on sync_notes failure**

When `with_anchor_retry` detects `ShieldedAnchorMismatch`, it calls `sync_notes()` (line 443). If that sync fails, the `?` at line 453 causes an early return — but the shielded wallet state was `remove()`d from the `shielded_states` map at line 434 and is never re-inserted. This permanently orphans the wallet's shielded state until the app restarts.

The sibling method `sync_shielded_notes()` (lines 275–307) handles this correctly: it always re-inserts state at line 298–301 before propagating errors at line 303. The same guard pattern must be applied here.

All callers — `shielded_transfer_task`, `unshield_credits_task`, and `shielded_withdrawal_task` — are affected.

<details>

```suggestion
    let result = if matches!(result, Err(TaskError::ShieldedAnchorMismatch { .. })) {
            tracing::info!(
                "Shielded anchor mismatch during {operation_name} — syncing notes and retrying"
            );
            match crate::backend_task::shielded::sync::sync_notes(
                self,
                seed_hash,
                &mut state,
                self.network,
            )
            .await
            {
                Ok(_) => {
                    state.last_notes_synced_at = Some(std::time::Instant::now());
                    operation(&state).await
                }
                Err(e) => {
                    tracing::warn!("Note sync after anchor mismatch failed: {e}");
                    Err(e)
                }
            }
        } else {
            result
        };
```

</details>

<sub>source: unknown</sub>

<details>

```

In `src/context/shielded.rs`:
- [BLOCKING] lines 439-458: with_anchor_retry permanently leaks shielded state on sync_notes failure
  When `with_anchor_retry` detects `ShieldedAnchorMismatch`, it calls `sync_notes()` (line 443). If that sync fails, the `?` at line 453 causes an early return — but the shielded wallet state was `remove()`d from the `shielded_states` map at line 434 and is never re-inserted. This permanently orphans the wallet's shielded state until the app restarts.

The sibling method `sync_shielded_notes()` (lines 275–307) handles this correctly: it always re-inserts state at line 298–301 before propagating errors at line 303. The same guard pattern must be applied here.

All callers — `shielded_transfer_task`, `unshield_credits_task`, and `shielded_withdrawal_task` — are affected.
```
</details>

src/ui/network_chooser_screen.rs
```diff
@@ -472,40 +481,76 @@ impl NetworkChooserScreen {
...
+                                );
                             }
+                        } else {
+                            MessageBanner::set_global(
+                                ui.ctx(),
+                                "Core RPC password saved successfully.",
+                                MessageType::Success,
+                            );
                         }
+                        } // else: config.save() succeeded
```
thepastaclaw: **🔴 Blocking: Config save failure path never applies password for current session**

When `config.save()` fails (line 494), the banner says "Your changes will apply for this session only." However, the in-memory config update (`*cfg_lock = updated_config` at line 521) and `reinit_core_client_and_sdk()` (line 527) are both inside the `else` branch (lines 501–553), which only runs on save *success*. So on the failure path, neither the file nor the live RPC client is updated — the password change is silently discarded despite the message promising session-level application.

This is a logic regression introduced by commit da4814a2. The in-memory update and reinit should run unconditionally; only the success/warning banner should be conditional on save outcome.

<sub>source: unknown</sub>

<details>

```

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] lines 494-553: Config save failure path never applies password for current session
  When `config.save()` fails (line 494), the banner says "Your changes will apply for this session only." However, the in-memory config update (`*cfg_lock = updated_config` at line 521) and `reinit_core_client_and_sdk()` (line 527) are both inside the `else` branch (lines 501–553), which only runs on save *success*. So on the failure path, neither the file nor the live RPC client is updated — the password change is silently discarded despite the message promising session-level application.

This is a logic regression introduced by commit da4814a2. The in-memory update and reinit should run unconditionally; only the success/warning banner should be conditional on save outcome.
```
</details>

src/backend_task/shielded/bundle.rs
```diff
@@ -483,7 +472,7 @@ pub async fn shield_from_asset_lock(
 
         wallet
             .recalculate_affected_address_balances(&used_utxos, app_context.as_ref())
-            .map_err(|e| TaskError::ShieldedTransitionBuildFailed { detail: e })?;
+            .map_err(shielded_build_error)?;
     }
```
thepastaclaw, line 531: **🟡 Suggestion: Shield-from-asset-lock skips DB persistence of asset lock before broadcast**

The existing asset lock flow uses `broadcast_and_commit_asset_lock()` (transaction_processing.rs:130–136) which stores the asset lock in the DB *before* broadcast and removes it on failure. The new shield-from-asset-lock path broadcasts directly via `core_client.send_raw_transaction` (line 447) without any DB persistence.

If the app crashes after broadcast but before the shielded state transition completes, the asset lock funds have left the wallet UTXO set but there's no DB record for the standard recovery path to reuse. `recover_asset_locks()` can find unspent asset locks on-chain via `list_unspent`, so funds aren't permanently lost, but the automatic recovery guarantee is weaker than the established pattern provides.

<sub>source: unknown</sub>

<details>

```

In `src/backend_task/shielded/bundle.rs`:
- [SUGGESTION] lines 433-476: Shield-from-asset-lock skips DB persistence of asset lock before broadcast
  The existing asset lock flow uses `broadcast_and_commit_asset_lock()` (transaction_processing.rs:130–136) which stores the asset lock in the DB *before* broadcast and removes it on failure. The new shield-from-asset-lock path broadcasts directly via `core_client.send_raw_transaction` (line 447) without any DB persistence.

If the app crashes after broadcast but before the shielded state transition completes, the asset lock funds have left the wallet UTXO set but there's no DB record for the standard recovery path to reuse. `recover_asset_locks()` can find unspent asset locks on-chain via `list_unspent`, so funds aren't permanently lost, but the automatic recovery guarantee is weaker than the established pattern provides.
```
</details>

src/context/shielded.rs
```diff
@@ -216,23 +217,50 @@ impl AppContext {
...
+        // We check ALL notes (including spent) to avoid a false positive when
+        // the user legitimately spent everything.
+        if state.last_synced_index > 0 && state.notes.is_empty() {
+            let all_notes = self.db.get_all_shielded_notes(&seed_hash, &network_str)?;
+            if all_notes.is_empty() {
+                tracing::info!(
+                    "Shielded init: tree synced to index {} but no notes in DB — forcing full resync",
+                    state.last_synced_index,
+                );
+                self.db.clear_commitment_tree_tables()?;
```
thepastaclaw: **🟡 Suggestion: Safety-net resync clears the shared commitment tree, affecting all wallets on the network**

`clear_commitment_tree_tables()` (line 253) deletes all rows from the `commitment_tree_*` tables with no wallet-scoping WHERE clause. The commitment tree is correctly shared per-network (it represents the global Orchard tree), but if one wallet triggers this safety-net path, every other wallet's shielded operations will encounter `AnchorMismatch` until the tree is re-synced.

The impact is limited — the anchor mismatch retry in `with_anchor_retry` would re-sync automatically — but it forces unnecessary re-syncs for unrelated wallets. Consider scoping the safety net more narrowly (e.g., only reset `last_synced_index` and `notes` for the affected wallet, preserving the shared tree).

<sub>source: unknown</sub>

<details>

```

In `src/context/shielded.rs`:
- [SUGGESTION] lines 246-253: Safety-net resync clears the shared commitment tree, affecting all wallets on the network
  `clear_commitment_tree_tables()` (line 253) deletes all rows from the `commitment_tree_*` tables with no wallet-scoping WHERE clause. The commitment tree is correctly shared per-network (it represents the global Orchard tree), but if one wallet triggers this safety-net path, every other wallet's shielded operations will encounter `AnchorMismatch` until the tree is re-synced.

The impact is limited — the anchor mismatch retry in `with_anchor_retry` would re-sync automatically — but it forces unnecessary re-syncs for unrelated wallets. Consider scoping the safety net more narrowly (e.g., only reset `last_synced_index` and `notes` for the affected wallet, preserving the shared tree).
```
</details>

src/backend_task/shielded/bundle.rs
```diff
@@ -433,7 +422,7 @@ pub async fn shield_from_asset_lock(
                         asset_lock_duffs,
                         false,
                     )
-                    .map_err(|e| TaskError::ShieldedTransitionBuildFailed { detail: e })?;
+                    .map_err(shielded_build_error)?;
```
thepastaclaw, line 480: **🟡 Suggestion: shielded_build_error used for non-build errors (reload_utxos, generic_asset_lock_transaction)**

Lines 416, 425, and 475 use `.map_err(shielded_build_error)` for errors from `reload_utxos()` and `generic_asset_lock_transaction()`, which are not shielded transaction builds. The `shielded_build_error()` function pattern-matches for 'amount + fee exceeds spendable' and 'AnchorMismatch'. While it's very unlikely a UTXO reload error would match, routing unrelated error strings through this parser is semantically wrong and could theoretically produce a misleading `ShieldedFeeExceedsBalance` or `ShieldedAnchorMismatch` variant. Consider using `TaskError::ShieldedTransitionBuildFailed { detail }` directly for these non-build callsites.

<sub>source: unknown</sub>

<details>

```

In `src/backend_task/shielded/bundle.rs`:
- [SUGGESTION] lines 413-425: shielded_build_error used for non-build errors (reload_utxos, generic_asset_lock_transaction)
  Lines 416, 425, and 475 use `.map_err(shielded_build_error)` for errors from `reload_utxos()` and `generic_asset_lock_transaction()`, which are not shielded transaction builds. The `shielded_build_error()` function pattern-matches for 'amount + fee exceeds spendable' and 'AnchorMismatch'. While it's very unlikely a UTXO reload error would match, routing unrelated error strings through this parser is semantically wrong and could theoretically produce a misleading `ShieldedFeeExceedsBalance` or `ShieldedAnchorMismatch` variant. Consider using `TaskError::ShieldedTransitionBuildFailed { detail }` directly for these non-build callsites.
```
</details>

src/backend_task/core/mod.rs
```diff
@@ -481,6 +469,25 @@ impl AppContext {
...
+        }
+        if is_rpc_connection_error(e) {
+            let url = config
+                .as_ref()
+                .map(|c| format!("{}:{}", c.core_host, c.core_rpc_port))
+                .unwrap_or_else(|| "unknown".to_string());
+            return Some(TaskError::CoreRpcConnectionFailed { url, source: None });
+        }
+        None
+    }
```
thepastaclaw, line 497: **🟡 Suggestion: Error chain lost in chain_lock_rpc_error — source: None discards diagnostic info**

`chain_lock_rpc_error` takes `e: &dashcore_rpc::Error` by reference and creates `CoreRpcConnectionFailed { url, source: None }` (line 486), dropping the original error. By contrast, `rpc_error_with_url()` in `context/mod.rs` takes `e` by value and preserves the chain via `source: Some(e)`. The actual I/O error (e.g., "connection refused at 127.0.0.1:9998") is lost from `Debug` output and `BannerHandle::with_details()`. If `dashcore_rpc::Error` implements `Clone`, the fix is straightforward; otherwise consider restructuring the caller to pass ownership.

<sub>source: unknown</sub>

<details>

```

- [SUGGESTION] lines 474-489: Error chain lost in chain_lock_rpc_error — source: None discards diagnostic info
  `chain_lock_rpc_error` takes `e: &dashcore_rpc::Error` by reference and creates `CoreRpcConnectionFailed { url, source: None }` (line 486), dropping the original error. By contrast, `rpc_error_with_url()` in `context/mod.rs` takes `e` by value and preserves the chain via `source: Some(e)`. The actual I/O error (e.g., "connection refused at 127.0.0.1:9998") is lost from `Debug` output and `BannerHandle::with_details()`. If `dashcore_rpc::Error` implements `Clone`, the fix is straightforward; otherwise consider restructuring the caller to pass ownership.
```
</details>

thepastaclaw (changes requested):
## Code Review

Two blocking bugs found: `with_anchor_retry` permanently leaks shielded wallet state from the in-memory map when `sync_notes` fails, and the new config-save error handling shows a 'session only' warning but never actually applies the password change in-memory. Several lower-severity issues around recovery durability and UX guard rails for the shielded pool.

*Reviewed commit: da4814a2*

🔴 2 blocking | 🟡 4 suggestion(s)

<details>

```

In `src/context/shielded.rs`:
- [BLOCKING] lines 439-458: with_anchor_retry permanently leaks shielded state on sync_notes failure
  When `with_anchor_retry` detects `ShieldedAnchorMismatch`, it calls `sync_notes()` (line 443). If that sync fails, the `?` at line 453 causes an early return — but the shielded wallet state was `remove()`d from the `shielded_states` map at line 434 and is never re-inserted. This permanently orphans the wallet's shielded state until the app restarts.

The sibling method `sync_shielded_notes()` (lines 275–307) handles this correctly: it always re-inserts state at line 298–301 before propagating errors at line 303. The same guard pattern must be applied here.

All callers — `shielded_transfer_task`, `unshield_credits_task`, and `shielded_withdrawal_task` — are affected.
- [SUGGESTION] lines 246-253: Safety-net resync clears the shared commitment tree, affecting all wallets on the network
  `clear_commitment_tree_tables()` (line 253) deletes all rows from the `commitment_tree_*` tables with no wallet-scoping WHERE clause. The commitment tree is correctly shared per-network (it represents the global Orchard tree), but if one wallet triggers this safety-net path, every other wallet's shielded operations will encounter `AnchorMismatch` until the tree is re-synced.

The impact is limited — the anchor mismatch retry in `with_anchor_retry` would re-sync automatically — but it forces unnecessary re-syncs for unrelated wallets. Consider scoping the safety net more narrowly (e.g., only reset `last_synced_index` and `notes` for the affected wallet, preserving the shared tree).

In `src/ui/network_chooser_screen.rs`:
- [BLOCKING] lines 494-553: Config save failure path never applies password for current session
  When `config.save()` fails (line 494), the banner says "Your changes will apply for this session only." However, the in-memory config update (`*cfg_lock = updated_config` at line 521) and `reinit_core_client_and_sdk()` (line 527) are both inside the `else` branch (lines 501–553), which only runs on save *success*. So on the failure path, neither the file nor the live RPC client is updated — the password change is silently discarded despite the message promising session-level application.

This is a logic regression introduced by commit da4814a2. The in-memory update and reinit should run unconditionally; only the success/warning banner should be conditional on save outcome.

In `src/backend_task/shielded/bundle.rs`:
- [SUGGESTION] lines 433-476: Shield-from-asset-lock skips DB persistence of asset lock before broadcast
  The existing asset lock flow uses `broadcast_and_commit_asset_lock()` (transaction_processing.rs:130–136) which stores the asset lock in the DB *before* broadcast and removes it on failure. The new shield-from-asset-lock path broadcasts directly via `core_client.send_raw_transaction` (line 447) without any DB persistence.

If the app crashes after broadcast but before the shielded state transition completes, the asset lock funds have left the wallet UTXO set but there's no DB record for the standard recovery path to reuse. `recover_asset_locks()` can find unspent asset locks on-chain via `list_unspent`, so funds aren't permanently lost, but the automatic recovery guarantee is weaker than the established pattern provides.
- [SUGGESTION] lines 413-425: shielded_build_error used for non-build errors (reload_utxos, generic_asset_lock_transaction)
  Lines 416, 425, and 475 use `.map_err(shielded_build_error)` for errors from `reload_utxos()` and `generic_asset_lock_transaction()`, which are not shielded transaction builds. The `shielded_build_error()` function pattern-matches for 'amount + fee exceeds spendable' and 'AnchorMismatch'. While it's very unlikely a UTXO reload error would match, routing unrelated error strings through this parser is semantically wrong and could theoretically produce a misleading `ShieldedFeeExceedsBalance` or `ShieldedAnchorMismatch` variant. Consider using `TaskError::ShieldedTransitionBuildFailed { detail }` directly for these non-build callsites.

- [SUGGESTION] lines 474-489: Error chain lost in chain_lock_rpc_error — source: None discards diagnostic info
  `chain_lock_rpc_error` takes `e: &dashcore_rpc::Error` by reference and creates `CoreRpcConnectionFailed { url, source: None }` (line 486), dropping the original error. By contrast, `rpc_error_with_url()` in `context/mod.rs` takes `e` by value and preserves the chain via `source: Some(e)`. The actual I/O error (e.g., "connection refused at 127.0.0.1:9998") is lost from `Debug` output and `BannerHandle::with_details()`. If `dashcore_rpc::Error` implements `Clone`, the fix is straightforward; otherwise consider restructuring the caller to pass ownership.
```
</details>

src/context/shielded.rs
lklimek: Fixed in commit e41a72d5. The `?` after `sync_notes()` is removed — sync failure is now captured and logged, but execution always falls through to the state re-insertion block. State is never orphaned.

src/ui/network_chooser_screen.rs
lklimek: Fixed in commit 6e775a4d. Restructured so in-memory config update + `reinit_core_client_and_sdk()` run unconditionally (when the network context exists). A `match (save_failed, reinit_failed)` block picks the correct banner for all four outcome combinations.

src/backend_task/shielded/bundle.rs
lklimek: Valid observation. This is not addressed in this PR — commented on PR #790 where the asset lock flow is being reworked: https://github.com/dashpay/dash-evo-tool/pull/790#issuecomment-4118050020

src/context/shielded.rs
lklimek: Fixed — the commitment tree is now per-wallet (separate SQLite files). `clear_commitment_tree_for_wallet()` only affects the target wallet's tree. Filed dashpay/grovedb#653 to properly abstract storage in the SDK.

src/backend_task/shielded/bundle.rs
lklimek: Fixed in commit 95782ae3. Added dedicated `WalletUtxoReloadFailed` and `WalletBalanceRecalculationFailed` variants to `TaskError`. Non-build callsites (`reload_utxos`, `recalculate_affected_address_balances`) now use these instead of routing through `shielded_build_error`.

src/backend_task/core/mod.rs
lklimek: Fixed in commit a12cc6ac. Added `detail: Option<String>` to `CoreRpcConnectionFailed` and populated it with `format!("{e}")` in `chain_lock_rpc_error()`. Also boxed the `source` field to keep the enum within clippy's size threshold.

src/database/initialization.rs
```diff
@@ -1065,4 +1158,148 @@ mod test {
...
+        let version: u16 = conn
+            .query_row(
+                "SELECT database_version FROM settings WHERE id = 1",
+                [],
+                |row| row.get(0),
+            )
+            .unwrap();
+        assert_eq!(version, DEFAULT_DB_VERSION);
+        assert_eq!(version, 34);
+
```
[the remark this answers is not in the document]
lklimek: Fixed in commit 344c2294 — renamed to `test_v34_migration_fresh_install`.

src/model/wallet/mod.rs
```diff
@@ -415,7 +418,7 @@ impl Wallet {
         // Derive the first receive address (m/44'/coin'/0'/0/0)
         let (known_addresses, watched_addresses) =
             Self::derive_first_address(&master_bip44_ecdsa_extended_public_key, network, &secp)
-                .map_err(|e| TaskError::WalletKeyDerivationFailed { detail: e })?;
+                .map_err(|e| TaskError::WalletKeyDerivationFailed { source: e.into() })?;
```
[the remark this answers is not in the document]
lklimek: This is correct as-is. `derive_first_address()` returns `Result<..., String>`, and at the callsite (line 421) it maps via:

```rust
.map_err(|e| TaskError::WalletKeyDerivationFailed { source: e.into() })?;
```

`String` implements `Into<Box<dyn Error + Send + Sync>>`, so `e.into()` converts the `String` into the boxed trait object expected by the `#[source]` field. This is a standard Rust pattern for bridging string errors into typed error chains. No type mismatch exists.

The other two callsites (lines 404, 411) use `Box::new(e)` on `bip32::Error` which also implements `Error + Send + Sync`. All three paths are type-safe.

src/model/address.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        // 1. Shielded (dash1z... / tdash1z...)
+        if trimmed.starts_with("dash1z") || trimmed.starts_with("tdash1z") {
+            return Some(AddressKind::Shielded);
+        }
+
+        // 2. Platform (Bech32m per DIP-18, but NOT shielded — already excluded above)
+        if crate::ui::helpers::is_platform_address_string(trimmed) {
+            return Some(AddressKind::Platform);
+        }
```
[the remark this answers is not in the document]
lklimek: Valid architectural observation. `AddressKind::detect()` in `src/model/address.rs:72` calls `crate::ui::helpers::is_platform_address_string()` — a function in the UI layer. This creates a model→UI dependency, which inverts the expected layering.

However, looking at the function itself (`src/ui/helpers.rs:21`), `is_platform_address_string()` is a pure string-matching utility (checks Bech32 HRP prefixes like `dash1`/`tdash1`). It has zero UI dependencies — no `egui`, no `Context`, no rendering. It's misplaced in `ui::helpers` but functionally correct.

The proper fix would be to move `is_platform_address_string()` to `src/model/address.rs` (or a `model::helpers` module) so the dependency direction is correct. This is a pre-existing issue — not introduced by this PR — and doesn't affect correctness or behavior.

src/database/mod.rs
```diff
@@ -188,14 +186,28 @@ impl Database {
...
+        let mut wallet_hashes: Vec<Vec<u8>> = Vec::new();
+        {
+            let mut stmt = tx.prepare("SELECT seed_hash FROM wallet WHERE network = ?1")?;
+            let rows = stmt.query_map(rusqlite::params![&network_str], |row| {
+                row.get::<_, Vec<u8>>(0)
+            })?;
+            for row in rows {
+                wallet_hashes.push(row?);
+            }
+        }
```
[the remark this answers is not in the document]
lklimek: No longer applicable. The per-wallet commitment tree approach (separate `.db` files under `shielded/`) was reverted in 712049bc. The current `clear_network_data()` is a single transaction that does not reference `seed_hash` or commitment trees at all — the SELECT-after-DELETE issue no longer exists.

src/database/shielded.rs
```diff
@@ -163,18 +163,13 @@ impl Database {
...
-        let _ = conn.execute("DELETE FROM commitment_tree_shards", []);
-        let _ = conn.execute("DELETE FROM commitment_tree_cap", []);
-        let _ = conn.execute("DELETE FROM commitment_tree_checkpoints", []);
-        let _ = conn.execute("DELETE FROM commitment_tree_checkpoint_marks_removed", []);
+        conn.execute("DELETE FROM commitment_tree_shards", [])?;
+        conn.execute("DELETE FROM commitment_tree_cap", [])?;
+        conn.execute("DELETE FROM commitment_tree_checkpoints", [])?;
+        conn.execute("DELETE FROM commitment_tree_checkpoint_marks_removed", [])?;
         Ok(())
     }
```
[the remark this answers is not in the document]
lklimek: No longer applicable. `src/database/shielded.rs` was removed from the codebase — the per-wallet commitment tree files approach was reverted in 712049bc, and `clear_commitment_tree_tables()` no longer exists. The concern about hard-failing on missing tables is moot.

src/database/mod.rs
```diff
@@ -188,14 +184,10 @@ impl Database {
...
-        // These tables are created by grovedb on first use, so they may not
-        // exist yet — ignore errors from missing tables.
-        let _ = tx.execute("DELETE FROM commitment_tree_shards", []);
-        let _ = tx.execute("DELETE FROM commitment_tree_cap", []);
-        let _ = tx.execute("DELETE FROM commitment_tree_checkpoints", []);
-        let _ = tx.execute("DELETE FROM commitment_tree_checkpoint_marks_removed", []);
+        tx.commit()?;
 
-        tx.commit()
+        self.clear_commitment_tree_tables()?;
```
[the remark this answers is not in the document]
lklimek: No longer applicable. The current `clear_network_data()` (previously `clear_all_data`) is a single transaction with one `tx.commit()` at the end — no commitment tree clearing happens outside the transaction. The non-atomicity concern is resolved by the removal of the separate commitment tree clearing step (reverted in 712049bc).

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1524 @@
...
+
+    fn validate_core(&self, trimmed: &str) -> (Option<String>, Option<ValidatedAddress>) {
+        match trimmed.parse::<Address<NetworkUnchecked>>() {
+            Ok(addr) => match addr.require_network(self.network) {
+                Ok(checked) => (None, Some(ValidatedAddress::Core(checked))),
+                Err(_) => (
+                    Some("This address belongs to a different network.".to_string()),
+                    None,
+                ),
+            },
```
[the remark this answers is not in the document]
lklimek: No longer applicable. `src/ui/components/address_input.rs` has been removed from the PR branch — the `AddressInput` component was merged separately via #787 and is not part of this PR's diff. The `require_network()` vs `networks_address_compatible()` concern does not affect any code in this PR.

src/context/shielded.rs
```diff
@@ -216,23 +217,55 @@ impl AppContext {
...
+                let fresh_tree = ClientPersistentCommitmentTree::open_on_shared_connection(
+                    self.db.shared_connection(),
+                    100,
+                )
+                .map_err(|e| TaskError::ShieldedTreeUpdateFailed {
+                    detail: e.to_string(),
+                })?;
+                state.commitment_tree = std::sync::Mutex::new(fresh_tree);
+                state.last_synced_index = 0;
             }
```
thepastaclaw, line 268: **🟡 Suggestion: Safety-net resync clears shared commitment tree for all wallets**

When `initialize_shielded_wallet()` detects a wallet with `last_synced_index > 0` but zero persisted notes, it calls `clear_commitment_tree_tables()` (line 254). Verified in `src/database/shielded.rs:170-188`: this function runs `DELETE FROM` on four shared tables (`commitment_tree_shards`, `commitment_tree_cap`, `commitment_tree_checkpoints`, `commitment_tree_checkpoint_marks_removed`) with no per-wallet filtering. If wallet A triggers this recovery, wallet B loses its commitment tree state and must resync. The trigger condition is narrow (only fires when DB state is internally inconsistent), so in practice this mainly affects edge-case recovery scenarios. The worst case is a forced resync (reconstructible from network), not data loss. Acknowledged as a conscious design trade-off after the revert in 712049bc from per-wallet files back to shared DB.

<details>

```

In `src/context/shielded.rs`:
- [SUGGESTION] lines 246-268: Safety-net resync clears shared commitment tree for all wallets
  When `initialize_shielded_wallet()` detects a wallet with `last_synced_index > 0` but zero persisted notes, it calls `clear_commitment_tree_tables()` (line 254). Verified in `src/database/shielded.rs:170-188`: this function runs `DELETE FROM` on four shared tables (`commitment_tree_shards`, `commitment_tree_cap`, `commitment_tree_checkpoints`, `commitment_tree_checkpoint_marks_removed`) with no per-wallet filtering. If wallet A triggers this recovery, wallet B loses its commitment tree state and must resync. The trigger condition is narrow (only fires when DB state is internally inconsistent), so in practice this mainly affects edge-case recovery scenarios. The worst case is a forced resync (reconstructible from network), not data loss. Acknowledged as a conscious design trade-off after the revert in 712049bc from per-wallet files back to shared DB.
```
</details>

thepastaclaw (commented):
## Code Review

Three prior blocking findings (anchor retry state leak, RPC password session-apply, and commitment tree missing-table crash) are fixed. The global commitment tree clear remains as a known design trade-off. The new wallet UI changes introduce a mine-dialog regression for fresh wallets and a narrow transaction-filter gap. Advanced platform sends lack network validation. No true blockers remain — all findings are suggestions or nitpicks.

*Reviewed commit: 810a6867*

🟡 6 suggestion(s) | 💬 2 nitpick(s)

### 2 additional findings

<details>
<summary>🟡 <b>suggestion:</b> Advanced platform sends discard the parsed network without validation</summary>

`src/ui/wallets/send_screen.rs (lines 2590-2637)`

Both `send_advanced_utxo_to_platform` (line 2590) and `send_advanced_platform_to_platform` (line 2635) parse platform addresses via `PlatformAddress::from_bech32m_string()` which returns `(PlatformAddress, Network)`, but both discard the network with `.map(|(addr, _)| addr)`. By contrast, `AddressInput.validate_platform()` (address_input.rs:526-537) checks the bech32m prefix against the expected network before parsing. The advanced paths bypass AddressInput and read from raw text fields. A user pasting a mainnet `dash1...` address while on testnet would get a confusing backend error instead of an immediate validation message. No actual cross-network funds risk (the SDK would reject it), but poor UX.

</details>
<details>
<summary>💬 <b>nitpick:</b> Developer 'Resync Notes' button also clears global commitment tree</summary>

`src/ui/wallets/shielded_tab.rs (line 558)`

Line 558 calls `clear_commitment_tree_tables()` which deletes shared table data for all wallets, same as the safety-net in initialize_shielded_wallet. Developer-mode only, so low user impact, but worth noting as part of the same global-clear pattern.

</details>

<details>

```

In `src/context/shielded.rs`:
- [SUGGESTION] lines 246-268: Safety-net resync clears shared commitment tree for all wallets
  When `initialize_shielded_wallet()` detects a wallet with `last_synced_index > 0` but zero persisted notes, it calls `clear_commitment_tree_tables()` (line 254). Verified in `src/database/shielded.rs:170-188`: this function runs `DELETE FROM` on four shared tables (`commitment_tree_shards`, `commitment_tree_cap`, `commitment_tree_checkpoints`, `commitment_tree_checkpoint_marks_removed`) with no per-wallet filtering. If wallet A triggers this recovery, wallet B loses its commitment tree state and must resync. The trigger condition is narrow (only fires when DB state is internally inconsistent), so in practice this mainly affects edge-case recovery scenarios. The worst case is a forced resync (reconstructible from network), not data loss. Acknowledged as a conscious design trade-off after the revert in 712049bc from per-wallet files back to shared DB.

In `src/ui/wallets/wallets_screen/dialogs.rs`:
- [SUGGESTION] lines 1288-1293: Mine dialog offers no addresses for fresh or zero-balance wallets
  Verified: `AddressInput::extract_wallet_entries()` (address_input.rs:352) pulls Core addresses exclusively from `wallet.address_balances`, which only contains addresses with current UTXOs. For a fresh wallet, `address_balances` is empty while `known_addresses` contains 48+ bootstrapped BIP44 addresses. Combined with `.with_selection_only(true)` at line 1292, the user cannot select OR type an address. This is a chicken-and-egg problem: the mine dialog exists to fund a wallet, but requires the wallet to already have funded addresses. Regtest-only impact (mining doesn't work on mainnet/testnet), but broken for its primary use case.

In `src/ui/wallets/wallets_screen/mod.rs`:
- [SUGGESTION] lines 1234-1247: Transaction filter drops send-all transactions with no change output
  The filter at line 1234 keeps only transactions with at least one output matching `wallet_guard.known_addresses`. A legitimate send-all transaction where total input equals output + fee (no change output back to the wallet) would be filtered out and invisible in the transaction list. This is display-only — no funds risk. The edge case is narrow (requires exact-amount sends or dust-to-fee scenarios), and the comment at line 1231-1233 acknowledges the design: outputs-only checking is intentional as a cross-wallet contamination guard. Worth noting for completeness but low real-world frequency.

In `src/ui/wallets/send_screen.rs`:
- [SUGGESTION] lines 2590-2637: Advanced platform sends discard the parsed network without validation
  Both `send_advanced_utxo_to_platform` (line 2590) and `send_advanced_platform_to_platform` (line 2635) parse platform addresses via `PlatformAddress::from_bech32m_string()` which returns `(PlatformAddress, Network)`, but both discard the network with `.map(|(addr, _)| addr)`. By contrast, `AddressInput.validate_platform()` (address_input.rs:526-537) checks the bech32m prefix against the expected network before parsing. The advanced paths bypass AddressInput and read from raw text fields. A user pasting a mainnet `dash1...` address while on testnet would get a confusing backend error instead of an immediate validation message. No actual cross-network funds risk (the SDK would reject it), but poor UX.

- [SUGGESTION] lines 489-494: chain_lock_rpc_error embeds raw error Display into user-facing url field
  Line 492 formats the url as `format!("{}:{} ({})", c.core_host, c.core_rpc_port, e)`, embedding the raw `dashcore_rpc::Error` Display text into the `url` field. Since `CoreRpcConnectionFailed`'s `#[error]` attribute includes `{url}` in the user-facing message, OS-level error text (e.g., "Connection refused (os error 111)") leaks to the user. Additionally, `source: None` is passed because the function borrows `&e` and can't move it, breaking the error chain. The function signature should take `e` by value to preserve the source.

- [SUGGESTION] lines 820-836: Fee-exceeds-spendable detection relies on fragile English substring parsing
  Verified: `parse_fee_exceeds_spendable` extracts numeric fields by searching for hard-coded English substrings ("amount ", "fee ", "exceeds total spendable value ") inside the SDK error text. If the upstream SDK changes wording, the parser silently fails and falls through to the generic `ShieldedTransitionBuildFailed` variant. The graceful degradation means this isn't a crash risk, but the user would see a generic error instead of the specific fee/balance breakdown. Not actionable until the SDK exposes structured error data — noting for awareness.
```
</details>

src/ui/wallets/wallets_screen/dialogs.rs
```diff
@@ -1280,38 +1285,20 @@ impl WalletsBalancesScreen {
             return;
         };
 
+        let address_input = AddressInput::new(self.app_context.network)
+            .with_label("Mine to address:")
+            .with_address_kinds(&[AddressKind::Core])
+            .with_wallets(&[wallet])
+            .with_selection_only(true)
+            .with_full_addresses(true);
```
thepastaclaw, line 1293: **🟡 Suggestion: Mine dialog offers no addresses for fresh or zero-balance wallets**

Verified: `AddressInput::extract_wallet_entries()` (address_input.rs:352) pulls Core addresses exclusively from `wallet.address_balances`, which only contains addresses with current UTXOs. For a fresh wallet, `address_balances` is empty while `known_addresses` contains 48+ bootstrapped BIP44 addresses. Combined with `.with_selection_only(true)` at line 1292, the user cannot select OR type an address. This is a chicken-and-egg problem: the mine dialog exists to fund a wallet, but requires the wallet to already have funded addresses. Regtest-only impact (mining doesn't work on mainnet/testnet), but broken for its primary use case.

<sub>source: ['codex']</sub>

<details>

```

In `src/ui/wallets/wallets_screen/dialogs.rs`:
- [SUGGESTION] lines 1288-1293: Mine dialog offers no addresses for fresh or zero-balance wallets
  Verified: `AddressInput::extract_wallet_entries()` (address_input.rs:352) pulls Core addresses exclusively from `wallet.address_balances`, which only contains addresses with current UTXOs. For a fresh wallet, `address_balances` is empty while `known_addresses` contains 48+ bootstrapped BIP44 addresses. Combined with `.with_selection_only(true)` at line 1292, the user cannot select OR type an address. This is a chicken-and-egg problem: the mine dialog exists to fund a wallet, but requires the wallet to already have funded addresses. Regtest-only impact (mining doesn't work on mainnet/testnet), but broken for its primary use case.
```
</details>

src/ui/wallets/wallets_screen/mod.rs
```diff
@@ -1181,16 +1200,61 @@ impl WalletsBalancesScreen {
...
+                .filter(|&i| {
+                    let tx = &wallet_guard.transactions[i];
+                    tx.transaction.output.iter().any(|output| {
+                        Address::from_script(&output.script_pubkey, self.app_context.network)
+                            .ok()
+                            .is_some_and(|addr| wallet_addresses.contains(&addr))
+                    })
+                })
+                .collect()
+        });
```
thepastaclaw, line 1247: **🟡 Suggestion: Transaction filter drops send-all transactions with no change output**

The filter at line 1234 keeps only transactions with at least one output matching `wallet_guard.known_addresses`. A legitimate send-all transaction where total input equals output + fee (no change output back to the wallet) would be filtered out and invisible in the transaction list. This is display-only — no funds risk. The edge case is narrow (requires exact-amount sends or dust-to-fee scenarios), and the comment at line 1231-1233 acknowledges the design: outputs-only checking is intentional as a cross-wallet contamination guard. Worth noting for completeness but low real-world frequency.

<sub>source: ['codex']</sub>

<details>

```

In `src/ui/wallets/wallets_screen/mod.rs`:
- [SUGGESTION] lines 1234-1247: Transaction filter drops send-all transactions with no change output
  The filter at line 1234 keeps only transactions with at least one output matching `wallet_guard.known_addresses`. A legitimate send-all transaction where total input equals output + fee (no change output back to the wallet) would be filtered out and invisible in the transaction list. This is display-only — no funds risk. The edge case is narrow (requires exact-amount sends or dust-to-fee scenarios), and the comment at line 1231-1233 acknowledges the design: outputs-only checking is intentional as a cross-wallet contamination guard. Worth noting for completeness but low real-world frequency.
```
</details>

src/backend_task/core/mod.rs
```diff
@@ -481,6 +477,25 @@ impl AppContext {
...
+    ) -> Option<TaskError> {
+        if is_rpc_auth_error(e) {
+            return Some(TaskError::CoreRpcAuthFailed);
+        }
+        if is_rpc_connection_error(e) {
+            let url = config
+                .as_ref()
+                .map(|c| format!("{}:{} ({})", c.core_host, c.core_rpc_port, e))
+                .unwrap_or_else(|| "unknown".to_string());
+            return Some(TaskError::CoreRpcConnectionFailed { url, source: None });
```
thepastaclaw, line 494: **🟡 Suggestion: chain_lock_rpc_error embeds raw error Display into user-facing url field**

Line 492 formats the url as `format!("{}:{} ({})", c.core_host, c.core_rpc_port, e)`, embedding the raw `dashcore_rpc::Error` Display text into the `url` field. Since `CoreRpcConnectionFailed`'s `#[error]` attribute includes `{url}` in the user-facing message, OS-level error text (e.g., "Connection refused (os error 111)") leaks to the user. Additionally, `source: None` is passed because the function borrows `&e` and can't move it, breaking the error chain. The function signature should take `e` by value to preserve the source.

<details>

```suggestion
fn chain_lock_rpc_error(
    config: &Option<NetworkConfig>,
    e: dashcore_rpc::Error,
) -> Option<TaskError> {
    if is_rpc_auth_error(&e) {
        return Some(TaskError::CoreRpcAuthFailed);
    }
    if is_rpc_connection_error(&e) {
        let url = config
            .as_ref()
            .map(|c| format!("{}:{}", c.core_host, c.core_rpc_port))
            .unwrap_or_else(|| "unknown".to_string());
        return Some(TaskError::CoreRpcConnectionFailed {
            url,
            source: Some(Box::new(e)),
        });
    }
    None
}
```

</details>

<sub>source: ['claude']</sub>

<details>

```

- [SUGGESTION] lines 489-494: chain_lock_rpc_error embeds raw error Display into user-facing url field
  Line 492 formats the url as `format!("{}:{} ({})", c.core_host, c.core_rpc_port, e)`, embedding the raw `dashcore_rpc::Error` Display text into the `url` field. Since `CoreRpcConnectionFailed`'s `#[error]` attribute includes `{url}` in the user-facing message, OS-level error text (e.g., "Connection refused (os error 111)") leaks to the user. Additionally, `source: None` is passed because the function borrows `&e` and can't move it, breaking the error chain. The function signature should take `e` by value to preserve the source.
```
</details>

src/backend_task/error.rs
```diff
@@ -698,6 +805,90 @@ pub fn is_instant_lock_proof_invalid(error: &SdkError) -> bool {
...
+    let fee_start = fee_marker + 4;
+    let fee_end = detail[fee_start..].find(' ')? + fee_start;
+    let fee: u64 = detail[fee_start..fee_end].parse().ok()?;
+
+    let spendable_marker = detail.find("exceeds total spendable value ")?;
+    let spendable_start = spendable_marker + 30;
+    let spendable: u64 = detail[spendable_start..].trim().parse().ok()?;
+
+    Some((amount, fee, spendable))
+}
```
thepastaclaw, line 958: **🟡 Suggestion: Fee-exceeds-spendable detection relies on fragile English substring parsing**

Verified: `parse_fee_exceeds_spendable` extracts numeric fields by searching for hard-coded English substrings ("amount ", "fee ", "exceeds total spendable value ") inside the SDK error text. If the upstream SDK changes wording, the parser silently fails and falls through to the generic `ShieldedTransitionBuildFailed` variant. The graceful degradation means this isn't a crash risk, but the user would see a generic error instead of the specific fee/balance breakdown. Not actionable until the SDK exposes structured error data — noting for awareness.

<sub>source: ['codex']</sub>

<details>

```

- [SUGGESTION] lines 820-836: Fee-exceeds-spendable detection relies on fragile English substring parsing
  Verified: `parse_fee_exceeds_spendable` extracts numeric fields by searching for hard-coded English substrings ("amount ", "fee ", "exceeds total spendable value ") inside the SDK error text. If the upstream SDK changes wording, the parser silently fails and falls through to the generic `ShieldedTransitionBuildFailed` variant. The graceful degradation means this isn't a crash risk, but the user would see a generic error instead of the specific fee/balance breakdown. Not actionable until the SDK exposes structured error data — noting for awareness.
```
</details>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+                        .to_string(),
+                ),
+                None,
+            );
+        }
+        use dash_sdk::dpp::address_funds::OrchardAddress;
+        match OrchardAddress::from_bech32m_string(trimmed) {
+            Ok((_, network)) => {
+                if network != self.network
+                    && !(self.network != Network::Mainnet && network != Network::Mainnet)
```
thepastaclaw, line 578: **💬 Nitpick: Double-negation in shielded network check is hard to read**

Line 578: `!(self.network != Network::Mainnet && network != Network::Mainnet)` is a double-negative that means "at least one is Mainnet". The intent is to allow cross-testnet/devnet/regtest addresses but reject mainnet/non-mainnet mismatches. A positive formulation or use of an existing helper would clarify the logic.

<details>

```suggestion
if network != self.network
                    && (self.network == Network::Mainnet || network == Network::Mainnet)
                {
```

</details>

<sub>source: ['claude']</sub>

src/backend_task/core/mod.rs
lklimek: Related to #797 (item 3 — typed SPV error variants). The raw error leaking into `url` will be addressed when `WalletPaymentFailed` string details are replaced with typed variants.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/backend_task/error.rs
lklimek: Related to #795 (item 2 — fee calculation correctness). The fragile substring parsing is a known limitation; the proper fix is upstream SDK providing structured error data. Tracked as part of the fee calculation issue.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/wallets_screen/dialogs.rs
lklimek: Fixed in PR #791 — `AddressInput::extract_wallet_entries()` now pulls from `known_addresses` (all derived) instead of `address_balances` (funded only). Zero-balance addresses appear in the dropdown; callers use `with_balance_range(1..)` for funded-only filtering.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/wallets_screen/mod.rs
lklimek: Fixed in PR #791 — transaction filter now uses the `is_ours` flag instead of output-only address matching. SPV reconcile path overrides upstream's send-only `is_ours` to `true` for all matched transactions (address ownership already verified by `check_transaction()`). E2E test added in `tests/backend-e2e/tx_is_ours.rs`.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
lklimek: Fixed in PR #791 — replaced double-negation with explicit `same_mainnet_class` comparison. Added comment documenting that shielded addresses only encode mainnet vs non-mainnet in the HRP (testnet/devnet/local all share `tdash1z`).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/context/connection_status.rs
```diff
@@ -422,6 +445,9 @@ impl ConnectionStatus {
                         devnet_chainlock,
                         local_chainlock,
                     );
+                    // INTENTIONAL(SEC-005): RPC error strings shown as-is in network
+                    // status tooltip. Acceptable for desktop app — helps debugging.
+                    self.set_rpc_last_error(rpc_error.clone());
```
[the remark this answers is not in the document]
lklimek: Tracked in #797 (item 7 — raw RPC errors in user-facing status).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/context/shielded.rs
```diff
@@ -216,23 +217,55 @@ impl AppContext {
...
+
+        // Safety net: if the tree has been synced but no notes exist at all
+        // (spent or unspent), force a full resync from index 0. This handles
+        // the case where change notes from prior operations were only in memory
+        // and the app restarted before the next sync persisted them.
+        // We check ALL notes (including spent) to avoid a false positive when
+        // the user legitimately spent everything.
+        if state.last_synced_index > 0 && state.notes.is_empty() {
+            let all_notes = self.db.get_all_shielded_notes(&seed_hash, &network_str)?;
+            if all_notes.is_empty() {
```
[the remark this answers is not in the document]
lklimek: Tracked in #796 (item 7 — resync safety net misses wallets with historical notes).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/shield_credits_screen.rs
```diff
@@ -359,6 +357,11 @@ impl ScreenLike for ShieldCreditsScreen {
             action = AppAction::BackendTask(task);
         }
 
+        // Dispatch pending refresh task (sync notes after successful shield)
+        if let Some(task) = self.pending_refresh_task.take() {
+            action |= AppAction::BackendTask(task);
+        }
```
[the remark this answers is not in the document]
lklimek: Tracked in #795 (item 7 — batch shielding skips note sync).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/shield_from_asset_lock_screen.rs
```diff
@@ -45,35 +52,26 @@ impl ShieldFromAssetLockScreen {
...
 }
 
 impl ScreenLike for ShieldFromAssetLockScreen {
     fn ui(&mut self, ctx: &Context) -> AppAction {
-        let mut action = add_top_panel(
+        let mut action = self
+            .pending_refresh_task
+            .take()
+            .map(AppAction::BackendTask)
+            .unwrap_or(AppAction::None);
```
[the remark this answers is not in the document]
lklimek: Tracked in #795 (item 8 — Done button drops queued sync task).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/shield_from_asset_lock_screen.rs
```diff
@@ -117,30 +115,21 @@ impl ScreenLike for ShieldFromAssetLockScreen {
...
-                    dash, credits
-                ));
-                if duffs > self.core_balance_duffs {
-                    ui.colored_label(
-                        Color32::from_rgb(255, 100, 100),
-                        "Exceeds core wallet balance",
-                    );
-                }
-            }
+            amount_input.set_max_amount(Some(max_credits));
```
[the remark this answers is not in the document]
lklimek: Tracked in #795 (item 9 — max amount ignores asset-lock fee).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/unshield_credits_screen.rs
```diff
@@ -54,63 +60,29 @@ impl UnshieldCreditsScreen {
...
 }
 
 impl ScreenLike for UnshieldCreditsScreen {
     fn ui(&mut self, ctx: &Context) -> AppAction {
-        let mut action = add_top_panel(
+        let mut action = self
+            .pending_refresh_task
+            .take()
+            .map(AppAction::BackendTask)
+            .unwrap_or(AppAction::None);
```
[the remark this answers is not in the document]
lklimek: Tracked in #795 (item 10 — Done drops queued sync in unshield).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/unshield_credits_screen.rs
```diff
@@ -148,63 +120,72 @@ impl ScreenLike for UnshieldCreditsScreen {
...
-            if let Some(credits) = self.parse_amount_credits() {
-                let dash = credits as f64 / CREDITS_PER_DUFF as f64 / 1e8;
-                ui.label(format!("= {:.8} DASH ({} credits)", dash, credits));
-                if credits > self.max_balance {
-                    ui.colored_label(Color32::from_rgb(255, 100, 100), "Exceeds shielded balance");
-                }
-            }
+            amount_input.set_max_amount(Some(self.max_balance));
+            let response = amount_input.show(ui);
+            response.inner.update(&mut self.amount);
```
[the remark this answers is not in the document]
lklimek: Tracked in #795 (item 11 — max amount ignores transition fees in unshield).

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+
+    fn validate_core(&self, trimmed: &str) -> (Option<String>, Option<ValidatedAddress>) {
+        match trimmed.parse::<Address<NetworkUnchecked>>() {
+            Ok(addr) => match addr.require_network(self.network) {
+                Ok(checked) => (None, Some(ValidatedAddress::Core(checked))),
+                Err(_) => (
+                    Some("This address belongs to a different network.".to_string()),
+                    None,
+                ),
+            },
```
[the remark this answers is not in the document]
lklimek: Dismissed — BIP-350 requires either all-uppercase or all-lowercase for bech32m. Our mixed-case rejection is correct per spec.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+                if network != self.network
+                    && !(self.network != Network::Mainnet && network != Network::Mainnet)
+                {
+                    (
+                        Some("This address belongs to a different network.".to_string()),
+                        None,
+                    )
+                } else {
+                    (None, Some(ValidatedAddress::Shielded(trimmed.to_string())))
+                }
```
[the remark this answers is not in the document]
lklimek: Dismissed — empty input in selection-only mode is the initial state (nothing selected yet), not an error. By design.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/backend_task/core/mod.rs
```diff
@@ -195,20 +184,26 @@ impl AppContext {
...
+                    _ => (&mainnet_result, maybe_mainnet_config),
+                };
+                let active_rpc_error = if let Err(e) = active_result {
+                    if let Some(task_err) = Self::chain_lock_rpc_error(active_config, e) {
+                        return Err(task_err);
+                    }
+                    // Non-auth, non-connection error — log the raw error but show
+                    // a sanitized message in the UI status display.
+                    tracing::warn!(network = ?self.network, error = %e, "Chain lock query failed on active network");
+                    Some("RPC error — check Dash Core status".to_string())
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 (commit 24256db5) — `set_rpc_last_error(None)` called on successful chain lock fetch to clear stale error state.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+            return (Some(msg.to_string()), None);
+        }
+
+        // Type-specific validation
+        match detected {
+            DetectedType::Core => self.validate_core(trimmed),
+            DetectedType::Platform => self.validate_platform(trimmed),
+            DetectedType::Shielded => self.validate_shielded(trimmed),
+            DetectedType::Identity => self.validate_identity(trimmed),
+            DetectedType::Unknown => unreachable!(),
```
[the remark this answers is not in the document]
lklimek: Acknowledged — Base58 overlap between Core addresses and Identity IDs is a known limitation of prefix-based detection. No current callers use identity-only mode, so this is theoretical. Will address when identity address input is needed.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+        assert_eq!(result, DetectedType::Unknown);
+    }
+
+    #[test]
+    fn detect_identity_when_enabled() {
+        // A 32-byte Base58 identifier that does not parse as a Core address
+        let id = Identifier::random();
+        let id_str = id.to_string(Encoding::Base58);
+        let result = detect_address_type(&id_str, true);
+        assert_eq!(result, DetectedType::Identity);
```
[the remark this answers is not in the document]
lklimek: Same root cause as the identity-vs-Core detection overlap — deferred together.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/model/address.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        // 1. Shielded (dash1z... / tdash1z...)
+        if trimmed.starts_with("dash1z") || trimmed.starts_with("tdash1z") {
+            return Some(AddressKind::Shielded);
+        }
+
+        // 2. Platform (Bech32m per DIP-18, but NOT shielded — already excluded above)
+        if crate::ui::helpers::is_platform_address_string(trimmed) {
+            return Some(AddressKind::Platform);
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — `is_platform_address_string()` moved from `src/ui/helpers.rs` to `src/model/address.rs` where `AddressKind::detect()` lives. Old location re-exports to avoid breaking callers.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/components/address_input.rs
```diff
@@ -0,0 +1,1522 @@
...
+                (
+                    None,
+                    Some(ValidatedAddress::Identity {
+                        id,
+                        dpns_name: dpns,
+                    }),
+                )
+            }
+            Err(_) => (
+                Some("This does not look like a valid address.".to_string()),
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — validation messages now include actionable next steps: "Please check for typos", "Please check you are using the correct network", "Please use all lowercase".

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/mod.rs
```diff
@@ -809,12 +809,16 @@ impl Screen {
...
             }
             Screen::ImportMnemonicScreen(screen) => {
                 screen.app_context = app_context;
                 screen.reset_core_wallets_cache();
             }
-            Screen::WalletSendScreen(screen) => screen.app_context = app_context,
+            Screen::WalletSendScreen(screen) => {
+                screen.app_context = app_context;
+                screen.invalidate_address_input();
+            }
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — `invalidate_address_input()` added to `TransferScreen`, `AddressBalanceScreen`, and `ShieldedSendScreen`, wired into `change_context()`.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/network_chooser_screen.rs
```diff
@@ -414,6 +415,15 @@ impl NetworkChooserScreen {
...
                                     app_action = AppAction::SwitchNetwork(Network::Regtest);
                                 }
+                                if self.current_network != prev_network
+                                    && let Ok(config) =
+                                        Config::load_from(&self.mainnet_app_context.data_dir)
+                                    && let Some(network_config) =
+                                        config.config_for_network(self.current_network)
+                                {
+                                    self.dashmate_password_input
+                                        .set_text(network_config.core_rpc_password.clone());
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — password field now cleared to empty when switching to a network with no saved config.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/send_screen.rs
```diff
@@ -471,8 +468,14 @@ impl WalletSendScreen {
         estimate_platform_fee(fee_estimator, usable_count)
     }
 
+    /// Clear the AddressInput widget so it picks up the new network on next frame.
+    pub(crate) fn invalidate_address_input(&mut self) {
+        self.address_input = None;
+    }
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — `invalidate_address_input()` now also clears `validated_destination = None` in both `send_screen.rs` and `unshield_credits_screen.rs`.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

src/ui/wallets/shielded_send_screen.rs
```diff
@@ -215,10 +219,30 @@ impl ScreenLike for ShieldedSendScreen {
...
+            BackendTaskSuccessResult::ShieldedNotesSynced {
+                seed_hash,
+                new_notes,
+                balance,
+            } if seed_hash == self.seed_hash => {
+                tracing::debug!(
+                    "ShieldedSendScreen: post-transfer sync complete, new_notes={}, balance={} credits",
+                    new_notes,
+                    balance,
+                );
```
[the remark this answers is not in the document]
lklimek: Fixed in PR #791 — `ShieldedNotesSynced` now updates `max_balance`, clears `balance_update_pending`, and appends remaining balance to the success message.

<sub>🤖 [Claudius the Magnificent](https://github.com/lklimek/claudius)</sub>

lklimek:
Review comments handled in #791

PR #792, opened 2026-03-24
fix(db): add migration v34 to drop global commitment tree tables

lklimek:
## Summary

~~Backports DB migration v34 (drop global commitment_tree_* tables) from `zk-fixes` to `v1.0-dev`.~~

**Closed** — Migration v34 was reverted in commit 712049bc on `zk-fixes`. The commitment tree now uses the shared main database instead of per-wallet files (accepted tradeoff until dashpay/grovedb#653 adds proper wallet-scoping to `ClientPersistentCommitmentTree`). No migration backport is needed.

### DIFF
--- a/src/database/initialization.rs
+++ b/src/database/initialization.rs
@@ -4,7 +4,7 @@ use rusqlite::{Connection, params};
 use std::fs;
 use std::path::Path;
 
-pub const DEFAULT_DB_VERSION: u16 = 33;
+pub const DEFAULT_DB_VERSION: u16 = 34;
 
 pub const DEFAULT_NETWORK: &str = "mainnet";
 
@@ -51,6 +51,18 @@ impl Database {
 
     fn apply_version_changes(&self, version: u16, tx: &Connection) -> rusqlite::Result<()> {
         match version {
+            34 => {
+                // Commitment trees are now stored in per-wallet SQLite files
+                // under <data_dir>/shielded/<hex>.db. Drop the old global
+                // tables that were shared across all wallets.
+                let _ = tx.execute(
+                    "DROP TABLE IF EXISTS commitment_tree_checkpoint_marks_removed",
+                    [],
+                );
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_checkpoints", []);
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_cap", []);
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_shards", []);
+            }
             // Versions 28-32 were consolidated into v33 to resolve migration
             // numbering conflicts between the zk and v1.0-dev branches.
             // If migrating from < 28, these are no-ops that just bump the version.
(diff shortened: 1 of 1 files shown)


src/database/initialization.rs
```diff
@@ -51,6 +51,18 @@ impl Database {
...
+                // Commitment trees are now stored in per-wallet SQLite files
+                // under <data_dir>/shielded/<hex>.db. Drop the old global
+                // tables that were shared across all wallets.
+                let _ = tx.execute(
+                    "DROP TABLE IF EXISTS commitment_tree_checkpoint_marks_removed",
+                    [],
+                );
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_checkpoints", []);
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_cap", []);
+                let _ = tx.execute("DROP TABLE IF EXISTS commitment_tree_shards", []);
```
thepastaclaw, line 64: **🟡 Suggestion: `let _ =` silently discards DROP TABLE errors — use `?` like every other migration**

All four `tx.execute()` calls discard their `Result` with `let _ =`. Since `IF EXISTS` already handles the "table doesn't exist" case at the SQL level, the only errors that could surface here are real failures: disk I/O errors, database corruption, or locking issues.

Because `try_perform_migration` calls `apply_version_changes()` then `update_database_version()` then `tx.commit()`, swallowing these errors means the schema version advances to 34 and the transaction commits even if the tables were **not** actually dropped. The stale tables would remain with no log entry or indication of failure.

This is inconsistent with every other migration in the file (v6–v33), including the identical `DROP TABLE IF EXISTS` pattern at lines 153 and 155 which correctly propagate via `?`.

<details>

```suggestion
                tx.execute(
                    "DROP TABLE IF EXISTS commitment_tree_checkpoint_marks_removed",
                    [],
                )?;
                tx.execute("DROP TABLE IF EXISTS commitment_tree_checkpoints", [])?;
                tx.execute("DROP TABLE IF EXISTS commitment_tree_cap", [])?;
                tx.execute("DROP TABLE IF EXISTS commitment_tree_shards", [])?;
```

</details>

<sub>source: ['claude-general', 'codex-general', 'claude-rust-quality', 'codex-rust-quality']</sub>

<details>

```

In `src/database/initialization.rs`:
- [SUGGESTION] lines 58-64: `let _ =` silently discards DROP TABLE errors — use `?` like every other migration
  All four `tx.execute()` calls discard their `Result` with `let _ =`. Since `IF EXISTS` already handles the "table doesn't exist" case at the SQL level, the only errors that could surface here are real failures: disk I/O errors, database corruption, or locking issues.

Because `try_perform_migration` calls `apply_version_changes()` then `update_database_version()` then `tx.commit()`, swallowing these errors means the schema version advances to 34 and the transaction commits even if the tables were **not** actually dropped. The stale tables would remain with no log entry or indication of failure.

This is inconsistent with every other migration in the file (v6–v33), including the identical `DROP TABLE IF EXISTS` pattern at lines 153 and 155 which correctly propagate via `?`.
```
</details>

thepastaclaw (commented):
## Code Review

Clean, well-scoped migration that drops 4 orphaned global commitment tree tables. Good test coverage for happy paths. However, the `let _ =` pattern silently discards execution errors, which is inconsistent with every other migration in the file (including the identical `DROP TABLE IF EXISTS` pattern at lines 153/155 which correctly use `?`).

*Reviewed commit: d6857344*

🟡 1 suggestion(s)

<details>

```

In `src/database/initialization.rs`:
- [SUGGESTION] lines 58-64: `let _ =` silently discards DROP TABLE errors — use `?` like every other migration
  All four `tx.execute()` calls discard their `Result` with `let _ =`. Since `IF EXISTS` already handles the "table doesn't exist" case at the SQL level, the only errors that could surface here are real failures: disk I/O errors, database corruption, or locking issues.

Because `try_perform_migration` calls `apply_version_changes()` then `update_database_version()` then `tx.commit()`, swallowing these errors means the schema version advances to 34 and the transaction commits even if the tables were **not** actually dropped. The stale tables would remain with no log entry or indication of failure.

This is inconsistent with every other migration in the file (v6–v33), including the identical `DROP TABLE IF EXISTS` pattern at lines 153 and 155 which correctly propagate via `?`.
```
</details>

Issue #652, opened 2026-02-24
Refresh button resets the nonces in address table


lklimek:
1.Open Wallet
2. Open Platform addresses section
3. Check if there are any nonces > 0. If not, do some transaciton
5. Click refresh button
Expected: at least 1 nonce > 0.
Actual: All nonces are 0.

PR #807, opened 2026-03-31
fix(mcp): use PlatformAddress in PlatformAddressBalances backend task result

lklimek:
## Summary

- **Bug**: `platform_addresses_list` MCP tool returned Core address format (`yXxx...`) instead of Platform bech32m format (`tdash1...` / `dash1...`)
- **Root cause**: `BackendTaskSuccessResult::PlatformAddressBalances` carried `BTreeMap<Address<NetworkChecked>, ...>` (Core addresses), forcing every consumer to convert independently
- **Fix**: Changed the backend task result to carry `BTreeMap<PlatformAddress, ...>` — conversion happens once at the source, not at every consumer

### Changes (4 files)

1. **`backend_task/mod.rs`** — `PlatformAddressBalances` variant now uses `PlatformAddress` keys
2. **`backend_task/wallet/fetch_platform_address_balances.rs`** — Converts Core→Platform at the source via `PlatformAddress::try_from()`
3. **`mcp/tools/wallet.rs`** — Simplified to direct `addr.to_bech32m_string(network)` call
4. **`ui/wallets/wallets_screen/mod.rs`** — Converts Platform→Core via `to_address_with_network()` for wallet storage (model still uses Core keys internally)

## Test plan

- [ ] Call `platform_addresses_list` MCP tool and verify addresses are returned in bech32m format (`tdash1...` on testnet, `dash1...` on mainnet)
- [ ] Verify wallet screen still displays platform address balances correctly after refresh

### DIFF
--- a/src/backend_task/wallet/fetch_platform_address_balances.rs
+++ b/src/backend_task/wallet/fetch_platform_address_balances.rs
@@ -5,6 +5,7 @@ use crate::model::wallet::{
     WalletSeedHash,
 };
 use dash_sdk::RequestSettings;
+use dash_sdk::dpp::address_funds::PlatformAddress;
 use dash_sdk::dpp::dashcore::Network;
 use dash_sdk::dpp::key_wallet::bip32::DerivationPath;
 use dash_sdk::platform::address_sync::AddressSyncConfig;
@@ -169,7 +170,11 @@ impl AppContext {
             wallet
                 .platform_address_info
                 .iter()
-                .map(|(addr, info)| (addr.clone(), (info.balance, info.nonce)))
+                .filter_map(|(addr, info)| {
+                    PlatformAddress::try_from(addr.clone())
+                        .ok()
+                        .map(|pa| (pa, (info.balance, info.nonce)))
+                })
                 .collect()
         };
 
--- a/src/ui/wallets/wallets_screen/mod.rs
+++ b/src/ui/wallets/wallets_screen/mod.rs
@@ -2881,9 +2881,11 @@ impl ScreenLike for WalletsBalancesScreen {
                     && let Ok(mut wallet) = selected.write()
                     && wallet.seed_hash() == seed_hash
                 {
-                    // Update balances in the wallet
-                    for (addr, (balance, nonce)) in balances {
-                        wallet.set_platform_address_info(addr, balance, nonce);
+                    // Convert PlatformAddress back to Core Address for wallet storage
+                    let network = self.app_context.network();
+                    for (platform_addr, (balance, nonce)) in balances {
+                        let core_addr = platform_addr.to_address_with_network(network);
+                        wallet.set_platform_address_info(core_addr, balance, nonce);
                     }
                 }
                 self.refresh_platform_sync_info_cache(&seed_hash);
--- a/src/backend_task/mod.rs
+++ b/src/backend_task/mod.rs
@@ -10,8 +10,7 @@ use crate::backend_task::platform_info::{PlatformInfoTaskRequestType, PlatformIn
 use crate::backend_task::system_task::SystemTask;
 use crate::backend_task::wallet::WalletTask;
 use crate::context::AppContext;
-use dash_sdk::dpp::dashcore::Address;
-use dash_sdk::dpp::dashcore::address::NetworkChecked;
+use dash_sdk::dpp::address_funds::PlatformAddress;
 use dash_sdk::dpp::dashcore::bls_sig_utils::BLSSignature;
 use dash_sdk::dpp::dashcore::network::message_qrinfo::QRInfo;
 use dash_sdk::dpp::dashcore::BlockHash;
@@ -193,8 +192,8 @@ pub enum BackendTaskSuccessResult {
     /// Platform address balances fetched from Platform
     PlatformAddressBalances {
         seed_hash: WalletSeedHash,
-        /// Map of address to (balance, nonce)
-        balances: BTreeMap<Address<NetworkChecked>, (u64, u32)>,
+        /// Map of platform address to (balance, nonce)
+        balances: BTreeMap<PlatformAddress, (u64, u32)>,
     },
     /// Platform credits transferred between addresses
     PlatformCreditsTransferred {
--- a/src/mcp/tools/wallet.rs
+++ b/src/mcp/tools/wallet.rs
@@ -351,10 +351,11 @@ impl AsyncTool<DashMcpService> for FetchPlatformBalances {
 
         match result {
             BackendTaskSuccessResult::PlatformAddressBalances { balances, .. } => {
+                let network = ctx.network();
                 let entries = balances
                     .into_iter()
                     .map(|(addr, (balance, nonce))| PlatformAddressBalance {
-                        address: addr.to_string(),
+                        address: addr.to_bech32m_string(network),
                         balance,
                         nonce,
                     })


thepastaclaw:
✅ Review complete (commit 6be5388c)

src/backend_task/wallet/fetch_platform_address_balances.rs
```diff
@@ -169,7 +170,11 @@ impl AppContext {
             wallet
                 .platform_address_info
                 .iter()
-                .map(|(addr, info)| (addr.clone(), (info.balance, info.nonce)))
+                .filter_map(|(addr, info)| {
+                    PlatformAddress::try_from(addr.clone())
+                        .ok()
+                        .map(|pa| (pa, (info.balance, info.nonce)))
+                })
```
thepastaclaw, line 177: **🟡 Suggestion: Silent address drop via filter_map + .ok() loses data without trace**

Convergent finding (claude + codex). Verified at lines 173-177: `PlatformAddress::try_from(addr.clone()).ok()` silently drops any address that isn't P2PKH/P2SH. For wallet-derived addresses this should never fail (they're always P2PKH), but if it did, the balance would vanish from the result while the task reports success. The old code used `.map()` which was infallible. Consider logging at warn level so the failure is observable.

Note: the logging block at lines 102-113 already has the exact same pattern (`PlatformAddress::try_from(...).unwrap_or_else(|_| addr.to_string())`) — it logs a fallback rather than dropping. The return-path should be at least as defensive.

<details>

```suggestion
.filter_map(|(addr, info)| {
                    match PlatformAddress::try_from(addr.clone()) {
                        Ok(pa) => Some((pa, (info.balance, info.nonce))),
                        Err(e) => {
                            tracing::warn!("Skipping address {addr} — not convertible to PlatformAddress: {e}");
                            None
                        }
                    }
                })
```

</details>

<sub>source: unknown</sub>

<details>

```

In `src/backend_task/wallet/fetch_platform_address_balances.rs`:
- [SUGGESTION] lines 173-177: Silent address drop via filter_map + .ok() loses data without trace
  Convergent finding (claude + codex). Verified at lines 173-177: `PlatformAddress::try_from(addr.clone()).ok()` silently drops any address that isn't P2PKH/P2SH. For wallet-derived addresses this should never fail (they're always P2PKH), but if it did, the balance would vanish from the result while the task reports success. The old code used `.map()` which was infallible. Consider logging at warn level so the failure is observable.

Note: the logging block at lines 102-113 already has the exact same pattern (`PlatformAddress::try_from(...).unwrap_or_else(|_| addr.to_string())`) — it logs a fallback rather than dropping. The return-path should be at least as defensive.
```
</details>

src/ui/wallets/wallets_screen/mod.rs
```diff
@@ -2881,9 +2881,11 @@ impl ScreenLike for WalletsBalancesScreen {
...
                     && wallet.seed_hash() == seed_hash
                 {
-                    // Update balances in the wallet
-                    for (addr, (balance, nonce)) in balances {
-                        wallet.set_platform_address_info(addr, balance, nonce);
+                    // Convert PlatformAddress back to Core Address for wallet storage
+                    let network = self.app_context.network();
+                    for (platform_addr, (balance, nonce)) in balances {
+                        let core_addr = platform_addr.to_address_with_network(network);
+                        wallet.set_platform_address_info(core_addr, balance, nonce);
```
thepastaclaw, line 2888: **💬 Medium: Network-mismatch regression: UI rebinds PlatformAddress with current network, not originating network**

Verified regression introduced by this PR. Before the change, `balances` was `BTreeMap<Address<NetworkChecked>, ...>` — the address carried the network from the backend task's AppContext, which is captured at dispatch time (app.rs:925). The old handler passed addresses directly to `wallet.set_platform_address_info()` with no network rebinding.

Now `PlatformAddress` is network-agnostic (stores only 20-byte hashes, confirmed in SDK source). Line 2885 reads `self.app_context.network()` which reflects the *current* screen context — `change_context()` swaps this Arc on network switch. If the user switches networks while the fetch is in flight, `to_address_with_network()` would encode the addresses under the wrong network.

The MCP path (wallet.rs:354) is safe because `ctx` is a snapshot captured at invocation start.

Fix: embed the originating `Network` in the `PlatformAddressBalances` result variant and use it in the UI handler instead of reading from the current context.

<sub>source: unknown</sub>

thepastaclaw (commented):
## Code Review

PR correctly moves Core→Platform address conversion to the producer side, fixing bech32m serialization in MCP. The `.map()` → `.filter_map(.ok())` change introduces silent balance drops on conversion failure (convergent finding across agents). The type change also introduces a network-mismatch regression: previously Address<NetworkChecked> carried the network from the backend; now PlatformAddress is network-agnostic and the UI rebinds from the current context, which can differ after a network switch during the async task.

*Reviewed commit: 6be5388c*

🟡 1 suggestion(s) | 💬 2 nitpick(s)

### 1 additional finding

<details>
<summary>💬 <b>nitpick:</b> Redundant scoped `use PlatformAddress` import — now imported at module level</summary>

`src/backend_task/wallet/fetch_platform_address_balances.rs (line 103)`

Verified. Line 8 imports `PlatformAddress` at module scope (added by this PR). The scoped `use dash_sdk::dpp::address_funds::PlatformAddress;` at line 103 inside the logging loop is now redundant. Pre-existing code, but this PR is the right place to clean it up since it added the module-level import.

</details>

<details>

```

In `src/backend_task/wallet/fetch_platform_address_balances.rs`:
- [SUGGESTION] lines 173-177: Silent address drop via filter_map + .ok() loses data without trace
  Convergent finding (claude + codex). Verified at lines 173-177: `PlatformAddress::try_from(addr.clone()).ok()` silently drops any address that isn't P2PKH/P2SH. For wallet-derived addresses this should never fail (they're always P2PKH), but if it did, the balance would vanish from the result while the task reports success. The old code used `.map()` which was infallible. Consider logging at warn level so the failure is observable.

Note: the logging block at lines 102-113 already has the exact same pattern (`PlatformAddress::try_from(...).unwrap_or_else(|_| addr.to_string())`) — it logs a fallback rather than dropping. The return-path should be at least as defensive.
```
</details>

thepastaclaw (commented):
## Code Review

Small, well-scoped fix that centralizes Core->Platform address conversion in the backend task producer. The type swap is sound and consumers are coherently updated. Two real concerns: a silent error-swallow in the new filter_map (both agents converge here), and missing regression coverage for the bech32m output contract. Two architecture nitpicks about duplicated/round-trip conversion logic are accurate but out of scope for this bugfix.

*Reviewed commit: 6be5388c*

### 2 additional findings

<details>
<summary>🟡 <b>suggestion:</b> No regression test for bech32m output and PlatformAddress round-trip</summary>

`src/mcp/tools/wallet.rs (lines 333-369)`

This PR fixes a user-visible serialization bug (MCP `platform_addresses_list` returning Core/Base58 instead of bech32m) by changing the type in `BackendTaskSuccessResult::PlatformAddressBalances`, converting in the backend, formatting via `to_bech32m_string()` here, and converting back in the wallet screen. That cross-module behavior can regress quietly without a test that locks in the contract. Add a focused test that asserts `FetchPlatformBalances` produces `tdash1...`/`dash1...` strings from a `PlatformAddressBalances` result and preserves the associated balance/nonce values, so a future refactor cannot reintroduce Core formatting at this boundary.

</details>
<details>
<summary>💬 <b>nitpick:</b> Producer-side logging duplicates the Core->Platform conversion done later</summary>

`src/backend_task/wallet/fetch_platform_address_balances.rs (lines 100-113)`

Lines 102-113 iterate `provider.found_balances()` (Core `Address`) and re-implement Core->Platform->bech32m via `PlatformAddress::try_from(addr.clone()).map(|p| p.to_bech32m_string(self.network)).unwrap_or_else(|_| addr.to_string())`. The block at lines 173-177 then performs the same conversion over a superset of the same wallet's addresses. Logging after `balances` is built — iterating over the already-typed `PlatformAddress` keys — would remove this duplication and the only place a non-bech32m address could appear in logs (the `unwrap_or_else` fallback). Pure refactor; no behavior change.

</details>

<details>

```

In `src/backend_task/wallet/fetch_platform_address_balances.rs`:
- [SUGGESTION] lines 170-178: filter_map silently drops addresses whose PlatformAddress conversion fails
  The new `.filter_map(|(addr, info)| PlatformAddress::try_from(addr.clone()).ok().map(...))` discards entries whose `try_from` returns `Err` with no log and no error. Previously this mapping was infallible, so every `wallet.platform_address_info` entry was guaranteed to round-trip back to consumers. After the change, an unsupported address type silently disappears from both the MCP `platform_addresses_list` response and the wallet UI refresh — exactly the failure mode that issue #652 (referenced two lines above) was meant to prevent (UI losing nonces for missing addresses).

In practice, DET only derives BIP32 P2PKH addresses for platform payment, so this should not fire today. But this function already returns `Result<_, TaskError>` and a dedicated `TaskError::AddressConversionFailed { #[source] source: Box<ProtocolError> }` variant exists at `src/backend_task/error.rs:342` for exactly this case. Either propagate the error via that variant, or at minimum log a `tracing::warn!` when an entry is dropped — matching the `tracing::warn!` style used elsewhere in this same block (lines 121, 149, 160).

In `src/mcp/tools/wallet.rs`:
- [SUGGESTION] lines 333-369: No regression test for bech32m output and PlatformAddress round-trip
  This PR fixes a user-visible serialization bug (MCP `platform_addresses_list` returning Core/Base58 instead of bech32m) by changing the type in `BackendTaskSuccessResult::PlatformAddressBalances`, converting in the backend, formatting via `to_bech32m_string()` here, and converting back in the wallet screen. That cross-module behavior can regress quietly without a test that locks in the contract. Add a focused test that asserts `FetchPlatformBalances` produces `tdash1...`/`dash1...` strings from a `PlatformAddressBalances` result and preserves the associated balance/nonce values, so a future refactor cannot reintroduce Core formatting at this boundary.
```
</details>

src/backend_task/wallet/fetch_platform_address_balances.rs
```diff
@@ -169,7 +170,11 @@ impl AppContext {
...
                 .collect()
```
thepastaclaw, line 178: **🟡 Suggestion: filter_map silently drops addresses whose PlatformAddress conversion fails**

The new `.filter_map(|(addr, info)| PlatformAddress::try_from(addr.clone()).ok().map(...))` discards entries whose `try_from` returns `Err` with no log and no error. Previously this mapping was infallible, so every `wallet.platform_address_info` entry was guaranteed to round-trip back to consumers. After the change, an unsupported address type silently disappears from both the MCP `platform_addresses_list` response and the wallet UI refresh — exactly the failure mode that issue #652 (referenced two lines above) was meant to prevent (UI losing nonces for missing addresses).

In practice, DET only derives BIP32 P2PKH addresses for platform payment, so this should not fire today. But this function already returns `Result<_, TaskError>` and a dedicated `TaskError::AddressConversionFailed { #[source] source: Box<ProtocolError> }` variant exists at `src/backend_task/error.rs:342` for exactly this case. Either propagate the error via that variant, or at minimum log a `tracing::warn!` when an entry is dropped — matching the `tracing::warn!` style used elsewhere in this same block (lines 121, 149, 160).

<details>

```suggestion
            wallet
                .platform_address_info
                .iter()
                .map(|(addr, info)| {
                    PlatformAddress::try_from(addr.clone())
                        .map(|pa| (pa, (info.balance, info.nonce)))
                        .map_err(|e| crate::backend_task::error::TaskError::AddressConversionFailed {
                            source: Box::new(e),
                        })
                })
                .collect::<Result<_, _>>()?
```

</details>

<details>

```

In `src/backend_task/wallet/fetch_platform_address_balances.rs`:
- [SUGGESTION] lines 170-178: filter_map silently drops addresses whose PlatformAddress conversion fails
  The new `.filter_map(|(addr, info)| PlatformAddress::try_from(addr.clone()).ok().map(...))` discards entries whose `try_from` returns `Err` with no log and no error. Previously this mapping was infallible, so every `wallet.platform_address_info` entry was guaranteed to round-trip back to consumers. After the change, an unsupported address type silently disappears from both the MCP `platform_addresses_list` response and the wallet UI refresh — exactly the failure mode that issue #652 (referenced two lines above) was meant to prevent (UI losing nonces for missing addresses).

In practice, DET only derives BIP32 P2PKH addresses for platform payment, so this should not fire today. But this function already returns `Result<_, TaskError>` and a dedicated `TaskError::AddressConversionFailed { #[source] source: Box<ProtocolError> }` variant exists at `src/backend_task/error.rs:342` for exactly this case. Either propagate the error via that variant, or at minimum log a `tracing::warn!` when an entry is dropped — matching the `tracing::warn!` style used elsewhere in this same block (lines 121, 149, 160).
```
</details>

src/ui/wallets/wallets_screen/mod.rs
```diff
@@ -2881,9 +2881,11 @@ impl ScreenLike for WalletsBalancesScreen {
...
                     }
                 }
```
thepastaclaw, line 2890: **💬 Nitpick: Platform->Core roundtrip in the UI consumer undercuts the stated simplification**

The PR description frames the change as "conversion happens once at the source, not at every consumer", but this consumer still converts Platform->Core via `to_address_with_network(network)` because `wallet.platform_address_info` and `Wallet::set_platform_address_info` are keyed by Core `Address`. So the producer does Core->Platform, the UI converts back Platform->Core, and `set_platform_address_info` (`src/model/wallet/mod.rs:2229`) immediately converts Address->PlatformAddress again to canonicalize/dedupe.

This is functionally correct (P2PKH/P2SH on a single network roundtrips losslessly) but means every new `PlatformAddressBalances` consumer that wants to update wallet state will repeat this dance. Out of scope for a bugfix PR — worth filing as follow-up to migrate `wallet.platform_address_info` to be keyed on `PlatformAddress` directly.

<sub>source: ['claude']</sub>

thepastaclaw:
Closing as already addressed on current `v1.0-dev`: `BackendTaskSuccessResult::PlatformAddressBalances` now carries `BTreeMap<PlatformAddress, (u64, u32)>`, and MCP formats those addresses from the Platform type. The old type-conversion patch is no longer needed.

PR #811, opened 2026-04-01
fix: preserve existing identity data when loading by wallet

lklimek:
## Summary

Imagine you're a user who registered a DPNS name for your identity. You go to Load Identity > By Wallet to reload it (maybe after switching networks or reinstalling). Previously, the reload would silently delete your alias — the identity entry was completely overwritten with fresh-but-incomplete data. Now the app merges new data with existing records, preserving your alias and wallet associations.

### Key changes

- **Alias derivation**: `load_identity_from_wallet.rs` now derives the alias from the first DPNS name (same logic as `load_identity.rs`), instead of hardcoding `alias: None`
- **Database merge**: `insert_local_qualified_identity()` changed from `INSERT OR REPLACE` (destructive overwrite) to `INSERT ... ON CONFLICT DO UPDATE` with `COALESCE` — preserves existing non-null `alias`, `wallet`, and `wallet_index` when the incoming value is NULL

### What gets preserved vs updated

| Field | Behavior |
|-------|----------|
| `data` (identity blob) | Always updated (fresh from Platform) |
| `alias` | Kept if existing is non-null and incoming is null |
| `wallet` / `wallet_index` | Kept if existing is non-null and incoming is null |
| `identity_type` | Always updated |
| `status` | Always updated |
| `network` | Always updated |

## Test plan

- [ ] `cargo test --all-features --workspace` passes (511 tests)
- [ ] Load an identity by ID with a DPNS name → alias is set
- [ ] Reload the same identity via By Wallet → alias is preserved
- [ ] Load a new identity via By Wallet with DPNS name → alias is derived from DPNS name
- [ ] Load a new identity via By Wallet without DPNS name → alias remains NULL (no regression)

### DIFF
--- a/src/backend_task/identity/load_identity_from_wallet.rs
+++ b/src/backend_task/identity/load_identity_from_wallet.rs
@@ -222,13 +222,20 @@ impl AppContext {
 
         let wallet_seed_hash = wallet_arc_ref.wallet.read()?.seed_hash();
 
+        // Derive alias from DPNS names if available (same logic as load_identity.rs)
+        let alias = if !maybe_owned_dpns_names.is_empty() {
+            Some(format!("{}.dash", maybe_owned_dpns_names[0].name))
+        } else {
+            None
+        };
+
         let mut qualified_identity = QualifiedIdentity {
             identity: identity.clone(),
             associated_voter_identity: None,
             associated_operator_identity: None,
             associated_owner_key_id: None,
             identity_type: IdentityType::User,
-            alias: None,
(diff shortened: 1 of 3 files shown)


thepastaclaw:
Closing as already addressed by the current architecture. `load_identity_from_wallet` now carries the user alias from an existing record and updates it in place, while identity persistence has moved from the old SQL `INSERT OR REPLACE` path to the k/v identity store. The old implementation is therefore both solved and structurally obsolete.

PR #813, opened 2026-04-01
fix(test): stop orphaned SPV on panic to prevent data directory lock

lklimek:
## Summary

Imagine you're running the backend E2E tests and the framework wallet is unfunded. The init panics at "Framework wallet has no spendable balance". The next test retries init — but hits "Data directory locked" because the first SPV instance is still running on the shared tokio runtime, holding the lock file. Every subsequent test fails the same way. You have to manually kill the process and delete the lock file.

Now a global panic hook cancels the SPV tasks immediately, releasing the lock so the retry can start fresh.

### Root cause

`tokio::sync::OnceCell` does not cache panicked inits — if `BackendTestContext::init()` panics, the next test retries from scratch. But the orphaned SPV tasks from the panicked init keep running on the shared tokio runtime, holding the `testnet.lock` file via `flock`. The retry creates a new `AppContext` + `SpvManager` that fails at `DiskStorageManager::new()` with "Data directory locked".

### Fix

- Clone the `TaskManager`'s `CancellationToken` before passing it to `AppContext`
- Stash it in a `static Mutex<Option<CancellationToken>>`
- Install a global panic hook (via `std::panic::set_hook`) that cancels the token on any panic — SPV's stop_token is a child, so it fires and the lock is released
- On init retry, check and cancel any leftover token first

## Test plan

- [x] `cargo clippy --all-features --all-targets -- -D warnings` passes
- [x] Parallel E2E run: 0 "Data directory locked" errors (was 2-5 before fix)
- [x] Panic hook fires correctly on test panics (verified in logs)
- [x] Single-threaded E2E run still works

## Summary by CodeRabbit

* **Tests**
  * Improved backend test framework lifecycle and cancellation handling during initialization. Tests now detect and stop leftover background tasks from prior runs, install a one-time panic handler to cancel task managers on panic, and ensure cleanup after successful setup—resulting in more reliable parallel test execution and reduced resource interference between runs.

### DIFF
--- a/tests/backend-e2e/framework/harness.rs
+++ b/tests/backend-e2e/framework/harness.rs
@@ -37,6 +37,17 @@ use std::time::Duration;
 /// runtime context (via `block_on`) rather than spawning a nested one.
 static CTX: tokio::sync::OnceCell<BackendTestContext> = tokio::sync::OnceCell::const_new();
 
+/// Cancellation token for the task manager that owns SPV tasks.
+///
+/// `tokio::sync::OnceCell` does not cache panicked inits — if `init()`
+/// panics (e.g. framework wallet unfunded), the next test retries from
+/// scratch. But the orphaned SPV tasks from the panicked init still run
+/// on the shared tokio runtime, holding the data directory lock. A global
+/// panic hook cancels this token, which stops SPV (its stop_token is a
+/// child) and releases the lock file.
+static SPV_CANCEL: std::sync::Mutex<Option<tokio_util::sync::CancellationToken>> =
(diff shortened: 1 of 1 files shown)


shumkov (approved):

thepastaclaw:
### Review Gate

Commit: `bec0bbdc`

- ⏳ **Debounce**: 5m ago (need 30m)
- ⏳ **CI checks**: checks still running (2 pending)
- ✅ **CodeRabbit review**: comment found

- [ ] Run review now (check to override)

PR #830, opened 2026-04-16
fix(spv): atomically extend BIP44 receive pool on wallet import

lklimek:
## Summary

Fixes the second of the four layers behind the zero-balance-on-import bug tracked in [#829](https://github.com/dashpay/dash-evo-tool/issues/829): addresses generated via the UI beyond the WalletManager's default BIP44 gap limit (30) are not re-registered into the live monitored set on startup, so SPV filter matching misses any UTXO at those higher indices.

This PR is now **narrowly focused on the bug-2 fix only**. The platform pin bump and `clear_data_dir` fix have been extracted into [#834](https://github.com/dashpay/dash-evo-tool/pull/834) and must land first.

## Reproduction

1. In an existing wallet, use the UI to generate 35+ receive addresses (indices 0–34).
2. Restart DET.
3. Send a testnet transaction to the address at an index ≥ 30 (beyond the WalletManager default gap limit).
4. Wait for the transaction to confirm, then observe the wallet balance in the UI.

**Expected:** The balance at the receiving address appears, because the WalletManager's monitored address set should include every index that exists in DET's local database.

**Actual:** The address shows zero balance. On restart, the WalletManager is reconstructed with `WalletAccountCreationOptions::Default` which pre-generates BIP44 indices 0–29 only. Any indices DET previously stored in `wallet_addresses` beyond 29 are never re-registered into the WalletManager, so the live monitored address set is narrower than DET's on-disk state. SPV filter matching misses any UTXO at indices ≥ 30.

Observed with a reference testnet wallet that has a multi-output transaction at BIP44 indices 0 and 32–40 (motivating tx: `56bdab0ac5dfefc47e136eedeede07a83ff0f2cb753683578fd41677975f8b32` in block 1,459,352). Only index 0 shows balance; 32–40 do not.

## Root cause

Upstream's BIP44 gap-limit reactive extension works as designed: if at least one address within the current gap is observed, the pool extends. But DET stores more addresses in its local DB than it restores to the live monitored set on startup, and there's no code path that reconciles the two.

There's also a second, subtler race: earlier drafts extended the pool *after* `load_wallet_from_seed` returned, meaning `wm.wallet_count()` could cross the threshold that `run_spv_loop` waits on while the pool was still at its default size — SPV's filter pipeline then started matching against an incomplete set.

## Fix

[...]

Issue #829, opened 2026-04-16
bug: Imported wallet shows zero balance — historical UTXOs invisible even after restart


lklimek:
## Summary

After importing a wallet into Dash Evo Tool (via mnemonic or from the local database on startup), the wallet shows **zero balance** even when it has confirmed funds on-chain — or shows only some addresses' balances while others remain invisible. Closing and reopening DET sometimes partially helps but is not reliable. In the worst case, the zero balance persists across multiple app restarts until the user clears all SPV data and lets the tool re-sync from scratch.

## Does this match what you're seeing?

- ✅ You imported a mnemonic for a testnet (or mainnet) wallet you know has funds, and DET displays the balance as **0 DASH**.
- ✅ You see **some addresses with balance but not others**, even though both have confirmed transactions on an explorer.
- ✅ You restarted DET and the balance **still didn't appear** (or only partially appeared).
- ✅ "Clear SPV Data" followed by a full re-sync eventually fixes it — but that takes minutes and disrupts any other wallets synced in the app.
- ✅ Log lines contain `Filters already synced to <height>` or `Loading stored filters <tip> to <tip> into current batch` shortly after importing.

If two or more of these match, this is the issue.

## Steps to reproduce

1. Start DET with SPV mode on testnet (or mainnet).
2. Either let SPV fully sync against any existing wallet, **or** start fresh with no wallets.
3. Import a mnemonic for a wallet that has historical transactions — ideally one with BIP44 external indices that include gaps wider than the default gap limit of 30 (e.g. indices 0 and 32–40).
4. Observe the wallet balance in the UI.

Alternative (gap-limit variant):
1. In an existing wallet, generate 35+ receive addresses via the UI.
2. Restart the app.
3. Send a transaction to the address at index 32+.
4. Observe zero balance for that address in the UI.

## Expected

All UTXOs on addresses belonging to the wallet appear in the balance, regardless of whether the wallet was imported mid-session, loaded from the local database at startup, or sits behind a gap wider than 30 between addresses.

## Actual

- Index 0 may appear after one or more app restarts.
- Indices 32–40 never appear (neither the addresses nor their balances are visible in the UI).
- In the worst case, the wallet shows a persistent zero balance across multiple app restarts, and only a full **"Clear SPV Data" + resync from checkpoint** reliably surfaces the historical UTXOs.

[...]

PR #651, opened 2026-02-24
fix(fees): consolidate Core and Platform fee estimation

lklimek:
## Summary

- **Fix Core (L1) tx fee estimation**: Replace hardcoded `DEFAULT_TX_FEE` of 1000 duffs with size-based fee calculation. Fixes "min relay fee not met" errors for larger transactions (e.g., multi-input asset locks). Deduplicates 3 copies of `estimate_p2pkh_tx_size` into a single utility in `fee_estimation.rs`.
- **Consolidate Platform (L2) fee estimation**: Add unified `PlatformFeeEstimator` methods that compute `max(legacy, transition-based)` fees internally, so callers don't need dual-estimation logic. Reduces public API surface by making internal helpers private and removing dead code.
- **UTXO removal consistency**: `FundWithUtxo` paths in identity registration/top-up now use `remove_selected_utxos()` instead of manual `retain()`.
- **Add comprehensive tests**: 9 tests covering the unified transition-based fee methods plus a regression test for the relay fee fix.

## Test plan

- Automated: `cargo test --all-features --workspace` — includes new fee estimation tests
- Manual test scenarios: [Fee consolidation](docs/ai-design/2026-02-24-fee-consolidation/manual-test-scenarios.md)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

## Summary by CodeRabbit

* **New Features**
  * Unified, iterative fee estimation across wallet, asset-lock and platform flows for more accurate fees and multi-recipient fee subtraction.

* **Bug Fixes**
  * More reliable fee reconciliation with automatic UTXO reloads and improved dust/change handling.
  * Better error propagation in send flows so failures are surfaced and handled gracefully.

* **Documentation**
  * Added comprehensive test scenarios and guidance for fee estimation, asset-lock, platform transfers, and UTXO handling.

PR #833, opened 2026-04-20
fix(spv): wait for wallet bootstrap before filter scan begins

lklimek:
## Summary

On app startup, imported wallets displayed a zero balance until the user manually cleared SPV data and re-synced. Root cause is an init-order + signaling race between two parallel async flows during `AppContext` bring-up. SPV's filter-scan wait loop was fed a stale reading of `expected_wallet_count`, interpreted it as "no wallets", skipped the wait, and ran the full filter scan against an empty monitored set. Fix: consult the database when the in-memory wallet count is zero.

## Reproduction

1. Launch DET with one or more wallets already in the local database (any scenario where DET was used before and quit cleanly).
2. On startup, don't touch anything — let SPV begin syncing.

**Expected:** SPV waits until the DB wallets are loaded into the in-memory set before it begins scanning compact block filters. Wallet transactions present on-chain are matched during the scan. Balances show correctly on the main screen.

**Actual:** SPV begins filter scanning with zero monitored addresses. Filter matches zero. Wallet transactions are permanently skipped for the session. Balances display as 0 until the user manually clears SPV data and relaunches.

## Root cause — init-order + signaling race

Two independent async flows fire during `AppContext` construction:

- **Flow A — SPV bring-up (`start_spv`)**: count wallets → `spv_manager.start(expected)` → `run_spv_loop` → filter scan.
- **Flow B — Wallet bootstrap (`bootstrap_loaded_wallets`)**: read DB → unlock seeds → register in-memory → load into SPV's monitored set.

Flow A needs Flow B's result (monitored addresses) before scanning. The code already has the right synchronisation primitive — `run_spv_loop` contains a wait loop that blocks until `wm.wallet_count() >= expected_wallet_count`. The bug is in the **signal** passed to that primitive.

### The stale signal

`start_spv()` computed `expected_wallet_count` like this:

```rust
let expected_wallets = self
    .wallets
    .read()
    .map(|guard| {
        guard.values()
            .filter(|w| {
                w.read().ok()
                 .is_some_and(|g| g.is_open() && g.seed_bytes().is_ok())
            })
            .count()
    })
    .unwrap_or(0);
```

Two filters: `is_open()` (seed decrypted) and `seed_bytes().is_ok()` (bytes accessible). At the moment `start_spv()` runs, Flow B hasn't decrypted any seeds yet — those happen later on a separate task. So `expected_wallets = 0`.

When `expected_wallets = 0`, `run_spv_loop` skips its wait loop:

```rust
if expected_wallet_count > 0 {
    // wait for wallet_count() to reach expected, then proceed
}
// otherwise fall straight through to build_client / filter scan
```

Downstream, `build_client(has_wallets=false)` sets `start_height = u32::MAX` as its "no wallets, use latest checkpoint" sentinel. The filter scan runs against an empty address set. Zero matches. Every wallet transaction on-chain is invisible until the next full rescan.

### Classic TOCTOU

Check was done too early to be meaningful. The wait was correctly skipped — but for the wrong reason.

## Fix shape

Signal correction, not reorder. Keep parallelism, but make `expected_wallets` reflect the *eventual* count rather than the *current* count.

```rust
// Fallback: if no open wallets in memory yet but the database has
// wallets for this network, expect at least 1.
// bootstrap_loaded_wallets will load them shortly and the SPV wait
// loop will block until load completes.
let expected_wallets = if expected_wallets == 0 {
    self.db
        .wallet_count_for_network(&self.network)
        .unwrap_or(0)
        .min(1)
} else {
    expected_wallets
};
```

Two key ideas:

1. **Consult the database when in-memory count is 0.** New helper `Database::wallet_count_for_network()` runs a parameterised `SELECT COUNT(*) FROM wallet WHERE network = ?`. Returns the count of wallets that *will* be loaded shortly, even if none are loaded right now.

2. **Clamp to a minimum of 1, not the full count.** SPV's wait condition is `wallet_count() >= expected_wallets`. As soon as `bootstrap_loaded_wallets` loads the **first** wallet, the wait unblocks. Reporting 1 (rather than N) means one successful unlock releases the wait without being blocked by wallets that might later fail to decrypt (corrupt seed, wrong password, etc.).

## Why signal-correction over reorder

- **Minimal risk.** Reordering startup has broader consequences — Flow A also registers reconcile/finality listeners that Flow B triggers; serialising the flows could create its own races or require adding signals in the opposite direction.
- **Matches the intent of the existing code.** The wait loop was clearly designed for this exact scenario. The bug was feeding it a stale reading of the state it was waiting on.
- **Defensive in the cold-start case.** On a genuinely fresh install with no DB rows, both paths return 0, the wait loop correctly skips, and the scan runs from the checkpoint as intended. The fix doesn't break that path.

## Files changed

| File | Lines | Change |
|------|------:|--------|
| `src/context/wallet_lifecycle.rs` | +12 | DB fallback block inserted immediately after the in-memory `expected_wallets` calculation in `start_spv()` |
| `src/database/wallet.rs` | +51 | New `Database::wallet_count_for_network()` helper + `insert_test_wallet` test helper + `test_wallet_count_for_network_counts_per_network` unit test |

Total: **+63 lines, 0 removed**. Zero changes to the SPV manager, wallet lifecycle outside `start_spv`, or any other subsystem.

### Automated
- [x] `cargo build --all-features` — clean
- [x] `cargo clippy --all-features --all-targets -- -D warnings` — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo test --all-features --workspace --lib` — 453 tests pass, including the new `test_wallet_count_for_network_counts_per_network` which verifies the DB helper counts per-network correctly.

### Manual on testnet
- [ ] **Fix validates.** Use any DB with at least one wallet already imported. Quit DET, relaunch. Expected: wallet balance appears on first load without manual clear-and-resync. Log contains `SPV: all wallets loaded with monitored addresses, proceeding expected=1 loaded=N` where N is the actual wallet count.
- [ ] **Fresh-install regression.** `rm -rf ~/.config/dash-evo-tool/data.db ~/.config/dash-evo-tool/spv/`, launch DET with no wallets. Expected: SPV starts from latest checkpoint without hanging on the wait loop (since `expected_wallets = 0`).
- [ ] **Single failed unlock regression.** With two wallets in DB, simulate one having a bad password or corrupt seed. Expected: the other wallet still loads, SPV wait releases on the first successful unlock (that's why we clamp to `.min(1)`).

## Scope

This PR carries **only** the Flow-A / Flow-B startup signaling fix (bug 1). Sibling PR [#830](https://github.com/dashpay/dash-evo-tool/pull/830) carries the atomic BIP44 receive-pool extension fix (bug 2) and the platform pin bump to rust-dashcore 309fac8.

Earlier drafts of #830 also carried a mid-session-wallet-import SPV restart workaround (bug 3). That workaround has since been stripped from #830 because it is blocked upstream by [rust-dashcore#651 constraint #4](https://github.com/dashpay/rust-dashcore/issues/651) (`SegmentCache::first_valid_offset` under-reports on reload, clamping `scan_start` back at the chain tip). Until upstream #651 lands, the sanctioned workaround for mid-session imports is **Clear SPV Data**. The stripped attempt is preserved on [`archive/spv-wallet-address-sync-2026-04-20`](https://github.com/dashpay/dash-evo-tool/tree/archive/spv-wallet-address-sync-2026-04-20) for future reference.

If this PR lands first, the `51bd044e` commit currently on #830 (which contains an earlier combined bug-1+bug-2 fix) will be rebased / dropped — its bug-1 portion is identical to what this PR ships.

## Related

- Part of the zero-balance fix chain tracked in [#829](https://github.com/dashpay/dash-evo-tool/issues/829).
- Sibling PR: [#830](https://github.com/dashpay/dash-evo-tool/pull/830) — atomic pool extension + platform bump.
- Upstream: [rust-dashcore#651](https://github.com/dashpay/rust-dashcore/issues/651) — tracks the upstream work needed to reinstate a mid-session-import rescan path.

### DIFF
--- a/src/context/wallet_lifecycle.rs
+++ b/src/context/wallet_lifecycle.rs
@@ -69,6 +69,18 @@ impl AppContext {
                     .count()
             })
             .unwrap_or(0);
+        // Fallback: if no open wallets in memory yet but the database has
+        // wallets for this network, expect at least 1.
+        // bootstrap_loaded_wallets will load them shortly and the SPV wait
+        // loop will block until load completes.
+        let expected_wallets = if expected_wallets == 0 {
+            self.db
+                .wallet_count_for_network(&self.network)
+                .unwrap_or(0)
+                .min(1)
+        } else {
+            expected_wallets
+        };
         // Register reconcile channel BEFORE starting SPV so the event handlers
         // (spawned inside run_spv_loop) always capture a valid sender.
         self.spv_setup_reconcile_listener();
--- a/src/context/wallet_lifecycle.rs
+++ b/src/context/wallet_lifecycle.rs
@@ -74,10 +74,7 @@ impl AppContext {
         // bootstrap_loaded_wallets will load them shortly and the SPV wait
         // loop will block until load completes.
         let expected_wallets = if expected_wallets == 0 {
-            self.db
-                .wallet_count_for_network(&self.network)
-                .unwrap_or(0)
-                .min(1)
+            self.db.wallet_count_for_network(&self.network)?.min(1)
         } else {
             expected_wallets
         };
--- a/src/database/wallet.rs
+++ b/src/database/wallet.rs
@@ -92,6 +92,16 @@ impl Database {
         tx.commit()
     }
 
+    /// Count how many wallets are stored in the database for a given network.
+    pub fn wallet_count_for_network(&self, network: &Network) -> rusqlite::Result<usize> {
+        let conn = self.conn.lock().unwrap();
+        conn.query_row(
+            "SELECT COUNT(*) FROM wallet WHERE network = ?",
+            [network.to_string()],
+            |row| row.get(0),
+        )
+    }
+
     /// Update the Dash Core wallet name for an HD wallet.
     ///
     /// Returns `Ok(true)` if exactly one row was updated, `Ok(false)` if no
@@ -1883,5 +1893,46 @@ mod tests {
             )
             .expect("Failed to query total_received");
         assert_eq!(total_received, 10_000_000);
+    }
+
+    fn insert_test_wallet(db: &Database, seed_hash: &[u8; 32], network: Network) {
+        let network_str = network.to_string();
+        let conn = db.conn.lock().unwrap();
+        conn.execute(
+            "INSERT INTO wallet (seed_hash, encrypted_seed, salt, nonce, master_ecdsa_bip44_account_0_epk, uses_password, network)
+             VALUES (?, ?, ?, ?, ?, 0, ?)",
+            rusqlite::params![
+                seed_hash.as_slice(),
+                vec![0u8; 64],
+                vec![0u8; 16],
+                vec![0u8; 12],
+                vec![0u8; 78],
+                network_str,
+            ],
+        )
+        .expect("Failed to insert test wallet");
+    }
+
+    #[test]
+    fn test_wallet_count_for_network_counts_per_network() {
+        let db = create_test_database().expect("create db");
+        let mut seed_a = create_test_seed_hash();
+        let mut seed_b = create_test_seed_hash();
+        seed_a[0] = 0xA0;
+        seed_b[0] = 0xB0;
+
+        insert_test_wallet(&db, &seed_a, Network::Testnet);
+        insert_test_wallet(&db, &seed_b, Network::Testnet);
+
+        assert_eq!(
+            db.wallet_count_for_network(&Network::Testnet)
+                .expect("count"),
+            2
+        );
+        assert_eq!(
+            db.wallet_count_for_network(&Network::Mainnet)
+                .expect("count"),
+            0
+        );
     }
 }


thepastaclaw:
✅ Review complete (commit 8c3c9aa5)

src/context/wallet_lifecycle.rs
```diff
@@ -69,6 +69,15 @@ impl AppContext {
...
             .unwrap_or(0);
+        // Fallback: if no open wallets in memory yet but the database has
+        // wallets for this network, expect at least 1.
+        // bootstrap_loaded_wallets will load them shortly and the SPV wait
+        // loop will block until load completes.
+        let expected_wallets = if expected_wallets == 0 {
+            self.db.wallet_count_for_network(&self.network)?.min(1)
+        } else {
+            expected_wallets
+        };
```
thepastaclaw, line 80: **🔴 Blocking: `.min(1)` fallback undercounts wallets and counts password-protected ones that cannot auto-unlock**

`run_spv_loop` waits until `wm.wallet_count() >= expected_wallet_count` before building the client and starting the initial filter scan (`src/spv/manager.rs:1045-1075`). The DB fallback here has two problems that both cause the race this PR is meant to fix to persist:

1. **Multi-wallet undercount.** Clamping the DB count with `.min(1)` means that when the database contains multiple wallets for this network, SPV only waits for the first one to load. As soon as wallet #1 is imported into the `WalletManager`, `wallet_count() >= 1` passes and the compact-filter scan starts, but wallets #2..N are still being loaded asynchronously by `bootstrap_loaded_wallets`, so their addresses are absent from `monitored_addresses()` when the scan begins. Their historical transactions are skipped for the session, reproducing the original bug for every wallet past the first.

2. **Password-protected wallets are counted but will never load.** The in-memory count above filters by `is_open() && seed_bytes().is_ok()`, meaning only wallets with decrypted seeds. The DB fallback is unfiltered, so password-protected wallets (still `WalletSeed::Closed` after bootstrap) raise the expected count but `handle_wallet_unlocked` → `wallet_seed_snapshot` returns `None` for them, so `queue_spv_wallet_load` is never called. `wallet_count()` stays at 0 and `run_spv_loop` burns its full 30-second deadline on every startup for users whose wallets are all password-protected.

Both issues share the same root cause: the fallback should count the full set of wallets that can actually auto-load into SPV, not collapse everything to 1 and not include locked wallets that can never satisfy the wait condition.

<details>

```suggestion
        // Fallback: if no open wallets in memory yet, expect as many wallets
        // as the database holds for this network that can auto-unlock on
        // bootstrap. Password-protected wallets are excluded because
        // bootstrap_loaded_wallets cannot unlock them, so waiting for them
        // would burn run_spv_loop's 30s deadline without progress.
        let expected_wallets = if expected_wallets == 0 {
            self.db
                .auto_openable_wallet_count_for_network(&self.network)?
        } else {
            expected_wallets
        };
```

</details>

<details>

```

In `src/context/wallet_lifecycle.rs`:
- [BLOCKING] lines 76-80: `.min(1)` fallback undercounts wallets and counts password-protected ones that cannot auto-unlock
  `run_spv_loop` waits until `wm.wallet_count() >= expected_wallet_count` before building the client and starting the initial filter scan (`src/spv/manager.rs:1045-1075`). The DB fallback here has two problems that both cause the race this PR is meant to fix to persist:

1. **Multi-wallet undercount.** Clamping the DB count with `.min(1)` means that when the database contains multiple wallets for this network, SPV only waits for the first one to load. As soon as wallet #1 is imported into the `WalletManager`, `wallet_count() >= 1` passes and the compact-filter scan starts, but wallets #2..N are still being loaded asynchronously by `bootstrap_loaded_wallets`, so their addresses are absent from `monitored_addresses()` when the scan begins. Their historical transactions are skipped for the session, reproducing the original bug for every wallet past the first.

2. **Password-protected wallets are counted but will never load.** The in-memory count above filters by `is_open() && seed_bytes().is_ok()`, meaning only wallets with decrypted seeds. The DB fallback is unfiltered, so password-protected wallets (still `WalletSeed::Closed` after bootstrap) raise the expected count but `handle_wallet_unlocked` → `wallet_seed_snapshot` returns `None` for them, so `queue_spv_wallet_load` is never called. `wallet_count()` stays at 0 and `run_spv_loop` burns its full 30-second deadline on every startup for users whose wallets are all password-protected.

Both issues share the same root cause: the fallback should count the full set of wallets that can actually auto-load into SPV, not collapse everything to 1 and not include locked wallets that can never satisfy the wait condition.
```
</details>

thepastaclaw, line 80: **🟡 Suggestion: DB error now aborts `start_spv` instead of falling back to 0**

This commit replaced `.unwrap_or(0).min(1)` with `?.min(1)`, converting a previously silent fallback into a hard failure. Any transient `rusqlite::Error` now propagates as `TaskError::Database` and prevents SPV from starting at all, even though the wait loop could safely proceed with `expected_wallets = 0`. The in-memory wallet read immediately above still uses `.unwrap_or(0)` for lock poisoning, so the two paths treat "couldn't determine count" asymmetrically. This is low-risk in practice because the DB is already open at this point, but it is still a behavior change worth making explicit or softening with a warning + fallback.

<details>

```suggestion
        let expected_wallets = if expected_wallets == 0 {
            self.db
                .wallet_count_for_network(&self.network)
                .inspect_err(|e| tracing::warn!(?e, "wallet_count_for_network failed, assuming 0"))
                .unwrap_or(0)
                .min(1)
        } else {
            expected_wallets
        };
```

</details>

<sub>source: ['claude']</sub>

<details>

```

In `src/context/wallet_lifecycle.rs`:
- [SUGGESTION] lines 76-80: DB error now aborts `start_spv` instead of falling back to 0
  This commit replaced `.unwrap_or(0).min(1)` with `?.min(1)`, converting a previously silent fallback into a hard failure. Any transient `rusqlite::Error` now propagates as `TaskError::Database` and prevents SPV from starting at all, even though the wait loop could safely proceed with `expected_wallets = 0`. The in-memory wallet read immediately above still uses `.unwrap_or(0)` for lock poisoning, so the two paths treat "couldn't determine count" asymmetrically. This is low-risk in practice because the DB is already open at this point, but it is still a behavior change worth making explicit or softening with a warning + fallback.
```
</details>

src/database/wallet.rs
```diff
@@ -1884,4 +1894,45 @@ mod tests {
...
+        insert_test_wallet(&db, &seed_b, Network::Testnet);
+
+        assert_eq!(
+            db.wallet_count_for_network(&Network::Testnet)
+                .expect("count"),
+            2
+        );
+        assert_eq!(
+            db.wallet_count_for_network(&Network::Mainnet)
+                .expect("count"),
```
thepastaclaw, line 1934: **🟡 Suggestion: Signaling fix has no automated coverage — only the DB helper is tested**

`test_wallet_count_for_network_counts_per_network` verifies that the new helper returns the right per-network count, but the actual bug fixed here — that `start_spv` passes a non-zero `expected_wallet_count` when DB wallets exist but memory is still empty — is only exercised by the manual test plan. The original race was a subtle TOCTOU between async flows and the fallback is easy to regress. A unit test that either calls `start_spv()` in a controlled state, or extracts the `expected_wallets` computation into a pure helper and tests its branches, would give cheap protection.

<sub>source: ['claude']</sub>

<details>

```

In `src/database/wallet.rs`:
- [SUGGESTION] lines 1898-1934: Signaling fix has no automated coverage — only the DB helper is tested
  `test_wallet_count_for_network_counts_per_network` verifies that the new helper returns the right per-network count, but the actual bug fixed here — that `start_spv` passes a non-zero `expected_wallet_count` when DB wallets exist but memory is still empty — is only exercised by the manual test plan. The original race was a subtle TOCTOU between async flows and the fallback is easy to regress. A unit test that either calls `start_spv()` in a controlled state, or extracts the `expected_wallets` computation into a pure helper and tests its branches, would give cheap protection.
```
</details>

src/database/wallet.rs
```diff
@@ -1884,4 +1894,45 @@ mod tests {
...
+                seed_hash.as_slice(),
+                vec![0u8; 64],
+                vec![0u8; 16],
+                vec![0u8; 12],
+                vec![0u8; 78],
+                network_str,
+            ],
+        )
+        .expect("Failed to insert test wallet");
+    }
```
thepastaclaw, line 1914: **💬 Nitpick: `insert_test_wallet` hand-writes INSERT SQL instead of going through `store_wallet`**

The new test helper covers only part of the `wallet` table schema and relies on the remaining columns staying NULLable. That creates a second source of truth for the write path: a future schema change can break this helper without exercising the real insertion logic. The surrounding DB tests already use real `Wallet` fixtures with `store_wallet` / `store_wallet_with_addresses`, which both avoids duplicated SQL and exercises a realistic path.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Small SPV startup race fix has a load-bearing logic bug: the `.min(1)` clamp on the DB-based wallet count both undercounts multi-wallet setups and includes password-protected wallets that `bootstrap_loaded_wallets` cannot auto-unlock. Secondary concerns are that the DB read now aborts `start_spv` on transient errors instead of falling back to 0, and that the signaling fix itself is only covered by manual testing; the new DB helper test uses a hand-written insert helper that may drift from the real write path.

*Reviewed commit: 8c3c9aa5*

🔴 1 blocking | 🟡 2 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/context/wallet_lifecycle.rs`:
- [BLOCKING] lines 76-80: `.min(1)` fallback undercounts wallets and counts password-protected ones that cannot auto-unlock
  `run_spv_loop` waits until `wm.wallet_count() >= expected_wallet_count` before building the client and starting the initial filter scan (`src/spv/manager.rs:1045-1075`). The DB fallback here has two problems that both cause the race this PR is meant to fix to persist:

1. **Multi-wallet undercount.** Clamping the DB count with `.min(1)` means that when the database contains multiple wallets for this network, SPV only waits for the first one to load. As soon as wallet #1 is imported into the `WalletManager`, `wallet_count() >= 1` passes and the compact-filter scan starts, but wallets #2..N are still being loaded asynchronously by `bootstrap_loaded_wallets`, so their addresses are absent from `monitored_addresses()` when the scan begins. Their historical transactions are skipped for the session, reproducing the original bug for every wallet past the first.

2. **Password-protected wallets are counted but will never load.** The in-memory count above filters by `is_open() && seed_bytes().is_ok()`, meaning only wallets with decrypted seeds. The DB fallback is unfiltered, so password-protected wallets (still `WalletSeed::Closed` after bootstrap) raise the expected count but `handle_wallet_unlocked` → `wallet_seed_snapshot` returns `None` for them, so `queue_spv_wallet_load` is never called. `wallet_count()` stays at 0 and `run_spv_loop` burns its full 30-second deadline on every startup for users whose wallets are all password-protected.

Both issues share the same root cause: the fallback should count the full set of wallets that can actually auto-load into SPV, not collapse everything to 1 and not include locked wallets that can never satisfy the wait condition.
- [SUGGESTION] lines 76-80: DB error now aborts `start_spv` instead of falling back to 0
  This commit replaced `.unwrap_or(0).min(1)` with `?.min(1)`, converting a previously silent fallback into a hard failure. Any transient `rusqlite::Error` now propagates as `TaskError::Database` and prevents SPV from starting at all, even though the wait loop could safely proceed with `expected_wallets = 0`. The in-memory wallet read immediately above still uses `.unwrap_or(0)` for lock poisoning, so the two paths treat "couldn't determine count" asymmetrically. This is low-risk in practice because the DB is already open at this point, but it is still a behavior change worth making explicit or softening with a warning + fallback.

In `src/database/wallet.rs`:
- [SUGGESTION] lines 1898-1934: Signaling fix has no automated coverage — only the DB helper is tested
  `test_wallet_count_for_network_counts_per_network` verifies that the new helper returns the right per-network count, but the actual bug fixed here — that `start_spv` passes a non-zero `expected_wallet_count` when DB wallets exist but memory is still empty — is only exercised by the manual test plan. The original race was a subtle TOCTOU between async flows and the fallback is easy to regress. A unit test that either calls `start_spv()` in a controlled state, or extracts the `expected_wallets` computation into a pure helper and tests its branches, would give cheap protection.
```
</details>

src/database/wallet.rs
```diff
@@ -92,6 +92,16 @@ impl Database {
...
 
+    /// Count how many wallets are stored in the database for a given network.
+    pub fn wallet_count_for_network(&self, network: &Network) -> rusqlite::Result<usize> {
+        let conn = self.conn.lock().unwrap();
+        conn.query_row(
+            "SELECT COUNT(*) FROM wallet WHERE network = ?",
+            [network.to_string()],
+            |row| row.get(0),
+        )
+    }
```
thepastaclaw, line 103: **🔴 Blocking: DB fallback counts password-locked wallets and forces a 30s startup stall on password-only profiles**

`wallet_count_for_network` counts every row in `wallet`, but `get_wallets()` reconstructs password-protected rows as `WalletSeed::Closed` (src/database/wallet.rs:549-553), and `wallet_seed_snapshot()` only yields a load when `is_open()` is true (src/context/wallet_lifecycle.rs:292-305). For a profile whose only wallets are password-protected, the new fallback at src/context/wallet_lifecycle.rs:76-78 sets `expected_wallets = 1` while `SpvManager::wallet_count()` stays at 0 because `bootstrap_loaded_wallets` cannot decrypt them. `run_spv_loop` then sleeps for the full 30s deadline at src/spv/manager.rs:1045-1069 before proceeding. Pre-PR this case proceeded immediately (with the empty-scan bug); post-PR it adds a deterministic 30s startup delay on top of still requiring a separate unlock-triggered SPV reload. The fix is to count only wallets bootstrap can actually load — filter the SQL by `uses_password = 0`. That also lets us drop the `.min(1)` clamp safely, since locked wallets are no longer in the total.

<details>

```suggestion
    /// Count how many wallets are stored in the database for a given network
    /// that will be auto-loaded into SPV at startup (i.e. not password-protected).
    pub fn wallet_count_for_network(&self, network: &Network) -> rusqlite::Result<usize> {
        let conn = self.conn.lock().unwrap();
        conn.query_row(
            "SELECT COUNT(*) FROM wallet WHERE network = ? AND uses_password = 0",
            [network.to_string()],
            |row| row.get(0),
        )
    }
```

</details>

<details>

```

In `src/database/wallet.rs`:
- [BLOCKING] lines 95-103: DB fallback counts password-locked wallets and forces a 30s startup stall on password-only profiles
  `wallet_count_for_network` counts every row in `wallet`, but `get_wallets()` reconstructs password-protected rows as `WalletSeed::Closed` (src/database/wallet.rs:549-553), and `wallet_seed_snapshot()` only yields a load when `is_open()` is true (src/context/wallet_lifecycle.rs:292-305). For a profile whose only wallets are password-protected, the new fallback at src/context/wallet_lifecycle.rs:76-78 sets `expected_wallets = 1` while `SpvManager::wallet_count()` stays at 0 because `bootstrap_loaded_wallets` cannot decrypt them. `run_spv_loop` then sleeps for the full 30s deadline at src/spv/manager.rs:1045-1069 before proceeding. Pre-PR this case proceeded immediately (with the empty-scan bug); post-PR it adds a deterministic 30s startup delay on top of still requiring a separate unlock-triggered SPV reload. The fix is to count only wallets bootstrap can actually load — filter the SQL by `uses_password = 0`. That also lets us drop the `.min(1)` clamp safely, since locked wallets are no longer in the total.
```
</details>

src/context/wallet_lifecycle.rs
thepastaclaw, line 80: **🟡 Suggestion: `.min(1)` clamps the count at most 1 — comment says "at least 1" and the clamp masks multi-wallet startup**

`expected_wallets = self.db.wallet_count_for_network(...)?.min(1)` returns 0 when the DB count is 0 and 1 for any positive count — i.e. clamps at most 1, not at least 1. Two issues: (1) the inline comment says "expect at least 1," which a future reader (or `clippy::redundant_min_max` autofix) may "correct" to `.max(1)`, silently regressing the cold-start path (DB empty → `expected = 1` → 30s wait, then empty scan); (2) the clamp also limits the multi-wallet startup case: with N>1 not-password-protected wallets in DB and 0 currently in memory, the SPV wait loop releases as soon as the first wallet finishes importing (src/spv/manager.rs:1051), even though wallets 2..N may still be deriving addresses, so they may miss the initial filter scan. If the DB query is filtered to `uses_password = 0` (see other finding), the count then represents exactly the loadable set and the `.min(1)` clamp can be removed entirely — `expected_wallets` becomes the precise number of wallets bootstrap will populate. If keeping the clamp, expand the comment to capture both (a) why we clamp (avoid stalling on a wallet that fails to decrypt) and (b) why 0 is preserved when the DB is empty.

<details>

```suggestion
        // Fallback: when no wallets are open in memory yet but the database
        // has wallets for this network, expect bootstrap to load all of them.
        // wallet_count_for_network already excludes password-protected rows
        // (which bootstrap cannot decrypt), so the count matches exactly what
        // run_spv_loop will see in monitored_addresses() once loads complete.
        let expected_wallets = if expected_wallets == 0 {
            self.db.wallet_count_for_network(&self.network)?
        } else {
            expected_wallets
        };
```

</details>

<details>

```

In `src/context/wallet_lifecycle.rs`:
- [SUGGESTION] lines 72-80: `.min(1)` clamps the count at most 1 — comment says "at least 1" and the clamp masks multi-wallet startup
  `expected_wallets = self.db.wallet_count_for_network(...)?.min(1)` returns 0 when the DB count is 0 and 1 for any positive count — i.e. clamps at most 1, not at least 1. Two issues: (1) the inline comment says "expect at least 1," which a future reader (or `clippy::redundant_min_max` autofix) may "correct" to `.max(1)`, silently regressing the cold-start path (DB empty → `expected = 1` → 30s wait, then empty scan); (2) the clamp also limits the multi-wallet startup case: with N>1 not-password-protected wallets in DB and 0 currently in memory, the SPV wait loop releases as soon as the first wallet finishes importing (src/spv/manager.rs:1051), even though wallets 2..N may still be deriving addresses, so they may miss the initial filter scan. If the DB query is filtered to `uses_password = 0` (see other finding), the count then represents exactly the loadable set and the `.min(1)` clamp can be removed entirely — `expected_wallets` becomes the precise number of wallets bootstrap will populate. If keeping the clamp, expand the comment to capture both (a) why we clamp (avoid stalling on a wallet that fails to decrypt) and (b) why 0 is preserved when the DB is empty.
```
</details>

src/database/wallet.rs
```diff
@@ -1884,4 +1894,45 @@ mod tests {
...
+            db.wallet_count_for_network(&Network::Testnet)
+                .expect("count"),
+            2
+        );
+        assert_eq!(
+            db.wallet_count_for_network(&Network::Mainnet)
+                .expect("count"),
+            0
+        );
+    }
```
thepastaclaw, line 1937: **💬 Nitpick: No test exercises the start_spv() fallback branch**

`test_wallet_count_for_network_counts_per_network` covers the new helper directly, which is appropriate at the DB layer. However the actual race-prevention logic — the `if expected_wallets == 0 { ... }` block in `start_spv()` — has no automated coverage and is verified only via the manual test plan. Given the bug class (TOCTOU between two parallel async flows during AppContext bring-up), a regression here is exactly what tests should catch. Either extract the small computation into a pure helper that takes `(in_memory_count, db_count) -> usize` and assert (0,0)→0, (0,N)→N, (M,_)→M, or add a kittest scenario that pre-populates the DB and asserts SPV starts with a non-zero `expected_wallets`. Optional but cheap insurance against a future `.min`/`.max` flip.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Targeted, well-scoped startup race fix. Two real issues with the new fallback: (1) the DB count includes password-locked wallets that bootstrap can't load, producing a guaranteed 30s startup stall for password-only profiles — a clear UX regression; (2) the `.min(1)` clamp leaves the multi-wallet startup case under-served and the inline comment misreads the operator. Both converge on a cleaner fix: filter the DB count by `uses_password = 0` and drop the clamp, so `expected_wallets` represents exactly the set that bootstrap will actually load. Test coverage on the new fallback branch in `start_spv()` is also missing.

*Reviewed commit: 8c3c9aa5*

🔴 1 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/database/wallet.rs`:
- [BLOCKING] lines 95-103: DB fallback counts password-locked wallets and forces a 30s startup stall on password-only profiles
  `wallet_count_for_network` counts every row in `wallet`, but `get_wallets()` reconstructs password-protected rows as `WalletSeed::Closed` (src/database/wallet.rs:549-553), and `wallet_seed_snapshot()` only yields a load when `is_open()` is true (src/context/wallet_lifecycle.rs:292-305). For a profile whose only wallets are password-protected, the new fallback at src/context/wallet_lifecycle.rs:76-78 sets `expected_wallets = 1` while `SpvManager::wallet_count()` stays at 0 because `bootstrap_loaded_wallets` cannot decrypt them. `run_spv_loop` then sleeps for the full 30s deadline at src/spv/manager.rs:1045-1069 before proceeding. Pre-PR this case proceeded immediately (with the empty-scan bug); post-PR it adds a deterministic 30s startup delay on top of still requiring a separate unlock-triggered SPV reload. The fix is to count only wallets bootstrap can actually load — filter the SQL by `uses_password = 0`. That also lets us drop the `.min(1)` clamp safely, since locked wallets are no longer in the total.

In `src/context/wallet_lifecycle.rs`:
- [SUGGESTION] lines 72-80: `.min(1)` clamps the count at most 1 — comment says "at least 1" and the clamp masks multi-wallet startup
  `expected_wallets = self.db.wallet_count_for_network(...)?.min(1)` returns 0 when the DB count is 0 and 1 for any positive count — i.e. clamps at most 1, not at least 1. Two issues: (1) the inline comment says "expect at least 1," which a future reader (or `clippy::redundant_min_max` autofix) may "correct" to `.max(1)`, silently regressing the cold-start path (DB empty → `expected = 1` → 30s wait, then empty scan); (2) the clamp also limits the multi-wallet startup case: with N>1 not-password-protected wallets in DB and 0 currently in memory, the SPV wait loop releases as soon as the first wallet finishes importing (src/spv/manager.rs:1051), even though wallets 2..N may still be deriving addresses, so they may miss the initial filter scan. If the DB query is filtered to `uses_password = 0` (see other finding), the count then represents exactly the loadable set and the `.min(1)` clamp can be removed entirely — `expected_wallets` becomes the precise number of wallets bootstrap will populate. If keeping the clamp, expand the comment to capture both (a) why we clamp (avoid stalling on a wallet that fails to decrypt) and (b) why 0 is preserved when the DB is empty.
```
</details>

Fresh dispatcher-requested review for exact queued SHA 8c3c9aa5ca3a53556106b6b9017bfb6f7d00541d. Existing/same-SHA review dedup was intentionally bypassed per coordinator instruction.

lklimek:
Closing as superseded by the platform-wallet backend rewrite.

This PR was cut against the pre-migration codebase. The mechanism it patches (`run_spv_loop` / `expected_wallet_count` wait-signal on the DET-side `SpvManager`) no longer exists — `src/spv/manager.rs` is gone. The init-order race it targets is now structurally closed by the `ensure_wallet_backend_and_start_spv` chokepoint (`src/context/wallet_lifecycle/spv.rs:141-158`), which wires the wallet backend before SPV start.

It also carries an outstanding blocking review that was never addressed: a `.min(1)` fallback that guarantees a 30s startup stall for password-locked wallets, a multi-wallet undercount that reproduces the underlying UTXO-visibility bug for wallets #2..N, a `?` change that lets a transient DB error abort `start_spv`, and zero automated coverage for the signaling fix.

The underlying "historical UTXOs invisible" symptom is tracked fresh against current code in #829 — a real fix will need to target `wallet_backend::register_wallet_from_seed` via the upstream `platform-wallet` crate, not this approach.

PR #839, opened 2026-04-23
fix(spv): no-op ensure_address_imported when backend is SPV

lklimek:
## Summary

- **Bug.** In SPV mode (default since #836), creating an identity through "Address with QR code" fails before rendering the QR with `Could not connect to Dash Core at 127.0.0.1:19998`. The same failure path affects identity top-up via QR and the Create Asset Lock wallet tool.
- **Root cause.** Two functions in `AppContext` unconditionally called Dash Core RPC to watch newly-derived receive addresses. In SPV mode there is no Dash Core node to reach, so those calls failed with `CoreRpcConnectionFailed` and the user never saw the QR code:
  - `ensure_address_imported` (`src/context/mod.rs`) — called via `get_address_info` + `import_address`
  - `try_import_address` (`src/context/wallet_lifecycle.rs`) — the async companion used in the wallet lifecycle path
- **Fix.** Both functions now early-return `Ok(())` when `core_backend_mode() != CoreBackendMode::Rpc`. The import is unnecessary in SPV mode because `Wallet::receive_address -> unused_bip_44_public_key -> register_address` already inserts the address into `known_addresses`; the SPV wallet manager (loaded from the same xprv via `WalletAccountCreationOptions::Default`) watches the entire standard BIP44 account. Incoming UTXOs reach `wallet.utxos` through `received_transaction_finality`, which is the exact map the QR flow polls via `capture_qr_funding_utxo_if_available`. Matches the precedent already in `Wallet::register_address` and the broadcast dispatcher shipped in #837.
- **Doc fix.** Updated the `ensure_address_imported` doc comment to accurately describe the SPV UTXO delivery path (previously the comment described only the RPC path).

**Diff scope:** 4 source files + 1 `.gitignore` entry — SPV no-op guards, doc correction, and the PasswordInput secondary changes below.

## Secondary changes — PasswordInput improvements

While working in the password input area, several small correctness and quality issues were addressed:

| Change | Commit | Detail |
|---|---|---|
| Default input width | `8ba789db` | Sets a sensible default width so the widget renders correctly before the caller provides an explicit size |
| Char limit 66 → 64 for private-key hex input | `9a0e0029` | A hex-encoded 32-byte private key is exactly 64 characters; 66 allowed impossible inputs that would silently fail later |
| Avoid double-alloc per frame in width measurement | `15a069c0` | `Typography::measure_text_width` now accepts `impl Into<String>` instead of `String`, eliminating one allocation per frame during rendering |
| Single source of truth for reveal-icon margin | `6e6cf2c4` | `PASSWORD_INPUT_REVEAL_ICON_WIDTH` constant now drives the `TextEdit` right-margin directly, removing a duplicated magic number |

## Verification performed

- **HD vs single-key wallet reachability.** All three affected screens (`AddNewIdentityScreen`, `TopUpIdentityScreen`, `CreateAssetLockScreen`) hold `Option<Arc<RwLock<Wallet>>>` — `Wallet` is structurally HD-only (has `master_bip44_ecdsa_extended_public_key`). `SingleKeyWallet` is a distinct type kept in a separate `AppContext.single_key_wallets` map, so single-key wallets cannot reach these flows. No risk of silently hiding a single-key-in-SPV bug.
- **rust-dashcore PR #554 interaction.** Orthogonal. #554 adds DIP-9 identity-authentication account types at `m/9'/coinType'/5'/subfeature'/index`. This PR only affects BIP44 funding addresses at `m/44'/coinType'/0'/0/index`, already covered by `WalletAccountCreationOptions::Default`.
- **`list_core_wallets` / `try_detect_core_wallet_for_address` siblings.** Audited all callers: `add_new_wallet_screen` and `import_mnemonic_screen` have explicit SPV short-circuits; `wallets_screen/mod.rs` and `with_wallet_recovery` only fire on `TaskError::CoreWalletNotConfigured`, which in SPV mode is preempted by `CoreRpcConnectionFailed`. No sibling fixes required.

## Test plan

**SPV QR funding (primary fix)**
- [ ] In SPV mode (default), open "Create Identity" → select a wallet → choose "Address with QR Code" → verify a QR code and pay-URI render instead of the connection error.
- [ ] In SPV mode, send a small Dash payment to the displayed address on testnet → verify the flow advances from "Waiting for funds" to "Funds received" without a banner error, and the identity registration proceeds.
- [ ] In SPV mode, repeat on the "Top Up Identity" → "Address with QR Code" flow.
- [ ] In SPV mode, open Tools → "Create Asset Lock" → verify the QR renders and funds landing advances state.
- [ ] In **Expert + Local Dash Core** mode, repeat the identity-creation QR flow → confirm `ensure_address_imported` still imports the address into Core's wallet (existing behaviour unchanged).

**PasswordInput secondary changes**
- [ ] Open any screen that shows a private-key import field — verify it accepts exactly 64 hex characters and rejects a 65th character.
- [ ] Visually confirm the password input renders at a sensible default width before any explicit size is set by the caller.
- [ ] `cargo +nightly fmt --all -- --check` → clean.
- [ ] `cargo clippy --all-features --all-targets -- -D warnings` → clean.

## Release Notes

* **Bug Fixes**
  * Fixed wallet address import behavior to properly handle SPV mode operations

* **Improvements**
  * Enhanced password input field width calculation based on character limits
  * Added 66-character maximum limit to private key input field

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ test_db*
 **/.DS_Store
 explorer.log
 .gitaipconfig
+.claude/worktrees
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -737,12 +737,18 @@ impl AppContext {
     }
 
     /// Import address into Core, ignoring errors. For best-effort registration.
+    ///
+    /// No-ops in SPV mode — mirroring [`Self::ensure_address_imported`] — because there is no
+    /// RPC client to talk to and every call would fail silently, wasting resources.
     pub fn try_import_address(
         &self,
         address: &Address,
         core_wallet_name: Option<&str>,
         label: Option<&str>,
     ) {
+        if self.core_backend_mode() != CoreBackendMode::Rpc {
+            return;
+        }
         if let Ok(client) = self.core_client_for_wallet(core_wallet_name) {
             let _ = client.import_address(address, label, Some(false));
         }
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -707,13 +707,14 @@ impl AppContext {
     /// Uses `core_wallet_name` to target the right wallet on multi-wallet nodes.
     /// No-op if the address is already watched/mine.
     ///
-    /// In SPV mode this is a no-op: there is no Dash Core node to import into,
-    /// and HD-derived addresses are already tracked by the SPV subsystem via
-    /// `Wallet::register_address` (which populates `known_addresses` and feeds
-    /// the SPV bloom filter). Incoming UTXOs on these addresses are delivered
-    /// through `received_transaction_finality()` from InstantLock/ChainLock
-    /// events — the same path the QR funding flows poll via
-    /// `capture_qr_funding_utxo_if_available`.
+    /// In SPV mode this is a no-op: there is no Dash Core node to import into.
+    /// HD-derived addresses are tracked by the SPV wallet manager watching the
+    /// BIP44 account derived from the same xprv — `Wallet::register_address`
+    /// records them in wallet state (`known_addresses`) but does not update the
+    /// SPV bloom filter directly. Incoming UTXOs for these addresses are
+    /// populated via the SPV reconciliation path (`reconcile_spv_wallets()`),
+    /// which is what downstream checks such as
+    /// `capture_qr_funding_utxo_if_available` observe.
     pub fn ensure_address_imported(
         &self,
         address: &Address,
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -706,12 +706,23 @@ impl AppContext {
     /// Import an address into the correct Core wallet if it's not already known.
     /// Uses `core_wallet_name` to target the right wallet on multi-wallet nodes.
     /// No-op if the address is already watched/mine.
+    ///
+    /// In SPV mode this is a no-op: there is no Dash Core node to import into,
+    /// and HD-derived addresses are already tracked by the SPV subsystem via
+    /// `Wallet::register_address` (which populates `known_addresses` and feeds
+    /// the SPV bloom filter). Incoming UTXOs on these addresses are delivered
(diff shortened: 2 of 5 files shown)


thepastaclaw:
✅ Review complete (commit c777f594)

thepastaclaw (commented):
@/tmp/review_839_body.txt

thepastaclaw (commented):
This rerun reviewed commit `764977c0` specifically.

Overall: the PR fixes the user-visible SPV QR funding failure in `ensure_address_imported()`, but the same SPV receive-address path still contains one unconditional Core import helper. I also confirmed that the add-key UI still allows impossible 66-character "hex" private-key input even though validation only accepts 32 decoded bytes.

Findings:

1. `src/context/mod.rs:739`
   - `try_import_address()` still performs Core RPC work in SPV mode.
   - `ensure_address_imported()` now short-circuits outside RPC mode, but `Wallet::unused_bip_44_public_key()` still reaches `try_import_address()` when deriving a fresh receive address. In SPV mode that means we still do a pointless synchronous Core RPC attempt whose error is silently discarded. `register_address()` already guards its own import path by backend mode, so `try_import_address()` should do the same.

2. `src/ui/identities/keys/add_key_screen.rs:79`
   - The add-key forms allow 65-66 character hex private keys that can never validate.
   - All three constructors use `with_char_limit(66)`, but submission only accepts `hex::decode(...)` output with decoded length exactly 32 bytes, i.e. 64 hex characters. The UI currently permits two extra characters that are guaranteed to fail later at submit time.

thepastaclaw (commented):
## Code Review

Both agents independently reviewed the four-file delta from 764977c0 to c777f594 and found no new issues. The two prior findings (SPV guard in try_import_address, char limit in add-key forms) are both fixed on the current head. The follow-up password-input cleanup is a benign refactor with no correctness issues.

*Reviewed commit: c777f594*

PR #859, opened 2026-05-15
perf: detect system theme off UI thread

### LINKED ISSUE #782: perf: move OS theme detection to a background task
### User Story

As a **user running Dash Evo Tool on Linux**, I want the app to remain responsive when the OS theme detection is slow, so that I don't experience frame drops or UI stutter every 2 seconds.

### Problem

`dark_light::detect()` queries the XDG Desktop Portal via D-Bus, which has a hardcoded ~100ms timeout. On systems without a portal daemon (minimal WMs, headless setups, some Wayland compositors), this blocks the main UI thread every 2 seconds during the theme polling cycle.

Currently the timeout is handled gracefully (previous theme is preserved, error is logged at debug level), but the blocking call can still cause frame drops.

### Proposed Solution

Move theme detection off the main thread:

1. Use `tokio::task::spawn_blocking` (or a dedicated thread) to run `dark_light::detect()` asynchronously
2. Send the result back via a channel or shared atomic state
3. The main `update()` loop reads the latest detected theme without blocking

This follows the existing app task pattern — the detection becomes a lightweight background poll rather than a synchronous call in the frame loop.

### Context

- `dark_light` v2.0.0 is the latest version; no upstream fix for the timeout
- The timeout is a known issue: [winit#2893](https://github.com/rust-windowing/winit/issues/2893), [egui#3061](https://github.com/emilk/egui/issues/3061)
- Related PR: #781 (throttle + cache theme detection)

thepastaclaw:
# Pull Request

## Summary

- Move recurring system theme detection out of the egui update path.
- Run the throttled detection probe in a `spawn_blocking` worker.
- Keep startup and explicit preference changes synchronous to preserve the
  existing no-flash behavior.
- Tag background detection results with a generation so stale probes cannot
  overwrite newer preference state.

Fixes dashpay/dash-evo-tool#782.

## Validation

- `cargo fmt --check`
- `cargo check`
- Pre-PR code review:
  `code-review dashpay/dash-evo-tool upstream/v1.0-dev
  tracker-782-theme-detection-bg "Move recurring OS theme detection off the
  egui update path by running the throttled dark_light::detect poll in a
  spawn_blocking task, returning generation-tagged results over a channel, and
  applying cached theme state without blocking periodic UI frames while
  preserving startup and explicit preference-change behavior"`
  - Result: Recommendation: ship

## Summary by CodeRabbit

* **Refactor**
  * System theme detection now runs asynchronously and non-blocking with smarter throttling and safeguards, improving responsiveness and reducing incorrect or late theme switches.
  * Preference changes apply the resolved theme immediately and avoid redundant background work for a smoother UX.
* **Tests**
  * Added unit tests to ensure reliable detection behavior, throttling, and concurrency handling.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -66,36 +66,141 @@ impl From<Result<BackendTaskSuccessResult, TaskError>> for TaskResult {
     }
 }
 
+/// Minimum interval between consecutive OS-theme probes when preference is
+/// `System`. Prevents thrashing the underlying detector library on every frame.
+const DETECTION_THROTTLE: Duration = Duration::from_secs(2);
+
+/// Maximum time an in-flight detection probe is allowed to occupy the slot
+/// before the watchdog forcibly releases it. Protects against a panicking or
+/// hung detector preventing all future probes.
+const DETECTION_WATCHDOG: Duration = Duration::from_secs(10);
+
 struct ThemeState {
     preference: ThemeMode,
     resolved: ThemeMode,
     last_applied: Option<ThemeMode>,
-    last_checked: Instant,
+    last_checked: Option<Instant>,
+    detection_in_flight: bool,
+    /// When the current in-flight probe was dispatched. Used by the watchdog
+    /// to release a stuck detection slot.
+    detection_dispatched_at: Option<Instant>,
+    detection_generation: u64,
+    detection_sender: mpsc::Sender<(u64, Option<ThemeMode>)>,
+    detection_receiver: mpsc::Receiver<(u64, Option<ThemeMode>)>,
 }
 
 impl ThemeState {
     fn new(preference: ThemeMode) -> Self {
+        let (detection_sender, detection_receiver) = mpsc::channel();
         Self {
             resolved: crate::ui::theme::resolve_theme_mode(preference),
             last_applied: None,
-            last_checked: Instant::now(),
+            last_checked: Some(Instant::now()),
+            detection_in_flight: false,
+            detection_dispatched_at: None,
+            detection_generation: 0,
+            detection_sender,
+            detection_receiver,
             preference,
         }
     }
 
-    /// Polls the OS for system theme changes (throttled to every 2s) and
-    /// applies the theme if it changed. Returns `true` if the theme was applied.
-    fn poll_and_apply(&mut self, ctx: &egui::Context) -> bool {
-        if self.preference == ThemeMode::System {
-            let now = Instant::now();
-            if now.duration_since(self.last_checked) >= Duration::from_secs(2) {
-                self.last_checked = now;
-                if let Some(detected) = crate::ui::theme::try_detect_system_theme()
-                    && detected != self.resolved
-                {
-                    self.resolved = detected;
-                }
+    /// Apply a single detection result. Stale-generation messages are
+    /// discarded (so they cannot overwrite `resolved` after a preference
+    /// change). A matching-generation `None` result clears the in-flight
+    /// slot but leaves `resolved` unchanged. Returns `true` iff `resolved`
+    /// changed.
+    fn handle_detection_result(&mut self, generation: u64, detected: Option<ThemeMode>) -> bool {
+        if generation != self.detection_generation {
+            return false;
+        }
+        self.detection_in_flight = false;
+        self.detection_dispatched_at = None;
+        if self.preference != ThemeMode::System {
+            return false;
+        }
+        match detected {
+            Some(detected) if detected != self.resolved => {
+                self.resolved = detected;
+                true
             }
+            _ => false,
+        }
+    }
+
+    fn drain_detection_results(&mut self) {
+        while let Ok((generation, detected)) = self.detection_receiver.try_recv() {
+            self.handle_detection_result(generation, detected);
+        }
+    }
+
+    /// If a detection probe has been in flight longer than [`DETECTION_WATCHDOG`],
+    /// release the slot so future probes can proceed even if the worker
+    /// panicked or hung. Advances the generation so a delayed result from the
+    /// stuck worker will be treated as stale. Returns `true` if the watchdog
+    /// fired.
+    fn check_detection_watchdog(&mut self, now: Instant) -> bool {
+        if !self.detection_in_flight {
+            return false;
+        }
+        let stuck = match self.detection_dispatched_at {
+            // Defensive: in-flight without a timestamp — treat as stuck.
+            None => true,
+            Some(dispatched_at) => now.duration_since(dispatched_at) >= DETECTION_WATCHDOG,
+        };
+        if stuck {
+            tracing::warn!(
+                "OS theme detection watchdog fired after {:?}; releasing in-flight slot",
+                DETECTION_WATCHDOG
+            );
+            self.detection_in_flight = false;
+            self.detection_dispatched_at = None;
+            self.detection_generation = self.detection_generation.wrapping_add(1);
+        }
+        stuck
+    }
+
+    fn request_system_detection(&mut self, ctx: &egui::Context, now: Instant) {
+        self.check_detection_watchdog(now);
+        if self.detection_in_flight {
+            return;
+        }
+        if self
+            .last_checked
+            .is_some_and(|last_checked| now.duration_since(last_checked) < DETECTION_THROTTLE)
+        {
+            return;
+        }
+
+        self.last_checked = Some(now);
+        self.detection_in_flight = true;
+        self.detection_dispatched_at = Some(now);
+        let sender = self.detection_sender.clone();
+        let generation = self.detection_generation;
+        let ctx = ctx.clone();
+        tokio::task::spawn_blocking(move || {
+            // Catch panics from the underlying `dark_light` crate so a panic
+            // still produces a generation-tagged result and a repaint request,
+            // preventing the in-flight slot from being held forever.
+            let detected = std::panic::catch_unwind(crate::ui::theme::try_detect_system_theme)
+                .unwrap_or_else(|_| {
+                    tracing::warn!("OS theme detection panicked; treating as no result");
+                    None
+                });
+            if sender.send((generation, detected)).is_err() {
+                tracing::debug!("Dropping OS theme detection result because receiver closed");
+            }
+            ctx.request_repaint();
+        });
+    }
+
+    /// Polls asynchronously for OS theme changes (throttled to every 2s) and
+    /// applies the latest detected theme if it changed. Returns `true` if the
+    /// theme was applied.
+    fn poll_and_apply(&mut self, ctx: &egui::Context) -> bool {
+        self.drain_detection_results();
+        if self.preference == ThemeMode::System {
+            self.request_system_detection(ctx, Instant::now());
         }
         if self.last_applied != Some(self.resolved) {
             crate::ui::theme::apply_theme(ctx, self.resolved);
@@ -108,6 +213,10 @@ impl ThemeState {
 
     fn apply_new_preference(&mut self, ctx: &egui::Context, new_theme: ThemeMode) -> bool {
         self.preference = new_theme;
+        self.detection_generation = self.detection_generation.wrapping_add(1);
+        self.detection_in_flight = false;
+        self.detection_dispatched_at = None;
+        let now = Instant::now();
         let mut detection_failed = false;
         self.resolved = if new_theme == ThemeMode::System {
             match crate::ui::theme::try_detect_system_theme() {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
@coderabbitai review

thepastaclaw:
✅ Review complete (commit ed0d022f)

src/app.rs
```diff
@@ -70,32 +70,78 @@ struct ThemeState {
...
+        let generation = self.detection_generation;
+        let ctx = ctx.clone();
+        tokio::task::spawn_blocking(move || {
+            let detected = crate::ui::theme::try_detect_system_theme();
+            if sender.send((generation, detected)).is_err() {
+                tracing::debug!("Dropping OS theme detection result because receiver closed");
+            }
+            ctx.request_repaint();
+        });
+    }
```
thepastaclaw, line 195: **🟡 Suggestion: `detection_in_flight` can latch true forever if the background probe panics or hangs**

`request_system_detection` sets `self.detection_in_flight = true` before `spawn_blocking`, and the flag is only cleared in `drain_detection_results` when a matching-generation result arrives, or in `apply_new_preference` when the user changes preference. Two failure modes leave the flag stuck:

1. **Worker panic** — if `dark_light::detect()` (called inside `try_detect_system_theme`) panics on some platform/version, the cloned sender is dropped via stack unwinding, but the original `detection_sender` remains owned by `ThemeState` (line 76). The channel therefore never becomes `Disconnected`; `try_recv` keeps returning `Empty`. No result tuple is ever delivered for this generation, and `detection_in_flight` stays true for the lifetime of the process.
2. **Worker hang** — if the OS call never returns, the same outcome applies without any panic.

From that point on, every subsequent `request_system_detection` returns at the `if self.detection_in_flight { return; }` guard, so automatic OS-theme tracking is silently disabled until the user manually toggles preference. This is a real resiliency regression versus the prior synchronous path (which couldn't latch). Cheap mitigations: wrap the worker body in `catch_unwind` and always send a generation-tagged result; or treat `last_checked` as a watchdog and reset `detection_in_flight` if a probe outstays a reasonable bound (e.g. 10–30s) before falling through to the throttle check.

<sub>source: ['claude-general', 'codex-rust-quality']</sub>

<details>

```

In `src/app.rs`:
- [SUGGESTION] lines 113-136: `detection_in_flight` can latch true forever if the background probe panics or hangs
  `request_system_detection` sets `self.detection_in_flight = true` before `spawn_blocking`, and the flag is only cleared in `drain_detection_results` when a matching-generation result arrives, or in `apply_new_preference` when the user changes preference. Two failure modes leave the flag stuck:

1. **Worker panic** — if `dark_light::detect()` (called inside `try_detect_system_theme`) panics on some platform/version, the cloned sender is dropped via stack unwinding, but the original `detection_sender` remains owned by `ThemeState` (line 76). The channel therefore never becomes `Disconnected`; `try_recv` keeps returning `Empty`. No result tuple is ever delivered for this generation, and `detection_in_flight` stays true for the lifetime of the process.
2. **Worker hang** — if the OS call never returns, the same outcome applies without any panic.

From that point on, every subsequent `request_system_detection` returns at the `if self.detection_in_flight { return; }` guard, so automatic OS-theme tracking is silently disabled until the user manually toggles preference. This is a real resiliency regression versus the prior synchronous path (which couldn't latch). Cheap mitigations: wrap the worker body in `catch_unwind` and always send a generation-tagged result; or treat `last_checked` as a watchdog and reset `detection_in_flight` if a probe outstays a reasonable bound (e.g. 10–30s) before falling through to the throttle check.
```
</details>

thepastaclaw (commented):
## Code Review

Refactor moves recurring OS theme detection off the egui thread via `spawn_blocking`, with generation-tagged results to discard stale probes. Both reviewers independently flag two genuine concerns: (1) `detection_in_flight` can latch true forever if the worker panics or hangs, since the channel keeps a live sender inside `ThemeState` and the flag is only ever cleared by a matching-generation result or a preference change; (2) the new state machine has no test coverage. Remaining items are minor nits.

*Reviewed commit: 4292e1fc*

### 3 additional findings

<details>
<summary>🟡 <b>suggestion:</b> New generation/throttle/in-flight state machine has no test coverage</summary>

`src/app.rs (lines 69-177)`

`ThemeState` now encodes several non-trivial invariants that are the entire point of this PR: the 2s throttle on probe dispatch, the `detection_in_flight` guard preventing concurrent probes, the generation-tagged discard semantics that protect against stale OS-detected values overwriting an explicit preference, and the `None`-detection fallback that preserves the previous resolved theme. None of this is covered by tests.

Most of it is testable without a tokio runtime or egui context by extracting `drain_detection_results` and the throttle/generation bookkeeping into pure functions taking `(now: Instant, incoming: (u64, Option<ThemeMode>))`. Two high-value cases to lock in: (a) a stale-generation result arriving after `apply_new_preference` must not overwrite `resolved`, and (b) a `None` detection result must leave the prior `resolved` value intact. Without coverage, a future refactor can silently reintroduce theme flashing or stale overwrites.

</details>
<details>
<summary>💬 <b>nitpick:</b> Explain why `apply_new_preference` keeps the synchronous probe</summary>

`src/app.rs (lines 155-176)`

The PR moves recurring detection off the UI thread, but `apply_new_preference` still calls `try_detect_system_theme()` synchronously when the user switches to `System`. This is intentional — it preserves the no-flash transition — but the asymmetry is exactly the kind of thing a future contributor will try to "fix" by making it async, reintroducing flicker. A one-line comment on this branch explaining the rationale would prevent that.

</details>
<details>
<summary>💬 <b>nitpick:</b> `apply_new_preference` updates `last_checked` even for non-System preferences</summary>

`src/app.rs (lines 155-176)`

After switching to Dark or Light, `last_checked` is still bumped to `now`. It's harmless today because `request_system_detection` is only called from `poll_and_apply` when `preference == ThemeMode::System`, so the throttle is never consulted while a fixed preference is active. But it means that on switching back to System, the throttle is measured from the time of the last preference change rather than the last actual detection — small surprise if the throttle is ever reused elsewhere. Either gate the update on `new_theme == System` or rename the field for clarity.

</details>

<details>

```

In `src/app.rs`:
- [SUGGESTION] lines 113-136: `detection_in_flight` can latch true forever if the background probe panics or hangs
  `request_system_detection` sets `self.detection_in_flight = true` before `spawn_blocking`, and the flag is only cleared in `drain_detection_results` when a matching-generation result arrives, or in `apply_new_preference` when the user changes preference. Two failure modes leave the flag stuck:

1. **Worker panic** — if `dark_light::detect()` (called inside `try_detect_system_theme`) panics on some platform/version, the cloned sender is dropped via stack unwinding, but the original `detection_sender` remains owned by `ThemeState` (line 76). The channel therefore never becomes `Disconnected`; `try_recv` keeps returning `Empty`. No result tuple is ever delivered for this generation, and `detection_in_flight` stays true for the lifetime of the process.
2. **Worker hang** — if the OS call never returns, the same outcome applies without any panic.

From that point on, every subsequent `request_system_detection` returns at the `if self.detection_in_flight { return; }` guard, so automatic OS-theme tracking is silently disabled until the user manually toggles preference. This is a real resiliency regression versus the prior synchronous path (which couldn't latch). Cheap mitigations: wrap the worker body in `catch_unwind` and always send a generation-tagged result; or treat `last_checked` as a watchdog and reset `detection_in_flight` if a probe outstays a reasonable bound (e.g. 10–30s) before falling through to the throttle check.
- [SUGGESTION] lines 69-177: New generation/throttle/in-flight state machine has no test coverage
  `ThemeState` now encodes several non-trivial invariants that are the entire point of this PR: the 2s throttle on probe dispatch, the `detection_in_flight` guard preventing concurrent probes, the generation-tagged discard semantics that protect against stale OS-detected values overwriting an explicit preference, and the `None`-detection fallback that preserves the previous resolved theme. None of this is covered by tests.

Most of it is testable without a tokio runtime or egui context by extracting `drain_detection_results` and the throttle/generation bookkeeping into pure functions taking `(now: Instant, incoming: (u64, Option<ThemeMode>))`. Two high-value cases to lock in: (a) a stale-generation result arriving after `apply_new_preference` must not overwrite `resolved`, and (b) a `None` detection result must leave the prior `resolved` value intact. Without coverage, a future refactor can silently reintroduce theme flashing or stale overwrites.
```
</details>

src/app.rs
```diff
@@ -120,13 +229,169 @@ impl ThemeState {
         } else {
             new_theme
         };
-        self.last_checked = Instant::now();
+        self.last_checked = Some(now);
         crate::ui::theme::apply_theme(ctx, self.resolved);
         self.last_applied = Some(self.resolved);
         detection_failed
     }
```
thepastaclaw, line 236: **💬 Nitpick: Document why apply_new_preference keeps a synchronous probe**

The PR's main goal is to move recurring OS theme detection off the egui thread, yet `apply_new_preference` still calls `crate::ui::theme::try_detect_system_theme()` synchronously at line 222 when the user switches to `System`. This is intentional — it preserves the no-flash transition by resolving the theme before the next frame paints — but the asymmetry with the async `request_system_detection` is exactly the kind of thing a future contributor will try to 'fix' by making it async, reintroducing flicker. Add a one-line comment on the `if new_theme == ThemeMode::System` branch noting that this synchronous detection is intentional, to avoid a Light→System theme flash.

<sub>source: ['claude-general', 'claude-rust-quality']</sub>

src/app.rs
```diff
@@ -120,13 +229,169 @@ impl ThemeState {
         } else {
             new_theme
         };
-        self.last_checked = Instant::now();
+        self.last_checked = Some(now);
```
thepastaclaw, line 232: **💬 Nitpick: apply_new_preference updates last_checked even for non-System preferences**

After switching to Dark or Light, `last_checked` is unconditionally bumped to `now` at line 232. It's harmless today because `request_system_detection` is only called from `poll_and_apply` when `preference == ThemeMode::System` (line 202), so the throttle is never consulted while a fixed preference is active. But it means that on switching from (say) Dark back to System, the next probe's throttle window is measured from when Dark was selected — not from the last actual detection. Either gate the assignment on `new_theme == ThemeMode::System`, or rename `last_checked` to something like `last_throttle_anchor` to make the semantics explicit.

<sub>source: ['claude-general', 'claude-rust-quality']</sub>

thepastaclaw (commented):
## Code Review

The refactor successfully addresses both prior blocking concerns: a 10s watchdog (line 142) plus catch_unwind (line 185) prevent the in-flight slot from latching, and a new test module (lines 239-336) covers stale-generation, None-result, in-flight guard, throttle, and watchdog behavior. Only two minor nits remain in apply_new_preference; this is a COMMENT-level review.

*Reviewed commit: ed0d022f*

💬 2 nitpick(s)

src/app.rs
```diff
@@ -1060,36 +1060,141 @@ async fn forward_backend_task_join_error(
...
+/// hung detector preventing all future probes.
+const DETECTION_WATCHDOG: Duration = Duration::from_secs(10);
+
 struct ThemeState {
     preference: ThemeMode,
     resolved: ThemeMode,
     last_applied: Option<ThemeMode>,
-    last_checked: Instant,
+    last_checked: Option<Instant>,
+    detection_in_flight: bool,
```
lklimek: we don't really want to grow the ThemeState. Is there no simpler solution? Also check architecture documents and put the detection logic into a correct place.

lklimek:
@thepastaclaw please apply comments

src/app.rs
```diff
@@ -1060,36 +1060,141 @@ async fn forward_backend_task_join_error(
...
+    /// to release a stuck detection slot.
+    detection_dispatched_at: Option<Instant>,
+    detection_generation: u64,
+    detection_sender: mpsc::Sender<(u64, Option<ThemeMode>)>,
+    detection_receiver: mpsc::Receiver<(u64, Option<ThemeMode>)>,
 }
 
 impl ThemeState {
     fn new(preference: ThemeMode) -> Self {
+        let (detection_sender, detection_receiver) = mpsc::channel();
```
thepastaclaw: **🔴 Blocking: Use a valid channel type and initialization**

`mpsc::Sender`, `mpsc::Receiver`, and `mpsc::channel()` are referenced without an `mpsc` import, so this commit does not compile. The intended API also needs to match the synchronous calls used below: `channel()` without a capacity and `Sender::send(...).is_err()` correspond to `std::sync::mpsc`, not the existing Tokio alias, whose channel requires a capacity and whose `send` is asynchronous.

```suggestion
use std::sync::{Arc, mpsc};
```

<sub>source: `gpt-6-astra` (phase2-reviewer: general)</sub>

src/app.rs
```diff
@@ -1060,36 +1060,141 @@ async fn forward_backend_task_join_error(
...
+            return;
+        }
+
+        self.last_checked = Some(now);
+        self.detection_in_flight = true;
+        self.detection_dispatched_at = Some(now);
+        let sender = self.detection_sender.clone();
+        let generation = self.detection_generation;
+        let ctx = ctx.clone();
+        tokio::task::spawn_blocking(move || {
```
thepastaclaw: **🔴 Blocking: Do not require a live Tokio runtime for every later UI frame**

`tokio::task::spawn_blocking` requires an active Tokio runtime at the point it is called. Production frames run inside the runtime created by `main`, but `tests/kittest/support.rs::mount_app` only enters its temporary runtime while constructing and settling the harness, then drops it before callers continue driving frames. With the default `System` preference, any later frame after `DETECTION_THROTTLE` reaches this call without a runtime and can panic. Use a runtime-independent worker mechanism or ensure the runtime remains entered for the full frame-driving lifetime.

<sub>source: `gpt-6-astra` (phase2-reviewer: rust-quality)</sub>

src/app.rs
```diff
@@ -1060,36 +1060,141 @@ async fn forward_backend_task_join_error(
...
+            let detected = std::panic::catch_unwind(crate::ui::theme::try_detect_system_theme)
+                .unwrap_or_else(|_| {
+                    tracing::warn!("OS theme detection panicked; treating as no result");
+                    None
+                });
+            if sender.send((generation, detected)).is_err() {
+                tracing::debug!("Dropping OS theme detection result because receiver closed");
+            }
+            ctx.request_repaint();
+        });
```
thepastaclaw: **🟡 Suggestion: Keep theme detection orchestration with the existing theme detector**

`src/ui/theme.rs` already owns `BoundedThemeDetector`, including a dedicated worker, single-flight protection, a bounded 25 ms caller wait, late-result handling, and panic recovery. The new `ThemeState` controller adds a second in-flight flag, watchdog, generation counter, and result channel around `try_detect_system_theme`, duplicating lifecycle and concurrency responsibilities in `app.rs`. Move the background detection orchestration into the theme module or extend its detector abstraction, leaving `AppState` responsible for applying the resulting theme state.

<sub>source: `gpt-6-astra` (phase2-reviewer: rust-quality)</sub>

thepastaclaw (commented):
## Re-review — Final validation — Phase 2 only (queue backlog)

The PR currently has a hard compilation failure because `ThemeState` uses an unresolved `mpsc` module. After that is corrected, the new polling path can panic in existing kittest harnesses because `spawn_blocking` is invoked after `mount_app` has dropped its Tokio runtime; the detector orchestration also duplicates responsibilities already owned by `ui::theme`.

<details>
<summary>Review provenance</summary>

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a substantial concurrency and UI-state refactor in `src/app.rs` involving background workers, generation handling, watchdog recovery, and synchronous preference updates, but it does not affect a critical surface such as funds, cryptography, consensus, networking, or storage.
- Phase 1 reviewers: **not run (skipped for throughput: 14 PRs queued, above the 10 limit)**
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — rust-quality (completed, effort high); agent `phase2-reviewer`

</details>

<details>

```

In `src/app.rs`:
- [BLOCKING] src/app.rs:1082-1088: Use a valid channel type and initialization
  `mpsc::Sender`, `mpsc::Receiver`, and `mpsc::channel()` are referenced without an `mpsc` import, so this commit does not compile. The intended API also needs to match the synchronous calls used below: `channel()` without a capacity and `Sender::send(...).is_err()` correspond to `std::sync::mpsc`, not the existing Tokio alias, whose channel requires a capacity and whose `send` is asynchronous.
- [BLOCKING] src/app.rs:1175: Do not require a live Tokio runtime for every later UI frame
  `tokio::task::spawn_blocking` requires an active Tokio runtime at the point it is called. Production frames run inside the runtime created by `main`, but `tests/kittest/support.rs::mount_app` only enters its temporary runtime while constructing and settling the harness, then drops it before callers continue driving frames. With the default `System` preference, any later frame after `DETECTION_THROTTLE` reaches this call without a runtime and can panic. Use a runtime-independent worker mechanism or ensure the runtime remains entered for the full frame-driving lifetime.
- [SUGGESTION] src/app.rs:1072-1188: Keep theme detection orchestration with the existing theme detector
  `src/ui/theme.rs` already owns `BoundedThemeDetector`, including a dedicated worker, single-flight protection, a bounded 25 ms caller wait, late-result handling, and panic recovery. The new `ThemeState` controller adds a second in-flight flag, watchdog, generation counter, and result channel around `try_detect_system_theme`, duplicating lifecycle and concurrency responsibilities in `app.rs`. Move the background detection orchestration into the theme module or extend its detector abstraction, leaving `AppState` responsible for applying the resulting theme state.
```
</details>

_Canonical verifier result: `REQUEST_CHANGES`. GitHub does not allow authors to approve or request changes on their own pull requests, so this review was submitted using `COMMENT` transport. The findings and blocker status above are unchanged._

PR #860, opened 2026-05-18
feat: rewrite Dash Evo Tool onto the new platform-wallet

lklimek:
## Goal

This PR started as a wallet-backend swap: move DET's SPV/wallet stack onto the upstream **`platform-wallet`** crate (`dashpay/platform`), which owns chain sync, address derivation, identity/asset-lock handling, and the shielded coordinator. That rewrite is still the foundation, and the framing below still holds: DET's bespoke `src/spv/` stack and the legacy RPC wallet mode are gone, chain sync and derivation come from upstream, and the `BackendTask` action/channel contract is preserved.

Since it was opened, ~30 additional PRs (#861–#900, excluding #862) stacked onto this branch and merged. Together they turned this from a backend swap into a **full rewrite of Dash Evo Tool** against the new backend — the UI layer, the identity/masternode/DPNS/DashPay workflows, the secret-storage model, and the progressive-disclosure interface were all rebuilt, re-verified, and in several cases substantially extended, not just re-plumbed onto the new engine. This PR is now the merge point for that entire rewrite into `v1.0-dev`.

### What shipped beyond the backend swap

- **Masternodes tab** (#876): a dedicated screen for loading, listing, and operating masternode/evonode identities by ProTxHash — key-role display, inline DPNS voting, Top Up/Withdraw/Transfer, per-node key management, evonode token-reward claiming, and a withdrawal key-selection fix so the app never pre-selects a signing key the wallet holds no private material for.
- **User-role / persona gating** (#879, folding in #880/#881): the old binary Expert-mode flag is replaced by a strict `Everyday < Power < Developer` role model (`UserRole`, `Capability`/`FeatureGate::checks()`) with a three-way "Interface mode" picker, so future features have a real place to attach beyond "developer or not."
- **v1.0 UI parity batch** (#882): closed a 31-item feature-parity audit against `v0.10-dev` — working Identity Hub Contacts (accept/decline/cancel/search/pay), a restored shielded receive-address view, settings/scheduled-vote/top-up-history migration on upgrade, interactive global nav pills, and a formal disclosure/closure record for the items intentionally not carried forward.
- **Legacy v0.9.3 identity migration** (#885): identities and their keys (including a masternode's owner/voting keys) loaded under v0.9.3 are now imported into the modern identity store on first launch, closing a silent data-loss path a mutation-tested regression test in #882 had exposed.
[...]

PR #862, opened 2026-05-29
chore(deps): bump platform to v3.1.0-dev.8 (9e00c8b894)

lklimek:
## Summary

Bumps `dash-sdk` and `rs-sdk-trusted-context-provider` from platform pin `54048b9352` to **`v3.1-dev` HEAD `9e00c8b894`** (v3.1.0-dev.8).

This is a prerequisite step ahead of the v0.10-alpha release. It is an **L-sized, semantic** migration (537 commits / 300 files upstream), dominated by an async `Signer<K>` rework, the reworked identity state-transition API, and a rust-dashcore/key-wallet `0.42 → 0.43` sideways move — **not** a mechanical version bump.

## What changed (27 files, +414 / −439)

| Theme | Kind | Notes |
|---|---|---|
| **A** `DocumentQuery` gained `select` / `group_by` / `having` | mechanical | 15 sites |
| **B** `Signer` is now `async` + generic `Signer<K>` with `sign_create_witness` | semantic | both DET impls → `#[async_trait]` + `async fn`; every `.sign()` caller now `.await` |
| **C** Identity ST API renames | semantic | `top_up_identity_with_private_key`, `put_to_platform_and_wait_for_response_with_private_key`, `try_from_identity_with_private_key`, `try_from_identity_with_signer_and_private_key` (async), `sign_external_with_options` (async) |
| **D** `TransactionBuilder` reshaped | semantic | `set_selection_strategy` / `set_current_height` / `add_inputs` + `build_unsigned()`; removed old string-classification |
| **E** `ManagedAccount*` / `WalletManager` | semantic | `managed_account_type()`, `next_change_address`, `next_receive_address`, `ManagedAccountTrait` import |
| **F** `AddressProvider` associated types | semantic | `Tag = AddressIndex`, `Address = PlatformAddress`; iterator returns + callbacks reshaped |
| **G** `WalletManager` per-wallet `synced_height`, `DashSpvClient<W,N,S>` 3 params, `run()`/`stop()` lifecycle | semantic | — |

## Verification

- ✅ `cargo build --all-features`
- ✅ `cargo clippy --all-features --all-targets -- -D warnings` (lib + bins + kittest + e2e + backend-e2e)
- ✅ `cargo +nightly fmt --all`
- ⚠️ Network/`#[ignore]` test suites **not** executed — see pre-merge gates below.

## ⚠️ Behavioural risk — the dashcore move is *sideways*, not forward

The dashcore bump is a divergent move (~90 commits theirs we lacked; 1 squashed commit ours they lack), so **compile-clean ≠ behaviour-equal**. The compiler cannot catch regressions in these areas — they require a runtime smoke test:

- SPV filter-height gating / rescan-from-genesis trigger
- Coin selection, fee, and change-output parity
- Address derivation
- Orchard `0.12 → 0.13` shielded path
- SPV client cancellation/lifecycle

### Behaviour-ambiguous sites flagged for audit (`TODO(v3.1-bump)`)

1. `src/spv/manager.rs:822` — global filter-height → per-wallet `synced_height` reset on data clear; verify rescan-from-genesis still fires.
2. `src/ui/wallets/shield_screen.rs:328` — Halo2 proof gen moved off `spawn_blocking` onto the async worker (builder is now async); watch for runtime starvation.
3. `backend_task/core/mod.rs` — change-address: index-peek → `next_change_address(…, add_to_state=false)`.
4. `model/wallet/mod.rs` `current_balances` — now skips stored balances whose index left the pending window.

## Pre-merge gates

- [ ] QA audit of the 4 behaviour-ambiguous sites + dashcore regression surface (Marvin, in progress)
- [ ] Regtest/testnet smoke run: wallet sync, send, identity top-up, shielded path
- [ ] **Release-gate note:** platform PR [#3625](https://github.com/dashpay/platform/pull/3625) (the SQLite `platform-wallet-storage` persister crate) is **still open** against `v3.1-dev`. This bump is *necessary-but-not-sufficient* if the upcoming architecture work (PR #860) requires that crate specifically. This PR does **not** unblock #860 on its own.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed missing async/await operations in identity creation, updates, and contract state transitions
  * Improved wallet address synchronization and per-wallet sync height tracking behavior
  * Enhanced document query field initialization for consistent data retrieval across operations
  * Corrected SPV wallet address derivation metadata and synchronization workflow

* **Chores**
  * Updated dash-sdk and trusted-context-provider dependencies to latest platform commit

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/dash-evo-tool/pull/862?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend_task/core/mod.rs
+++ b/src/backend_task/core/mod.rs
@@ -708,11 +708,11 @@ impl AppContext {
         // change address (key-wallet 0.43 derives addresses directly rather than
         // exposing a bare next-change index).
         let account_xpub = {
-            let wallet = wm
-                .get_wallet(wallet_id)
-                .ok_or_else(|| TaskError::WalletPaymentFailed {
-                    detail: "Wallet object not found".to_string(),
-                })?;
+            let wallet =
+                wm.get_wallet(wallet_id)
+                    .ok_or_else(|| TaskError::WalletPaymentFailed {
+                        detail: "Wallet object not found".to_string(),
+                    })?;
             let wallet_account = wallet
                 .accounts
                 .standard_bip44_accounts
@@ -741,12 +741,11 @@ impl AppContext {
                 })?;
 
             let utxos: Vec<_> = account.utxos.values().cloned().collect();
-            let change_addr =
-                account
-                    .next_change_address(Some(&account_xpub), false)
-                    .map_err(|e| TaskError::WalletPaymentFailed {
-                        detail: format!("Failed to derive change address: {e}"),
-                    })?;
+            let change_addr = account
+                .next_change_address(Some(&account_xpub), false)
+                .map_err(|e| TaskError::WalletPaymentFailed {
+                    detail: format!("Failed to derive change address: {e}"),
+                })?;
             (utxos, change_addr)
         };
 
@@ -756,7 +755,7 @@ impl AppContext {
                 .map(|(addr, amt)| (addr.clone(), (*amt as f64 * scale_factor) as u64))
                 .collect();
 
-            let build_result = (|| -> Result<Transaction, BuilderError> {
+            let build_result: Result<Transaction, BuilderError> = {
                 let mut builder = TransactionBuilder::new()
                     .set_fee_rate(FeeRate::normal())
                     .set_change_address(change_addr.clone())
@@ -769,7 +768,7 @@ impl AppContext {
                 }
 
                 builder.build_unsigned().map(|(tx, _fee)| tx)
-            })();
+            };
 
             match build_result {
                 Ok(tx) => return Ok(tx),
--- a/src/backend_task/core/mod.rs
+++ b/src/backend_task/core/mod.rs
@@ -25,7 +25,7 @@ use dash_sdk::dpp::dashcore::{
 };
 use dash_sdk::dpp::fee::Credits;
 use dash_sdk::dpp::key_wallet::Network as WalletNetwork;
-use dash_sdk::dpp::key_wallet::account::ECDSAAddressDerivation;
+use dash_sdk::dpp::key_wallet::managed_account::managed_account_trait::ManagedAccountTrait;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::ManagedWalletInfo;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::coin_selection::SelectionStrategy;
 use dash_sdk::dpp::key_wallet::wallet::managed_wallet_info::fee::FeeRate;
@@ -704,43 +704,51 @@ impl AppContext {
         let mut scale_factor = 1.0f64;
         let mut attempted_fallback = false;
 
-        // Get UTXOs and change address from the wallet account
-        let (utxos, change_index) = {
-            let managed_info =
-                wm.get_wallet_info(wallet_id)
-                    .ok_or_else(|| TaskError::WalletPaymentFailed {
-                        detail: "Wallet info unavailable".to_string(),
-                    })?;
-            let account = managed_info
-                .accounts()
+        // The account xpub is needed by the managed account to derive the next
+        // change address (key-wallet 0.43 derives addresses directly rather than
+        // exposing a bare next-change index).
+        let account_xpub = {
+            let wallet = wm
+                .get_wallet(wallet_id)
+                .ok_or_else(|| TaskError::WalletPaymentFailed {
+                    detail: "Wallet object not found".to_string(),
+                })?;
+            let wallet_account = wallet
+                .accounts
                 .standard_bip44_accounts
                 .get(&DEFAULT_BIP44_ACCOUNT_INDEX)
+                .ok_or_else(|| TaskError::WalletPaymentFailed {
+                    detail: "BIP44 wallet account missing".to_string(),
+                })?;
+            wallet_account.account_xpub
+        };
+
+        // Get UTXOs and the next (unused) change address from the managed account.
+        // `add_to_state = false` peeks the next change address without advancing
+        // the pool, preserving the prior "derive at next change index" behaviour.
+        let (utxos, change_addr) = {
+            let managed_info = wm.get_wallet_info_mut(wallet_id).ok_or_else(|| {
+                TaskError::WalletPaymentFailed {
+                    detail: "Wallet info unavailable".to_string(),
+                }
+            })?;
+            let account = managed_info
+                .accounts_mut()
+                .standard_bip44_accounts
+                .get_mut(&DEFAULT_BIP44_ACCOUNT_INDEX)
                 .ok_or_else(|| TaskError::WalletPaymentFailed {
                     detail: "BIP44 account missing".to_string(),
                 })?;
 
             let utxos: Vec<_> = account.utxos.values().cloned().collect();
-            let change_index = account.get_next_change_address_index().unwrap_or(0);
-            (utxos, change_index)
+            let change_addr =
+                account
+                    .next_change_address(Some(&account_xpub), false)
+                    .map_err(|e| TaskError::WalletPaymentFailed {
+                        detail: format!("Failed to derive change address: {e}"),
+                    })?;
+            (utxos, change_addr)
         };
-
-        let wallet = wm
-            .get_wallet(wallet_id)
-            .ok_or_else(|| TaskError::WalletPaymentFailed {
-                detail: "Wallet object not found".to_string(),
-            })?;
-        let wallet_account = wallet
-            .accounts
-            .standard_bip44_accounts
-            .get(&DEFAULT_BIP44_ACCOUNT_INDEX)
-            .ok_or_else(|| TaskError::WalletPaymentFailed {
-                detail: "BIP44 wallet account missing".to_string(),
-            })?;
-        let change_addr = wallet_account
-            .derive_change_address(change_index)
-            .map_err(|e| TaskError::WalletPaymentFailed {
-                detail: format!("Failed to derive change address: {e}"),
-            })?;
 
         loop {
             let scaled_recipients: Vec<(Address, u64)> = recipients
@@ -751,20 +759,16 @@ impl AppContext {
             let build_result = (|| -> Result<Transaction, BuilderError> {
                 let mut builder = TransactionBuilder::new()
                     .set_fee_rate(FeeRate::normal())
-                    .set_change_address(change_addr.clone());
+                    .set_change_address(change_addr.clone())
+                    .set_selection_strategy(SelectionStrategy::LargestFirst)
+                    .set_current_height(current_height)
+                    .add_inputs(utxos.iter().cloned());
 
                 for (addr, amt) in &scaled_recipients {
-                    builder = builder.add_output(addr, *amt)?;
+                    builder = builder.add_output(addr, *amt);
                 }
 
-                builder = builder.select_inputs(
-                    &utxos,
-                    SelectionStrategy::LargestFirst,
-                    current_height,
-                    |_| None, // No private keys for unsigned tx
-                )?;
-
-                builder.build()
+                builder.build_unsigned().map(|(tx, _fee)| tx)
             })();
 
             match build_result {
@@ -871,30 +875,22 @@ impl AppContext {
         // Build the transaction using TransactionBuilder
         let mut builder = TransactionBuilder::new()
             .set_fee_rate(FeeRate::normal())
-            .set_change_address(change_address.clone());
+            .set_change_address(change_address.clone())
+            .set_selection_strategy(SelectionStrategy::OptimalConsolidation)
+            .set_current_height(current_height)
+            .add_inputs(all_utxos);
 
         for (address, amount) in recipients {
-            builder = builder
-                .add_output(&address, amount)
-                .map_err(|e: BuilderError| WalletError::TransactionBuild(e.to_string()))?;
+            builder = builder.add_output(&address, amount);
         }
 
-        builder = builder
-            .select_inputs(
-                &all_utxos,
-                SelectionStrategy::OptimalConsolidation,
-                current_height,
-                |_| None, // No private keys for unsigned transaction
-            )
-            // TODO(RUST-002): String-based error classification — see #660
-            .map_err(|e: BuilderError| match e.to_string() {
-                msg if msg.contains("Insufficient") => WalletError::InsufficientFunds,
-                msg => WalletError::TransactionBuild(msg),
-            })?;
-
         builder
-            .build()
-            .map_err(|e: BuilderError| WalletError::TransactionBuild(e.to_string()))
+            .build_unsigned()
+            .map(|(tx, _fee)| tx)
+            .map_err(|e: BuilderError| match e {
+                BuilderError::InsufficientFunds { .. } => WalletError::InsufficientFunds,
+                other => WalletError::TransactionBuild(other.to_string()),
+            })
     }
 
     fn sign_spv_transaction(
--- a/src/backend_task/core/mod.rs
+++ b/src/backend_task/core/mod.rs
@@ -724,8 +724,14 @@ impl AppContext {
         };
 
         // Get UTXOs and the next (unused) change address from the managed account.
-        // `add_to_state = false` peeks the next change address without advancing
-        // the pool, preserving the prior "derive at next change index" behaviour.
+        // `add_to_state = false` does not advance the address pool, but key-wallet
+        // 0.43's `next_change_address` still bumps the managed account's monitor
+        // revision unconditionally. There is no read-only "peek next change address"
+        // on ManagedCoreFundsAccount, so this fee-estimation peek takes a write lock
+        // and nudges the monitor revision. That only signals the mempool bloom filter
+        // it may be stale (a one-time rebuild), so it is acceptable here.
+        // TODO(v3.1-bump): switch to a non-mutating change-address peek if key-wallet
+        // exposes one upstream.
         let (utxos, change_addr) = {
             let managed_info = wm.get_wallet_info_mut(wallet_id).ok_or_else(|| {
                 TaskError::WalletPaymentFailed {
@@ -755,6 +761,10 @@ impl AppContext {
                 .map(|(addr, amt)| (addr.clone(), (*amt as f64 * scale_factor) as u64))
                 .collect();
 
+            // build_unsigned requires input >= output + fee and returns
+            // InsufficientFunds otherwise; the old build() let change shrink and
+            // could return a tx that underpaid the fee. The fallback loop below
+            // now drives off that earlier (correct) InsufficientFunds.
             let build_result: Result<Transaction, BuilderError> = {
                 let mut builder = TransactionBuilder::new()
                     .set_fee_rate(FeeRate::normal())
--- a/src/backend_task/shielded/bundle.rs
+++ b/src/backend_task/shielded/bundle.rs
@@ -114,23 +114,59 @@ pub async fn build_shield_credit(
     let fee_strategy: AddressFundsFeeStrategy =
         vec![AddressFundsFeeStrategyStep::DeductFromInput(0)];
 
-    // Clone the wallet so the signer is owned across the async build (the
-    // std RwLock guard is not held across an await point).
+    // Clone the wallet so the signer is owned by the offloaded build.
     let wallet = { wallet_arc.read()?.clone() };
+    let platform_version = sdk.version();
     // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
-    build_shield_transition(
-        &recipient_addr,
+    build_shield_transition_offloaded(
+        recipient_addr,
         amount,
         inputs,
         fee_strategy,
-        &wallet,
-        0,
-        &prover,
-        [0u8; 36],
-        sdk.version(),
+        wallet,
+        prover,
+        platform_version,
     )
     .await
-    .map_err(|e| shielded_build_error(e.to_string()))
+}
+
+/// Run the (CPU-heavy, Halo2-proving) shield build on the blocking pool.
+///
+/// `build_shield_transition` is async but its proving step runs synchronously
+/// inline before the first await, so awaiting it directly would block a tokio
+/// worker for seconds. We drive the whole future on a `spawn_blocking` thread
+/// (where `block_on` is permitted) so the proof generation stays off the async
+/// worker pool.
+async fn build_shield_transition_offloaded(
+    recipient_addr: OrchardAddress,
+    amount: u64,
+    inputs: BTreeMap<PlatformAddress, (u32, u64)>,
+    fee_strategy: AddressFundsFeeStrategy,
+    wallet: crate::model::wallet::Wallet,
+    prover: CachedProver,
+    platform_version: &'static PlatformVersion,
+) -> Result<dash_sdk::dpp::state_transition::StateTransition, TaskError> {
+    let handle = tokio::runtime::Handle::current();
+    tokio::task::spawn_blocking(move || {
+        handle.block_on(async {
+            // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
+            build_shield_transition(
+                &recipient_addr,
+                amount,
+                inputs,
+                fee_strategy,
+                &wallet,
+                0,
+                &prover,
+                [0u8; 36],
+                platform_version,
+            )
+            .await
+            .map_err(|e| shielded_build_error(e.to_string()))
+        })
+    })
+    .await
+    .map_err(|e| shielded_build_error(format!("shield proof task panicked: {e}")))?
 }
 
 /// Build and broadcast a Shield transition (transparent -> shielded pool).
@@ -198,23 +234,19 @@ pub async fn shield_credits(
     }
 
     let state_transition = {
-        // Clone the wallet so the signer is owned across the async build (the
-        // std RwLock guard is not held across an await point).
+        // Clone the wallet so the signer is owned by the offloaded build.
         let wallet = { wallet_arc.read()?.clone() };
-        // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
-        build_shield_transition(
-            &recipient_addr,
+        let platform_version = sdk.version();
+        build_shield_transition_offloaded(
+            recipient_addr,
             amount,
             inputs,
             fee_strategy,
-            &wallet,
-            0,
-            &prover,
-            [0u8; 36],
-            sdk.version(),
+            wallet,
+            prover,
+            platform_version,
         )
-        .await
-        .map_err(|e| shielded_build_error(e.to_string()))?
+        .await?
     };
 
     if let Some(s) = &stage {
--- a/src/backend_task/shielded/bundle.rs
+++ b/src/backend_task/shielded/bundle.rs
@@ -21,7 +21,7 @@ use dash_sdk::grovedb_commitment_tree::{
 };
 use dash_sdk::platform::transition::broadcast::BroadcastStateTransition;
 use std::collections::BTreeMap;
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 
 /// Wrapper around a cached `ProvingKey` that implements `OrchardProver`.
 struct CachedProver {
@@ -114,8 +114,6 @@ pub async fn build_shield_credit(
     let fee_strategy: AddressFundsFeeStrategy =
         vec![AddressFundsFeeStrategyStep::DeductFromInput(0)];
 
-    // Clone the wallet so the signer is owned by the offloaded build.
-    let wallet = { wallet_arc.read()?.clone() };
     let platform_version = sdk.version();
     // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
     build_shield_transition_offloaded(
@@ -123,7 +121,7 @@ pub async fn build_shield_credit(
         amount,
         inputs,
         fee_strategy,
-        wallet,
+        wallet_arc,
         prover,
         platform_version,
     )
@@ -136,18 +134,23 @@ pub async fn build_shield_credit(
 /// inline before the first await, so awaiting it directly would block a tokio
 /// worker for seconds. We drive the whole future on a `spawn_blocking` thread
 /// (where `block_on` is permitted) so the proof generation stays off the async
-/// worker pool.
+/// worker pool. The wallet read guard is taken inside that dedicated thread and
+/// never crosses a thread boundary, so we pass the cheap `Arc` rather than
+/// deep-cloning the wallet (and its seed material).
 async fn build_shield_transition_offloaded(
     recipient_addr: OrchardAddress,
     amount: u64,
     inputs: BTreeMap<PlatformAddress, (u32, u64)>,
     fee_strategy: AddressFundsFeeStrategy,
-    wallet: crate::model::wallet::Wallet,
+    wallet_arc: Arc<RwLock<crate::model::wallet::Wallet>>,
     prover: CachedProver,
     platform_version: &'static PlatformVersion,
 ) -> Result<dash_sdk::dpp::state_transition::StateTransition, TaskError> {
     let handle = tokio::runtime::Handle::current();
     tokio::task::spawn_blocking(move || {
+        let wallet = wallet_arc
+            .read()
+            .map_err(|_| shielded_build_error("wallet lock poisoned".to_string()))?;
         handle.block_on(async {
             // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
             build_shield_transition(
@@ -155,7 +158,7 @@ async fn build_shield_transition_offloaded(
                 amount,
                 inputs,
                 fee_strategy,
-                &wallet,
+                &*wallet,
                 0,
                 &prover,
                 [0u8; 36],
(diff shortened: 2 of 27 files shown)


thepastaclaw:
✅ Review complete (commit d510bcd3)

src/model/wallet/mod.rs
```diff
@@ -2647,19 +2652,31 @@ impl WalletAddressProvider {
...
 
-    fn pending_addresses(&self) -> Vec<(AddressIndex, AddressKey)> {
+    fn pending_addresses(&self) -> impl Iterator<Item = (Self::Tag, Self::Address)> + '_ {
         self.pending
             .iter()
             .filter(|(index, _)| !self.resolved.contains(index))
-            .map(|(index, (key, _))| (*index, key.clone()))
-            .collect()
+            .filter_map(|(index, (_, core_address))| {
+                PlatformAddress::try_from(core_address.clone())
```
lklimek, line 2667: it is confusing - why we convert core_address to PlatformAddress? These addresses are on different chain. Maybe just naming issue, but double-check.

lklimek:
## QA audit (Marvin) — 1 CRITICAL blocker, fixes in progress

Full audit traced the behaviour-ambiguous sites. **Not mergeable as-is** — one critical regression, proven by the branch's own failing unit test.

### 🔴 QA-001 (CRITICAL) — "Clear SPV Data" no longer triggers rescan-from-genesis — `src/spv/manager.rs:833-841`
The old unconditional `update_filter_committed_height(0)` was replaced with per-wallet `update_wallet_synced_height(&id, 0)`. In key-wallet 0.43 that setter is **forward-only monotonic** (`wallet_interface.rs:127-129`: *"a value below the current is silently ignored"*; `process_block.rs:269` guards `if height > info.synced_height()`). So `0 > N` is false → the reset is a **no-op** for any synced wallet. After clearing data, the in-memory height stays stale, the next session declares itself synced, skips genesis rescan, and balance reads stale/zero. `clear_data_dir` clears `det_wallets` but never removes the wallet from the WalletManager, so the monotonic guard defeats the reset. **Proven**: the branch's own `test_clear_data_dir_resets_filter_committed_height` fails `5000 ≠ 0`.

### 🟠 QA-002 (HIGH) — failing unit test shipped — `src/spv/tests.rs:660-695`
Plain `#[tokio::test]` (not `#[ignore]`) → `cargo test --lib` is red. The change set was compiled (`clippy --all-targets`) but its non-ignored tests were never run.

### 🟡 QA-003 (MEDIUM, suspected) — Halo2 shielded proof blocks a tokio worker — `shield_screen.rs:328`, `bundle.rs:88`
Async `build_shield_transition` runs multi-second CPU proving inline (no `spawn_blocking`) on the awaiting worker → runtime-starvation risk under batch shielding.

### Cleared (PASS)
Change-address derivation, `current_balances` window, TransactionBuilder fee/change parity, AddressProvider iteration, and the async `Signer` migration (byte-identical signing; `sign_create_witness` is a real impl) all verified behaviour-preserving.

### Low / informational
QA-004 (latent balance-window drop), QA-005 (change-peek mutates account state on fee estimation), QA-006 (insufficient-funds now includes fee — arguably a fix).

**Status:** QA-001/002/003 sent back to dev; branch will be updated. Runtime smoke test (regtest/testnet: sync, send, identity top-up, shielded path) remains a hard pre-merge gate given the sideways dashcore move.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/backend_task/core/mod.rs
```diff
@@ -704,68 +704,71 @@ impl AppContext {
...
                 })?;
 
             let utxos: Vec<_> = account.utxos.values().cloned().collect();
-            let change_index = account.get_next_change_address_index().unwrap_or(0);
-            (utxos, change_index)
+            let change_addr = account
+                .next_change_address(Some(&account_xpub), false)
+                .map_err(|e| TaskError::WalletPaymentFailed {
+                    detail: format!("Failed to derive change address: {e}"),
+                })?;
```
thepastaclaw, line 754: **🟡 Suggestion: Consider add_to_state=true now that the API exposes the choice — current `false` preserves a latent issue**

The new `next_change_address(account_xpub, add_to_state)` makes the address-pool registration explicit. With `add_to_state=false`, the address-pool path in key-wallet generates a new change address (when no pre-generated unused change address exists) without inserting it into `addresses` / `address_index` / `script_pubkey_index` (`address_pool.rs:507-516`). Because DET's SPV registration walks `account.all_addresses()` (`src/context/wallet_lifecycle.rs:518`), and that returns only entries actually in `self.addresses`, a freshly-derived change address used in a send is not added to the SPV watch set.

Important context for the author and reviewers: this is NOT a fresh regression. The prior code (`wallet_account.derive_change_address(change_index)`) was an immutable `&` borrow and did not register either. The new API simply makes the trade-off visible. Two observations:

1. The comment correctly notes parity with the prior behavior, and the PR description marks this for smoke testing. If smoke runs confirm that subsequent UTXO scans recover the change output (because the wallet records its own outgoing tx outputs), the existing `false` is defensible.
2. If the smoke test reveals the change output is not picked up by SPV until a manual recovery/rescan, the simplest fix is `add_to_state=true`. Re-using the same change address across retries inside the `loop` would still work (the address-pool returns the same unused entry on subsequent calls until it's marked used).

At minimum, the outcome of this smoke check should be documented in the PR before merge, since this site sits right on the boundary the new API was designed to clarify.

src/model/wallet/mod.rs
```diff
@@ -2647,19 +2652,31 @@ impl WalletAddressProvider {
...
+                    .ok()
+                    .map(|pa| (*index, pa))
+            })
     }
```
thepastaclaw, line 2671: **🟡 Suggestion: pending_addresses() and current_balances() silently drop conversion failures — possible silent sync hang**

Both `pending_addresses()` (2662-2671) and `current_balances()` (2730-2743) use `filter_map(... .ok())` to drop entries where `PlatformAddress::try_from(core_address)` fails. The SDK consumes `pending_addresses()` to learn what is still to be resolved, but `has_pending()` (2724-2728) iterates `self.pending` directly — so any entry that fails conversion never reaches the SDK, never gets `on_address_found` / `on_address_absent`, never gets inserted into `self.resolved`, and `has_pending()` stays `true` forever. Result: silent sync hang with no error and no log line.

In practice the conversion should not fail (each `pending` entry came from `derive_address_at_index`, which already round-tripped `PlatformAddress::try_from` at insertion), but the silent-drop branch is the kind of invariant-breakage that is very hard to diagnose later. At minimum, log at `error!` when conversion fails and either force-resolve the offending index (so the loop makes progress) or `unwrap()` to surface the invariant violation. The same applies to `current_balances()` at 2738-2742, where the silent skip on a missing-from-`pending` index would corrupt the SDK's view of the stored baseline for incremental catch-up.

<sub>source: ['claude']</sub>

src/spv/manager.rs
```diff
@@ -1177,16 +1176,23 @@ impl SpvManager {
...
                 _ = stop_token.cancelled() => {
-                    run_cancel.cancel();
+                    if let Err(e) = client.stop().await {
+                        tracing::warn!("client.stop() during cancellation failed: {e}");
+                    }
+                    // Let run() observe the stop flag and unwind its monitors.
+                    let _ = (&mut run_future).await;
                     Outcome::Cancelled
                 },
             }
```
thepastaclaw, line 1209: **🟡 Suggestion: Cancellation path can block forever if client.stop() errors without flipping the run flag**

The prior code used `run_cancel.cancel()` — a synchronous, infallible signal — that guaranteed `run()` would unwind. The new path now relies on `client.stop().await` to flip the client's internal running flag, then `(&mut run_future).await` to let `run()` observe it. If `stop()` returns an error in a path that did not actually flip the flag, the subsequent unconditional `(&mut run_future).await` blocks indefinitely, hanging the SPV worker on shutdown or network switch. The error from `stop()` is only logged at `warn!` level today.

Two straightforward defenses: (a) wrap the post-stop `(&mut run_future).await` in a `tokio::time::timeout` and drop/abort on expiry, or (b) when `stop()` errored, return `Outcome::RunCompleted(Err(...))` immediately without awaiting `run_future`. Adding a `TODO(v3.1-bump)` here — matching the four already-flagged sites — would also give the smoke-test pass a hook to verify the cancellation contract explicitly.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Large platform v3.1.0-dev.8 migration with careful TODO markers and explicit smoke-test gating on the behavior-ambiguous wallet/SPV/shielded paths. The mechanical adaptations (async Signer awaits, DocumentQuery field expansion, TransactionBuilder reshape, identity ST renames) look consistent. I found no verified blocking regression, but there are a few non-blocking follow-ups worth addressing or explicitly covering in the planned smoke run.

🟡 3 suggestion(s) | 💬 1 nitpick(s)

<details>

```

- [SUGGESTION] src/backend_task/core/mod.rs:744-748: Consider add_to_state=true now that the API exposes the choice — current `false` preserves a latent issue
  The new `next_change_address(account_xpub, add_to_state)` makes the address-pool registration explicit. With `add_to_state=false`, the address-pool path in key-wallet generates a new change address (when no pre-generated unused change address exists) without inserting it into `addresses` / `address_index` / `script_pubkey_index` (`address_pool.rs:507-516`). Because DET's SPV registration walks `account.all_addresses()` (`src/context/wallet_lifecycle.rs:518`), and that returns only entries actually in `self.addresses`, a freshly-derived change address used in a send is not added to the SPV watch set.

  Important context for the author and reviewers: this is NOT a fresh regression. The prior code (`wallet_account.derive_change_address(change_index)`) was an immutable `&` borrow and did not register either. The new API simply makes the trade-off visible. Two observations:

  1. The comment correctly notes parity with the prior behavior, and the PR description marks this for smoke testing. If smoke runs confirm that subsequent UTXO scans recover the change output (because the wallet records its own outgoing tx outputs), the existing `false` is defensible.
  2. If the smoke test reveals the change output is not picked up by SPV until a manual recovery/rescan, the simplest fix is `add_to_state=true`. Re-using the same change address across retries inside the `loop` would still work (the address-pool returns the same unused entry on subsequent calls until it's marked used).

  At minimum, the outcome of this smoke check should be documented in the PR before merge, since this site sits right on the boundary the new API was designed to clarify.

In `src/model/wallet/mod.rs`:
- [SUGGESTION] src/model/wallet/mod.rs:2662-2671: pending_addresses() and current_balances() silently drop conversion failures — possible silent sync hang
  Both `pending_addresses()` (2662-2671) and `current_balances()` (2730-2743) use `filter_map(... .ok())` to drop entries where `PlatformAddress::try_from(core_address)` fails. The SDK consumes `pending_addresses()` to learn what is still to be resolved, but `has_pending()` (2724-2728) iterates `self.pending` directly — so any entry that fails conversion never reaches the SDK, never gets `on_address_found` / `on_address_absent`, never gets inserted into `self.resolved`, and `has_pending()` stays `true` forever. Result: silent sync hang with no error and no log line.

  In practice the conversion should not fail (each `pending` entry came from `derive_address_at_index`, which already round-tripped `PlatformAddress::try_from` at insertion), but the silent-drop branch is the kind of invariant-breakage that is very hard to diagnose later. At minimum, log at `error!` when conversion fails and either force-resolve the offending index (so the loop makes progress) or `unwrap()` to surface the invariant violation. The same applies to `current_balances()` at 2738-2742, where the silent skip on a missing-from-`pending` index would corrupt the SDK's view of the stored baseline for incremental catch-up.

In `src/spv/manager.rs`:
- [SUGGESTION] src/spv/manager.rs:1188-1198: Cancellation path can block forever if client.stop() errors without flipping the run flag
  The prior code used `run_cancel.cancel()` — a synchronous, infallible signal — that guaranteed `run()` would unwind. The new path now relies on `client.stop().await` to flip the client's internal running flag, then `(&mut run_future).await` to let `run()` observe it. If `stop()` returns an error in a path that did not actually flip the flag, the subsequent unconditional `(&mut run_future).await` blocks indefinitely, hanging the SPV worker on shutdown or network switch. The error from `stop()` is only logged at `warn!` level today.

  Two straightforward defenses: (a) wrap the post-stop `(&mut run_future).await` in a `tokio::time::timeout` and drop/abort on expiry, or (b) when `stop()` errored, return `Outcome::RunCompleted(Err(...))` immediately without awaiting `run_future`. Adding a `TODO(v3.1-bump)` here — matching the four already-flagged sites — would also give the smoke-test pass a hook to verify the cancellation contract explicitly.
```
</details>

lklimek:
## ✅ QA re-verify (Marvin) — both blockers resolved at the mechanism level

Re-checked commits `6d1dfe93` (QA-001) and `84ab1a50` (QA-003). `cargo test --all-features --lib` → **468 passed, 0 failed**.

**QA-001 (CRITICAL) — CONFIRMED-FIXED** (`src/spv/manager.rs:837-844`): direct rewind of `synced_height`/`last_processed_height` (the monotonic setter cannot express a downward reset). Traced: `FiltersManager::tick()` reads `wallets_behind(committed)` → `reset_for_rescan()` → re-downloads from the rewound height; cold-start seeds committed height from the cross-wallet min `synced_height`. Zeroing both heights is consistent (on-disk UTXO/tx state is wiped, so from-genesis reprocessing rebuilds idempotently). `fs::remove_dir_all` deletes the disk store, so no persisted height can re-defeat the rescan. Rewritten test asserts the production predicate `wallets_behind(PRESEED).contains(id)` — genuine, not tautological.

**QA-003 (MEDIUM) — CONFIRMED-FIXED** (`src/backend_task/shielded/bundle.rs:137-170`): `spawn_blocking` + `Handle::block_on` runs the heavy Halo2 prove off the worker pool on a blocking thread (where `block_on` is legal — panic path unreachable). Captured inputs are owned/`'static`; no Send/deadlock issue.

**Remaining pre-merge gates (external to code):**
- [ ] Runtime regtest/testnet smoke: wallet sync, send, identity top-up, shielded path (sideways dashcore move — compile-green ≠ behaviour-green).
- [ ] Release-gate note: platform #3625 still open; this bump is necessary-but-not-sufficient for PR #860.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

thepastaclaw (commented):
## Code Review

Latest delta (commits 6d1dfe93 and 84ab1a50) addresses two QA items cleanly: clear_data_dir now rewinds per-wallet SPV sync metadata with a regression test, and shielded Halo2 proving is offloaded onto the blocking pool via spawn_blocking + Handle::block_on. No new defects in the delta. Three prior findings from the cumulative PR remain valid and carried forward (change-address peek not registering state, SPV cancellation hang on stop() error, dead highest_found field). One prior finding (PlatformAddress conversion drops in pending_addresses/current_balances) is now mitigated by the explanatory comment plus the P2PKH-only construction invariant; dropped as outdated.

🟡 2 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/spv/manager.rs`:
- [SUGGESTION] src/spv/manager.rs:1194-1200: SPV cancellation can hang indefinitely when client.stop() returns Err
  Carried forward from the prior review — still valid in the current source. On cancellation the code calls `client.stop().await`, logs any error via `tracing::warn!`, and then unconditionally awaits `(&mut run_future).await` to let `run()` observe the stop flag and unwind. The whole termination guarantee rides on `stop()` having flipped the client's internal running flag before it returned. If `stop()` returns `Err` before clearing that flag (e.g. an internal guard-acquisition failure or shutdown handshake error in the SDK), `run_future` has no other cancellation signal to observe and this await blocks forever — stalling SPV shutdown, network switches, and clear_data_dir on the exact error path where graceful shutdown matters most. The previous design used `run_cancel.cancel()`, a synchronous infallible signal that guaranteed unwind. Treat a `stop()` error as a hard cancellation: either `tokio::select!` the final await against `stop_token` again with a short timeout, or skip the await entirely on `Err` and let `run_future` drop. At minimum, verify upstream `DashSpvClient::stop()` always clears the running flag before any fallible work.

- [SUGGESTION] src/backend_task/core/mod.rs:750-754: Change-address peek with add_to_state=false leaves newly derived change unregistered
  Carried forward from the prior review — still valid. This site still calls `next_change_address(Some(&account_xpub), false)`. In key-wallet 0.43, `next_change_address` delegates to `address_pool::next_unused(..., add_to_state)` which will derive a brand-new change address whenever the internal pool has no pre-generated unused entry. With `add_to_state = false`, that fresh address is returned but not inserted into the account's internal address pool, so the transaction can pay change to an address the wallet has never reserved/registered for monitoring. The wallet can then miss tracking its own change output until some later rescan or unrelated address-generation step discovers it. The new TODO comment correctly explains why the existing `true` path has an unwanted monitor-revision side effect, but does not address this correctness risk — and because this PR migrates DET onto the new explicit `add_to_state` API, closing this hole is part of the migration surface. Either flip to `true` and accept the bloom-filter rebuild, or take a different non-mutating peek that still registers the address before use.
```
</details>

src/spv/manager.rs
```diff
@@ -1177,16 +1180,23 @@ impl SpvManager {
...
             tokio::select! {
                 result = &mut run_future => Outcome::RunCompleted(result),
                 _ = stop_token.cancelled() => {
-                    run_cancel.cancel();
+                    if let Err(e) = client.stop().await {
+                        tracing::warn!("client.stop() during cancellation failed: {e}");
+                    }
+                    // Let run() observe the stop flag and unwind its monitors.
+                    let _ = (&mut run_future).await;
                     Outcome::Cancelled
```
thepastaclaw, line 1207: **🟡 Suggestion: SPV cancellation can hang indefinitely when client.stop() returns Err**

Carried forward from the prior review — still valid in the current source. On cancellation the code calls `client.stop().await`, logs any error via `tracing::warn!`, and then unconditionally awaits `(&mut run_future).await` to let `run()` observe the stop flag and unwind. The whole termination guarantee rides on `stop()` having flipped the client's internal running flag before it returned. If `stop()` returns `Err` before clearing that flag (e.g. an internal guard-acquisition failure or shutdown handshake error in the SDK), `run_future` has no other cancellation signal to observe and this await blocks forever — stalling SPV shutdown, network switches, and clear_data_dir on the exact error path where graceful shutdown matters most. The previous design used `run_cancel.cancel()`, a synchronous infallible signal that guaranteed unwind. Treat a `stop()` error as a hard cancellation: either `tokio::select!` the final await against `stop_token` again with a short timeout, or skip the await entirely on `Err` and let `run_future` drop. At minimum, verify upstream `DashSpvClient::stop()` always clears the running flag before any fallible work.

```suggestion
                _ = stop_token.cancelled() => {
                    match client.stop().await {
                        Ok(()) => {
                            let _ = (&mut run_future).await;
                        }
                        Err(e) => {
                            tracing::warn!("client.stop() during cancellation failed: {e}");
                        }
                    }
                    Outcome::Cancelled
                },
```

src/backend_task/core/mod.rs
```diff
@@ -704,68 +704,81 @@ impl AppContext {
...
```
thepastaclaw, line 754: **🟡 Suggestion: Change-address peek with add_to_state=false leaves newly derived change unregistered**

Carried forward from the prior review — still valid. This site still calls `next_change_address(Some(&account_xpub), false)`. In key-wallet 0.43, `next_change_address` delegates to `address_pool::next_unused(..., add_to_state)` which will derive a brand-new change address whenever the internal pool has no pre-generated unused entry. With `add_to_state = false`, that fresh address is returned but not inserted into the account's internal address pool, so the transaction can pay change to an address the wallet has never reserved/registered for monitoring. The wallet can then miss tracking its own change output until some later rescan or unrelated address-generation step discovers it. The new TODO comment correctly explains why the existing `true` path has an unwanted monitor-revision side effect, but does not address this correctness risk — and because this PR migrates DET onto the new explicit `add_to_state` API, closing this hole is part of the migration surface. Either flip to `true` and accept the bloom-filter rebuild, or take a different non-mutating peek that still registers the address before use.

```suggestion
            let change_addr = account
                .next_change_address(Some(&account_xpub), true)
                .map_err(|e| TaskError::WalletPaymentFailed {
                    detail: format!("Failed to derive change address: {e}"),
                })?;
```

<sub>source: ['codex']</sub>

lklimek:
## Testnet smoke (Marvin) — core behaviour GREEN, one test-infra regression

Live testnet, serial backend-e2e on the bump branch.

### Functionally PASS (dashcore 0.42→0.43 behaviour preserved)
- `test_spv_sync_and_create_wallet`, `test_spv_transactions_is_ours_flag` — SPV sync + address derivation
- `test_send_and_receive_funds` — coin selection / fee / change parity (reworked `TransactionBuilder`)
- `test_tc001/tc002_refresh_wallet_info` — balance reporting
- `test_create_identity` — async-`Signer` registration (the **production** registration path)
- `test_tc005_create_top_up_asset_lock` — functionally correct (passes with adequate stack)
- `test_withdraw_from_identity`
- **SPV health GREEN**: peers connect ~2s, full sync ~10-20s, balance resolved across ~8 harness restarts.

### 🟠 New regression (test-infrastructure only) — being fixed
`test_tc005_create_top_up_asset_lock` overflows the default 2 MB thread stack (SIGABRT) when nested inside the suite's `SHARED_IDENTITY.get_or_init` fixture. Bisected: passes on `v1.0-dev`, overflows on the bump branch; `RUST_MIN_STACK=32M` makes it pass → it's stack *size* from the deepened async-`Signer<K>` future graph, not recursion or a logic bug. **The production registration path is NOT affected** (`test_create_identity` runs the real task without fixture nesting and passes at default stack). Fix: `Box::pin` the deep fixture futures (test-harness).

### Not validated
- **Shielded lifecycle (QA-003 / orchard 0.13)**: BLOCKED — shielded ops are feature-gated OFF on this testnet. Needs a shielded-enabled network. Static re-verify earlier confirmed the offload is structurally correct.
- **QA-001 (Clear-SPV-Data rescan)**: no e2e coverage — manual GUI verification pending.

### Remaining pre-merge work
- [ ] Stack fix + full suite re-run to completion (never finished end-to-end on this branch)
- [ ] Shielded path on a shielded-enabled network
- [ ] Manual GUI check of Clear-SPV-Data rescan

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

lklimek:
## Correction: the "stack overflow" was a missed run-flag, not a regression

Follow-up to the smoke report above. The earlier SIGABRT was **operator error in the smoke run, not a defect**:

- `tests/backend-e2e/main.rs:7-12` documents `RUST_MIN_STACK=16777216` (16 MB) as **required** for backend-e2e (the SDK uses deep stacks for state-transition construction; default 8 MB is insufficient). The first smoke run omitted it.
- Re-run **at the documented 16 MB**: `test_tc005_create_top_up_asset_lock` **PASSES** (asset lock broadcast, TX `54df4898…`, 22 s, no SIGABRT).
- **The bump did NOT push past the documented 16 MB requirement.** No code change, no docs change, no production impact. Branch tip unchanged (`84ab1a50`).

Net: the dashcore 0.42→0.43 behaviour is sound on testnet, and there is **no stack regression**. Full serial suite re-run (with the required env var) is in progress for end-to-end confirmation.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

lklimek:
## ✅ Full backend-e2e suite GREEN — zero new regressions from the bump

Full serial run, live testnet, documented `RUST_MIN_STACK=16777216`. Suite reached completion end-to-end (first time on this branch), no overflow/SIGABRT.

**Tally: 55 passed / 4 failed / 0 ignored (647 s).**

### The 4 failures — all pre-existing or environmental (bisected against `v1.0-dev`, fail identically there)
| Test | Bucket |
|---|---|
| `tc_037`, `tc_043` | KNOWN-PREEXISTING — documented "incoming contact request not associated with sender" |
| `tc_014_wallet_platform_lifecycle` | Pre-existing test-data/funding artifact (`AddressesNotEnoughFunds`) — reproduces on `v1.0-dev` |
| `tc_018_fund_platform_address_from_asset_lock` | Environmental — InstantLock proof propagation timeout (broadcast succeeded); reproduces on `v1.0-dev` |

**None are caused by the bump.** The dashcore 0.42→0.43 InstantLock/SPV path did not regress.

### Priority tests (exercise the changed code) — all PASS
SPV sync, send/receive (coin-selection/fee/change via reworked `TransactionBuilder`), is-ours/address derivation, balance reporting (TC001/TC002), identity registration (async `Signer`), **top-up asset lock (tc005)**, withdrawal.

### Remaining manual gates (coverage gaps, not failures)
- [ ] **Shielded lifecycle** (QA-003 / orchard 0.13) — BLOCKED here: shielded ops are feature-gated OFF on this testnet. Needs a shielded-enabled network. (Offload statically re-verified correct.)
- [ ] **Clear-SPV-Data rescan** (QA-001) — unit-verified; manual GUI check pending.

### Net
Code-complete; build/clippy/lib green; QA-audited + re-verified; **runtime-validated on testnet with zero new regressions** to the extent this network allows. Outstanding items are the two manual gates above.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/model/wallet/mod.rs
lklimek: Good catch on the naming — checked it. `PlatformAddress` here is `dash_sdk::dpp::address_funds::PlatformAddress` — an SDK/DPP wrapper type used by the `AddressProvider` / address-funds interface, **not** an Evolution/Platform-chain address. Both sides are the same Core L1 address; `PlatformAddress::try_from(core_address)` is a representation conversion (validates/wraps a Core P2PKH address into the type the SDK's address-resolution API expects), not a cross-chain conversion. So no chain boundary is being crossed — the `Platform` in the name refers to the `dpp` (Dash **Platform** Protocol) crate, which is admittedly confusing since it reads like the Evolution chain. The conversion is infallible in practice for the P2PKH addresses this provider derives.

No code change needed for correctness; the naming is imposed by the upstream SDK type. Flagging upstream that `dpp::address_funds::PlatformAddress` is an easy name to misread would be reasonable if we want to push on it.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/spv/manager.rs
lklimek: Fixed in `d510bcd3`. On `stop()` `Err` we now `warn!` and **drop** `run_future` (no unconditional await) — treating a failed stop as hard cancellation, so the worker can't hang when the run flag wasn't flipped. The `Ok(())` path still awaits so `run()` unwinds its monitors cleanly. Resolving.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/backend_task/shielded/bundle.rs
```diff
@@ -114,20 +114,59 @@ pub fn build_shield_credit(
...
+                &prover,
+                [0u8; 36],
+                platform_version,
+            )
+            .await
+            .map_err(|e| shielded_build_error(e.to_string()))
+        })
+    })
+    .await
+    .map_err(|e| shielded_build_error(format!("shield proof task panicked: {e}")))?
```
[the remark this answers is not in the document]
lklimek: Fixed in `d510bcd3`. The `spawn_blocking` JoinError now branches on `e.is_cancelled()` → "shield proof task was cancelled" vs "…panicked: {e}". Single offload helper, covers both shield paths. Resolving.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/backend_task/shielded/bundle.rs
```diff
@@ -114,20 +114,59 @@ pub fn build_shield_credit(
     let fee_strategy: AddressFundsFeeStrategy =
         vec![AddressFundsFeeStrategyStep::DeductFromInput(0)];
 
-    let wallet = wallet_arc.read()?;
+    // Clone the wallet so the signer is owned by the offloaded build.
+    let wallet = { wallet_arc.read()?.clone() };
+    let platform_version = sdk.version();
     // memo: 36-byte structured memo (4-byte type tag + 32-byte payload); all zeros = empty memo
```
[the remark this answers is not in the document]
lklimek: Fixed in `d510bcd3` — and the deep clone is now gone entirely. `build_shield_transition_offloaded` takes `Arc<RwLock<Wallet>>` (cheap Arc clone, no seed/key duplication); the read guard is taken *inside* the `spawn_blocking` closure and lives only on that dedicated blocking thread across `block_on` (`block_on` doesn't require a `Send` future, so the `!Send` guard is fine there). Applies to both `build_shield_credit` and `shield_credits`. Resolving this and the `shield_credits` duplicate.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

src/spv/tests.rs
```diff
@@ -641,48 +641,68 @@ async fn test_live_testnet_sync_and_shutdown() {
...
 ///
-/// Given a manager with a non-zero filter_committed_height,
+/// Given a loaded wallet synced to a non-zero height,
 /// When clear_data_dir() is called,
-/// Then filter_committed_height is reset to 0.
+/// Then the wallet's synced height is 0 AND `wallets_behind(tip)` reports it as needing
+/// filter coverage (i.e. a genesis rescan will actually run).
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn test_clear_data_dir_resets_filter_committed_height() {
     let (manager, _tm, _tmp_dir) = create_test_manager();
```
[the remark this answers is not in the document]
lklimek: Fixed in `d510bcd3`. Renamed `test_clear_data_dir_resets_filter_committed_height` → `test_clear_data_dir_rewinds_wallet_synced_height` to match the per-wallet `synced_height` + `wallets_behind()` assertion. Resolving.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

thepastaclaw (commented):
## Code Review

Latest delta (d510bcd3) fixes prior finding #1 (SPV cancellation hang): src/spv/manager.rs:1192-1208 now matches client.stop() and only awaits run_future on Ok, treating Err as hard cancellation with a warn log. The shielded offload (src/backend_task/shielded/bundle.rs:140-179) also distinguishes JoinError::is_cancelled from a panic and passes Arc<RwLock<Wallet>> instead of deep-cloning. Prior findings #2 (change-address peek with add_to_state=false) and #3 (highest_found dead state) remain STILL VALID: the new TODO comment at src/backend_task/core/mod.rs:726-734 explicitly acknowledges that the pool is not advanced but does not change runtime behavior, and the derived change address is still used as the real change output on line 771 (not just for fee estimation). Carrying both forward.

🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

- [SUGGESTION] src/backend_task/core/mod.rs:749-755: Change-address derived with add_to_state=false is used as the real change output, leaving it unregistered for monitoring
  Carried forward from the prior review and re-validated at d510bcd3. `account.next_change_address(Some(&account_xpub), false)` is called at line 750-751, and the returned address is then passed to `builder.set_change_address(change_addr.clone())` at line 771 and broadcast as the actual change output — this is not just a fee-estimation peek. In key-wallet 0.43, `next_change_address` forwards `add_to_state` to `address_pool::next_unused`; when no pre-generated unused internal address remains, the fallback `generate_address_at_index(next_index, key_source, add_to_state)` only inserts the new address into the managed account's pool when `add_to_state` is `true`. The new comment at lines 726-734 acknowledges the pool is not advanced and argues the monitor-revision bump is acceptable for bloom-filter rebuild reasons, but a bloom rebuild does not help if the address is not in the wallet's tracked set; the monitoring concern is unresolved. Two consecutive sends before SPV sees the first change tx can also derive the same change index and reuse it. The TODO(v3.1-bump) tag is appropriate, but as long as the runtime behavior persists this stays a correctness/privacy risk.
```
</details>

src/backend_task/core/mod.rs
```diff
@@ -704,68 +704,81 @@ impl AppContext {
...
+            (utxos, change_addr)
```
thepastaclaw, line 755: **🟡 Suggestion: Change-address derived with add_to_state=false is used as the real change output, leaving it unregistered for monitoring**

Carried forward from the prior review and re-validated at d510bcd3. `account.next_change_address(Some(&account_xpub), false)` is called at line 750-751, and the returned address is then passed to `builder.set_change_address(change_addr.clone())` at line 771 and broadcast as the actual change output — this is not just a fee-estimation peek. In key-wallet 0.43, `next_change_address` forwards `add_to_state` to `address_pool::next_unused`; when no pre-generated unused internal address remains, the fallback `generate_address_at_index(next_index, key_source, add_to_state)` only inserts the new address into the managed account's pool when `add_to_state` is `true`. The new comment at lines 726-734 acknowledges the pool is not advanced and argues the monitor-revision bump is acceptable for bloom-filter rebuild reasons, but a bloom rebuild does not help if the address is not in the wallet's tracked set; the monitoring concern is unresolved. Two consecutive sends before SPV sees the first change tx can also derive the same change index and reuse it. The TODO(v3.1-bump) tag is appropriate, but as long as the runtime behavior persists this stays a correctness/privacy risk.

```suggestion
            let utxos: Vec<_> = account.utxos.values().cloned().collect();
            let change_addr = account
                .next_change_address(Some(&account_xpub), true)
                .map_err(|e| TaskError::WalletPaymentFailed {
                    detail: format!("Failed to derive change address: {e}"),
                })?;
            (utxos, change_addr)
```

PR #865, opened 2026-06-19
feat(wallet): unify wallet secret storage on one vault seam with per-secret at-rest encryption

lklimek:
## Why this PR exists
- **Problem**: Wallet secrets were exposed at rest. Identity private keys were written **plaintext** into `det-app.sqlite`; `ClosedSingleKey`'s derived `Debug` could print raw key bytes; no-password key material sat in non-zeroized memory. The storage layer freely serialized plaintext secrets through serde.
- **What breaks without it**: A host-local attacker (another account, malware with file-read, a leaked backup, swap/heap forensics) reading `det-app.sqlite` recovers identity private keys verbatim — full control of the on-chain identity — with **no user mitigation available** (`PrivateKeyData::Encrypted` had no constructor).
- **Blocking relationship**: **Stacked on #860** (`docs/platform-wallet-migration-design`). It depends on that branch's `SecretStore` / `SecretAccess` / `wallet_backend` engine, so the base is set accordingly. Merge after (or onto) #860.

## What was done
- **One secret seam** (`src/wallet_backend/secret_seam.rs`): all three secret classes — HD seed, imported single key, identity private key — store **raw bytes** through a single `put/get/delete_secret` chokepoint into the upstream `SecretStore` vault. No DET-side serialization of plaintext.
- **No-serialization invariant, compiler-enforced**: `SecretBytes` has no `Serialize`, so no persisted struct can embed a plaintext secret. Guarded by `compile_fail` doctests + a source-audit test.
- **Per-secret at-rest encryption (Tier-2)**: password-protected secrets are sealed with the upstream **per-secret** envelope — Argon2id KDF + XChaCha20-Poly1305 AEAD, a **fresh per-object salt**, and AAD bound to `wallet_id‖label`. Same password on two secrets derives different keys; a copied envelope fails the tag under a different scope. DET's own per-wallet AES-GCM envelopes are dropped (retained **decode-only** as legacy migration readers).
- **Per-secret passwords**: each encrypted secret is prompted and (optionally) remembered **independently**. The just-in-time `SecretAccess` cache and prompt are keyed by `SecretScope` (HD `seed_hash` / single-key address / identity label), so one secret's remembered password can never satisfy another. Remember is opt-in / off by default.
- **Identity keys never resident**: `PrivateKeyData::InVault` placeholder; fetched per-signature through the async `Signer` → `resolve_private_key_bytes` → `SecretAccess::with_secret(IdentityKey)` chokepoint and zeroized on scope exit.
[...]

PR #867, opened 2026-06-25
fix(backend-e2e): stabilize PR #860 platform-wallet e2e suite + SEC-001 hardening

lklimek:
## Why this PR exists

- **Problem**: PR #860 (the platform-wallet backend rewrite) shipped a large new wallet backend, but its `backend-e2e` suite was red against live testnet — a mix of real regressions, stale test expectations, and test-infra gaps that made it impossible to tell genuine bugs from noise. We could not gate the rewrite on a suite we couldn't trust.
- **What breaks without it**: #860 ships with a real **funds bug** — Core sends were gated on the `confirmed` balance while the coin selector *and* the "Max" button use spendable, so users get *"insufficient confirmed balance"* rejections on funds they can actually spend; plus wrong error semantics (unknown wallet → `WalletNotLoaded` instead of `WalletNotFound`), a top-up fee that misreports after a backend reload, and an e2e suite too flaky/red to validate the migration.
- **Blocking relationship**: Stacked atop **PR #860** (`docs/platform-wallet-migration-design`). Stabilizes its e2e suite and folds in the SEC-001 at-rest identity-key encryption hardening.

## What was done

**Product fixes**
- **QA-002** — unknown seed hash now returns typed `WalletNotFound` (was `WalletNotLoaded`).
- **QA-006** — meaningful top-up actual-fee after a backend reload, centralized in `model/fee_estimation.rs` (+ a stale-HIGH `balance_before` guard so it can't report the full minted amount as the "fee").
- **QA-010** — gate Core sends on **spendable** balance (= confirmed + unconfirmed), matching the coin selector + Max button, across `send_screen`, dashpay `send_payment`, and the send dialog. Fixes false "insufficient funds" rejections on IS-locked/spendable funds.
- **SEC-001 hardening** (reconciled from the #860 branch and extended): fail-closed when opt-in identity-key protection would leave resident plaintext; defensive guard for legacy `Encrypted` keys; zeroize the migration rollback clone.

**Test / harness fixes**
- QA-003 restart-in-place reconnect contract; QA-004 poll for key visibility (vs fixed sleep); QA-005 tc_012 address-advance marked PENDING; QA-008 real curve-point fixture; QA-013 bounded registration retry; QA-015 `stop_spv` rustdoc + inline comment; QA-016 cold-boot funding buffer; QA-017 funded-wallet SPV-pickup budget 30s→120s.
- Dashpay backend-e2e module deferred pending upstream `dashpay/platform#3841`.

**Quality** — a parallel grumpy-review (security / Rust / consistency) returned **9 findings, all LOW**; 8 were applied (fee fallback, key zeroize, error copy, stale comments/docs, test asserts).

## Testing

- `cargo build --all-features`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo +nightly fmt` — all clean.
- Unit tests pass (incl. 18/18 fee, 12/12 identity-protect).
- **Backend-e2e final run** vs live testnet (3 healthy DAPI nodes): **54 passed / 12 failed**. All 12 reds are **environmental/upstream, not code regressions**:
  - QA-009 asset-lock IS finality (`tc_004`, `tc_018`) + fixture cascade (2): upstream IS-proof not arriving within the hardcoded 300s wait (unchanged from old DET) — testnet Core-P2P/IS degradation.
  - 7× SPV wallet-registration timeout at 120s: testnet Core-P2P degradation (budget bump cut the cascade ~18→7).
  - `tc_014`: pre-existing environmental failure (DAPI exhaustion at the withdrawal step).
- SEC-001 reconciled commits independently security-audited: fail-closed verified, no bypass, no secret leak.

## Known gaps (deferred with upstream references)

- **QA-005** receive-address advancement → `dashpay/rust-dashcore#818` (`next_unused_and_reserve`); tc_012 advance assertion PENDING.
- **dashpay backend-e2e** → `dashpay/platform#3841` ("complete dashpay"); module gated out, re-enable on land.
- **QA-009** asset-lock IS finality → upstream platform-wallet SPV IS-detection / testnet IS health; not DET-tunable (300s hardcoded upstream).
- Two LOW pre-existing at-rest-residual edges (SEC-001-B/C) tracked for follow-up.

## Reviewing this PR

Because **#865 was squash-merged** into the base, the commit *list* looks long (~60 commits) — but the **Files changed** diff is clean: only the e2e-suite stabilization + the grumpy-review fixes, since the SEC-001 content already matches the base. Please review via **Files changed**, not the commit list.

## Attribution

## Summary by CodeRabbit

* **New Features**
  * Wallet send screens now show and use spendable balance, so “Max” and send validation better match what can actually be spent.
  * Identity protection now gives clearer guidance when protection can’t be completed, including a new message for identities stored in an older format.

* **Bug Fixes**
  * Improved identity protection safety to avoid partial changes on incorrect passwords or legacy key formats.
  * Receive address generation now checks that the wallet exists before continuing.
  * SPV reconnect now preserves the same wallet backend across restarts for smoother recovery.


thepastaclaw:
✅ Review complete (commit 9218eb51)

src/backend_task/identity/protect_identity_keys.rs
```diff
@@ -0,0 +1,736 @@
...
+    password: &SecretString,
+) -> Result<usize, TaskError> {
+    let mut reverted = 0usize;
+    for (target, key_id) in keys {
+        if view.scheme(target, *key_id)? == SecretScheme::Protected {
+            let raw = view
+                .get_protected(target, *key_id, password)?
+                .ok_or(TaskError::IdentityKeyMissing)?;
+            view.store_unprotected(target, *key_id, &raw)?;
+            reverted += 1;
```
thepastaclaw, line 276: **🔴 Blocking: Verify every protected key before downgrading any of them**

`seal_identity_keys` first calls `verify_existing_protection_password()` so a mixed-state opt-in retry proves the supplied password opens every protected key before it mutates any label. `unseal_identity_keys` does not have that preflight: it decrypts one key, immediately writes it back with `store_unprotected`, and only then moves to the next protected key. If a partial/corrupt identity ever has a later key sealed under a different password, unprotect returns `IdentityKeyPassphraseIncorrect` after earlier keys have already been downgraded to Tier-1, and `unprotect_identity_keys` aborts before deleting the hint sidecar. That gives the user a failed opt-out while silently stripping protection from a prefix of keys. Please mirror the opt-in flow by proving the supplied password opens all protected labels before any `store_unprotected` write.

thepastaclaw (changes requested):
## Code Review

Degraded review coverage note: the requested ACP reviewer sessions were unavailable in this environment, so I recovered with native Claude-family and Codex reviewer passes and verified the result locally. The spendable-balance and fee-reporting fixes look in-scope, but the SEC-001 opt-out migration has one blocking atomicity gap: it can remove protection from earlier keys before discovering the supplied identity password cannot open a later protected key.

🔴 1 blocking

<details>

```

In `src/backend_task/identity/protect_identity_keys.rs`:
- [BLOCKING] src/backend_task/identity/protect_identity_keys.rs:248-260: Verify every protected key before downgrading any of them
  `seal_identity_keys` first calls `verify_existing_protection_password()` so a mixed-state opt-in retry proves the supplied password opens every protected key before it mutates any label. `unseal_identity_keys` does not have that preflight: it decrypts one key, immediately writes it back with `store_unprotected`, and only then moves to the next protected key. If a partial/corrupt identity ever has a later key sealed under a different password, unprotect returns `IdentityKeyPassphraseIncorrect` after earlier keys have already been downgraded to Tier-1, and `unprotect_identity_keys` aborts before deleting the hint sidecar. That gives the user a failed opt-out while silently stripping protection from a prefix of keys. Please mirror the opt-in flow by proving the supplied password opens all protected labels before any `store_unprotected` write.
```
</details>

thepastaclaw (changes requested):
## Code Review

Incremental review for d63ad9f5: latest delta from 7918d821 only adds a wallet_lifecycle TODO documenting dashpay/rust-dashcore#824, and I found no new latest-delta defects. Carried-forward prior findings: the previous blocking SEC-001 opt-out atomicity issue is STILL VALID on current head because unseal_identity_keys still decrypts and downgrades each protected key before validating later protected keys. The carried-forward inline finding already has an open review thread from the prior SHA, so this exact-SHA review keeps the blocking action without duplicating that thread.

🔴 1 blocking

<details>

```

In `src/backend_task/identity/protect_identity_keys.rs`:
- [BLOCKING] src/backend_task/identity/protect_identity_keys.rs:248-260: Verify every protected key before downgrading any of them
  Carried-forward prior finding: STILL VALID at d63ad9f5. `seal_identity_keys` first calls `verify_existing_protection_password()`, so a mixed-state opt-in retry proves the supplied password opens every protected key before it mutates any label. `unseal_identity_keys` still does not have that preflight: it decrypts one key, immediately writes it back with `store_unprotected`, and only then moves to the next protected key. If a partial/corrupt identity ever has a later key sealed under a different password, unprotect can return `IdentityKeyPassphraseIncorrect` after earlier keys have already been downgraded to Tier-1, and `unprotect_identity_keys` aborts before deleting the hint sidecar. That gives the user a failed opt-out while silently stripping protection from a prefix of keys. Please mirror the opt-in flow by proving the supplied password opens all protected labels before any `store_unprotected` write.
```
</details>

src/backend_task/identity/protect_identity_keys.rs
lklimek: **Fixed in `b98515d9`.** Good catch — the asymmetry was real. `unseal_identity_keys` (opt-out) lacked the all-keys password preflight that `seal_identity_keys` (opt-in) runs, so on a mixed-password identity a prefix of keys could be downgraded via `store_unprotected` before a later-key mismatch aborted, leaving those keys unprotected at rest while `unprotect_identity_keys` returned an error and skipped the hint-sidecar delete — a non-atomic partial strip on the failure path.

The fix adds the identical `verify_existing_protection_password(view, keys, password)?` as the **first statement** of `unseal_identity_keys`, before the downgrade loop. Any password that doesn't open every protected label now returns with **zero `store_unprotected` writes** — the opt-out is atomic by construction rather than relying on the emergent one-password-per-identity invariant. Added a regression test (`unseal_mixed_password_aborts_without_partial_downgrade`) that seals two keys under different passwords, unseals with a password opening only the first, and asserts both keys remain protected with bytes intact. `cargo clippy -D warnings` clean; identity-protect module tests green.

thepastaclaw (commented):
## Code Review

Latest delta (d63ad9f5 → b98515d9) is a focused commit that resolves the prior blocking SEC-001 finding: `unseal_identity_keys` now calls `verify_existing_protection_password` as a preflight before any per-key downgrade, mirroring `seal_identity_keys`, and a Mixed-password regression test (`unseal_mixed_password_aborts_without_partial_downgrade`) proves the CWE-460 partial-downgrade window is closed for opt-out. No carried-forward prior blockers remain (only one prior finding was in flight; both general agents and both rust-quality agents agree it is FIXED). Two new in-scope suggestions hold up under verification: (1) the QA-006 top-up fee helper only falls back at the exact stale-LOW / stale-HIGH boundaries, so partial-stale offsets still produce a wrong displayed actual fee; (2) the SEC-001 plaintext guard now triggers for legacy `Encrypted` keys too, but reuses an error whose recovery copy (close/reopen) cannot resolve that branch, leaving the user in a retry loop. Neither is a blocker — the first is a UX-accuracy gap in a display path, the second a copy mismatch for a no-current-producer code path — so review_action is COMMENT.

🟡 2 suggestion(s)

<details>

```

In `src/backend_task/identity/protect_identity_keys.rs`:
- [SUGGESTION] src/backend_task/identity/protect_identity_keys.rs:174-179: Legacy `Encrypted` keys hit a guard whose recovery action cannot help them
  `reject_resident_identity_plaintext` now returns `TaskError::IdentityKeyProtectionIncomplete` for BOTH residual plaintext (`has_plaintext_for_vault`) AND legacy `Encrypted` keys (`has_encrypted_legacy_keys`). The error's user-facing copy in `src/backend_task/error.rs:280-283` is `"Some of this identity's keys are not fully protected yet. Close and reopen the application, then try protecting this identity again."` That action is correct for the residual-plaintext branch (the eager load-path vault migration retries on next start), but the doc comments on `has_encrypted_legacy_keys` describe these as decode-only with no current producer and no current vault migration path — so closing and reopening will keep hitting the same guard, leaving the user in a retry loop with no self-service exit. Per the project's error-message rules every message must include a concrete action the user can take themselves. Either split the legacy-encrypted case into its own `TaskError` variant with copy that points to a path the user can actually take (e.g. recreate the identity from the wallet, or that this identity needs a developer-assisted migration), or drop the `has_encrypted_legacy_keys` branch from this guard if the eager load-path will already keep these keys from reaching protect.

In `src/model/fee_estimation.rs`:
- [SUGGESTION] src/model/fee_estimation.rs:436-452: Top-up actual-fee helper only catches exact-boundary stale balances
  `resolve_identity_topup_actual_fee` falls back to the estimate only when `delta_fee == 0` (stale-LOW saturates the increase to the full mint) or `delta_fee == expected_credits` (stale-HIGH saturates the increase to zero). Any partial stale offset — for example a cached `balance_before` that is stale-LOW by less than the real fee, or stale-HIGH by less than the mint amount — passes the `0 < delta_fee && delta_fee < expected_credits` guard and is returned as the "actual" fee even though the subtraction is invalid unless `balance_before` is the true pre-top-up value. Concretely, for a 500 000-duff top-up that really paid a 3 000 000-credit fee but whose cached `balance_before` is stale-LOW by 1 000 000 credits, the helper reports 2 000 000 credits as the actual fee. The doc comment frames the staleness mode as "collapses the computed fee to zero", which matches the implementation but understates the gap: non-boundary stale offsets are silently trusted. Impact is bounded (display only, no funds at risk) but the helper does not fully deliver QA-006's stale-safe guarantee. Consider either refreshing `balance_before` from the live source before the call, sanity-checking the delta against the deterministic estimate's plausible range, or documenting the residual so the UI knows the value is only trustworthy when caches are coherent.
```
</details>

src/model/fee_estimation.rs
```diff
@@ -406,6 +406,51 @@ impl PlatformFeeEstimator {
...
+        //  • `delta_fee == expected_credits` — the balance did not grow at all
+        //    (`balance_after <= balance_before`), meaning `balance_before` was stale-HIGH;
+        //    `balance_increase` saturates to 0, so `delta_fee` equals the full minted
+        //    amount and is returned as the "fee", which is nonsensical.
+        if 0 < delta_fee && delta_fee < expected_credits {
+            delta_fee
+        } else {
+            self.estimate_identity_topup()
+        }
+    }
```
thepastaclaw: **🟡 Suggestion: Top-up actual-fee helper only catches exact-boundary stale balances**

`resolve_identity_topup_actual_fee` falls back to the estimate only when `delta_fee == 0` (stale-LOW saturates the increase to the full mint) or `delta_fee == expected_credits` (stale-HIGH saturates the increase to zero). Any partial stale offset — for example a cached `balance_before` that is stale-LOW by less than the real fee, or stale-HIGH by less than the mint amount — passes the `0 < delta_fee && delta_fee < expected_credits` guard and is returned as the "actual" fee even though the subtraction is invalid unless `balance_before` is the true pre-top-up value. Concretely, for a 500 000-duff top-up that really paid a 3 000 000-credit fee but whose cached `balance_before` is stale-LOW by 1 000 000 credits, the helper reports 2 000 000 credits as the actual fee. The doc comment frames the staleness mode as "collapses the computed fee to zero", which matches the implementation but understates the gap: non-boundary stale offsets are silently trusted. Impact is bounded (display only, no funds at risk) but the helper does not fully deliver QA-006's stale-safe guarantee. Consider either refreshing `balance_before` from the live source before the call, sanity-checking the delta against the deterministic estimate's plausible range, or documenting the residual so the UI knows the value is only trustworthy when caches are coherent.

<sub>source: ['codex']</sub>

thepastaclaw (approved):
Clearing my stale CHANGES_REQUESTED state. The prior SEC-001 blocker is fixed at b98515d9; only the nonblocking suggestions from my latest review remain.

src/model/fee_estimation.rs
lklimek: Addressed in `20421533`.

`resolve_identity_topup_actual_fee` now adds a plausibility-band check on top of the two existing physical bounds: the measured `delta_fee` is trusted only when it falls at or below `estimate × 2`, otherwise it falls back to the deterministic estimate. This catches the partial-stale `balance_before` case that previously slipped past the two exact-boundary checks (positive and `< expected_credits`, yet still wrong).

The band is upper-only by design — `estimate_identity_topup()` over-predicts the real fee by ~16× on small top-ups (it bills the full asset-lock processing cost), so a symmetric/tight band would reject legitimate small fees. The doc comment now describes the band honestly, and a regression test (`test_identity_topup_actual_fee_rejects_partial_stale_inflated_delta`) covers an inflated partial-stale delta falling back to the estimate.

lklimek:
Re: the review note (review #4578554863) that `IdentityKeyProtectionIncomplete`'s "close and reopen" instruction is a no-exit retry loop for legacy `Encrypted` keys — addressed in `d78b9306`.

Split into a dedicated `TaskError::IdentityKeyProtectionLegacyFormat` variant for the legacy-`Encrypted` branch, with an honest, code-verified recovery action: *"Some of this identity's keys are saved in an older format that cannot be protected. Load this identity again using its recovery phrase or private key, then try protecting it."* `reject_resident_identity_plaintext` now routes per branch (legacy checked first, since re-loading also clears resident plaintext); the resident-plaintext branch keeps the original "close and reopen" message, which is correct there.

Verified the action is real: `Encrypted` is decode-only with no migration, but re-loading goes through `insert/update_local_qualified_identity`, which overwrites the stored blob, and load paths never emit `Encrypted` keys — so the legacy entries get replaced with protect-compatible ones via the existing "Load Identity" flow. Regression test `protect_rejects_legacy_encrypted_key_with_distinct_error` added.

lklimek:
@coderabbitai review

thepastaclaw (approved):
## Code Review

Latest delta (b98515d9 → d78b9306) cleanly resolves both prior findings and introduces no in-scope defects. prior-b985-1 is FIXED: a dedicated `TaskError::IdentityKeyProtectionLegacyFormat` variant with a concrete self-service recovery instruction is returned for legacy `Encrypted` keys, guarded ahead of the resident-plaintext branch, with a new `protect_rejects_legacy_encrypted_key_with_distinct_error` regression test. prior-b985-2 is FIXED: `resolve_identity_topup_actual_fee` now adds an `estimate × 2` plausibility band that catches partial-stale `balance_before` values inflating the measured delta, with a dedicated `test_identity_topup_actual_fee_rejects_partial_stale_inflated_delta` test; the residual on the stale-LOW deflation side is documented in code with a justification. No new in-scope findings: codex-rust-quality's lone suggestion to swap `PlatformFeeEstimator::new()` for `self.fee_estimator()` in `top_up_identity.rs` is a pre-existing project-wide pattern (30+ call sites) that this PR did not introduce or change, and the same agent already records it as an out-of-scope follow-up; it is dropped here. No CodeRabbit comments.

lklimek:
All 5 CodeRabbit items from review #4579124870 are addressed (pushed in `829c1d30`):

- **`top_up_identity.rs:20`** — now uses `self.fee_estimator()` instead of `PlatformFeeEstimator::new()`, so the displayed estimate *and* the `estimate × 2` fee-plausibility band track the active fee multiplier (`5957c1a2`). Scoped to this site only — it's user-facing and feeds the band; the project-wide `::new()` pattern stays deferred.
- **`identity_db.rs:324`** — `drop(taken);` before `persist()` to minimize the in-memory residency of the migrated plaintext key bytes (`e97592e0`).
- **`tests/backend-e2e/framework/wait.rs:102`** — corrected the stale comment ("confirmed and total" → "spendable and total") to match the code (`829c1d30`).
- **`tests/backend-e2e/core_tasks.rs` (TC-009)** — removed the dead single-key funding-send + 5s sleep; forward-traced and verified it had zero effect on the assertion (`RefreshSingleKeyWalletInfo` returns `SingleKeyWalletsUnsupported` unconditionally) and was wasting real testnet funds (`829c1d30`).
- **`generate_receive_address.rs`** — added regression test asserting an unknown seed hash returns `WalletNotFound` (the #860 `WalletNotFound`/`WalletNotLoaded` split) (`ad8738a3`).

Gate: `build` + `clippy --all-features --all-targets -D warnings` + `+nightly fmt` clean; `fee_estimation` 19/19, `identity_db` 23/23, new receive-address test 1/1, `backend-e2e` compiles. The three inline threads (fee estimator, key residency, stale comment) are resolved.

lklimek:
@coderabbitai review

tests/backend-e2e/core_tasks.rs
```diff
@@ -199,11 +199,12 @@ async fn test_tc005_create_top_up_asset_lock() {
...
-// single-key send can reach the network in both RPC and SPV modes. UTXO
-// discovery still requires Dash Core; in SPV mode the test verifies that
-// `RefreshSingleKeyWalletInfo` returns `OperationRequiresDashCore` and stops
-// before attempting the send (no spendable UTXOs available).
+// Single-key wallets are intentionally unsupported this release (PROJ-007 /
+// single-key-mock.md, Decision #7): every single-key task arm returns the typed
+// `SingleKeyWalletsUnsupported`. The funding step still exercises a real send
+// from the framework HD wallet to the single-key address, then the test
+// verifies that `RefreshSingleKeyWalletInfo` returns `SingleKeyWalletsUnsupported`
+// and stops before attempting the single-key send.
```
thepastaclaw: **💬 Nitpick: TC-009 header comment still describes the removed funding step**

The latest delta removed the framework-wallet funding send from TC-009, and the test body now goes straight from constructing the `SingleKeyWallet` to asserting that `RefreshSingleKeyWalletInfo` returns `SingleKeyWalletsUnsupported`. The header comment still claims "The funding step still exercises a real send from the framework HD wallet to the single-key address", which contradicts the test body and will mislead anyone re-enabling this flow into thinking funding coverage is still present. The accompanying commit (`test(backend-e2e): drop dead single-key funding block, fix stale comment`) intended to address this — the comment update is incomplete.

```suggestion
// TC-009: SendSingleKeyWalletPayment
//
// Single-key wallets are intentionally unsupported this release (PROJ-007 /
// single-key-mock.md, Decision #7): every single-key task arm returns the typed
// `SingleKeyWalletsUnsupported`. The test verifies that
// `RefreshSingleKeyWalletInfo` returns `SingleKeyWalletsUnsupported` and stops
// before attempting any funding or single-key send.
```

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Incremental delta from d78b9306 to 829c1d30 is small and focused, with no carried-forward prior findings from that review. Verification confirms two nonblocking in-scope items: a QA-006 sibling in `top_up_identity_from_platform_addresses` still uses a hardcoded `PlatformFeeEstimator::new()` despite the same-PR fix to its sibling, and the TC-009 header comment still describes the funding step that this PR removed.

🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>
<summary>Findings not posted inline (1)</summary>

These findings could not be anchored to the current diff, but they are still part of this review.

- **[SUGGESTION] `src/backend_task/identity/mod.rs:896`: QA-006 sibling: top_up_identity_from_platform_addresses still ignores active fee multiplier** — This PR's fix in `src/backend_task/identity/top_up_identity.rs` correctly routes the top-up estimate through `self.fee_estimator()` so the reported `FeeResult.estimated` tracks the cached `fee_multiplier_permille`. The sibling entry point `top_up_identity_from_platform_addresses` calls the same `estimate_identity_topup()` method on a hardcoded `PlatformFeeEstimator::new()`, so under any non-default fee multiplier the platform-addresses top-up will misreport its estimate in exactly the way QA-006 describes. Same domain, same function, same one-line fix; `AppContext::fee_estimator()` is already in scope here.

</details>

<details>

```

In `src/backend_task/identity/mod.rs`:
- [SUGGESTION] src/backend_task/identity/mod.rs:896: QA-006 sibling: top_up_identity_from_platform_addresses still ignores active fee multiplier
  This PR's fix in `src/backend_task/identity/top_up_identity.rs` correctly routes the top-up estimate through `self.fee_estimator()` so the reported `FeeResult.estimated` tracks the cached `fee_multiplier_permille`. The sibling entry point `top_up_identity_from_platform_addresses` calls the same `estimate_identity_topup()` method on a hardcoded `PlatformFeeEstimator::new()`, so under any non-default fee multiplier the platform-addresses top-up will misreport its estimate in exactly the way QA-006 describes. Same domain, same function, same one-line fix; `AppContext::fee_estimator()` is already in scope here.

In `tests/backend-e2e/core_tasks.rs`:
- [NITPICK] tests/backend-e2e/core_tasks.rs:200-207: TC-009 header comment still describes the removed funding step
  The latest delta removed the framework-wallet funding send from TC-009, and the test body now goes straight from constructing the `SingleKeyWallet` to asserting that `RefreshSingleKeyWalletInfo` returns `SingleKeyWalletsUnsupported`. The header comment still claims "The funding step still exercises a real send from the framework HD wallet to the single-key address", which contradicts the test body and will mislead anyone re-enabling this flow into thinking funding coverage is still present.
```
</details>

lklimek:
Addressed the two remaining items from the latest review in `9218eb51`:

- **Fee estimator (suggestion)** — completed the top-up estimator wiring: `top_up_identity_from_platform_addresses` (`mod.rs:896`) now uses `self.fee_estimator()`, matching `top_up_identity.rs` from the prior round. Scope is deliberately **top-up estimates only** — they feed the `resolve_identity_topup_actual_fee` plausibility band, so they must track the active multiplier. The transfer/other identity estimate sites (`transfer_to_addresses`, `estimate_credit_transfer`, `register_identity`, `add_key`, `register_dpns`, `withdraw`) stay on the default multiplier as part of the **project-wide fee-multiplier cleanup**, which is deferred and tracked separately — drawing the line at top-ups avoids a partial, inconsistent transfer-path fix in this PR.
- **TC-009 comment (nitpick)** — rewrote the header to present-state, dropping the stale "still exercises a real send…" line (the funding step was removed earlier in this PR); thread resolved.

Gate: `build` + `clippy --all-features --all-targets -D warnings` + `+nightly fmt` clean; `fee_estimation` 19/19, `--lib identity` 132/132, `backend-e2e` compiles.

src/backend_task/identity/mod.rs
```diff
@@ -889,11 +889,11 @@ impl AppContext {
...
         wallet_seed_hash: WalletSeedHash,
     ) -> Result<BackendTaskSuccessResult, TaskError> {
-        use crate::model::fee_estimation::PlatformFeeEstimator;
         use dash_sdk::platform::transition::top_up_identity_from_addresses::TopUpIdentityFromAddresses;
 
-        // Estimate fee for top-up from platform addresses
-        let estimated_fee = PlatformFeeEstimator::new().estimate_identity_topup();
+        // Estimate the top-up fee with the active network fee multiplier
+        // (context estimator) so the figure shown to the user is accurate.
+        let estimated_fee = self.fee_estimator().estimate_identity_topup();
```
thepastaclaw, line 896: **🟡 Suggestion: Use the platform-address top-up estimator on the address-funded path**

`top_up_identity_from_platform_addresses` calls `estimate_identity_topup()`, which is the wallet-funded variant and ignores per-input costs. The model already provides `estimate_identity_topup_from_addresses(input_count)` (model/fee_estimation.rs:376), which factors in `address_funding_asset_lock_cost`, `address_funds_transfer_input_cost * inputs`, plus a 20% safety buffer — and the matching UI screen already calls it (`src/ui/identities/top_up_identity_screen/by_platform_address.rs:103`). Because the backend returns `FeeResult::new(estimated_fee, estimated_fee)` from the same `u64`, using the wallet-funded estimator can under-report fees for this flow and silently diverges the UI estimate from the backend estimate, especially with >1 input.

```suggestion
        let estimated_fee = self
            .fee_estimator()
            .estimate_identity_topup_from_addresses(inputs.len());
```

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Both prior findings are FIXED in the latest delta: `top_up_identity_from_platform_addresses` now routes through `self.fee_estimator()` (prior-829-1), and the TC-009 header comment now matches the present-state test body (prior-829-2). One new in-scope suggestion from codex-rust-quality survives verification: the same line at src/backend_task/identity/mod.rs:896 still uses the wallet-funded `estimate_identity_topup()` even though the function is explicitly the platform-address top-up, and a dedicated `estimate_identity_topup_from_addresses(input_count)` exists in `model/fee_estimation.rs` and is already used by the matching UI screen. No CodeRabbit inline comment IDs were supplied, so coderabbit_reactions is empty.

🟡 1 suggestion(s)

<details>

```

In `src/backend_task/identity/mod.rs`:
- [SUGGESTION] src/backend_task/identity/mod.rs:896: Use the platform-address top-up estimator on the address-funded path
  `top_up_identity_from_platform_addresses` calls `estimate_identity_topup()`, which is the wallet-funded variant and ignores per-input costs. The model already provides `estimate_identity_topup_from_addresses(input_count)` (model/fee_estimation.rs:376), which factors in `address_funding_asset_lock_cost`, `address_funds_transfer_input_cost * inputs`, plus a 20% safety buffer — and the matching UI screen already calls it (`src/ui/identities/top_up_identity_screen/by_platform_address.rs:103`). Because the backend returns `FeeResult::new(estimated_fee, estimated_fee)` from the same `u64`, using the wallet-funded estimator can under-report fees for this flow and silently diverges the UI estimate from the backend estimate, especially with >1 input.
```
</details>

PR #869, opened 2026-07-01
feat(identity): fund-first onboarding — friendly labels, F1 avatar, empty-wallet guard

lklimek:
## Why this PR exists
- **Problem**: The Identity onboarding didn't match the redesign spec (`docs/ai-design/2026-04-22-identity-dashpay-redesign/design-spec.md` §B.1/§B.9/§B.10/§B.11). The F1 onboarding had no avatar, the create wizard was alias-first instead of fund-first, and the funding/load controls used engineering jargon ("Wallet Balance", "Unused Evo Funding Locks", "By Identity ID", "asset lock") that the Everyday-User persona (`docs/personas/everyday-user.md`) can't parse.
- **What breaks without it**: A first-time user clicking "Create my first identity" had to name the identity *before* choosing how to fund it, then met developer terminology on the flagship onboarding path. Two follow-up review rounds also found: a zero-balance wallet could get stranded on a dead-end funding screen with no explanation; the balance gate itself could be bypassed by unspendable (immature/CoinJoin-locked) funds; and two "Max amount" buttons could offer more than was actually spendable. A later audit also found that important banners (e.g. "this saved wallet couldn't be reopened") could silently auto-dismiss after a few seconds, before the user had a chance to read them. Separately, the "Send to" address field offered no way to narrow a long autocomplete list beyond plain substring text, making it hard to find a specific address once a user had several wallets and address categories loaded.

## What was done
- **A1 — F1 avatar**: abstract avatar silhouette + soft Dash-blue radial gradient above "Welcome to Identities." (`paint_abstract_avatar` in `src/ui/identity/avatar.rs`, wired in `onboarding.rs`). §B.1.
- **A2 — friendly labels + default**: `FundingMethod::Display` → "From your wallet (recommended)" / "Recover an unfinished funding" / "Use a Platform address"; Load-Identity tabs → "Identity ID & private key" / "From my wallet" / "My username" with a jargon-free caption; `UseWalletBalance` becomes the recommended pre-selected default **when the wallet has balance**. §B.9/§B.11.
- **A4 — fund-first reorder**: Create wizard is now 1. Choose funding → 2. amount/selection → 3. optional local alias → Create. No username step (deferred per §B.10). §B.10.
- **QA hardening** (from an adversarial review pass): balance-gated default with §B.1 insufficient-funds / no-wallet banners so an empty wallet never lands on a dead-end; de-jargoned the "unfinished funding" recovery screen (raw TxID/Vout now Advanced-only); fixed a stale "pick a username" CTA tooltip; disabled the recovery-screen Create button when nothing is selected; Top-Up screen pre-select parity; restored all three funding methods in `docs/user-stories.md` IDN-001.
- **Second review round** (13 automated findings triaged, 10 confirmed and fixed, 3 retracted as false positives): the insufficient-funds gate now checks `.spendable()` credits instead of `.total` via a shared `spendable_covers_minimum()` helper; the Top-Up wizard gained its own insufficient-funds banner (previously silently no-op'd on low balance); Create-Identity's funding pre-selection now requires fee-sufficiency, not just a non-zero balance; "alias" terminology standardized; Top-Up gets a context-appropriate funding-method label; a Title-Case label inconsistency fixed; Top-Up's own asset-lock recovery screen got the same de-jargoning pass as its sibling; design-spec doc synced with shipped copy.
- **Third review round** (re-verification with a call-tree walk found one fix incomplete plus one new instance of an already-fixed bug class): the RwLock panic-on-poison fix only patched the exact flagged lines — sibling reads on the *same* locks, one call away, still panicked; extended the defensive handling to those 3 sites (`show_wallet_balance` in both wizards, a second `funding_method` read in Top-Up). Also found and fixed both wizards' "Max amount" buttons still computing from `.total` + a bare `* 1000` instead of `.spendable()` + the shared `max_amount_after_fee_reserve()` helper.
- **Fourth round — frontend TODO triage**: the Create-Identity wallet picker (`render_wallet_selection` in `add_new_identity_screen/mod.rs`) listed only each wallet's alias, so users chose a wallet blind and discovered insufficient funds only later, deep in the funding flow. It now shows each wallet's **spendable** balance next to its alias (`alias — X DASH`), in both the ComboBox's selected text and every entry — same `.spendable()`-not-`.total` convention as the Max-amount fixes above.
- **Fifth round — banner auto-dismiss audit**: audited every `MessageBanner` call site across the PR (~300 sites, 64 files) and classified each by whether it reports something that genuinely affects the user (data loss, a major feature not working) versus routine noise (success confirmations, validation hints, in-progress status). Added `BannerHandle::disable_auto_dismiss()` and `OptionBannerExt::raise_persistent()` so a banner can stay on screen until manually dismissed instead of vanishing on the default 5s/9s timer. Made 49 sites persistent — dominated by a single recurring class: any `try_open_wallet_no_password` "this wallet's saved data looks damaged" error (38 of the 49), plus failed identity/key removal, failed key persistence, lost connection, and the generic backend-task-error fallthrough in `app.rs` that catches send/broadcast/asset-lock failures for screens that don't override `display_task_error`. Left alone: successes, validation nudges, and `.with_elapsed()` in-progress waits.
- **Sixth round — search tags + pills on the address picker**: the shared `AddressInput` autocomplete (`src/ui/components/address_input.rs`, used by Send, Shield, Unshield, and a wallets-screen dialog) gained a github-style search DSL: `type:core|platform|shielded|identity` and `wallet:<name>`, repeatable per key (OR), combinable across keys (AND), plain words still do free-text search as before. Each suggestion row now shows a wallet pill and a type pill instead of a plain "[alias] " text prefix, plus the address and an optional name in parens. The empty-state hint is now a live legend of the tags actually usable in that instance (capped at 5 wallets, with a visible "+N more" past that), replacing a static "enter an address" hint. Independently adversarially reviewed (parser edge cases, panic risk, the 5-wallet boundary, the two other callers' compatibility, i18n) — 4 low-severity polish items found, none blocking.

## Also included
- **Unrelated tokens-screen bugfix**, bundled into this PR at the maintainer's request rather than the identity-onboarding scope above: `src/ui/tokens/add_token_by_id_screen.rs` drove its status machine off substring-matching banner text (`msg.contains("DataContract successfully saved")`, `"Contract not found"`, etc.) — a "never parse error strings" violation. Investigation found the branches were dead code under the current `BackendTaskSuccessResult` routing: `SavedContract`/`ContractNotFound`/`TokenNotFound`/`FetchedTokenBalances` all bypass `display_message` and land in `display_task_result`, which had no handling for them. Net effect before the fix: clicking "Import Token" never reached the success screen, and the contract-ID → token-ID fallback search never fired. Replaced with typed-variant dispatch: `SavedContract` → `Complete`; `ContractNotFound` disambiguated via a new `SearchKind` (in-flight contract-ID lookup → retry as token ID; contract missing behind an already-found token → genuine error) so the two cases no longer collide; `TokenNotFound` → a calm, jargon-free error banner. `display_message` now reacts only to message *type*, never text.

## Testing
- `cargo build`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo +nightly fmt --all -- --check` — all clean.
- `cargo test --all-features --workspace`: 1329 lib tests passed (0 failed, 1 ignored), 192 kittest passed, 10 backend-e2e-adjacent/e2e passed, 3 legacy-table-surface passed, all doc-tests passed.
- Banner-timer audit specifically re-verified with the targeted kittest suites (`migration_banner`, `progress_overlay`, `skipped_wallets_banner`): 66/66 passed, included in the totals above.
- Address-search-tags feature specifically re-verified with its dedicated kittest suite (6/6) plus 72 inline unit tests covering the tag parser, wallet/type pill population, and the 5-wallet trim boundary, included in the totals above.
- Manually verified on testnet (headless GUI): F1 avatar, fund-first step order, de-jargoned recovery screens on both wizards, corrected tooltip, and the empty-wallet default (no dead-end) on both Create-Identity and Top-Up.
- The two frontend-TODO fixes (wallet-balance picker, add-token typed outcomes) were each built and verified independently in isolated worktrees, then re-verified together on the merged branch — all passing.

## Breaking changes
None — presentation-layer plus balance-source correctness fixes (labels, layout, an enum default, added banners, `.total` → `.spendable()` at several call sites, banner persistence, address-picker search/rendering). No backend/DB/secret/finality logic touched.

## Known deviations / follow-ups
- **Persona-gating not implemented** (spec §B.11 would hide the wallet-derivation load mode behind Advanced for Alex) — intentionally deferred by product decision; all load modes stay visible.
- **No username step** in the create flow — deferred per §B.10's "username can be registered later from Settings".
- **Funding pre-selection staleness**: the fee-sufficiency check that drives pre-selection is computed once when the screen/wallet is set (`new_with_wallet`) and not recomputed on a later wallet switch — marked with `TODO(bilby)` in `add_new_identity_screen/mod.rs`. Safely fixing this needs to track whether the user manually chose a funding method, so a wallet switch never clobbers an explicit choice — **open GitHub review thread awaiting a product decision** on the intended behavior before implementing.
- **Test-coverage gaps** (deferred): no kittest/e2e exercising the full `AddNewIdentityScreen` wizard end-to-end; avatar-wiring in `onboarding::render` not asserted.
- **Address-search-tags polish** (non-blocking, low severity): the "change" address annotation and the "+N more" trim indicator are hardcoded English literals rather than i18n-extractable placeholders; tracked as a follow-up.

## Base / stacking
Targets `docs/platform-wallet-migration-design`. PR #860 (through its subsequent platform-pin bump to `44c20e3`, its simplification-audit fixes, and its balance-derivation single-sourcing fix — the latter independently adversarially reviewed twice) and PR #871 are both fully merged into this branch's history via that base — current base commit `8bd1d1d2`. No pending overlap remains; re-synced and re-verified (fmt/clippy/full test suite: 1327 lib + 187 kittest + 10 e2e + doctests, 0 failed) after each base update, most recently for this merge.

## Checklist
- [x] Conventional commits
- [x] Build / clippy / fmt clean
- [x] Tests pass (1329 lib + 192 kittest + doc-tests, re-verified after each base sync and after each of the last two rounds)
- [x] User stories updated (IDN-001, IDH-001)
- [x] Rebase check vs base branch — resolved, base fully merged into this branch
- [ ] Product decision needed on funding-pre-selection staleness (open review thread) before that item can close


src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
```diff
@@ -23,6 +24,49 @@ impl AddNewIdentityScreen {
...
+    /// the Wallets screen (design-spec calls it "Go to Receive"; this app has
+    /// no separate top-level Receive screen, so the link goes to Wallets,
+    /// where the user's receiving address lives) and report that the caller
+    /// should stop rendering this step. Returns `None` when the balance is
+    /// sufficient, or when no wallet is selected (handled earlier by the
+    /// caller's own no-wallet gate).
+    fn render_insufficient_wallet_balance_banner(&self, ui: &mut egui::Ui) -> Option<AppAction> {
+        let selected_wallet = self.selected_wallet.as_ref()?;
+        let seed_hash = selected_wallet.read().unwrap().seed_hash();
+        let available_credits = self.app_context.snapshot_balance(&seed_hash).total * 1000; // duffs -> credits
```
Claudius-Maginificent: **Balance source: `.total` includes unspendable funds**

This reads `DetWalletBalance.total`, which per the struct's own doc comment includes immature-coinbase and CoinJoin-locked funds the coin selector cannot actually spend. The "not enough Dash" banner above this line can be skipped by an under-funded wallet, letting identity creation proceed and fail downstream. Use `.spendable()`/`.confirmed` instead.

A sibling commit (`ed5f6b55`, on `fix/ddda6e88-confirmed-funding`) already fixes an equivalent bug elsewhere and is not merged into this branch — worth pulling in the same fix here.

Also: the `* 1000` duffs→credits conversion on this line duplicates a magic number; a `CREDITS_PER_DUFF` constant already exists in a sibling module and is used elsewhere in the codebase.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
```diff
@@ -23,6 +24,49 @@ impl AddNewIdentityScreen {
         }
     }
 
+    /// If the selected wallet can't cover even the estimated identity-creation
+    /// fee, render design-spec §B.1's "not enough Dash" banner with a link to
+    /// the Wallets screen (design-spec calls it "Go to Receive"; this app has
```
Claudius-Maginificent, line 36: **Doc comment flags a design-spec deviation, but only here**

This comment correctly explains why the link goes to "Wallets" instead of design-spec §B.1's literal "Go to Receive" wording — but the deviation isn't reflected in `docs/ai-design/.../design-spec.md`, which still shows the old label verbatim. Worth a one-line update there so the spec and implementation don't silently diverge for the next reader.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
```diff
@@ -23,6 +24,49 @@ impl AddNewIdentityScreen {
...
+    /// no separate top-level Receive screen, so the link goes to Wallets,
+    /// where the user's receiving address lives) and report that the caller
+    /// should stop rendering this step. Returns `None` when the balance is
+    /// sufficient, or when no wallet is selected (handled earlier by the
+    /// caller's own no-wallet gate).
+    fn render_insufficient_wallet_balance_banner(&self, ui: &mut egui::Ui) -> Option<AppAction> {
+        let selected_wallet = self.selected_wallet.as_ref()?;
+        let seed_hash = selected_wallet.read().unwrap().seed_hash();
```
Claudius-Maginificent: **New `.read().unwrap()` on a shared `RwLock` — panics on poison**

Same pattern recurs in `top_up_identity_screen/mod.rs:164` (`wallet.read().unwrap()`). A panic anywhere else while holding this lock poisons it for every reader, including this new code path, taking down the UI thread. Consider `.unwrap_or_else(PoisonError::into_inner)` or the existing safe-read helper already used elsewhere in this file.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
```diff
@@ -23,6 +24,49 @@ impl AddNewIdentityScreen {
...
+        let available_credits = self.app_context.snapshot_balance(&seed_hash).total * 1000; // duffs -> credits
+
+        let key_count = self.identity_keys.others.len() + 1; // +1 for master key
+        let minimum_credits = self
+            .app_context
+            .fee_estimator()
+            .estimate_identity_create(key_count);
+
```
Claudius-Maginificent, line 60: **No test coverage for this balance-gate arithmetic**

`render_insufficient_wallet_balance_banner` and its `available_credits` computation (including the `.total` vs `.spendable()` question flagged above) have no unit test. A boundary test at "balance == estimated fee ± 1" would have caught that issue directly.

src/ui/identities/add_new_identity_screen/by_using_unused_asset_lock.rs
```diff
@@ -146,7 +144,18 @@ impl AddNewIdentityScreen {
             });
         ui.add_space(10.0);
 
-        if ui.button("Create Identity").clicked() {
+        self.render_alias_input(ui, step_number + 1);
+
+        let can_create = self.funding_asset_lock.is_some();
+        let button = egui::Button::new(RichText::new("Create Identity").color(Color32::WHITE))
```
Claudius-Maginificent, line 150: **Create-button affordance drifts across the three funding screens**

This screen disables the button (`add_enabled(can_create, ...)`); the sibling `by_using_unused_balance.rs` and `by_platform_address.rs` screens hide/show it instead. Pick one convention — inconsistent affordance (disabled vs. absent) across three parallel steps of the same wizard is a small but real UX inconsistency. Also worth double-checking the `action |=` vs `action =` accumulation pattern across these three screens for a possible dropped-action bug.

src/ui/identities/add_new_identity_screen/mod.rs
```diff
@@ -60,17 +60,42 @@ pub enum FundingMethod {
...
 }
 
+/// The funding-method chooser's starting state for a wallet that either does
+/// or doesn't have spendable balance (§B.9: pre-select `UseWalletBalance`
+/// only when it is actually available; otherwise start unselected rather
+/// than land on a method the ComboBox itself wouldn't offer). Shared by
+/// `AddNewIdentityScreen` and `TopUpIdentityScreen`, which both render this
+/// same chooser. A pure function so the decision is testable without
+/// constructing a real wallet/balance snapshot.
+pub(crate) fn default_funding_state(
```
Claudius-Maginificent, line 102: **"Presentation-layer only, safe to land independently of #860" doesn't hold**

`default_funding_state` here and its call sites (used by both `AddNewIdentityScreen` and `TopUpIdentityScreen`) are exactly the kind of shared logic PR #860 is also touching — confirmed #860's current branch already modifies every UI file this PR modifies except two. Please re-check for conflicts/behavioral overlap before merge, not just at rebase time.

src/ui/identities/add_new_identity_screen/mod.rs
```diff
@@ -157,13 +182,24 @@ impl AddNewIdentityScreen {
             }
         }
 
+        // §B.9: `UseWalletBalance` is the recommended, pre-selected primary
+        // path — but only when the wallet actually has funds to offer it
+        // with, matching the same availability gate the ComboBox itself
+        // applies in `render_funding_method` (`has_balance`), so the visibly
+        // "selected" method is never one the dropdown wouldn't actually offer.
+        let wallet_has_balance = selected_wallet.as_ref().is_some_and(|wallet| {
```
Claudius-Maginificent: **Pre-selection threshold doesn't match the sufficiency threshold**

`wallet_has_balance` here pre-selects `UseWalletBalance`/`ReadyToCreate` on *any* positive balance, but the new insufficient-funds banner requires the balance to cover the estimated fee. A dust-balance wallet gets pre-selected into "ready to create" and then immediately blocked by the banner — a confusing, self-contradicting first impression.

Separately: this is computed once at construction time (`new_with_wallet`), so if the screen opens before wallet sync completes, the pre-selection can go stale relative to the real balance.

src/ui/identities/add_new_identity_screen/mod.rs
```diff
@@ -1024,6 +1060,51 @@ impl AddNewIdentityScreen {
         ui.add_space(10.0);
     }
 
+    /// The optional local-alias step (design-spec §B.10: fund-first).
+    ///
+    /// Rendered by each funding-method branch just before its Create/Register
+    /// button, once the amount or lock for that method is chosen. This is a
+    /// Dash Evo Tool nickname stored locally, not a DPNS username.
+    fn render_alias_input(&mut self, ui: &mut egui::Ui, step_number: u32) {
```
Claudius-Maginificent, line 1160: **"Alias" vs "nickname" terminology drift**

This function is `render_alias_input`, its doc comment calls it "a Dash Evo Tool nickname," and the UI heading says "Set a local alias (optional)." Three different terms for the same concept in a handful of lines — pick one (the doc comment's clarification that it's *not* a DPNS username is good and worth keeping, just align the noun).

src/ui/identities/top_up_identity_screen/mod.rs
```diff
@@ -259,15 +274,15 @@ impl TopUpIdentityScreen {
...
                 );
 
                 ui.add_enabled_ui(has_any_unused_asset_lock, |ui| {
                     if ui
                         .selectable_value(
                             &mut *funding_method,
                             FundingMethod::UseUnusedAssetLock,
-                            "Unused Asset Locks",
+                            format!("{}", FundingMethod::UseUnusedAssetLock),
                         )
```
Claudius-Maginificent, line 294: **Shared `FundingMethod::Display` text doesn't fit the Top-Up context**

This reuses the same `Display` impl the Create-identity wizard uses, but at least one variant's label ("Recover an unfinished funding...") reads oddly in a top-up flow that was never mid-funding to begin with. Consider a context-aware label or a Top-Up-specific wrapper.

src/ui/identities/top_up_identity_screen/mod.rs
```diff
@@ -156,6 +156,21 @@ impl TopUpIdentityScreen {
                 true
             } else if let Some(wallet) = wallets.values().next() {
                 if self.wallet.is_none() {
+                    // §B.9 / QA-006: the very first time a wallet resolves with
+                    // nothing chosen yet, apply the same pre-selection the
+                    // create-identity wizard uses (`default_funding_state`) —
+                    // recommend `UseWalletBalance` only when this wallet
+                    // actually has funds to offer it with.
+                    if *self.funding_method.read().unwrap() == FundingMethod::NoSelection {
```
Claudius-Maginificent: **Top-Up wizard has no equivalent of the new insufficient-funds banner**

This PR adds `render_insufficient_wallet_balance_banner` to the Create-identity screen (`by_using_unused_balance.rs`), but the Top-Up wizard's auto-recommendation logic here has no equivalent check. A low/zero-balance user is still auto-steered onto `UseWalletBalance`, and clicking "Top Up Identity" silently no-ops with no feedback — the exact "dead-end funding screen" failure mode this PR just fixed on the other flow. Recommend porting the same banner/gate here.

Claudius-Maginificent:
Automated review — consolidated from two independent passes (Opus: 11 findings, Sonnet 5: 9 findings), cross-verified against source and git history, 0 confirmed false positives. 10 inline comments posted on this PR's diff; 3 additional findings couldn't be anchored to a changed line and are noted here instead:

- **Residual jargon**: `by_using_unused_balance.rs:20` — the `"Wallet Balance: {:.8} DASH"` label wasn't included in this PR's friendly-copy relabeling pass, unlike its siblings.
- **Title-Case nit**: `by_platform_address.rs:106` — `ComboBox::from_label("Platform Address")` retains Title-Case, out of step with the PR's new casing convention elsewhere.
- **Top-Up sub-screens keep old jargon**: `top_up_identity_screen/by_using_unused_asset_lock.rs` (not touched by this PR) still says "Select an unused asset lock" / "asset lock" throughout — if the friendly-copy pass is meant to eventually cover the Top-Up wizard's own sub-screens too, this file is still pending it.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Verified: `.total` includes immature/CoinJoin-locked funds that `.spendable()` excludes (per the field's own doc comment) — confirmed real, this can let an under-funded wallet pass the gate. `CREDITS_PER_DUFF` is indeed used in 20+ other call sites (even locally shadowed in `by_platform_address.rs:14`), so the `* 1000` here is a genuine duplicated magic number. Note: `ed5f6b55` / `fix/ddda6e88-confirmed-funding` don't exist anywhere in this repo's history — that reference doesn't corroborate anything, disregard it.

Recommend: swap `.total` → `.spendable()` and use the existing `CREDITS_PER_DUFF` constant before merge. Not fixed yet — leaving open.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Verified: `docs/ai-design/2026-04-22-identity-dashpay-redesign/design-spec.md:185-186` still shows the literal "[Go to Receive]" wording while the shipped code intentionally uses "Go to Wallets" (the doc comment explains why). Real doc/code drift — recommend a one-line spec update to match. Not fixed yet — leaving open.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Verified: both `.read().unwrap()` sites panic on lock poisoning as described. Correcting my own finding: the "safe-read helper already used elsewhere in this file" claim was wrong — that defensive `match .read() { Ok/Err }` pattern actually lives in the sibling `by_using_unused_asset_lock.rs`, not in this file. Recommend porting that pattern here and to `top_up_identity_screen/mod.rs:164`. Not fixed yet — leaving open.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Verified: no `#[cfg(test)]` module exists in this file at all — `render_insufficient_wallet_balance_banner`/`available_credits` have zero coverage. Contrast with `default_funding_state` in `mod.rs`, which this PR did extract and unit-test. Recommend a boundary test once the `.total`/`.spendable()` fix (see the sibling thread on this same file) lands, so the test exercises correct semantics. Not fixed yet — leaving open.

src/ui/identities/add_new_identity_screen/by_using_unused_asset_lock.rs
lklimek: Re-verified and retracting this one: `by_platform_address.rs` actually disables its button identically to `by_using_unused_asset_lock.rs` (`ui.add_enabled(can_create, button)`) — I misattributed it as hide/show. Only `by_using_unused_balance.rs` differs, via an early return rather than disabling. Also traced `action |=` vs `action =` across all three files: `action` is always `AppAction::None` at the assignment point in current code, so no dropped-action bug exists. Resolving — no fix needed.

src/ui/identities/add_new_identity_screen/mod.rs
lklimek: Re-verified and retracting the "except two" figure: PR #869's 9 changed files have full 9/9 overlap with PR #860's changed files, not partial. More importantly, the underlying concern is moot — PR #860's head (`e8d40d8c`) is already an ancestor of this branch (merged in via `c9ba1fc9`), so there's no pending overlap left to reconcile. Resolving.

src/ui/identities/add_new_identity_screen/mod.rs
lklimek: Verified both sub-claims: `snapshot_has_balance` triggers on any positive balance including dust, with no fee comparison, so a dust-balance wallet gets pre-selected into `ReadyToCreate` and then immediately blocked by the new insufficient-funds banner — a confusing first impression. Also confirmed pre-selection is computed once in `new_with_wallet` and never recomputed on wallet switch or sync completion. Recommend gating pre-selection on the same fee-sufficiency check the banner uses. Not fixed yet — leaving open.

src/ui/identities/add_new_identity_screen/mod.rs
lklimek: Verified, and it's slightly worse than originally flagged: the on-screen caption directly under the "Set a local alias" heading and "Alias:" label literally reads "this is a Dash Evo Tool nickname, not a DPNS username" — so both terms are simultaneously visible to the user on the same screen. Recommend picking one noun (suggest "alias", since the two labels already use it) and adjusting the caption to match. Not fixed yet — leaving open.

src/ui/identities/top_up_identity_screen/mod.rs
lklimek: Verified: `FundingMethod::UseUnusedAssetLock`'s shared `Display` text ("Recover an unfinished funding") is reused verbatim in the Top-Up wizard, but "recover"/"unfinished" framing doesn't fit a flow that tops up an already-fully-created identity. Low-severity copy nit — recommend a context-aware label or Top-Up-specific wrapper when convenient; not blocking. Not fixed yet — leaving open.

src/ui/identities/top_up_identity_screen/mod.rs
lklimek: Verified as a real gap: `top_up_identity_screen` has zero occurrences of any insufficient-funds messaging, and the "Top Up Identity" button's `UseWalletBalance` arm just does `if amount == 0 { return AppAction::None; }` with no user feedback — a silent no-op. This is the same dead-end-funding-screen failure class this PR fixes on the create-identity flow, just left unaddressed on top-up. Recommend porting `render_insufficient_wallet_balance_banner` (or an equivalent gate) here — flagging as should-fix-before-merge since it's a user-visible dead end, not just polish. Not fixed yet — leaving open.

lklimek:
Re-verification of the 3 unanchored findings from the consolidated review comment above:

- **`by_using_unused_balance.rs:20` (Wallet Balance label)** — retracting: the premise was wrong. No sibling screen's balance label was relabeled either (`by_platform_address.rs`'s equivalent label is equally untouched) — this is uniform non-treatment across all three funding screens, not an inconsistency singling this one out. No action needed.
- **`by_platform_address.rs:106` (Title-Case ComboBox label)** — confirmed present, but it's pre-existing code this PR didn't touch (the string predates this PR). Real minor inconsistency, but fixing it here would be scope creep — better as a follow-up cleanup.
- **`top_up_identity_screen/by_using_unused_asset_lock.rs` (stale jargon)** — confirmed: file is untouched by this PR and still says "asset lock" throughout, unlike the friendlier wording its create-identity sibling got. Confirmed out of this PR's scope; worth a follow-up if the de-jargoning pass is meant to eventually cover the Top-Up wizard's own sub-screens.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Confirmed fixed in `59e010de`: the balance gate now reads `.spendable()` via the shared `spendable_covers_minimum()` helper (real `CREDITS_PER_DUFF` constant, no bare `* 1000`). Traced all 3 call sites of the new helper — no unit-mismatch or double-conversion bugs. Resolving.

Note for the record: a separate, untouched call site — the "Max" amount button in `top_up_identity_screen/mod.rs`'s wallet-balance input — still computes from `.total` + a bare `* 1000`. Same bug class, not part of this thread's original scope; fixing it now as a follow-up.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Confirmed fixed in `65b1fd8f`: design-spec now shows `[Go to Wallets]` with the "no separate Receive screen" rationale, matching the shipped button text exactly. Resolving.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Re-verified with a call-tree walk, and this one isn't actually closed: the two lines you flagged are no longer bare `.read().unwrap()` (commit `db2b79d7`), but the panic risk survives via sibling reads on the *same* lock, one call away:
- `show_wallet_balance()` — in both this file and its Top-Up counterpart — still does a bare `.read().unwrap()` on `self.selected_wallet`, and runs immediately *before* the now-safe banner code in the render order.
- `top_up_identity_screen/mod.rs:179` does a bare `.read().unwrap()` on the *same* `self.funding_method` lock, a few lines after the line that was fixed.

Since `RwLock` poisoning is sticky, either sibling read still panics the UI thread under the exact condition this thread was raised about. Leaving open — extending the fix to cover these now.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
lklimek: Confirmed fixed: `funding_common.rs` now has boundary tests on `spendable_covers_minimum` (exact match, ±1 credit, zero, overflow) — ran them directly, all pass, and they'd catch a broken comparison or wrong constant. One nuance for the record: these tests exercise the pure arithmetic, not the call sites' choice of `.spendable()` vs `.total` — that field choice is covered by direct code verification instead (see the sibling thread). Resolving.

src/ui/identities/add_new_identity_screen/mod.rs
lklimek: Confirmed sub-claim (a) fixed in `d4c1728e`: pre-selection now requires fee-sufficiency (via `spendable_covers_minimum`), not just any positive balance — verified the code path and its call-tree.

Sub-claim (b), staleness on wallet-switch, is deliberately deferred — a `TODO(bilby)` documents the exact gap: recomputing safely needs to track whether the user manually picked a funding method, so a wallet switch never clobbers an explicit choice, and that's a UX call rather than something safe to guess at in a bug-fix pass.

Leaving this thread open pending your direction: should switching wallets ever silently change an already-selected funding method, or only ever refresh the pre-selection while the user hasn't touched it yet? Happy to implement either once there's a call.

src/ui/identities/add_new_identity_screen/mod.rs
lklimek: Confirmed fixed in `52a1d57f`: doc comment, on-screen caption, heading, and field label all now say "alias" — zero "nickname" occurrences left in this function or its rendered output. Resolving.

src/ui/identities/top_up_identity_screen/mod.rs
lklimek: Confirmed fixed in `d3d94056`: new `FundingMethod::top_up_label()` gives Top-Up its own wording ("Use an existing funding transaction" vs. Create-Identity's "Recover an unfinished funding"), and all 4 label call sites in the Top-Up wizard use it — verified no leftover raw `Display` usage anywhere in that screen. Backed by a unit test (`top_up_label_differs_only_for_asset_lock`) that would catch a regression. Resolving.

src/ui/identities/top_up_identity_screen/mod.rs
lklimek: Confirmed fixed: `top_up_identity_screen/by_using_unused_balance.rs` now has its own insufficient-funds banner using `spendable_covers_minimum`, and the "Top Up Identity" click path is unreachable with an insufficient balance (gated by both the new banner and the pre-existing amount check). Resolving the core ask.

Two adjacent issues surfaced during verification, tracked separately rather than blocking this thread: (1) this file's `show_wallet_balance()` has the same lock-poisoning gap as the sibling RwLock thread — fixing alongside that one now; (2) the "Max" amount button nearby still computes from `.total` + a bare `* 1000` — same bug class as the very first finding on this PR, fixing now too.

thepastaclaw:
✅ Review complete (commit f39d08b0)

thepastaclaw (commented):
## Code Review

Incremental review at b75179de. The latest delta (cold-start wallet-preparation timeout banner + skipped-wallets banner + OptionBannerExt::raise rename + concurrent-open guard test) is well-scoped, well-tested, and internally consistent — no in-scope findings against those changes. Carried forward from bc104b52: the `show_wallet_balance` label still reads `.total` in both funding wizards, contradicting every sibling gate on the same screen. New in the full-PR sweep: the Top-Up asset-lock recovery screen was not brought to parity with the Create-Identity fix that disables the primary button when no funding is selected, and the Top-Up wallet pre-selection recommends `UseWalletBalance` for wallets with any positive balance (including dust) instead of using the same `spendable_covers_minimum` gate the Create-Identity path just adopted.

Source: reviewers: `opus` (general), `gpt-5.5` (general), `opus` (rust-quality), `gpt-5.5` (rust-quality); verifier: `opus`; specialists: `rust-quality`

Prior reconciliation:
- P1 from `bc104b52` is STILL VALID and carried forward: both funding-wizard `show_wallet_balance` labels still read `.total` while sibling gates use `.spendable()`.

Latest delta (`bc104b52..b75179de`):
- No new in-scope findings against the cold-start wallet-preparation timeout banner, skipped-wallets banner, `OptionBannerExt::raise`, or concurrent-open guard changes.

New cumulative findings in the current head:
- Top-Up asset-lock recovery still allows a no-op primary action when no funding is selected.
- Top-Up wallet pre-selection still uses the looser positive-balance predicate instead of the spendable/fee-sufficient gate.


🟡 3 suggestion(s)

<details>

```

In `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22: `show_wallet_balance` still reads `.total` while every sibling gate on the same screen uses `.spendable()`
  Carried forward from the prior review at bc104b52 — the latest delta did not touch these files and the mismatch remains. This PR intentionally converts the wallet-picker ComboBox label (`format_wallet_picker_label` in mod.rs), the insufficient-funds banner (`render_insufficient_wallet_balance_banner` in this same file at line 48), the pre-selection gate (mod.rs `wallet_can_afford_creation` at line 232), and the Max-amount reserve to `.spendable()`, precisely so users never see a 'plenty of balance' number contradicted by a 'not enough Dash' banner when part of the total is immature or CoinJoin-locked and coin selection can't touch it. `show_wallet_balance` is the very label the user reads directly above the funding amount, and it is still calling `.total` (line 22 here; mirror at `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`). With any locked/immature funds present the user now sees three inconsistent numbers on one screen: picker (spendable), 'Wallet Balance:' label (total), and the insufficient-funds banner (spendable). Switch both call sites to `.spendable()` (optionally surfacing a locked portion separately) so all four surfaces agree.

In `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142: Top-Up asset-lock 'Top Up Identity' button not disabled when no funding is selected
  This PR added a `can_create` / `add_enabled(can_create, button)` gate to the sibling `AddNewIdentityScreen::render_ui_by_using_unused_asset_lock` at `src/ui/identities/add_new_identity_screen/by_using_unused_asset_lock.rs:149-160`, so the Create button is visibly disabled until an asset lock is picked. The Top-Up sibling here was not updated: the 'Top Up Identity' button is unconditionally added, and `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None` when `self.funding_asset_lock` is still `None` (`src/ui/identities/top_up_identity_screen/mod.rs:337-366`). Result: clicking the primary action before picking a funding does nothing and surfaces no feedback — the exact broken behavior the PR eliminated on the Create-Identity side. Bring this one to parity by gating on `self.funding_asset_lock.is_some()` and using `add_enabled` with the same disabled-fill styling.

In `src/ui/identities/top_up_identity_screen/mod.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/mod.rs:166-171: Top-Up pre-selection recommends `UseWalletBalance` for dust balances instead of using the same spendable/fee gate the Create path adopted
  The Create-Identity pre-selection at `src/ui/identities/add_new_identity_screen/mod.rs:230-240` was tightened in this PR to only recommend `UseWalletBalance` when the wallet's spendable balance covers the estimated identity-creation fee (`spendable_covers_minimum(spendable_duffs, minimum_credits)`), specifically so a wallet with dust doesn't get a stale recommendation the insufficient-funds banner immediately rejects. The Top-Up pre-selection here still passes `snapshot_has_balance()` (any positive balance, including immature/locked/dust) into `default_funding_state`, so a Top-Up user with 1 duff still gets `UseWalletBalance` recommended and then hits `render_insufficient_wallet_balance_banner` — the exact stale-recommendation UX this PR set out to eliminate for identity onboarding. Mirror the Create path: compute `spendable_covers_minimum(spendable, fee_estimator().estimate_identity_topup())` and pass that to `default_funding_state`.
```
</details>

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
```diff
@@ -1,15 +1,23 @@
...
-            let wallet = selected_wallet.read().unwrap(); // Read lock on the wallet
+            let wallet = match selected_wallet.read() {
+                Ok(w) => w,
+                Err(_) => {
+                    ui.label("Wallet is busy. Try again in a moment.");
+                    return;
+                }
+            };
 
             let total_balance: u64 = self.app_context.snapshot_balance(&wallet.seed_hash()).total;
```
thepastaclaw, line 22: **🟡 Suggestion: `show_wallet_balance` still reads `.total` while every sibling gate on the same screen uses `.spendable()`**

Carried forward from the prior review at bc104b52 — the latest delta did not touch these files and the mismatch remains. This PR intentionally converts the wallet-picker ComboBox label (`format_wallet_picker_label` in mod.rs), the insufficient-funds banner (`render_insufficient_wallet_balance_banner` in this same file at line 48), the pre-selection gate (mod.rs `wallet_can_afford_creation` at line 232), and the Max-amount reserve to `.spendable()`, precisely so users never see a 'plenty of balance' number contradicted by a 'not enough Dash' banner when part of the total is immature or CoinJoin-locked and coin selection can't touch it. `show_wallet_balance` is the very label the user reads directly above the funding amount, and it is still calling `.total` (line 22 here; mirror at `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`). With any locked/immature funds present the user now sees three inconsistent numbers on one screen: picker (spendable), 'Wallet Balance:' label (total), and the insufficient-funds banner (spendable). Switch both call sites to `.spendable()` (optionally surfacing a locked portion separately) so all four surfaces agree.

src/ui/identities/top_up_identity_screen/mod.rs
```diff
@@ -156,27 +156,50 @@ impl TopUpIdentityScreen {
...
+                    // recommend `UseWalletBalance` only when this wallet
+                    // actually has funds to offer it with.
+                    let funding_method_snapshot = self.funding_method.read().ok().map(|m| *m);
+                    if funding_method_snapshot == Some(FundingMethod::NoSelection) {
+                        let wallet_has_balance = {
+                            let wallet_read = wallet.read().unwrap();
+                            self.app_context
+                                .snapshot_has_balance(&wallet_read.seed_hash())
+                        };
+                        let (recommended, _) = default_funding_state(wallet_has_balance);
```
thepastaclaw, line 171: **🟡 Suggestion: Top-Up pre-selection recommends `UseWalletBalance` for dust balances instead of using the same spendable/fee gate the Create path adopted**

The Create-Identity pre-selection at `src/ui/identities/add_new_identity_screen/mod.rs:230-240` was tightened in this PR to only recommend `UseWalletBalance` when the wallet's spendable balance covers the estimated identity-creation fee (`spendable_covers_minimum(spendable_duffs, minimum_credits)`), specifically so a wallet with dust doesn't get a stale recommendation the insufficient-funds banner immediately rejects. The Top-Up pre-selection here still passes `snapshot_has_balance()` (any positive balance, including immature/locked/dust) into `default_funding_state`, so a Top-Up user with 1 duff still gets `UseWalletBalance` recommended and then hits `render_insufficient_wallet_balance_banner` — the exact stale-recommendation UX this PR set out to eliminate for identity onboarding. Mirror the Create path: compute `spendable_covers_minimum(spendable, fee_estimator().estimate_identity_topup())` and pass that to `default_funding_state`.

<sub>source: ['codex']</sub>

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
thepastaclaw, line 22: **🟡 Suggestion: STILL VALID: wallet funding label shows total instead of spendable balance**

Prior finding P1 is STILL VALID. `show_wallet_balance()` still displays `snapshot_balance(...).total`, while the surrounding funding gates use `.spendable()` to decide whether the wallet can actually fund identity creation. The mirror Top-Up screen has the same issue at `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`, so users with immature or locked funds can see a positive Wallet Balance immediately before the flow says the wallet does not have enough Dash. Both labels should display the spendable balance used by the funding checks.

src/ui/identities/top_up_identity_screen/mod.rs
thepastaclaw, line 171: **🟡 Suggestion: STILL VALID: Top-Up pre-selects wallet funding for dust balances**

Prior finding P3 is STILL VALID. The single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, which treats any positive balance as enough to recommend `UseWalletBalance`. The later insufficient-funds banner uses `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`, so a wallet with dust, immature funds, or otherwise insufficient spendable funds can be recommended and then immediately blocked. Use the same spendable and fee-sufficient predicate for the default recommendation.

thepastaclaw (commented):
## Code Review

Incremental review for `866bbf5f`. The latest delta (`b75179de..866bbf5f`) is the platform-wallet/settings/backend adaptation and I did not verify any new in-scope findings in that delta. Cumulatively, the three prior suggestions from `b75179de` are still valid at the current head and are carried forward below.

Source: reviewers: `opus` (general, failed: usage limit), `gpt-5.5` (general), `opus` (rust-quality, failed: usage limit), `gpt-5.5` (rust-quality); verifier: `gpt-5.5` (recovered after `opus` verifier failed: usage limit); specialists: `rust-quality`.

Prior reconciliation:
- P1 from `b75179de` is STILL VALID and carried forward: both funding-wizard `show_wallet_balance` labels still read `.total` while sibling gates use `.spendable()`.
- P2 from `b75179de` is STILL VALID and carried forward: Top-Up asset-lock recovery still leaves the primary button enabled when no funding is selected.
- P3 from `b75179de` is STILL VALID and carried forward: Top-Up wallet pre-selection still uses `snapshot_has_balance()` instead of the spendable/fee-sufficient predicate.

Latest delta (`b75179de..866bbf5f`):
- No new in-scope findings verified in the platform-wallet/settings/backend delta.
- Targeted validation passed: `cargo test v38_drops_column_and_preserves_other_settings --lib` and `cargo test reserved_core_backend_mode_byte_preserves_wire_layout --lib`.

Carried-forward prior findings:
- `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22` and `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`: `show_wallet_balance()` still displays `snapshot_balance(...).total`, while the surrounding funding gates use `.spendable()` to decide whether the wallet can actually fund identity creation/top-up. Users with immature or locked funds can see a positive Wallet Balance immediately before the flow says the wallet does not have enough Dash.
- `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142`: the Top-Up asset-lock flow always enables the `Top Up Identity` button, but `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None` when `self.funding_asset_lock` is `None`, making the primary action silently do nothing until the user selects funding.
- `src/ui/identities/top_up_identity_screen/mod.rs:166-171`: the single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, so dust/immature/otherwise insufficient funds can recommend `UseWalletBalance` and then immediately hit the spendable/fee-sufficient insufficient-funds banner.

🟡 3 suggestion(s)

<details>

```

In `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22: STILL VALID: wallet funding label shows total instead of spendable balance
  Prior finding P1 is STILL VALID. `show_wallet_balance()` still displays `snapshot_balance(...).total`, while the surrounding funding gates use `.spendable()` to decide whether the wallet can actually fund identity creation. The mirror Top-Up screen has the same issue at `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`, so users with immature or locked funds can see a positive Wallet Balance immediately before the flow says the wallet does not have enough Dash. Both labels should display the spendable balance used by the funding checks.

In `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142: STILL VALID: Top-Up asset-lock button remains enabled with no selection
  Prior finding P2 is STILL VALID. The Top-Up asset-lock flow always enables the `Top Up Identity` button, but `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None` when `self.funding_asset_lock` is `None` at `src/ui/identities/top_up_identity_screen/mod.rs:337-366`. This makes the primary button silently do nothing until the user selects funding. Gate the button with `self.funding_asset_lock.is_some()` so the UI communicates the required selection state.

In `src/ui/identities/top_up_identity_screen/mod.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/mod.rs:166-171: STILL VALID: Top-Up pre-selects wallet funding for dust balances
  Prior finding P3 is STILL VALID. The single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, which treats any positive balance as enough to recommend `UseWalletBalance`. The later insufficient-funds banner uses `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`, so a wallet with dust, immature funds, or otherwise insufficient spendable funds can be recommended and then immediately blocked. Use the same spendable and fee-sufficient predicate for the default recommendation.
```
</details>

thepastaclaw (commented):
## Code Review

Incremental review for `ec2d2e5d`. The latest delta (`866bbf5f..ec2d2e5d`) adds the wallet balance categorization work, persistent-banner API/call sites, backend-e2e retest docs/tests, and changelog/formatting updates. Cumulatively, all three prior suggestions from `866bbf5f` are still valid at the current head and are carried forward; the latest delta adds one new in-scope suggestion for account-summary/address-table consistency.

Source: reviewers: `opus` (claude-general failed: usage limit), `gpt-5.5` (codex-general), `opus` (claude-rust-quality failed: usage limit), `gpt-5.5` (codex-rust-quality); verifier: `gpt-5.5` (recovered after `opus` verifier failed: usage limit); specialists: `rust-quality`.

Prior reconciliation:
- P1 from `866bbf5f` is STILL VALID and carried forward: both funding-wizard `show_wallet_balance` labels still read `.total` while sibling gates use `.spendable()`.
- P2 from `866bbf5f` is STILL VALID and carried forward: Top-Up asset-lock recovery still leaves the primary button enabled when no funding is selected.
- P3 from `866bbf5f` is STILL VALID and carried forward: Top-Up wallet pre-selection still uses `snapshot_has_balance()` instead of the spendable/fee-sufficient predicate.
- Correction: the poster initially emitted two auto-resolution replies for P1/P3; those replies were corrected because both findings remain valid at `ec2d2e5d`.

New findings in latest delta:
- `src/ui/wallets/account_summary.rs:322-340`: snapshot-only funded addresses are now counted in account tab totals, but the selected account address table still renders only `wallet.known_addresses`, so the expanded row list can omit the address that explains the new tab total.

Carried-forward prior findings:
- `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24` and `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`: `show_wallet_balance()` still displays `snapshot_balance(...).total`, while the surrounding funding gates use `.spendable()` to decide whether identity creation/top-up can proceed.
- `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142`: the Top-Up asset-lock flow always enables the `Top Up Identity` button, but `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None` when `self.funding_asset_lock` is `None`.
- `src/ui/identities/top_up_identity_screen/mod.rs:166-171`: the single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, so dust/immature/otherwise insufficient funds can recommend `UseWalletBalance` and then immediately hit the spendable/fee-sufficient insufficient-funds banner.

🟡 4 suggestion(s)

<details>

```text

In `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24: Carried forward: wallet funding label shows total instead of spendable balance
  Prior finding P1 is STILL VALID. `show_wallet_balance()` displays `snapshot_balance(...).total`, while the surrounding funding gate uses `.spendable()` to decide whether identity creation can proceed. The Top-Up mirror has the same issue at `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21`, so wallets with immature or locked funds can show a positive Wallet Balance immediately before the flow says the wallet does not have enough Dash. Display the same spendable balance that the funding checks use.

In `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142: Carried forward: Top-Up asset-lock button remains enabled with no selection
  Prior finding P2 is STILL VALID. The Top-Up asset-lock screen always enables the `Top Up Identity` button, but `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None` when `self.funding_asset_lock` is `None`. That lets the primary action silently do nothing until the user selects unfinished funding. Gate the button on `self.funding_asset_lock.is_some()` so the required selection state is visible in the UI.

In `src/ui/identities/top_up_identity_screen/mod.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/mod.rs:166-171: Carried forward: Top-Up pre-selects wallet funding for dust balances
  Prior finding P3 is STILL VALID. The single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, which only means the wallet has a positive total balance. The later insufficient-funds banner uses `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`, so a wallet with dust, immature funds, or otherwise insufficient spendable funds can be recommended and then immediately blocked. Use the same spendable and fee-sufficient predicate for the default recommendation.

In `src/ui/wallets/account_summary.rs`:
- [SUGGESTION] src/ui/wallets/account_summary.rs:322-340: New: snapshot-only address totals have no matching table rows
  The latest delta adds a second summary pass for funded addresses present in the wallet-backend snapshot but missing from `wallet.watched_addresses`, so those balances now appear in the account tab totals. The selected account's address table still builds rows only from `wallet.known_addresses.iter()` in `src/ui/wallets/wallets_screen/address_table.rs:135-137` and then filters that list by account. Any address counted only by this new snapshot pass can make a tab show funds while the expanded address list for that tab omits the row that accounts for them. Extend the table's data source with the same snapshot address balances/paths used by the summary, or register those snapshot addresses into the wallet metadata before rendering.
```
</details>

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
thepastaclaw: Correction: this prior finding is still valid at `ec2d2e5d`. `show_wallet_balance()` still displays `snapshot_balance(...).total`, while the surrounding funding gates use `.spendable()`. The previous auto-resolution reply was incorrect; keep this thread open until the label uses the same spendable balance as the funding checks.

<sub>Auto-reconciliation correction for review `ec2d2e5d`.</sub>

src/ui/identities/top_up_identity_screen/mod.rs
thepastaclaw: Correction: this prior finding is still valid at `ec2d2e5d`. The single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, while the insufficient-funds path uses the spendable/fee-sufficient predicate. The previous auto-resolution reply was incorrect; keep this thread open until the default recommendation uses the same spendable check.

<sub>Auto-reconciliation correction for review `ec2d2e5d`.</sub>

thepastaclaw (commented):
## Code Review

Incremental review for `5671b9c4`. The latest delta (`ec2d2e5d..5671b9c4`) adds the AddressInput tag-search UI/test work plus changelog/docs updates. Cumulatively, all four prior suggestions from `ec2d2e5d` are still valid at the current head and are carried forward; the latest delta adds one new in-scope suggestion for tag-query validation.

Source: reviewers: `opus` (claude-general, failed: usage limit), `gpt-5.5` (codex-general), `opus` (claude-rust-quality, failed: usage limit), `gpt-5.5` (codex-rust-quality); verifier: `gpt-5.5` (recovered after `opus` verifier failed: usage limit); specialists: `rust-quality`.

Prior reconciliation:
- P1 from `ec2d2e5d` is STILL VALID and carried forward: the identity-create and Top-Up wallet-balance labels still display total balance while their funding gates use spendable balance.
- P2 from `ec2d2e5d` is STILL VALID and carried forward: the Top-Up asset-lock button is still enabled when no asset lock is selected and dispatches to a no-op.
- P3 from `ec2d2e5d` is STILL VALID and carried forward: the single-wallet Top-Up default still uses a total-positive balance predicate instead of the spendable/fee-sufficient predicate.
- P4 from `ec2d2e5d` is STILL VALID and carried forward: account summaries still count snapshot-only funded addresses that the selected account address table cannot render.

Carried-forward prior findings:
- `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24`: Wallet funding label shows total instead of spendable balance.
- `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142`: Top-Up asset-lock button remains enabled with no selection.
- `src/ui/identities/top_up_identity_screen/mod.rs:166-171`: Top-Up pre-selects wallet funding for insufficient balances.
- `src/ui/wallets/account_summary.rs:322-340`: Snapshot-only address totals have no matching table rows.

New findings in latest delta:
- `src/ui/components/address_input.rs:1207-1208`: Tag searches are still validated as destination addresses.

🟡 5 suggestion(s)

<details>

```text

In `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24: Wallet funding label shows total instead of spendable balance
  This funding step displays `snapshot_balance(...).total`, while the insufficient-funds gate immediately below it checks `snapshot_balance(...).spendable()` with `spendable_covers_minimum()`. `DetWalletBalance::total` intentionally includes immature or locked funds that coin selection cannot use, so the screen can show a positive Wallet Balance and then block the same wallet as unable to fund identity creation. The Top-Up wallet-balance screen has the same mismatch, so funding-adjacent labels should display the spendable amount used by the gate.

In `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142: Top-Up asset-lock button remains enabled with no selection
  The required funding choice is stored as `Option<OutPoint>`, but the primary `Top Up Identity` button is always enabled. When `funding_asset_lock` is `None`, `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)` returns `AppAction::None`, so clicking the main action silently does nothing instead of showing that a funding item must be selected. Gate the button on the option state so the UI enforces the same invariant required by task construction.

In `src/ui/identities/top_up_identity_screen/mod.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/mod.rs:166-171: Top-Up pre-selects wallet funding for insufficient balances
  The single-wallet Top-Up default still passes `snapshot_has_balance()` into `default_funding_state`, which only proves that confirmed or unconfirmed total balance is positive. Later in the same flow, wallet funding is correctly blocked with `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`. A wallet with dust, immature, locked, or otherwise insufficient spendable funds can therefore be recommended for wallet funding and then immediately rejected. Use the same spendable and fee-sufficient predicate for the default recommendation.

In `src/ui/wallets/account_summary.rs`:
- [SUGGESTION] src/ui/wallets/account_summary.rs:322-340: Snapshot-only address totals have no matching table rows
  The summary pass counts funded addresses present in the wallet-backend snapshot but missing from `wallet.watched_addresses`, so those balances can appear in account tab totals. The selected account's address table still builds rows only from `wallet.known_addresses.iter()` before filtering by account. Any address counted only by this snapshot pass can make a tab show funds while the expanded address list omits the row that accounts for them. Extend the table data source with the same snapshot address/path data or register those addresses into wallet metadata before rendering.

In `src/ui/components/address_input.rs`:
- [SUGGESTION] src/ui/components/address_input.rs:1207-1208: Tag searches are still validated as destination addresses
  The latest delta makes `type:` and `wallet:` valid search tags for filtering autocomplete rows, but the validation state still treats the same `input_text` as a final address after `has_blurred` is set. Typing a query such as `type:core wallet:alpha` is longer than three characters, so the text-change path sets `has_blurred = true`; this branch then calls `validate_input()`, `detect_address_type()` returns `Unknown`, and the component displays the invalid-address warning while matching rows are being shown. Suppress final address validation while the focused input contains recognized search tags, or validate only after the user selects a row or leaves the field without a matching selection.
```
</details>

src/ui/identities/top_up_identity_screen/mod.rs
```diff
@@ -156,27 +156,50 @@ impl TopUpIdentityScreen {
...
+                        *self.funding_method.write().unwrap() = recommended;
```
thepastaclaw, line 172: **🟡 Suggestion: Top-Up preselects wallet funding without enough spendable balance**

The single-wallet Top-Up default feeds `snapshot_has_balance()` into `default_funding_state`, but that predicate only proves the wallet has some confirmed or unconfirmed total. The later Top-Up wallet flow blocks on `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`, so a wallet with dust, locked funds, immature funds, or too little spendable balance can be recommended and then immediately rejected. Use the same spendable and fee-aware predicate before recommending wallet funding.

```suggestion
let wallet_can_afford_top_up = match wallet.read() {
                            Ok(wallet_read) => {
                                let spendable_duffs = self
                                    .app_context
                                    .snapshot_balance(&wallet_read.seed_hash())
                                    .spendable();
                                let minimum_credits = self.app_context.fee_estimator().estimate_identity_topup();
                                crate::ui::identities::funding_common::spendable_covers_minimum(
                                    spendable_duffs,
                                    minimum_credits,
                                )
                            }
                            Err(_) => false,
                        };
                        let (recommended, _) = default_funding_state(wallet_can_afford_top_up);
                        *self.funding_method.write().unwrap() = recommended;
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers: claude-general (opus, failed rc=1), codex-general (gpt-5.5), claude-rust-quality (opus, failed rc=1), codex-rust-quality (gpt-5.5); verifier: gpt-5.5; specialists: rust-quality

At f39d08b071cd72572c21ce127a350b7b8a57b084, all five prior findings are still valid and remain in scope for this PR's cumulative UI and wallet-backend migration changes. The latest delta only removes a blank line from CHANGELOG.md, so there are no new latest-delta code findings; one additional cumulative Top-Up balance-label mismatch is also valid.

Prior reconciliation:
- prior-1: STILL VALID — Create Identity wallet funding label still displays total balance while the funding gate uses spendable balance.
- prior-2: STILL VALID — Top-Up asset-lock action remains enabled with no selected unfinished funding and still no-ops.
- prior-3: STILL VALID — Top-Up still recommends wallet funding from a positive-balance predicate instead of the spendable/fee-sufficient predicate.
- prior-4: STILL VALID — account summaries still count snapshot-only funded addresses that the expanded address table cannot render.
- prior-5: STILL VALID — AddressInput tag queries are still validated as final destination addresses.

Carried-forward prior findings:
- `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24`: wallet funding label shows total instead of spendable balance.
- `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142`: Top-Up asset-lock button remains enabled with no selection.
- `src/ui/identities/top_up_identity_screen/mod.rs:166-172`: Top-Up pre-selects wallet funding for insufficient balances.
- `src/ui/wallets/account_summary.rs:280-293`: snapshot-only address totals have no matching table rows.
- `src/ui/components/address_input.rs:1207-1208`: tag searches are validated as destination addresses.

New findings in latest delta (`67b81031..f39d08b0`):
- None. The latest code delta only removes one line from `CHANGELOG.md`.

Additional cumulative finding:
- `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21-23`: the Top-Up wallet-balance label has the same total-vs-spendable mismatch as the carried-forward Create Identity finding.


🟡 6 suggestion(s)

_5 additional finding(s) omitted (not in diff)._

<details>

```

In `src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs:22-24: Create Identity balance label shows unusable funds
  This panel displays `snapshot_balance(...).total`, but the immediately following insufficient-funds gate uses `snapshot_balance(...).spendable()` with the fee-aware minimum check. `total` can include immature or locked funds that coin selection cannot spend, so the screen can show a positive wallet balance and then block the same wallet as unable to create the identity. Display the spendable amount used by the gate.

In `src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_balance.rs:21-23: Top-Up balance label shows unusable funds
  The Top-Up wallet-balance panel has the same mismatch: it displays `snapshot_balance(...).total`, while the funding gate below checks `snapshot_balance(...).spendable()` against the estimated top-up fee. A wallet with locked or immature funds can therefore look funded in the label and still fail the next check. Use the spendable balance consistently in this funding UI.

In `src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/by_using_unused_asset_lock.rs:136-142: Top-Up asset-lock action stays enabled with no selection
  The required unfinished-funding choice is stored as `Option<OutPoint>`, but the primary `Top Up Identity` button is always enabled. When `funding_asset_lock` is `None`, clicking it calls `top_up_identity_clicked(FundingMethod::UseUnusedAssetLock)`, which returns `AppAction::None`, leaving a dead primary action with no feedback. Disable the action until an unfinished funding item is selected.

In `src/ui/identities/top_up_identity_screen/mod.rs`:
- [SUGGESTION] src/ui/identities/top_up_identity_screen/mod.rs:166-172: Top-Up preselects wallet funding without enough spendable balance
  The single-wallet Top-Up default feeds `snapshot_has_balance()` into `default_funding_state`, but that predicate only proves the wallet has some confirmed or unconfirmed total. The later Top-Up wallet flow blocks on `spendable_covers_minimum(spendable_duffs, estimate_identity_topup())`, so a wallet with dust, locked funds, immature funds, or too little spendable balance can be recommended and then immediately rejected. Use the same spendable and fee-aware predicate before recommending wallet funding.

In `src/ui/wallets/account_summary.rs`:
- [SUGGESTION] src/ui/wallets/account_summary.rs:280-293: Snapshot-only address totals have no matching table rows
  `collect_account_summaries()` now counts funded addresses present in the wallet-backend snapshot but missing from `address_paths` by bucketing them under `Other(Unknown)`. The expanded address table still builds rows only from `wallet.known_addresses`, then filters that row set by account. A snapshot-only funded address can therefore make the Other/System tab show a non-zero balance while the table has no row explaining which address holds those funds.

In `src/ui/components/address_input.rs`:
- [SUGGESTION] src/ui/components/address_input.rs:1207-1208: Tag searches are validated as destination addresses
  The component accepts `type:` and `wallet:` as autocomplete search tags, but after blur it validates the same non-empty query text as a final destination address unless the user selected an autocomplete entry. A query like `type:core wallet:alpha` can filter suggestions correctly and still show the invalid-address warning because `detect_address_type()` treats the tag syntax as unknown. Keep tag-containing input on the search path unless the user selects a suggestion or enters an actual address.
```
</details>

PR #871, opened 2026-07-02
feat(wallet_backend): surface skipped persisted wallets via MessageBanner

Claudius-Maginificent:
## Why this PR exists
- **Problem**: When persisted platform wallets fail integrity checks at load, they are silently skipped — the user sees fewer wallets with no explanation.
- **What breaks without it**: Imagine you are a DET user whose store holds 3 wallets, one with a corrupt row. On next launch only 2 wallets appear and nothing tells you why — you assume funds/config are lost and start debugging blind. With this PR, a warning banner reports &#34;1 wallet was skipped during loading…&#34; immediately.
- **Blocking relationship**: stacked atop #860 (`docs/platform-wallet-migration-design`); completes `TODO(PROJ-010-T6)`.

## What was done
- New `skipped_wallets_banner_text(usize) -&gt; Option` in `wallet_backend/mod.rs` (singular/plural wording), wired into `register_persisted_wallets` — raises `MessageBanner::set_global(..., MessageType::Warning)` when any wallet is skipped.
- Replaced the stale `TODO(PROJ-010-T6)` comment in `event_bridge.rs` with a 2-line present-state note.
- `tests/kittest/skipped_wallets_banner.rs`: egui_kittest harness tests (singular Warning banner with ⚠ icon and ❌ dismiss, plural wording, zero skips → no banner), mirroring the `migration_banner.rs` pattern.
- Follow-up (review feedback): extracted the banner logic into `raise_skipped_wallets_banner`, a directly unit-testable free function, and retained its `BannerHandle` behind a `Mutex` on `Inner` so a re-entrant load pass (`ensure_wallets_registered`, after the migration engine) **supersedes** the previous banner instead of stacking a second, possibly contradictory one — and a later zero-skip pass clears it. Replaced the redundant `no_skips_renders_no_banner` kittest case (previously a verbatim duplicate of an inline unit test, no harness) with a harness-based render assertion. Corrected the `register_persisted_wallets` doc comment to note the banner.
- Follow-up (API cleanup): `OptionBannerExt::replace` collides with the inherent `Option::replace(value)`, which always wins method-call resolution — `opt.replace(ctx, msg, msg_type)` fails to compile against it (E0061), so the trait method was only reachable fully-qualified (and the trait's own doc example silently demonstrated code that doesn't compile, hidden by a `` ```ignore `` block). Added `OptionBannerExt::raise` as the collision-free preferred method (matching the sibling `OptionOverlayExt::raise`, which already avoided this name for the same reason), kept `replace` as a backward-compatible alias, and fixed the doc example.

## Testing
- `cargo test --all-features --lib wallet_backend` → 293/293 pass, incl. a test driving `raise_skipped_wallets_banner` through the rendered `MessageBanner` surface across three passes (singular → plural supersede → zero-skip clear); mutation-checked by temporarily reverting to the old fire-and-forget call and confirming the test fails.
- `cargo test --test kittest --all-features skipped_wallets_banner` → 3/3 pass.
- `cargo test --all-features --lib message_banner` and `--lib progress_overlay` → pass (unaffected by the rename).
- `cargo +nightly fmt --all -- --check` and `cargo clippy --all-features --all-targets -- -D warnings` (whole workspace, covering every screen using `OptionBannerExt`) both clean.
- The previously-noted limitation (banner wire-up not directly covered by a test) is now resolved.

## Breaking changes
None — `replace` is preserved as a deprecated alias, not removed.

## Checklist
- [x] Tests added and passing
- [x] fmt/clippy clean
- [x] Stacked on #860's branch; no changes to that branch itself
- [x] All review feedback addressed

## Attribution

### DIFF
--- a/src/wallet_backend/mod.rs
+++ b/src/wallet_backend/mod.rs
@@ -116,6 +116,8 @@ use crate::context::connection_status::ConnectionStatus;
 use crate::model::selected_identity::SelectedIdentity;
 use crate::model::selected_wallet::SelectedWallet;
 use crate::model::wallet::{PlatformAddressEntry, WalletSeedHash};
+use crate::ui::MessageType;
+use crate::ui::components::message_banner::{BannerHandle, OptionBannerExt};
 use crate::utils::egui_mpsc::SenderAsync;
 
 /// The upstream persister DET consumes. Authored upstream (PR #3625) — DET
@@ -217,6 +219,13 @@ struct Inner {
     /// dispatch is user-initiated and rare relative to lock acquisition
     /// cost.
     dashpay_address_index_lock: std::sync::Mutex<()>,
+    /// Handle to the warning banner raised when persisted wallets are skipped
+    /// on load. A re-entrant load pass (via [`WalletBackend::ensure_wallets_registered`])
+    /// supersedes this banner rather than stacking a second, possibly
+    /// contradictory one. Interior mutability because the backend is
+    /// `Arc`-shared and `register_persisted_wallets` runs on `&self`. See
+    /// [`raise_skipped_wallets_banner`].
+    skipped_wallets_banner: std::sync::Mutex<Option<BannerHandle>>,
     /// Encrypted secret vault. Holds imported single-key WIFs
     /// (`single_key_priv.*` labels, see [`single_key`]) and HD-wallet
     /// BIP-39 seeds (`envelope.v1` labels under `WalletId(seed_hash)`, see
@@ -365,6 +374,7 @@ impl WalletBackend {
                 network,
                 spv_storage_dir,
                 dashpay_address_index_lock: std::sync::Mutex::new(()),
+                skipped_wallets_banner: std::sync::Mutex::new(None),
                 secret_store,
                 single_key_index: std::sync::RwLock::new(std::collections::BTreeMap::new()),
                 app_kv,
@@ -445,7 +455,8 @@ impl WalletBackend {
 
     /// Run the configured loader to bring back persisted wallets watch-only.
     /// Identity-funding re-provision is deferred to the asset-lock chokepoint
-    /// (which obtains the seed just-in-time), so this pass only loads and logs.
+    /// (which obtains the seed just-in-time), so this pass loads, logs, and
+    /// raises a warning banner for any skipped wallet.
     async fn register_persisted_wallets(&self, ctx: &Arc<AppContext>) -> Result<(), TaskError> {
         let loader = Arc::clone(&self.inner.loader);
         let outcome = loader.load(self, ctx).await?;
@@ -461,12 +472,13 @@ impl WalletBackend {
                 "Skipped a corrupt persisted wallet row on load"
             );
         }
-        if let Some(text) = skipped_wallets_banner_text(outcome.skipped.len()) {
-            crate::ui::components::message_banner::MessageBanner::set_global(
-                ctx.egui_ctx(),
-                text,
-                crate::ui::MessageType::Warning,
-            );
+        {
+            let mut handle = self
+                .inner
+                .skipped_wallets_banner
+                .lock()
+                .unwrap_or_else(|e| e.into_inner());
+            raise_skipped_wallets_banner(ctx.egui_ctx(), outcome.skipped.len(), &mut handle);
         }
 
         // `load()` rebuilds `IdentityRegistration` from the manifest, but
@@ -3409,6 +3421,25 @@ pub fn skipped_wallets_banner_text(skipped: usize) -> Option<String> {
     }
 }
 
+/// Raise the skipped-persisted-wallets warning banner for this load pass,
+/// superseding any banner a previous pass left behind so a re-entrant call
+/// (e.g. via [`WalletBackend::ensure_wallets_registered`]) with a different
+/// skip count never stacks a second, contradictory banner alongside the
+/// first. A pass with zero skips clears a stale banner from an earlier pass
+/// rather than leaving it stuck.
+fn raise_skipped_wallets_banner(
+    ctx: &egui::Context,
+    skipped_count: usize,
+    banner_handle: &mut Option<BannerHandle>,
+) {
+    // Fully qualified: the trait's `replace` is shadowed by the inherent
+    // `Option::replace` under method syntax.
+    match skipped_wallets_banner_text(skipped_count) {
+        Some(text) => OptionBannerExt::replace(banner_handle, ctx, text, MessageType::Warning),
+        None => banner_handle.take_and_clear(),
+    }
+}
+
 /// Bucket for `PlatformWalletError`s coming out of identity register / top-up.
 enum IdentityOpErrorKind {
     /// Network or broadcast rejected the submission (SDK error or asset-lock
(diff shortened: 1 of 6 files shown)


thepastaclaw:
✅ Review complete (commit ba8ad715)

lklimek (approved):

src/wallet_backend/mod.rs
```diff
@@ -461,6 +461,13 @@ impl WalletBackend {
                 "Skipped a corrupt persisted wallet row on load"
             );
         }
+        if let Some(text) = skipped_wallets_banner_text(outcome.skipped.len()) {
+            crate::ui::components::message_banner::MessageBanner::set_global(
+                ctx.egui_ctx(),
+                text,
+                crate::ui::MessageType::Warning,
+            );
+        }
```
thepastaclaw: **🟡 Suggestion: Banner wire-up in register_persisted_wallets is not covered by any test**

The three kittest cases and the inline unit test only exercise `skipped_wallets_banner_text` and its rendering through `MessageBanner::set_global`. None assert that `register_persisted_wallets` actually invokes `set_global` with `outcome.skipped.len()` and `MessageType::Warning` — the two halves are recomposed manually in the tests. A regression that deletes lines 464–470, swaps the MessageType to `Info`, or reads the wrong length would leave every test green even though the invariant this PR was added to establish is gone. The PR body flags this as a known limitation shared with `migration_banner.rs`, but the wire-up is the actual behavior change and deserves coverage. A small backend test driving a stub loader that returns `LoadOutcome { skipped: vec![...] }`, calling `register_persisted_wallets`, and asserting a Warning banner lands on `ctx.egui_ctx()` would close the seam and also protect the `ensure_wallets_registered` re-run path.

<sub>source: ['claude']</sub>

tests/kittest/skipped_wallets_banner.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// Zero skips yields no banner text, so no banner is rendered.
+#[test]
+fn no_skips_renders_no_banner() {
+    assert!(
+        skipped_wallets_banner_text(0).is_none(),
+        "zero skips must not produce banner text",
+    );
+}
```
thepastaclaw, line 86: **🟡 Suggestion: `no_skips_renders_no_banner` never builds a Harness — it duplicates the unit test**

Unlike the two preceding cases, `no_skips_renders_no_banner` never constructs an `egui_kittest::Harness`, never calls `MessageBanner::set_global`/`show_global`, and never queries rendered output. Its body is `assert!(skipped_wallets_banner_text(0).is_none())`, which is already covered verbatim by `skipped_wallets_banner_text_singular_plural_and_none` in `src/wallet_backend/mod.rs:3855-3866`. The test name and module doc comment promise an integration-level check that the Harness renders no banner when the skip count is zero, but nothing is rendered — so a regression where `register_persisted_wallets` incorrectly calls `set_global` for a zero-skip outcome (or where `show_global` mishandles the empty case) would not be caught here. Either build a Harness that runs the same conditional-set/show pipeline as the other tests and asserts the ⚠ label is absent, or drop this case as redundant.

```suggestion
#[test]
fn no_skips_renders_no_banner() {
    let mut harness = Harness::builder()
        .with_size(egui::vec2(600.0, 200.0))
        .build_ui(|ui| {
            if let Some(text) = skipped_wallets_banner_text(0) {
                MessageBanner::set_global(ui.ctx(), text, MessageType::Warning);
            }
            MessageBanner::show_global(ui);
        });
    harness.run();

    assert!(
        harness.query_by_label("\u{26A0}").is_none(),
        "zero skips must not render a warning banner",
    );
}
```

<sub>source: ['claude']</sub>

src/wallet_backend/mod.rs
thepastaclaw: **💬 Nitpick: Re-runs with a different skip count stack two contradictory banners**

`register_persisted_wallets` is also reachable via `ensure_wallets_registered` (mod.rs:1951), which is invoked by the one-time migration engine after `register_persisted_wallets` has already run at construction. `MessageBanner::set_global` dedupes by exact text, so identical skip counts are idempotent, but if the two passes observe different counts (e.g. one skipped row on the first pass, none — or a different number — on the migrated pass) the second call adds a new banner without dismissing the first, so the user sees both '1 saved wallet couldn't be opened…' and '2 saved wallets couldn't be opened…' side by side until each 9 s auto-dismiss fires. In practice `load_from_persistor` is meant to be idempotent over persistor state so this is an edge case, but the banner surface has no memory of the previous count. Retaining a `BannerHandle` in the backend and `.take_and_clear()`-ing it before setting the new one (matching the `refresh_banner` pattern in CLAUDE.md) would close the seam — happy to defer to a follow-up if judged unreachable in practice.

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

Small, well-scoped PR that adds a Warning-typed banner when persisted wallets are skipped on load, backed by a pure `skipped_wallets_banner_text` helper (unit-tested) and kittest rendering coverage. Implementation follows the project's user-facing-message rules and the `set_global` chokepoint. Non-blocking findings: the actual wire-up in `register_persisted_wallets` is not itself covered by any test, one of the kittest cases doesn't exercise rendering, the doc comment above `register_persisted_wallets` is now stale, and a re-entry into the load pass with a different skip count would stack two contradictory banners because `set_global` dedupes only by exact text.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed); verifier: `opus`; specialists: `rust-quality`

<details>
<summary>Findings not posted inline (1)</summary>

These findings could not be anchored to the current diff, but they are still part of this review.

- **[NITPICK] `src/wallet_backend/mod.rs:446-448`: Doc comment on register_persisted_wallets no longer matches its behavior** — The doc comment states `so this pass only loads and logs`, but this PR adds a `MessageBanner::set_global` call inside the function, so it now also raises a user-facing warning banner as a side effect. Update the comment so future readers of this async boot-path function see the accurate behavior.

</details>

<details>

```

- [SUGGESTION] src/wallet_backend/mod.rs:464-470: Banner wire-up in register_persisted_wallets is not covered by any test
  The three kittest cases and the inline unit test only exercise `skipped_wallets_banner_text` and its rendering through `MessageBanner::set_global`. None assert that `register_persisted_wallets` actually invokes `set_global` with `outcome.skipped.len()` and `MessageType::Warning` — the two halves are recomposed manually in the tests. A regression that deletes lines 464–470, swaps the MessageType to `Info`, or reads the wrong length would leave every test green even though the invariant this PR was added to establish is gone. The PR body flags this as a known limitation shared with `migration_banner.rs`, but the wire-up is the actual behavior change and deserves coverage. A small backend test driving a stub loader that returns `LoadOutcome { skipped: vec![...] }`, calling `register_persisted_wallets`, and asserting a Warning banner lands on `ctx.egui_ctx()` would close the seam and also protect the `ensure_wallets_registered` re-run path.

In `tests/kittest/skipped_wallets_banner.rs`:
- [SUGGESTION] tests/kittest/skipped_wallets_banner.rs:70-76: `no_skips_renders_no_banner` never builds a Harness — it duplicates the unit test
  Unlike the two preceding cases, `no_skips_renders_no_banner` never constructs an `egui_kittest::Harness`, never calls `MessageBanner::set_global`/`show_global`, and never queries rendered output. Its body is `assert!(skipped_wallets_banner_text(0).is_none())`, which is already covered verbatim by `skipped_wallets_banner_text_singular_plural_and_none` in `src/wallet_backend/mod.rs:3855-3866`. The test name and module doc comment promise an integration-level check that the Harness renders no banner when the skip count is zero, but nothing is rendered — so a regression where `register_persisted_wallets` incorrectly calls `set_global` for a zero-skip outcome (or where `show_global` mishandles the empty case) would not be caught here. Either build a Harness that runs the same conditional-set/show pipeline as the other tests and asserts the ⚠ label is absent, or drop this case as redundant.
```
</details>

src/wallet_backend/mod.rs
lklimek: Confirmed still open at HEAD (e1c357e8): `register_persisted_wallets`'s `set_global` call (mod.rs:464-470) has no direct test coverage — the kittest cases and the inline unit test only exercise `skipped_wallets_banner_text` + rendering, never `register_persisted_wallets` itself. Agreed this is worth closing; tracking as a follow-up (a backend test with a stub loader returning `LoadOutcome { skipped: vec![...] }` asserting a Warning banner lands on `ctx.egui_ctx()`), not blocking this PR.

tests/kittest/skipped_wallets_banner.rs
lklimek: Confirmed still open at HEAD (e1c357e8): `no_skips_renders_no_banner` (tests/kittest/skipped_wallets_banner.rs:71-76) still just asserts `skipped_wallets_banner_text(0).is_none()` with no `Harness`, duplicating the inline unit test (`skipped_wallets_banner_text_singular_plural_and_none`, mod.rs:3855-3866). Agreed with the suggested diff — tracking as a follow-up to either build the Harness and assert the ⚠ label is absent, or drop the redundant case.

src/wallet_backend/mod.rs
lklimek: Confirmed still open at HEAD (e1c357e8): `register_persisted_wallets` doesn't retain a `BannerHandle`, so a re-run via `ensure_wallets_registered` with a different skip count would stack banners instead of replacing the prior one. Agreed this is a low-priority edge case (relies on `load_from_persistor` observing a different skip count across passes) — tracking as a follow-up to adopt the `refresh_banner`/`take_and_clear()` pattern if/when it's addressed, not blocking this PR.

lklimek: Fixed in d9567049: extracted the banner logic into `raise_skipped_wallets_banner`, a directly unit-testable free function, and added a test that drives it through the rendered `MessageBanner` surface — mutation-checked (reverted to the old fire-and-forget call, confirmed the test fails, restored).

tests/kittest/skipped_wallets_banner.rs
lklimek: Fixed in d9567049 — replaced with your suggested diff verbatim: builds a `Harness`, renders through the real `set_global`/`show_global` pipeline, and asserts the ⚠ label is absent.

src/wallet_backend/mod.rs
lklimek: Fixed in d9567049: `register_persisted_wallets` now retains the `BannerHandle` behind a `Mutex` on `Inner` and supersedes the previous banner on each pass (via `OptionBannerExt::replace`) instead of stacking; a zero-skip pass clears it.

thepastaclaw (approved):
## Code Review

Incremental delta at `d9567049` resolves all four prior findings from `e1c357e8`. The new `raise_skipped_wallets_banner` helper is now exercised across singular render, plural supersede, and zero-skip clear; `Inner` retains a `Mutex<Option<BannerHandle>>` so re-entrant load passes replace/clear the previous banner; the zero-skip kittest now uses a real `Harness`; and the `register_persisted_wallets` doc comment describes the banner side effect. Carry-forward prior findings: none. New latest-delta findings: none.

Source: reviewers: `opus` (general), `gpt-5.5` (general; parseable output, process rc=5), `opus` (rust-quality), `gpt-5.5` (rust-quality); verifier: `opus`; specialists: `rust-quality`.

Prior reconciliation:
- prior-1 FIXED: banner behavior was extracted into `raise_skipped_wallets_banner` and covered by the three-pass rendered-banner test; the remaining call site is a mechanical `outcome.skipped.len()` pass-through.
- prior-2 FIXED: `no_skips_renders_no_banner` now builds/runs an `egui_kittest::Harness` and asserts the warning icon is absent.
- prior-3 FIXED: the `register_persisted_wallets` doc comment now says the pass loads, logs, and raises a skipped-wallet warning banner.
- prior-4 FIXED: `Inner.skipped_wallets_banner` plus `OptionBannerExt::replace` / `take_and_clear` prevents contradictory stacked banners across re-runs.

PR #872, opened 2026-07-04
fix: resolve main control group update authorization

thepastaclaw:
# Fix Main Control Group Authorization

## Summary

- Fix the token config update screen to authorize `MainControlGroup` changes from
  `main_control_group_can_be_modified`.
- Keep the existing SDK helper for every other token config item.
- Add focused regression tests for the MainControlGroup authorization path.

## Motivation

The current DET SDK dependency reports `NoOne` from
`authorized_action_takers_for_configuration_item(MainControlGroup(_))`, even
when the token config allows the change through
`main_control_group_can_be_modified`.

Platform validation uses `main_control_group_can_be_modified`, so DET was
blocking a valid group-managed main control group update before it reached the
backend transition builder.

## Validation

- `cargo test --lib main_control_group_authorization -- --nocapture`
- `cargo test --lib other_change_items_delegate_to_sdk_helper -- --nocapture`
- `cargo test --lib test_token_creator_ui_builds_correct_contract -- --nocapture`
- `cargo +nightly fmt --check`
- `cargo clippy --lib -- -D warnings`
- Pre-PR `code-review` gate: `ship`

Note: plain `cargo test <filter>` on `v1.0-dev` also tries to compile the
currently-gated backend-e2e harness and fails on pre-existing gated helper and
private API errors. The focused library tests above exercise this change without
that unrelated harness compile path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token configuration changes for the main control group so permissions are now resolved correctly.
  * Changes to the main control group now behave as expected when the configuration can or cannot be modified.
  * Other token configuration change types continue to work as before.

### DIFF
--- a/src/ui/tokens/update_token_config.rs
+++ b/src/ui/tokens/update_token_config.rs
@@ -24,6 +24,7 @@ use crate::ui::{MessageType, Screen, ScreenLike};
 use dash_sdk::dpp::data_contract::GroupContractPosition;
 use dash_sdk::dpp::data_contract::accessors::v0::DataContractV0Getters;
 use dash_sdk::dpp::data_contract::accessors::v1::DataContractV1Getters;
+use dash_sdk::dpp::data_contract::associated_token::token_configuration::TokenConfiguration;
 use dash_sdk::dpp::data_contract::associated_token::token_configuration::accessors::v0::TokenConfigurationV0Getters;
 use dash_sdk::dpp::data_contract::associated_token::token_configuration_convention::TokenConfigurationConvention;
 use dash_sdk::dpp::data_contract::associated_token::token_configuration_item::TokenConfigurationChangeItem;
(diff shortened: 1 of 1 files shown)


thepastaclaw:
@coderabbitai review

thepastaclaw:
✅ Review complete (commit 30c9a1c6)

thepastaclaw (commented):
## Code Review

Small, well-scoped fix that adds a wrapper around the SDK's `authorized_action_takers_for_configuration_item` so `MainControlGroup` authorization is resolved from `main_control_group_can_be_modified` — the field Platform actually validates. Wrapper is used at the single call site, three focused unit tests cover the branches, and the only other reference in the codebase is inside the fallback branch. One suggestion: the now-live MainControlGroup editor coerces `None` to `Some(0)`, breaking the clear/remove flow, though the change/set happy path works.

Source: reviewers `opus` for general, `gpt-5.5` for general; verifier `opus`.

🟡 1 suggestion(s)

<details>

```

In `src/ui/tokens/update_token_config.rs`:
- [SUGGESTION] src/ui/tokens/update_token_config.rs:659-666: MainControlGroup editor coerces a missing group to group 0
  This PR makes the `MainControlGroup` update path actually reachable at Platform (previously authorization always resolved to `NoOne` via the SDK helper, short-circuiting the flow). The editor payload is `Option<GroupContractPosition>`, and Platform accepts `MainControlGroup(None)` as the representation for clearing the main control group, but this UI runs `opt_grp.map(|g| g).unwrap_or_default()` and unconditionally writes back `Some(grp_txt)`. A token with no current main control group, or a user clearing the textbox to remove it, will submit `Some(0)` instead of `None` — silently targeting group 0 if it exists, or triggering a validation failure if it doesn't. The change/set happy path is unaffected, but the clear path is now broken. Since this edge case only became observable at Platform because of this PR's authorization fix, it's worth fixing in the same change.
```
</details>

thepastaclaw:
@coderabbitai review

thepastaclaw:
Closing this because it was opened from the wrong interpretation of the group-mutation task. DET is unrelated to the reported app flow, so this PR should not stay in the review queue as a proposed fix.

PR #874, opened 2026-07-09
fix(deps)!: pin platform-wallet/dash-sdk to PR #3968 head, adapt to fail-fast load-skip API

Claudius-Maginificent:
## Why this PR exists

- **Problem**: as part of evaluating `dashpay/platform` PR #3968 (`feat/platform-wallet-storage-rehydration`, wallet persistence-reader work), dash-evo-tool's platform dependencies (`dash-sdk`, `rs-sdk-trusted-context-provider`, `platform-wallet`, `platform-wallet-storage`) are pinned to PR #3968's exact head commit — kept in step as that PR moves.
- **Current pin**: `93b967f9c7ab0164b47fe825d2bae58b3974625c` (branch `feat/platform-wallet-storage-rehydration`, base `v4.1-dev`). Previously pinned at `0503f80df5529b1d4f91e58c49abe08cde7b283f`.
- **What breaks without it**: nothing — this is a dependency-pin change plus the source adaptation it requires, not a bug fix. Without the adaptation, the build fails outright against the pinned rev.
- **Status/scope note for reviewers**: PR #3968 is itself open and unmerged on the platform side. This PR captures a deliberate, tested adaptation to that specific rev — not a claim that `dash-evo-tool` should stay pinned there long-term.
- **Blocking relationship**: stacked atop PR #873 (round-2 architecture audit) — base is `fix/round-2-architecture-audit`.

### Initial pin (0503f80d)

- Re-pinned the four platform git dependencies in `Cargo.toml`/`Cargo.lock` to PR #3968's then-head commit.
- Adapted `src/wallet_backend/` to a fail-fast wallet-load model, matching that rev's actual shape (`load_from_persistor()` returns a plain `Result<(), PlatformWalletError>`, not a tri-state `LoadOutcome`): removed the `SkipReason`/`CorruptKind` → `PersistedLoadSkip` mapping, the "N wallets skipped" banner machinery, and the `on_wallet_skipped_on_load` event-handler override. Any load failure — including a wallet that was already registered — is treated as fatal, per product decision (a corrupted or duplicate row is a bug to surface loudly, not silently skip).
- Commented out the `pwm.shutdown()` clean-shutdown report check (returns `()` at that rev, no report type yet) with a `TODO(platform-pr3968)` marker to restore once platform re-adds it.
- Pruned 6 `PlatformWalletError` match arms that didn't exist at that rev (`RehydrationTopologyUnsupported`, `RehydrationPoolMismatch`, `RehydrationPoolTypeMismatch`, `PersisterLoad`, `AddressNonceMismatch`, `ShieldedShutdownIncomplete`) across two exhaustive matches.
- Replaced 2 `From<PersistenceError>` conversions (which didn't exist at that rev) with explicit `PlatformWalletError::Persistence(e.to_string())` construction.
- Removed the kittest coverage (`tests/kittest/skipped_wallets_banner.rs`, 3 tests) that existed solely to cover the now-deleted banner.

### Re-pin to new head (93b967f9)

Upstream PR #3968 moved forward; re-pinned to its new head and re-diffed the relevant crates (`dash-sdk`, `rs-sdk-trusted-context-provider`, `platform-wallet`, `platform-wallet-storage`) between the two SHAs to find the actual public-API delta, rather than assuming the prior adaptation still applied. Cargo.lock also picked up a transitive `rust-dashcore` bump (`48a07d3f` → `b3fb8244`) that platform now requires.

Two of the six previously-pruned pieces came back, forced by the compiler on the now-non-exhaustive matches:

- `PlatformWalletError::PersisterLoad(#[from] PersistenceError)` restored (upstream `728f480f`) — the two contact-request / dashpay-payment persister record sites go back to `e.into()` (typed `PersisterLoad`) instead of the manual `Persistence(String)` stringification the old pin forced.
- `PlatformWalletError::AddressNonceMismatch` restored (upstream `aae79a86`) — re-added to both exhaustive matches (`map_shielded_op_error`, `identity_op_error_kind`), bucketed as a generic wallet-backend/`Other` error, unchanged behavior.

Everything else from the initial adaptation still stands at the new head: `load_from_persistor()` is still fail-fast (no `LoadOutcome`/`SkipReason`, so the skip-banner machinery stays removed — there's no per-wallet skip data to surface), `shutdown()` is still `()` (report check stays commented with its TODO), and the remaining four `PlatformWalletError` variants (`RehydrationTopologyUnsupported`, `RehydrationPoolMismatch`, `RehydrationPoolTypeMismatch`, `ShieldedShutdownIncomplete`) are still absent upstream.

## Testing

- `cargo +nightly fmt --all -- --check` — clean
- `cargo test --all-features --workspace --no-fail-fast` — 1372/1372 lib, 10/10 e2e, 184/184 kittest, 3/3 legacy-table-surface, doctests clean

## Breaking changes

None for end users. Internally: wallet-load behavior is "any load failure is fatal" — a deliberate product decision made at the initial pin, unchanged by the re-pin.

## Checklist

- [x] Adapted to PR #3968's current head, full workspace clippy + tests green
- [x] Platform-side follow-ups filed (duplicate-row protection, orphaned module file — from the initial pin)
- [ ] Human review
- [ ] Confirm whether/when to keep this pin vs. wait for platform to catch `dash-evo-tool` up to include #3968's work

## Attribution


src/wallet_backend/mod.rs
```diff
@@ -2301,8 +2283,15 @@ fn map_shielded_op_error(e: platform_wallet::error::PlatformWalletError) -> Task
         },
 
         // Every remaining variant → generic WalletBackend wrapper.
+        //
+        // TODO(platform-pr3968): `RehydrationTopologyUnsupported`,
```
lklimek: only AddressNonceMismatch, PersisterLoad, and maybe ShieldedShutdownIncomplete  will be restored, others are permanently dropped.

src/wallet_backend/mod.rs
```diff
@@ -2586,17 +2511,13 @@ fn identity_op_error_kind(e: &platform_wallet::error::PlatformWalletError) -> Id
...
-        // A rehydration structural invariant broke (a discovery probe and its
-        // real address pool disagreed on chain order); fail-closed as a
-        // precondition, unrelated to identity registration. Bucket as Other.
-        | P::RehydrationPoolTypeMismatch { .. }
-        // Address nonce desync and an incomplete shielded-worker shutdown are
-        // both precondition/state faults unrelated to identity registration.
-        | P::AddressNonceMismatch { .. }
-        | P::ShieldedShutdownIncomplete { .. } => IdentityOpErrorKind::Other,
+        | P::ShieldedSpendUnconfirmed { .. } => IdentityOpErrorKind::Other,
+        // TODO(platform-pr3968): `PersisterLoad`, `RehydrationTopologyUnsupported`,
```
lklimek: only AddressNonceMismatch, PersisterLoad, and maybe ShieldedShutdownIncomplete  will be restored, others are permanently dropped.

thepastaclaw:
✅ Review complete (commit bac274b9)

src/wallet_backend/mod.rs
lklimek: Fixed in bac274b9. Verified against the pinned rev (93b967f): \`PersisterLoad\` and \`AddressNonceMismatch\` already exist and are routed in the match arm — nothing to restore there. Trimmed both TODOs (the shielded-op map at ~L2281 and the identity-op-kind map at ~L2512) to list only \`ShieldedShutdownIncomplete\` as the variant that *may* return; dropped the three \`Rehydration*\` names since they are permanently gone upstream. The exhaustive (no `_` arm) match still forces a review here the moment platform re-adds any variant. Bilby out. 🐾

thepastaclaw (commented):
## Code Review

Source: opus general + opus rust-quality (initial lanes superseded after base-scope correction; corrective lanes ok); gpt-5.6-sol general + rust-quality (initial and corrective lanes failed: OAuth refresh token invalidated); opus verifier (corrective ok; initial verifier superseded).

PR #874 re-pins dash-sdk/platform-wallet to a fixed rev and adapts the wallet-backend seam to the upstream fail-fast load API, removing the persisted-wallet skip/banner path. The authoritative diff is exactly seven files; findings the prior pass raised against src/backend_task/error.rs and src/backend_task/dashpay/incoming_payments.rs are out of scope (those files are unchanged in this PR) and are dropped. The fail-fast whole-load behavior is a deliberate product decision, so findings critiquing it (and the commented-out report block, which carries a documenting TODO) are dropped. One valid in-scope nitpick remains: a doc comment that still references the removed skip banner.

**Prior Reconciliation**
- `045908b0`: **NOT APPLICABLE** — the checkpoint was skipped before reviewer launch and has no verifier output or trusted findings.

**Carried-Forward Prior Findings**
- None

**New Findings In Latest Delta**
- [nitpick] `src/wallet_backend/mod.rs:464-465` — `register_persisted_wallets` still claims it raises a skipped-wallet warning banner.

💬 1 nitpick(s)

<details>
<summary>Finding details</summary>

- **[NITPICK] `src/wallet_backend/mod.rs:464-465` — `register_persisted_wallets` doc still claims it raises a skip banner**

  The doc comment still says this pass “loads, logs, and raises a warning banner for any skipped wallet.” This PR removes all skip/banner handling: the pinned upstream load is fail-fast and the per-wallet skip concept no longer exists. The function now loads and logs, while any failure aborts the whole pass via `TaskError`. Update the comment to remove the obsolete banner/skip language so it matches the new behavior.

</details>

PR #878, opened 2026-07-10
fix(wallets): link top-nav wallet selector to the Wallets tab

lklimek:
## Why this PR exists

- **Problem**: the top-navigation wallet selector (breadcrumb pill) and the Wallets tab's own wallet picker are two independent selections that don't stay in sync.
- **What breaks without it**: Imagine you are Priya, a power user with several wallets. You switch your active wallet using the top-nav pill while on the Identities page, then click over to the Wallets tab expecting to see that wallet's balances — instead you see whichever wallet the tab happened to remember from before, because the pill is inert on the Wallets page itself (`Unwired`, an explicit un-actioned `TODO` in the code) and the tab never re-reads the shared selection when you arrive. Two controls that both claim to answer "which wallet am I looking at" disagree.
- **Blocking relationship**: stacked on top of PR #860 (branch `docs/platform-wallet-migration-design`) — targets that branch, not `v1.0-dev`.

## What was done

- Made the top-nav wallet pill interactive (`Consumed`) on the Wallets tab — previously subdued/inert there, everywhere else it already worked. Reuses the existing effect-dispatch wiring, no new plumbing.
- Fixed the Wallets tab's cached wallet handle to re-sync from the shared app-scoped selection every time the tab is arrived at, instead of only at construction/network-switch.
- **Closed a dual-hash ambiguity invariant** discovered during the fix: the app tracks an HD-wallet selection and a single-key-wallet selection as two separate hashes, and none of the three setters that touch them (`set_selected_hd_wallet`, `set_selected_single_key_wallet`, `set_selected_identity`) consistently cleared the other on an explicit pick — meaning a stale selection could silently win over a freshly-chosen one. All three setters now self-enforce "at most one active wallet selection" regardless of which one is called.
- Removed a pill-spec constructor that lost its only production caller in this change (dead code cleanup), trimmed a couple of over-length comments.
- `docs/user-stories.md`: added `WAL-028`, extended `WAL-004`.

## Testing

- `cargo clippy -p dash-evo-tool --all-features --all-targets -- -D warnings` — clean.
- `cargo test`/`nextest -p dash-evo-tool --all-features` (wallets/top_panel/hub_screen/global_nav/selected scope) — 46 passed, 0 failed, including a RED-then-GREEN test proving the dual-hash fix is actually necessary (it fails without the fix) and an invariant guard across all three setters (`setters_never_leave_both_hashes_set`).
- `cargo +nightly fmt --check` — clean.
- Independent QA pass: adversarial test execution against the committed tree (not the implementer's report), a security/correctness audit of the shared-setter change (no CRITICAL/HIGH/MEDIUM findings; confirmed no secret material touches this path, no persistence/atomicity issue, no cross-network leakage), and a project-consistency/dedup/docs review (no blockers). Two independent reviewers converged on the same residual asymmetry (the mirror direction of the fix), which was folded in before merge rather than deferred.
- **Known, accepted residual**: no `kittest` test drives a literal pill-click on the Wallets tab specifically (the underlying effect-dispatch logic is unit-tested, and the pill's render/click path is covered generically by an existing `global_nav_switcher` kittest) — a manual GUI spot-check covers cross-network visual behavior, locked-wallet UX, and confirms no Wallets-tab dialog needs a "peek at a different wallet" capability this change would break.

## Breaking changes

None. Internal state-sync fix; no public API, schema, or backend-task changes.

## Checklist

- [x] `cargo build` / `cargo clippy --all-features --all-targets -- -D warnings` clean
- [x] `cargo +nightly fmt --all -- --check` clean
- [x] Unit tests added and green, including a test that fails without the fix
- [ ] Manual GUI spot-check (cross-network, locked-wallet UX) — planned as a follow-up alongside the sibling QR-funding PR's GUI pass

## Attribution


lklimek:
## First live GUI pass — wallet-selector linking

Tested on Testnet with 3 wallets loaded (qrtest [funded, has identities], wallet2, wallet3-locked [password-protected]), built from an isolated genuine recompile of HEAD `c9cd89b3`. Screenshots (LAN): `http://agentic/dash-evo-tool/2026-07-10/partB-wallet-selector/`

| # | Test | Result | Evidence |
|---|------|--------|----------|
| 1 | Pill interactive on Wallets tab | ✅ PASS | Clicking the top-nav pill on the Wallets tab opens the wallet switcher and switching updates the tab's own display (in-content picker, header, balance, address list) **in place**, no navigation. (`B01`→`B02`) |
| 2 | Re-sync on arrival | ✅ PASS | Switched wallet via the pill on **Identity Hub** (a non-Wallets Consumed-pill page) → navigated to Wallets tab → it shows the just-picked wallet (qrtest), not the stale prior one. (`B04`→`B05`) |
| 3 | Two-way regression | ✅ PASS | Switched via the Wallets tab's own in-content `ComboBox` → the top-nav pill reflects it on another page. (`B03`) |
| 4 | Single-wallet (Alex) regression | ✅ PASS | Pill is count-gated (`global_nav_switcher.rs:65`: 0→placeholder, 1→subdued, ≥2→interactive). With one wallet the pill is subdued/info-only everywhere — unchanged by this PR (which only swaps the Wallets spec to `Consumed`). Observed: single-wallet pill opened no switcher. |
| 5 | Cross-network | ✅ PASS | Switched to Mainnet → Wallets tab shows **"No wallets yet"** and the pill shows the **"(no wallet yet)" placeholder** — the 3 testnet wallets do not leak. Selection is per-network `AppContext`-scoped. (`B07`) |
| 6 | Locked-wallet UX | ✅ PASS (core assertion) | Locked wallet3-locked, switched away, then switched **to** it via the pill — **no unexpected password dialog** appeared; the switch is a pure selection op. (`B06`) Caveat: the password was session-cached from creation, so a fully-cold locked-state test would need an app restart; the switching path touches no secrets so it cannot itself prompt. |
| 7 | UX consistency | ✅ PASS | The now-interactive Wallets pill opens the identical wallet-switcher dropdown (wallet list + "Set up another wallet") as the Identity Hub / Masternodes Consumed pills, with uniform count-based behavior. The everyday pages (e.g. Identities) intentionally keep an `Unwired` subdued pill — a deliberate progressive-disclosure choice, not an inconsistency introduced here. |

**Verdict: ready to un-draft.** The core linking (items 1–3) works correctly — the two selectors are now linked and sync both ways, immediately, across pages. Single-wallet, cross-network, and locked-wallet paths all behave correctly. Two minor follow-up-if-desired notes (not blockers): (a) item 6's fully-cold locked-state and (b) any additional edge of item 5 could be re-confirmed after an app restart, but neither is expected to change given the code paths.

> Env note: the shared `/data/target` binary was clobbered by a concurrent worktree build during Part A of this pass; I built Part B into an isolated target with a forced sccache-off recompile and verified the feature is present before testing.

thepastaclaw:
🕓 Ready for review — 8 ahead in queue (commit 820c9e8f)
_Queue position: 9/9_

PR #879, opened 2026-07-10
feat(user-role): persona/capability-gating — role model, callsite reclassification, and role-picker UI

Claudius-Maginificent:
## Why this PR exists
- **Problem**: The app currently gates advanced UI behind a single binary `is_developer_mode()` flag that conflates two unrelated axes — how much a user has chosen to see, and what the connected Platform network actually supports (see `FeatureGate::Shielded`) — and offers no room to distinguish "power user" from "platform developer," even though the README already describes a cumulative three-tier experience.
- **What breaks without it**: Every future feature that should sit strictly between "everyday" and "full developer" access (e.g. masternode operation) has nowhere natural to attach, so it either gets bundled into the overloaded dev-mode flag (wrong audience) or bypasses gating entirely. `docs/ai-design/2026-07-10-persona-capability-gating/design.md` (included in this PR) has the full design rationale, including Invariant I1 and the sentinel-safe migration this phase implements.
- **Blocking relationship**: Base for #880 (callsite reclassification) and #881 (new role-picker UI). **Both have since been merged into this branch** — this PR now represents the complete three-phase rollout, not just Phase 1 in isolation. It stacks on `docs/platform-wallet-migration-design`.

## What was done
Phase 1 — role model and compat shim:
- New `UserRole` type (`src/model/user_role.rs`): a strict total order `Everyday < Power < Developer` (Invariant I1 — feature availability is monotonic in role).
- New `Capability`/`Check`/`FeatureGate::checks()` mechanism (`src/context/feature_gate.rs`): `FeatureGate::is_available()` becomes `checks().all(is_met)`, where each check is `MinRole`, `Capability`, or `Experimental` — the single composition/extension point for all future gating.
- Sentinel-safe persistence migration: the orphaned `AppSettingsWire.user_mode` wire slot is repurposed to carry the new role. Legacy values (`"Advanced"`, `"Beginner"`, unknown/empty) decode to `None`, never to a guessed role.

Phase 2 (#880) — callsite reclassification: `is_developer_mode()`/`enable_developer_mode()` compat shims removed; every former dev-mode callsite reclassified onto explicit `MinRole`/`FeatureGate` checks per a four-bucket rubric. This is where behavior actually changes — see Breaking changes below.

Phase 3 (#881) — role-picker UI: a three-way "Interface mode" selector on Network Settings and during first-time onboarding (`docs/user-roles.md`), replacing the old binary Expert-mode toggle; single-source persistence via `AppSettings.user_role`.

Round 2 (post-review fixes): `DEVELOPER_MODE` support dropped entirely — every account without a chosen role, fresh install or upgraded, now starts in Detailed view (Power) unconditionally, with no `.env` seeding of any kind; the shared role atomic is now encapsulated behind a dedicated `UserRoleCell` type instead of a bare `AtomicU8`; a settings-cache-poisoning bug (a transient read failure could permanently stick the app on default settings, including the wrong network) is fixed by making the settings loader read-only — the write-back path it depended on is gone, not just patched.

## Testing
- `cargo test --all-features --workspace` and `cargo clippy --all-features --all-targets -- -D warnings`: clean at merge time for all three phases (see individual commits).
- Independent QA passes (Marvin) on Phase 2 and Phase 3, including a real bug caught in `src/mcp/server.rs`'s standalone boot path reading `.env DEVELOPER_MODE` directly instead of the new `AppSettings.user_role` source of truth — fixed and re-verified.
- Round 2: independently re-verified by Marvin, including hand-reverting each production fix and confirming its regression test fails for real before restoring it (not just reading the assertions) — zero surviving issues.

## Breaking changes
**Corrected from an earlier "None" claim, which was only true for Phase 1 in isolation.** Now that Phase 2/3 are merged into this branch:
- The single Expert-mode toggle is replaced by three interface levels (Default / Detailed / Developer tools).
- A few of the most advanced actions (state-transition signing overrides, proceeding without a locally held key, clearing Platform addresses) now require the new Developer tools level — Detailed view (formerly "expert mode") no longer implies these.
- `DEVELOPER_MODE` in `.env` is no longer read for role selection at all (superseding an earlier claim in this section that it still seeded the initial role) — every role-unset account starts in Detailed view (Power) unconditionally. The only remaining `.env` `DEVELOPER_MODE` reader is the one-time v34 SPV-database upgrade for pre-v34 installs, an unrelated migration detail.

Full detail: [`CHANGELOG.md`](https://github.com/dashpay/dash-evo-tool/blob/feat/user-role-gating/CHANGELOG.md) § "Expert mode replaced by three interface levels", and [`docs/user-roles.md`](https://github.com/dashpay/dash-evo-tool/blob/feat/user-role-gating/docs/user-roles.md).

## Checklist
- [x] Tests pass
- [x] Clippy clean
- [x] `cargo +nightly fmt` applied
- [ ] Reviewed by a human

## Attribution


thepastaclaw:
✅ Review complete (commit 60fd0915)

src/context/settings_db.rs
```diff
@@ -118,7 +131,46 @@ impl AppContext {
...
+            // itself stays — the v34 SPV migration still reads it.)
+            if let Err(e) = self
+                .app_kv
+                .put(DetScope::Global, AppSettings::KV_KEY, &settings)
+            {
+                tracing::warn!(
+                    error = ?e,
+                    "Failed to persist the seeded user role; it will be re-seeded on the next load"
+                );
+            }
```
thepastaclaw: **🔴 Blocking: Role seeding overwrites settings after a read failure**

Every `DetKv::get` error is converted to `AppSettings::default()`. The new seeding block then sees a missing role and writes those defaults back to the same key. A schema-version mismatch, corrupt value, or transient read failure can therefore destroy the original settings, including network, onboarding, and SPV preferences. Preserve the distinction between an absent or successfully decoded role-less value and a read error; never persist seeded settings after an error.

<sub>source: ['codex']</sub>

src/context/settings_db.rs
```diff
@@ -63,6 +64,18 @@ impl AppContext {
...
+    /// Set the app-global user role and persist it as the canonical string in
+    /// [`AppSettings`] — the single source of truth the runtime atomic is seeded
+    /// from at the next boot. Both role-setting UI surfaces (Settings selector,
+    /// onboarding row) go through here so the runtime atomic and the persisted
+    /// value never diverge.
+    pub fn set_and_persist_user_role(&self, role: UserRole) {
+        self.set_user_role(role);
+        if let Err(e) = self.update_app_settings(|s| s.user_role = Some(role)) {
+            tracing::warn!(error = ?e, "Failed to persist the selected user role");
+        }
```
thepastaclaw: **🔴 Blocking: Failed persistence leaves runtime and stored roles inconsistent**

`set_and_persist_user_role` publishes the new runtime role before writing settings, then reduces a write failure to a log warning. Both UI selectors consequently appear to accept a choice that silently reverts after restart, contradicting the method's documented invariant. Return the typed persistence error, update the runtime atomic only after a successful write, and let both callers retain the previous selection and display an actionable banner.

<sub>source: ['codex']</sub>

src/context/mod.rs
```diff
@@ -754,7 +782,9 @@ impl AppContext {
     }
 
     pub fn state_transition_options(&self) -> Option<StateTransitionCreationOptions> {
-        if self.is_developer_mode() {
+        // Signing override: only the Developer role may relax the security-level
+        // and purpose checks when signing a state transition.
+        if self.user_role().at_least(UserRole::Developer) {
```
thepastaclaw, line 787: **🔴 Blocking: Migrated expert users lose signing overrides**

The migration deliberately maps `DEVELOPER_MODE=true` to `UserRole::Power`, but `state_transition_options` now requires `Developer`. Before this change, the same legacy flag enabled the permissive signing options, so affected users lose existing signing behavior after upgrading. This conflicts with PR #879's explicit zero-behavior-change contract; exclude the Phase-2 reclassification from this PR or provide a migration that preserves the prior capability.

<sub>source: ['codex']</sub>

src/context/feature_gate.rs
```diff
@@ -49,8 +40,102 @@ impl FeatureGate {
...
+impl FeatureGate {
+    /// The checks that must ALL pass for this feature. An empty list means the
+    /// feature is always available (the empty conjunction is `true`).
+    fn checks(self) -> &'static [Check] {
+        match self {
+            FeatureGate::Shielded => &[Check::Capability(Capability::ShieldedProtocol)],
+            FeatureGate::DashPay => &[],
+            FeatureGate::Masternodes => &[Check::MinRole(UserRole::Power)],
+            FeatureGate::DeveloperTools => &[Check::MinRole(UserRole::Developer)],
+        }
```
thepastaclaw, line 150: **🟡 Suggestion: Experimental features bypass the composition mechanism**

No `FeatureGate::checks()` entry constructs `Check::Experimental`; shielded and DashPay UI call `experimental_enabled()` directly. Shielded send destinations can therefore be exposed without also checking `Capability::ShieldedProtocol`, and the advertised single composition point does not govern these multi-callsite features. Add named operation gates that compose the applicable experimental and capability checks, while retaining a separate capability-only predicate where backend initialization needs it.

<sub>source: ['codex']</sub>

src/context/feature_gate.rs
```diff
@@ -49,8 +40,102 @@ impl FeatureGate {
...
+    }
+
+    /// Evaluate whether this feature is available in the current context.
+    pub fn is_available(self, ctx: &AppContext) -> bool {
+        self.checks().iter().all(|c| c.is_met(ctx))
+    }
+}
```
thepastaclaw, line 157: **🟡 Suggestion: Feature-gate composition has no direct unit tests**

The PR's central new logic—heterogeneous checks combined with `all()`—has no focused unit coverage. Existing tests cover roles, persistence, and some UI behavior, but do not lock down the gate table, empty-conjunction behavior, capability handling when the protocol version is unavailable, or AND semantics. Add inline tests for each gate across relevant roles and protocol states, including a multi-check case.

<sub>source: ['claude']</sub>

src/context/mod.rs
```diff
@@ -395,10 +405,26 @@ impl AppContext {
...
+    pub fn user_role(&self) -> UserRole {
+        UserRole::from_u8(self.user_role.load(Ordering::Relaxed))
+    }
+
+    /// Set the app-global user role. Animations are disabled for Power and
+    /// Developer roles (they clutter an operator/developer workflow).
+    pub fn set_user_role(&self, role: UserRole) {
+        self.user_role.store(role as u8, Ordering::Relaxed);
+        self.enable_animations(!role.at_least(UserRole::Power));
+    }
```
thepastaclaw, line 418: **💬 Nitpick: Inactive network contexts retain stale animation state**

The role atomic is shared across all network contexts, but `animate` is stored separately in each context. Changing the role updates only the active context's animation flag. If another previously created context exists, switching back to it restores its old animation behavior even though it observes the new shared role. Synchronize animation state across contexts on role changes or derive the role-dependent animation decision when repainting.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The role model and sentinel-safe migration are well structured, and the current head fixes the earlier boot-source and `.env.example` gaps. However, the current cumulative head can overwrite unreadable settings, silently diverge the selected runtime role from durable settings, and removes signing overrides from users migrated to Power, so changes are required.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_opus_sample, bucket 0): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); opus/general=claude-opus-4-8(completed); codex/rust-quality=gpt-5.6-sol(completed); sonnet5/rust-quality=claude-sonnet-5(completed); opus/rust-quality=claude-opus-4-8(completed); verifier=verifier-codex-fallback-879-1783719869=gpt-5.6-sol fallback; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

🔴 3 blocking | 🟡 2 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/context/settings_db.rs`:
- [BLOCKING] src/context/settings_db.rs:135-150: Role seeding overwrites settings after a read failure
  Every `DetKv::get` error is converted to `AppSettings::default()`. The new seeding block then sees a missing role and writes those defaults back to the same key. A schema-version mismatch, corrupt value, or transient read failure can therefore destroy the original settings, including network, onboarding, and SPV preferences. Preserve the distinction between an absent or successfully decoded role-less value and a read error; never persist seeded settings after an error.
- [BLOCKING] src/context/settings_db.rs:72-76: Failed persistence leaves runtime and stored roles inconsistent
  `set_and_persist_user_role` publishes the new runtime role before writing settings, then reduces a write failure to a log warning. Both UI selectors consequently appear to accept a choice that silently reverts after restart, contradicting the method's documented invariant. Return the typed persistence error, update the runtime atomic only after a successful write, and let both callers retain the previous selection and display an actionable banner.

In `src/context/mod.rs`:
- [BLOCKING] src/context/mod.rs:787: Migrated expert users lose signing overrides
  The migration deliberately maps `DEVELOPER_MODE=true` to `UserRole::Power`, but `state_transition_options` now requires `Developer`. Before this change, the same legacy flag enabled the permissive signing options, so affected users lose existing signing behavior after upgrading. This conflicts with PR #879's explicit zero-behavior-change contract; exclude the Phase-2 reclassification from this PR or provide a migration that preserves the prior capability.

In `src/context/feature_gate.rs`:
- [SUGGESTION] src/context/feature_gate.rs:125-134: Experimental features bypass the composition mechanism
  No `FeatureGate::checks()` entry constructs `Check::Experimental`; shielded and DashPay UI call `experimental_enabled()` directly. Shielded send destinations can therefore be exposed without also checking `Capability::ShieldedProtocol`, and the advertised single composition point does not govern these multi-callsite features. Add named operation gates that compose the applicable experimental and capability checks, while retaining a separate capability-only predicate where backend initialization needs it.
- [SUGGESTION] src/context/feature_gate.rs:125-141: Feature-gate composition has no direct unit tests
  The PR's central new logic—heterogeneous checks combined with `all()`—has no focused unit coverage. Existing tests cover roles, persistence, and some UI behavior, but do not lock down the gate table, empty-conjunction behavior, capability handling when the protocol version is unavailable, or AND semantics. Add inline tests for each gate across relevant roles and protocol states, including a multi-check case.
```
</details>

src/context/mod.rs
Claudius-Maginificent: This is intentional, not a regression — confirmed against the merged design and documentation, not just this PR's original framing.

`DEVELOPER_MODE=true` deliberately migrates to **Power** ("Detailed view"), not **Developer**. Power already restores everything the old expert mode exposed (account details, address tables, masternode tools) — see `docs/user-roles.md` § *Migration from an older install*. The additional restriction of `state_transition_options`'s signing overrides to `Developer`-only is a **deliberate Phase 2 tightening**, explicitly called out in `CHANGELOG.md` under "Expert mode replaced by three interface levels": *"a few of the most advanced actions (state-transition signing overrides, proceeding without a locally held key, clearing Platform addresses) now require the new Developer tools level."*

You're right that this PR's original "Breaking changes: None" / zero-behavior-change framing is stale, though — Phase 2 (#880) and Phase 3 (#881) have since been merged into this branch, so the diff you're reviewing is no longer an isolated zero-behavior-change Phase 1. I'm updating the PR description to reflect that accurately instead of leaving the outdated claim in place.

Leaving this thread unresolved for a human pass since it's a design-intent judgment call, not a code fix.

src/context/mod.rs
Claudius-Maginificent: Confirmed real — `AppContext::user_role` is a shared `Arc<AtomicU8>` across every per-network context, but `animate: AtomicBool` is per-context and only updated for the context `set_user_role` was called on. Switching back to a previously-created, currently-inactive network context can show stale animation behavior relative to the (shared) role.

Deferring rather than fixing in this PR: deriving the repaint decision live from `user_role()` instead of a stored flag (your suggestion) would be the clean fix, but `animate`/`enable_animations` is also the mechanism ~30 kittest/e2e call sites use via `AppState::with_animations(false)` to force animations off for deterministic tests, independent of role. Collapsing the two without care risks destabilizing that test suite. Tracking as a low-priority follow-up (memcan:todo, project=dash-evo-tool) rather than rushing a fix into this PR.

Not resolving — low severity, tracked separately.

src/context/settings_db.rs
Claudius-Maginificent: Fixed in `116cc9ca`. `load_app_settings_uncached` now returns `Result<AppSettings, KvAdapterError>` and only performs the `.env`-role-seed write-back inside the `Ok` path — a k/v read error propagates instead of degrading to defaults-and-persist. `update_app_settings` was also hardened the same way (it does its own read-modify-write on top of the same function and had the identical clobber risk one layer up).

Regression tests `read_error_does_not_overwrite_persisted_settings` and `update_on_unreadable_store_fails_without_clobbering` (`src/context/settings_db.rs`) exercise a genuine `Err` from the store (a `FailingKv` test double), confirmed RED before the fix and green after. Independently re-verified by a second reviewer pass (read the diff, ran the tests myself).

src/context/settings_db.rs
Claudius-Maginificent: Fixed in `116cc9ca`. `set_and_persist_user_role` now persists to the k/v store first and only publishes to the runtime atomic on success, returning `Result<(), KvAdapterError>`. Both callers handle the error: `network_chooser_screen.rs` reverts its stored `selected_role` field to the previous value, `welcome_screen.rs`'s role is re-derived fresh from `user_role()` each frame so it naturally reflects the unchanged value; both show an actionable error banner via `MessageBanner` (no raw error text, details attached via `.with_details(e)`).

Independently re-verified by a second reviewer pass, including reading both call sites' revert logic directly.

src/context/feature_gate.rs
Claudius-Maginificent: Fixed in `116cc9ca`. Added composed `FeatureGate::ShieldedOperations` (`Capability::ShieldedProtocol` AND `Check::Experimental(Shielded)`) and `FeatureGate::DashPayOperations` (`Check::Experimental(DashPay)`), and routed every raw `experimental_enabled(...)` callsite through them: both `send_screen.rs` shielded sites, `shielded_tab.rs`, and all 4 DashPay UI sites. A repo-wide grep confirms no raw `experimental_enabled(ExperimentalFeature::{Shielded,DashPay})` call remains outside `feature_gate.rs` itself.

Worth noting for anyone reviewing the diff: at the currently pinned `dpp` rev, no `PlatformVersion` defines any shielded state-transition type, so `Capability::ShieldedProtocol` is unmet everywhere today — meaning the shielded send options now correctly disappear for Power/Developer-role users (previously shown on role alone, regardless of network support). That's this fix doing its job, not a new regression; independently confirmed against the upstream `dpp` source at the pinned commit.

src/context/feature_gate.rs
Claudius-Maginificent: Fixed in `116cc9ca`. Added a `#[cfg(test)]` module to `src/context/feature_gate.rs` covering: per-role gates (`Masternodes` at Power, `DeveloperTools` at Developer), the empty-conjunction case (`DashPay`'s always-available `checks() == []`), AND-semantics for the new multi-check `ShieldedOperations` gate (a dedicated test proves the gate is false when only the experimental axis passes — a regression to a bare `experimental_enabled()` check would wrongly pass this), and a tripwire test that fails loudly the day any `PlatformVersion` actually defines a shielded state transition, forcing a re-check of the true-case coverage when that happens.

src/context/mod.rs
lklimek: Let's drop support for DEVELOPER_MODE env var.

src/context/settings_db.rs
```diff
@@ -95,30 +127,70 @@ impl AppContext {
...
+        Ok(settings)
+    }
+
+    /// Seeds `user_role` from the legacy `.env DEVELOPER_MODE` flag when the
+    /// persisted blob recorded no explicit role (`None`) — every pre-migration
+    /// user and every fresh install. `DEVELOPER_MODE=true` maps to Power
+    /// (today's "developer mode" is power-user mode); anything else maps to
+    /// Everyday. Mirrors the impure dash-qt autodetect fallback: decoding stays
+    /// pure, the one-time `.env` read happens here at the load site.
+    fn seed_user_role_from_env(&self, mut settings: AppSettings) -> AppSettings {
```
lklimek: Drop DEVELOPER_MODE support,  we don't want to migrate. Set default user role to Power instead for migrated accounts.

src/ui/tokens/tokens_screen/mod.rs
```diff
@@ -3086,7 +3087,7 @@ mod tests {
             egui::Context::default(),
             app_kv,
             secret_store,
-            std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
+            std::sync::Arc::new(std::sync::atomic::AtomicU8::new(0)),
```
lklimek: Architecturally, I don't like the way we use AtomicU8 to store UserRole. I think we should just have some custom type or struct that will hide that atomic usage as internal implementation detail.

.env.example
```diff
@@ -34,6 +34,12 @@ LOCAL_core_rpc_user=dashmate
 LOCAL_core_rpc_password=password
 LOCAL_core_zmq_endpoint=tcp://127.0.0.1:50298
 
+# Interface mode seed (Default view / Detailed view / Developer tools).
+# Read only once, on the very first launch, when no mode has been chosen yet.
+# Once a mode is chosen (via this seed or the in-app selector under Network
+# Settings), this value is never consulted again. See docs/user-roles.md.
+DEVELOPER_MODE=false
```
lklimek: Drop support for DEVELOPER_MODE

src/context/settings_db.rs
```diff
@@ -95,30 +127,70 @@ impl AppContext {
...
-        let loaded = self.load_app_settings_uncached();
+        let loaded = self.load_app_settings_uncached().unwrap_or_else(|e| {
+            tracing::warn!(
+                error = ?e,
+                "Could not read the stored app settings; serving defaults for this read and leaving the stored value untouched"
+            );
+            self.seed_user_role_from_env(AppSettings::default())
+        });
         *guard = Some(loaded.clone());
         loaded
```
thepastaclaw: **🔴 Blocking: Fallback settings on read/seed-write error are permanently cached, discarding real data**

This is the fix for prior-1, and it does stop the store itself from being overwritten. But the in-memory fallback still has two compounding problems verified in the current code:

1. **Permanent cache poisoning**: on error, `loaded` becomes `self.seed_user_role_from_env(AppSettings::default())`, and line 137 unconditionally writes `*guard = Some(loaded.clone())`. `get_app_settings()`'s fast path (line 116) then serves this cached default forever — the only way to invalidate it is an explicit write (`invalidate_settings_cache()`, called from `update_app_settings`/`set_app_settings`). A read-only session (e.g. the MCP stdio boot path, or any UI session where the user never changes a setting) has no self-healing path: one transient read failure (a poisoned lock, a momentary SQLite hiccup) permanently resets the displayed network, theme, onboarding-completed flag, and user role in memory for the rest of the process's life, even though the on-disk value is intact and the store recovers immediately.
2. **Discards already-decoded data on the seed write-back path specifically**: `load_app_settings_uncached` (lines 156-173) can successfully `get` and decode the stored blob, and only fail later at the one-time `.env` seed write-back (`self.app_kv.put(...)?` at line 169-170) for a role-less legacy blob — every pre-migration user hits this exact path on first launch after upgrading. That `put` failure propagates via `?`, discarding the already-decoded `settings` (real network, theme, onboarding, SPV prefs — only the role was being seeded) and falling all the way back to `AppSettings::default()` in `get_app_settings`'s error handler, which then gets cached per point 1.

The doc comment's claim that a transient failure 'costs the user nothing once the store recovers' is only true if some other code happens to call an `update_*` method afterward, which is not guaranteed. Consider not caching the fallback value on error (return it without populating `guard`), so the next call retries the real read instead of pinning defaults into memory indefinitely.

```suggestion
let loaded = match self.load_app_settings_uncached() {
    Ok(settings) => settings,
    Err(e) => {
        tracing::warn!(
            error = ?e,
            "Could not read the stored app settings; serving defaults for this read only"
        );
        return self.seed_user_role_from_env(AppSettings::default());
    }
};
*guard = Some(loaded.clone());
loaded
```

<sub>source: ['sonnet5-rust-quality', 'codex-general']</sub>

thepastaclaw (changes requested):
## Code Review

This delta (79914cf5..fc3317c0) fixes four of the six prior blocking/suggestion findings with well-tested changes: settings read failures no longer write defaults back to the store, role persistence is now atomic-and-reported via a typed `Result` with both UI callers correctly reverting on failure, and the new `ShieldedOperations`/`DashPayOperations` gates route every former direct `experimental_enabled()` callsite through the composed `FeatureGate` mechanism with solid unit tests. Prior-3 (signing-override loss for migrated Power users) is confirmed downgraded to an intentionally documented breaking change (CHANGELOG.md:102-107, docs/user-roles.md). Prior-6 (stale per-context animation state) is confirmed unchanged and still valid. The fix for prior-1 introduces a residual issue of its own: `get_app_settings()` (src/context/settings_db.rs:130-138) now caches `AppSettings::default()` into the in-memory cache on *any* read or seed-write-back error and never retries until an unrelated write invalidates the cache — this both permanently poisons the session's settings after a transient hiccup, and, on the specific seed-write-back failure path (line 169-170), discards settings that were already successfully decoded. Separately, the Shielded tab's Shield/Send/Unshield buttons (src/ui/wallets/shielded_tab.rs:548-603) never consult the `ShieldedOperations` gate that was computed for this exact purpose — this predates the current delta (present since the file's introduction) and is currently unreachable because no protocol version yet grants the underlying capability, but it defeats the composition mechanism this PR built and should be closed as part of finishing the reclassification.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 3): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); codex/rust-quality=gpt-5.6-sol(completed); sonnet5/rust-quality=claude-sonnet-5(completed); verifier=verifier-sonnet5-879-1783940033=claude-sonnet-5(primary); orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

<details>

```

In `src/context/settings_db.rs`:
- [BLOCKING] src/context/settings_db.rs:130-138: Fallback settings on read/seed-write error are permanently cached, discarding real data
  This is the fix for prior-1, and it does stop the store itself from being overwritten. But the in-memory fallback still has two compounding problems verified in the current code:

  1. **Permanent cache poisoning**: on error, `loaded` becomes `self.seed_user_role_from_env(AppSettings::default())`, and line 137 unconditionally writes `*guard = Some(loaded.clone())`. `get_app_settings()`'s fast path (line 116) then serves this cached default forever — the only way to invalidate it is an explicit write (`invalidate_settings_cache()`, called from `update_app_settings`/`set_app_settings`). A read-only session (e.g. the MCP stdio boot path, or any UI session where the user never changes a setting) has no self-healing path: one transient read failure (a poisoned lock, a momentary SQLite hiccup) permanently resets the displayed network, theme, onboarding-completed flag, and user role in memory for the rest of the process's life, even though the on-disk value is intact and the store recovers immediately.
  2. **Discards already-decoded data on the seed write-back path specifically**: `load_app_settings_uncached` (lines 156-173) can successfully `get` and decode the stored blob, and only fail later at the one-time `.env` seed write-back (`self.app_kv.put(...)?` at line 169-170) for a role-less legacy blob — every pre-migration user hits this exact path on first launch after upgrading. That `put` failure propagates via `?`, discarding the already-decoded `settings` (real network, theme, onboarding, SPV prefs — only the role was being seeded) and falling all the way back to `AppSettings::default()` in `get_app_settings`'s error handler, which then gets cached per point 1.

  The doc comment's claim that a transient failure 'costs the user nothing once the store recovers' is only true if some other code happens to call an `update_*` method afterward, which is not guaranteed. Consider not caching the fallback value on error (return it without populating `guard`), so the next call retries the real read instead of pinning defaults into memory indefinitely.

In `src/ui/wallets/shielded_tab.rs`:
- [SUGGESTION] src/ui/wallets/shielded_tab.rs:548-603: Shield/Send/Unshield buttons never check the ShieldedOperations gate
  `render_address_section` computes `shielded_enabled = FeatureGate::ShieldedOperations.is_available(&self.app_context)` at line 197, but only uses it to set the address-table's `default_open` state. The Shield, Send (Private), and Unshield buttons at lines 548-603 are enabled purely on `!self.syncing && !spend_locked` / `can_spend` — neither the role axis nor the experimental axis is consulted before opening the send flow. The tab itself is only reachable when `FeatureGate::Shielded` (capability-only) is satisfied (`wallets_screen/mod.rs:1214`), so today this is unreachable: no known protocol version defines the shielded state transitions yet (confirmed by the tripwire test in feature_gate.rs:272-286). This predates the current delta — the same gap existed with the old `experimental_enabled()` call before this fix commit swapped it for the composed gate — so it isn't a regression introduced here, but it does defeat the exact purpose `ShieldedOperations` was built for (this PR's own doc comment: 'shielded send options must not be offered on a network that cannot settle them' / experimental-axis gating). Worth closing while this exact composition is being wired up, so an Everyday user can't reach these flows the moment the capability ships.
```
</details>

src/context/mod.rs
thepastaclaw: Still valid on `fc3317c0`: `user_role` is shared across network contexts, but `animate` remains per-context, so inactive contexts can retain stale animation state. This finding is carried forward unchanged.

src/context/mod.rs
thepastaclaw: Still valid on `fc3317c0`: the branch still supports `DEVELOPER_MODE` and maps `true` to Power while signing overrides require Developer. Lukasz has now explicitly requested dropping `DEVELOPER_MODE` support, so this remains an outstanding requested change rather than an intentionally deferred product decision.

src/context/settings_db.rs
Claudius-Maginificent: Done in `f34d8d2f`. `Config.developer_mode`, its `.env` write, and its `DEVELOPER_MODE` env read are removed entirely — confirmed it was the only consumer of that field (the unrelated v34 SPV-database migration reads `DEVELOPER_MODE` independently via its own parser, untouched).

Role-unset now maps unconditionally to `UserRole::Power` — this applies to every account without a chosen role, not only migrated ones. We considered narrowing it to migrated accounts only (distinguishing via `onboarding_completed`), but the decision was to keep it simple and universal: no role migration from the previous version, just a global Power default, full stop. Independently re-verified.

src/ui/tokens/tokens_screen/mod.rs
Claudius-Maginificent: Done in `f34d8d2f`. Added `UserRoleCell` (`src/model/user_role.rs`) — the bare `AtomicU8` is now a private implementation detail behind a `get()`/`set()`/`Clone` API, and `UserRole::from_u8` is private to the cell. Repo-wide `AtomicU8` grep (including `tests/backend-e2e/`) now returns only this cell's internals plus one unrelated field in `connection_status.rs`. Independently re-verified.

.env.example
Claudius-Maginificent: Done in `f34d8d2f` — the `DEVELOPER_MODE=false` block is removed from `.env.example`.

src/context/settings_db.rs
Claudius-Maginificent: Fixed in `f34d8d2f`, and structurally rather than by patching the symptom. `load_app_settings_uncached` no longer writes back at all — the one-time `.env` role seed it used to reconcile is gone (see the DEVELOPER_MODE-removal thread), so there's no `put` left to fail and no reason to cache a fallback on error. Both failure modes you flagged are closed by elimination:

1. The error path now returns the fallback without populating the cache guard, so a transient read failure no longer permanently poisons in-memory settings — the next call retries the real read.
2. There's no seed write-back left to discard already-decoded settings on failure.

Two regression tests lock this in, both confirmed RED against the pre-fix code before being confirmed GREEN against the fix: `read_error_does_not_poison_the_settings_cache` and `role_less_blob_keeps_real_settings_and_is_not_written_back`. Independently re-verified by a second pass, including hand-reverting the fix and confirming both tests fail for real before restoring it.

src/context/feature_gate.rs
```diff
@@ -49,8 +40,286 @@ impl FeatureGate {
                     && st.shield_from_asset_lock_state_transition.max_version > 0
                     && st.shielded_withdrawal_state_transition.max_version > 0
```
thepastaclaw, line 41: **🟡 Suggestion: Capability::ShieldedProtocol misuses max_version==0 as an absence sentinel**

`is_met` treats `st.shield_state_transition.max_version > 0` (and the other four shielded transitions) as "the feature has been defined in this protocol version." I verified this against the vendored `versioned-feature-core` crate: `FeatureVersionBounds::check_version(v)` is `v >= min_version && v <= max_version`, so `{min_version: 0, max_version: 0}` is a perfectly valid, checkable bound for version 0 — not an absence sentinel. Worse, I confirmed in the vendored `rs-platform-version` source (`dpp_state_transition_serialization_versions/v1.rs` and `v2.rs`) that `identity_create_state_transition` — a definitely-live, working feature — uses the exact same `{min_version: 0, max_version: 0, default_current_version: 0}` bounds as the shielded transitions in both known protocol versions. If the shielded transitions ship as v0 features (the same pattern every other v0 DPP feature uses), this check will remain permanently false and the capability will never register as available, even though the feature is live. It happens to correctly read as "not yet defined" today only because no known protocol version has shipped shielded transitions at all yet — not because the check is measuring the right thing. Use an explicit activation signal (e.g. a dedicated bool/enum on the version struct, or comparing against the protocol version the transitions were actually introduced in) rather than `max_version > 0`.

docs/ai-design/2026-07-10-persona-capability-gating/design.md
```diff
@@ -0,0 +1,426 @@
...
+
+**R6 — Backward-compatible persistence.**
+Existing users' saved settings must not break — both the positional bincode `AppSettings` blob (`settings.rs:267` — field order *is* the on-disk format) and the `.env DEVELOPER_MODE` key. The migration must not silently change any existing user's effective role (`[REVIEW PROJ-001]`).
+
+**R7 — Single source of truth.**
+The user role must have one authoritative persisted home and one runtime accessor. Today it has two persisted homes (`.env` live, `AppSettings.user_mode` dead) and the wrong one wins.
+
+**R8 — Default to the simplest role.**
+Fresh installs default to Everyday (progressive disclosure's baseline). This matches today's behaviour (`developer_mode` defaults to `false`, `mod.rs:313`), so no existing user regresses.
```
thepastaclaw, line 103: **🟡 Suggestion: Design doc's R6/R8 requirements contradict the Round-2 decisions this PR actually shipped**

This design doc is included in and referenced by the PR description as the architectural rationale. R6 (line 96-97) requires preserving `.env DEVELOPER_MODE` as a persistence source and says the migration "must not silently change any existing user's effective role," and R8 (line 102-103) requires fresh installs to default to Everyday. Both are now obsolete: this delta (and the immediately preceding one) removed `.env DEVELOPER_MODE` from role resolution entirely and set `UserRole::WHEN_UNSET = Power`, a deliberate departure from R8 documented in `UserRole::WHEN_UNSET`'s own doc comment (`model/user_role.rs:34-42`) and in CHANGELOG.md/docs/user-roles.md. Since this file is the design rationale new engineers will read to understand the role model, it should get a Round 2 addendum (or an update to R6/R8 themselves) so it doesn't send future readers looking for a migration path that no longer exists.

<sub>source: ['codex-general']</sub>

src/app.rs
```diff
@@ -636,6 +633,17 @@ impl AppState {
...
+        // Seed the shared role cell from AppSettings — the single source of truth.
+        // `get_app_settings` resolves an account that never chose a role to
+        // `UserRole::WHEN_UNSET`. `set_user_role` publishes the value to every
+        // context holding the cell and syncs animation gating.
+        active_context.set_user_role(
+            active_context
+                .get_app_settings()
+                .user_role
+                .unwrap_or(crate::model::user_role::UserRole::WHEN_UNSET),
+        );
```
thepastaclaw, line 645: **💬 Nitpick: Dead .unwrap_or(UserRole::WHEN_UNSET) fallback after the settings fix**

I traced every path through `get_app_settings()` (settings_db.rs:115-146): the success path, the first-run/absent-blob path, and the error path all now funnel through `with_default_user_role`, which does `settings.user_role.get_or_insert(UserRole::WHEN_UNSET)` unconditionally. That guarantees `AppSettings::user_role` is `Some(_)` for every value `get_app_settings()` can return, making `.unwrap_or(crate::model::user_role::UserRole::WHEN_UNSET)` here — and the identical one at `src/mcp/server.rs:394-399` — provably unreachable. Not a functional bug (the fallback value matches the guaranteed one), but it's a duplicated decision: if `with_default_user_role`'s default ever changes, these two callsites silently keep the stale value instead of surfacing the mismatch.

<sub>source: ['sonnet5-rust-quality']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers — gpt-5.6-sol (Sol reviewer: general, rust-quality) + claude-sonnet-5 (Sonnet reviewer: general, rust-quality); verifier — claude-sonnet-5; orchestrator — openai/gpt-5.6-sol (orchestration-only, not a finding source). Experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 2 (not Opus-sampled).

This delta (fc3317c0..60fd0915) genuinely fixes the settings cache-poisoning bug (the fallback is no longer cached at all, and the seed write-back path that used to trade decoded settings for defaults is deleted outright) and completes the human-directed removal of DEVELOPER_MODE role-seeding, resolving the earlier signing-override migration concern by eliminating the mechanism that caused it. It also cleanly encapsulates the role atomic behind UserRoleCell. The Shielded action-buttons gap is confirmed still open and, on inspection, deeper than previously described: preset send flows (Shield/Send Private/Unshield) never execute the code path that contains the only two ShieldedOperations checks in send_screen.rs, and there is no backend-task-level enforcement either — though this remains dormant today since no known protocol version defines the shielded state transitions. The stale per-context animation state is confirmed unchanged but is now backed by an explicit, reasoned author deferral (it doubles as a kittest/e2e animation-off seam), so it is treated as intentionally deferred rather than re-flagged. Four new findings surfaced on independent verification: a real latent bug in `Capability::ShieldedProtocol` where `max_version > 0` is not a valid "feature is defined" sentinel (confirmed against the vendored `dpp`/`platform-version` source — v0 is a valid, checkable bound, and even the live `identity_create_state_transition` uses `{0,0,0}`), a stale design doc that still requires the removed DEVELOPER_MODE seed and an Everyday default, a left-nav label that still says undifferentiated "Expert" despite this PR's own three-tier reclassification, and a dead `.unwrap_or(WHEN_UNSET)` fallback now provably unreachable after the settings fix.

🟡 4 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/ui/wallets/shielded_tab.rs`:
- [SUGGESTION] src/ui/wallets/shielded_tab.rs:197-603: Shielded action buttons and preset send flows fully bypass the ShieldedOperations gate
  `FeatureGate::ShieldedOperations` (capability AND experimental-role composition) is computed at shielded_tab.rs:197 but used only to set the addresses section's `default_open` state — the Shield, Send (Private), and Unshield buttons at lines 548-603 are enabled purely by sync/spend-lock/balance state. Tracing the click handlers further confirms the gap is deeper than a missing button check: each button opens a preset `SendFlow` (`open_send_flow`, shielded_tab.rs:116), and `SendScreen::ui()` (send_screen.rs:3541) routes any preset flow (`flow.is_preset()` is true for Shield/ShieldedSend/Unshield) straight to `render_flow_send`, which never calls `render_source_selection` — the only function in send_screen.rs that checks `FeatureGate::ShieldedOperations` (lines 2058, 2110), and that check only fires for the manual/General flow's source and destination pickers. There is also no backend-task-level enforcement (no `ShieldedOperations` reference anywhere under `src/backend_task/`). The tab itself is admitted only by the weaker, capability-only `FeatureGate::Shielded` (wallets_screen/mod.rs:1214). Currently unreachable in practice: both known protocol versions (v1, v2) define every shielded state transition with `max_version: 0` in the vendored `dpp` crate, so `Capability::ShieldedProtocol` (and therefore `FeatureGate::Shielded` itself) is unmet on any live network today — but the moment a protocol version supplies the capability, an Everyday/Power user without the experimental flag can complete a full Shield/Send/Unshield flow with nothing enforcing the composition this PR built for exactly this purpose.

In `src/context/feature_gate.rs`:
- [SUGGESTION] src/context/feature_gate.rs:34-41: Capability::ShieldedProtocol misuses max_version==0 as an absence sentinel
  `is_met` treats `st.shield_state_transition.max_version > 0` (and the other four shielded transitions) as "the feature has been defined in this protocol version." I verified this against the vendored `versioned-feature-core` crate: `FeatureVersionBounds::check_version(v)` is `v >= min_version && v <= max_version`, so `{min_version: 0, max_version: 0}` is a perfectly valid, checkable bound for version 0 — not an absence sentinel. Worse, I confirmed in the vendored `rs-platform-version` source (`dpp_state_transition_serialization_versions/v1.rs` and `v2.rs`) that `identity_create_state_transition` — a definitely-live, working feature — uses the exact same `{min_version: 0, max_version: 0, default_current_version: 0}` bounds as the shielded transitions in both known protocol versions. If the shielded transitions ship as v0 features (the same pattern every other v0 DPP feature uses), this check will remain permanently false and the capability will never register as available, even though the feature is live. It happens to correctly read as "not yet defined" today only because no known protocol version has shipped shielded transitions at all yet — not because the check is measuring the right thing. Use an explicit activation signal (e.g. a dedicated bool/enum on the version struct, or comparing against the protocol version the transitions were actually introduced in) rather than `max_version > 0`.

In `src/ui/components/left_panel.rs`:
- [SUGGESTION] src/ui/components/left_panel.rs:318-329: Left nav still shows the undifferentiated "Expert" label this PR's reclassification replaced
  This PR's Phase 2 commit (79914cf5) reclassified this exact gate from `is_developer_mode()` to `app_context.user_role().at_least(UserRole::Power)`, but left the displayed text unchanged: `"🔧 Expert"` with tooltip `"Expert mode is enabled — shows advanced options"`. Since the gate is `>= Power`, both Power ("Detailed view") and Developer ("Developer tools") users see the identical label and tooltip, with no way to tell which of the two new tiers is active — directly at odds with the three-tier vocabulary this same PR establishes everywhere else (Network Settings selector, onboarding row, CHANGELOG, docs/user-roles.md). This is the one user-facing surface still using the retired binary "Expert mode" framing; render the role's own `label()` instead.

In `docs/ai-design/2026-07-10-persona-capability-gating/design.md`:
- [SUGGESTION] docs/ai-design/2026-07-10-persona-capability-gating/design.md:96-103: Design doc's R6/R8 requirements contradict the Round-2 decisions this PR actually shipped
  This design doc is included in and referenced by the PR description as the architectural rationale. R6 (line 96-97) requires preserving `.env DEVELOPER_MODE` as a persistence source and says the migration "must not silently change any existing user's effective role," and R8 (line 102-103) requires fresh installs to default to Everyday. Both are now obsolete: this delta (and the immediately preceding one) removed `.env DEVELOPER_MODE` from role resolution entirely and set `UserRole::WHEN_UNSET = Power`, a deliberate departure from R8 documented in `UserRole::WHEN_UNSET`'s own doc comment (`model/user_role.rs:34-42`) and in CHANGELOG.md/docs/user-roles.md. Since this file is the design rationale new engineers will read to understand the role model, it should get a Round 2 addendum (or an update to R6/R8 themselves) so it doesn't send future readers looking for a migration path that no longer exists.
```
</details>

src/context/mod.rs
thepastaclaw: Intentionally deferred on `60fd0915`: the stale per-context animation state remains present, but it is display-only and `animate` / `enable_animations` also serves as the independent animation-off seam used by kittest and e2e coverage. The author is tracking the cleaner separation as a low-priority follow-up rather than changing that test seam in this PR.

<sub>This thread remains open; it was not fixed by `f34d8d2`.</sub>

src/context/mod.rs
thepastaclaw: Resolved in `f34d8d2` — _Migrated expert users lose signing overrides_ no longer present.

src/app.rs
```diff
@@ -636,6 +633,17 @@ impl AppState {
...
 
+        // Seed the shared role cell from AppSettings — the single source of truth.
+        // `get_app_settings` resolves an account that never chose a role to
+        // `UserRole::WHEN_UNSET`. `set_user_role` publishes the value to every
+        // context holding the cell and syncs animation gating.
+        active_context.set_user_role(
+            active_context
+                .get_app_settings()
+                .user_role
+                .unwrap_or(crate::model::user_role::UserRole::WHEN_UNSET),
```
thepastaclaw, line 644: **🔴 Blocking: A transient settings read still elevates the runtime role for the session**

`get_app_settings()` now correctly returns an uncached fallback on error, but this boot call immediately publishes that fallback. The error fallback is `with_default_user_role(AppSettings::default())`, so its role is `Power`. If the persisted role is `Everyday` and this read transiently fails, the shared `UserRoleCell` becomes `Power`; after the store recovers, later settings reads return `Everyday`, but nothing reseeds the cell, so UI and feature gates remain at Power until restart. The MCP boot path has the same sequence. This leaves the runtime role inconsistent with the recovered settings and can expose Power-only operations after a transient read failure.

Keep the least-privileged pre-seed role on read errors, or make the boot seed distinguish a successful/absent read from an error instead of publishing the fallback. Please add a regression that stores `Everyday`, fails the boot role read once, recovers the store, and asserts the runtime role is not left at `Power`.

thepastaclaw (changes requested):
Supplemental exact-head check: the cache fix is real, but one boot-time consumer still turns the one-read fallback into session-wide role state.

src/context/mod.rs
Claudius-Maginificent: Superseded by an explicit product decision, not left unaddressed. See the other thread on this PR (`src/context/settings_db.rs`, "Drop DEVELOPER_MODE support... Set default user role to Power instead for migrated accounts") — DEVELOPER_MODE was fully dropped from role resolution, not selectively migrated. There is no code path anywhere that maps `DEVELOPER_MODE=true` to any role, so the specific regression this finding describes (a `DEVELOPER_MODE=true` user silently losing `Developer`-gated signing because they get remapped to `Power`) cannot occur — every legacy account, `DEVELOPER_MODE=true` or not, now lands on `Power` uniformly via `UserRole::WHEN_UNSET`, same as a fresh install. A user who wants the permissive signing override self-selects `Developer` in the role picker post-upgrade. This is a deliberate, already-implemented scope decision (see `UserRole::WHEN_UNSET` doc comment, `CHANGELOG.md`, `docs/user-roles.md`), confirmed by the repo owner directly on this PR.

src/context/settings_db.rs
Claudius-Maginificent: Done — `DEVELOPER_MODE` was dropped entirely from role resolution (not migrated), and migrated accounts default to `Power` via `UserRole::WHEN_UNSET`. Implemented in `f34d8d2f` on this PR (now merged into #860).

src/ui/tokens/tokens_screen/mod.rs
Claudius-Maginificent: Done — the bare `AtomicU8` is now hidden behind `UserRoleCell` (`src/model/user_role.rs`), a small type exposing `get()`/`set()`/`Clone`. `UserRole::from_u8` was made private. Implemented in `f34d8d2f` on this PR (now merged into #860).

.env.example
Claudius-Maginificent: Done — the `DEVELOPER_MODE=false` block is removed from `.env.example`. Implemented in `f34d8d2f` on this PR (now merged into #860).

PR #885, opened 2026-07-13
feat(migration): import legacy v0.9.3 identities and their keys

Claudius-Maginificent:
## Why this PR exists
- **Problem**: A new mutation-tested integration test added in #882 (`src/backend_task/migration/v093_upgrade.rs`) proved that dash-evo-tool's v0.9.3 → v1.0 upgrade path silently drops a user's legacy identities — the `data.db` `identity` table survives the schema ladder intact, but nothing imports it into the modern identity store.
- **What breaks without it**: A v0.9.3 user upgrading loses every identity they had loaded — including a masternode's owner/voting identity and its private keys. They'd have to reload it manually, if they even realize it's gone. This directly undermines the upcoming masternode-owner-edition release, whose target audience is exactly the people upgrading from v0.9.3 to use it.
- **Blocking relationship**: Stacked on `feat/v1.0-parity-batch` (#882) — this PR needs the migration-test infrastructure #882 introduced and targets that branch directly rather than the shared integration branch, so this diff stays scoped to the identity-import work.

## What was done
- `legacy_import::read_identities`: reads the legacy `identity` table, filtering `is_local = 1 AND data IS NOT NULL` (v0.9.3 also stores observed/cached identities with `is_local = 0` and a NULL blob — those are cache, not user data, and are correctly skipped).
- New `MigrationStep::Identities`, wired into `finish_unwire::run()` after the wallet drain (so the vault is reachable and wallets are already hydrated for wallet-linked keys). Its own idempotency sentinel (`det:migration:identities:<net>:v1`) — deliberately not reusing the wallet-drain step's sentinel, which would have silently skipped identity import for anyone who'd already run the drain before this ships.
- `identity` added to `LEGACY_TABLES`, so an identity-only v0.9.3 install is actually detected as needing migration.
- Import goes through the existing `AppContext::insert_local_qualified_identity` — the same vault-first path normal identity loading already uses. **No new secret-handling code was written**; verified by grep that no new production code constructs or persists raw key bytes.
- Skip-if-already-present before insert, so a retry doesn't silently overwrite a user's post-migration edits (e.g. a renamed alias) with the stale legacy blob.
- A wallet link on an absent or currently-locked wallet is preserved, not nulled — it's what lets the identity re-attach once the wallet is unlocked later.
- An undecodable identity row is counted and warned about, without blocking the rows that decode successfully; the sentinel is only withheld when at least one row failed, so a future decoder fix can still pick it up on retry.

## Testing
- Six-row fixture in `v093_upgrade.rs`: a masternode with `Clear` owner+voting keys, a DPNS user identity with a wallet-derived key, a wallet-less evonode identity, an `is_local=0` row (must be skipped), an identity on a locked protected wallet, and an undecodable-blob row.
- **Golden-blob wire-compatibility test**: a real v0.9.3 `QualifiedIdentity::to_bytes()` blob (produced by an actual v0.9.3 build in a throwaway standalone crate, not committed) is pasted in as a constant and asserted to decode correctly against the current codebase's bincode `2.0.1` — closing the one real cross-version risk (v0.9.3 pinned bincode `2.0.0-rc.3`) empirically rather than by code inspection alone. It decodes byte-for-byte.
- Security assertion reads the raw stored bytes directly (`det:identity:v1`), not through a load path — an earlier version of this test accidentally checked *after* a load call that itself repairs any resident plaintext, which would have hidden a real bug; caught via mutation testing and rewritten to check the raw bytes with no repair path in between.
- A dedicated retry test proves skip-if-present is actually load-bearing: force the sentinel-withheld path (an unreadable row), rename an identity, re-run the full migration, and confirm the rename survives.
- `cargo test --lib --all-features`: 1563 passed, 0 failed (1547 base + 16 new). `cargo clippy --all-features --all-targets -- -D warnings`: clean. `cargo +nightly fmt --all`: applied.
- Independent QA in progress at the time of this PR — will follow up with a comment once it completes; pushing now so review can start in parallel.

## Breaking changes
None. Purely additive — a new one-time migration step, gated by its own sentinel, that only runs for accounts with legacy v0.9.3 identity data present.

## Checklist
- [x] Tests pass
- [x] Clippy clean
- [x] `cargo +nightly fmt` applied
- [x] Independent QA pass (complete — see comments; found and fixed 1 medium, 1 low)
- [ ] Reviewed by a human

## Attribution


Claudius-Maginificent:
Independent QA (adversarial, not just review) complete on `453cd151`. All five implementation claims independently reconfirmed — including rebuilding the bincode `2.0.0-rc.3` ↔ `2.0.1` cross-version proof from scratch (the original golden-blob generator crate was deleted after use, so its constant's provenance was otherwise unfalsifiable from the repo alone; that risk is now closed independently) and reproducing both of the implementer's self-caught false-green test bugs via revert-and-confirm-red. No new secret-handling code paths, confirmed by an independent grep.

Two real bugs were found and are now fixed in `ce1205aa`:

- **QA-001 (MEDIUM)**: the identity-import pass ran *after* the vote/top-up pass unwrapped its result with `?`, so a single corrupt scheduled-vote entry would permanently prevent identity import — including a masternode owner's private keys — on every retry. Reproduced with a real poisoned-index test, fixed by running identity import before either DET-owned result is unwrapped and folding both outcomes at the terminal-state step. Regression test added.
- **QA-002 (LOW)**: an out-of-range `status`/`wallet_index` column value hard-failed the entire identity read instead of being skipped like every other row-level corruption case in the same loop. Fixed to follow the same skip-and-count-unreadable policy. Regression test added.

Both fixes verified with fresh ledger evidence (clippy clean, 1565/1565 lib tests passing, new test names confirmed present in the log against a forced rebuild — not a cached replay).

thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 3326548a)

src/database/legacy_import.rs
```diff
@@ -222,6 +255,133 @@ pub(crate) fn read_scheduled_votes(
...
+            continue;
+        };
+
+        // Neither field is in the bincode blob — the legacy encoder skipped both
+        // and kept them in columns. Without this, every imported identity reads
+        // back as `Unknown` status on mainnet.
+        qi.status = IdentityStatus::from(status);
+        qi.network = network;
+
+        out.identities.push(LegacyIdentityRow { id, qi, wallet });
```
thepastaclaw, line 439: **🔴 Blocking: Alias-column fallback required by the design doc is never implemented**

The `SELECT` at line 278 omits `alias`, and nothing in the function ever reads it, so an imported identity's alias always comes from whatever is baked into the `data` blob — even when it's `None`. This PR's own design doc (`docs/ai-design/2026-07-13-legacy-identity-migration/design.md:196`) is explicit about the required behavior: "Column is a denormalised copy. Prefer the blob; the column is a fallback if `qi.alias` is `None`." That fallback path is entirely missing from the implementation. Any legacy identity whose blob alias is `None` but whose column carries a value — a realistic v0.9.3 state, since renaming an identity plausibly touched the cheap alias column without re-serializing the whole key-bearing blob — silently imports with no alias, contradicting the documented contract. Read the nullable `alias` column and apply it as a fallback when `qi.alias.is_none()`.

```suggestion
    let mut stmt = conn.prepare(
        "SELECT id, data, status, wallet, wallet_index, alias FROM identity \
         WHERE is_local = 1 AND data IS NOT NULL AND network IN (?1, ?2)",
    )?;
    let mut rows = stmt.query(rusqlite::params![
        network.to_string(),
        mainnet_alias_for(network)
    ])?;

    let mut out = LegacyIdentities::default();
    while let Some(row) = rows.next()? {
        let id: Vec<u8> = row.get(0)?;
        let data: Vec<u8> = row.get(1)?;
        // SQLite stores both of these as signed 64-bit integers and the legacy
        // schema puts no `CHECK` on either, so a corrupted row can hold a value
        // that does not fit. Widen the read and convert explicitly: a narrow
        // `row.get::<u8>` would raise `IntegralValueOutOfRange` through `?` and
        // take the whole identity read down — every other row-level corruption
        // here is counted and skipped, and this one must behave the same.
        let status: i64 = row.get(2)?;
        let wallet: Option<Vec<u8>> = row.get(3)?;
        let wallet_index: Option<i64> = row.get(4)?;
        let alias: Option<String> = row.get(5)?;

        let Ok(id) = <[u8; 32]>::try_from(id.as_slice()) else {
            tracing::warn!(
                target = "database::legacy_import",
                blob_len = id.len(),
                "Skipping legacy identity with a non-32-byte id",
            );
            out.unreadable = out.unreadable.saturating_add(1);
            continue;
        };

        let Ok(status) = u8::try_from(status) else {
            tracing::warn!(
                target = "database::legacy_import",
                identity = %hex::encode(id),
                "Skipping legacy identity with an out-of-range status value",
            );
            out.unreadable = out.unreadable.saturating_add(1);
            continue;
        };

        let wallet_index = match wallet_index.map(u32::try_from) {
            None => None,
            Some(Ok(index)) => Some(index),
            Some(Err(_)) => {
                tracing::warn!(
                    target = "database::legacy_import",
                    identity = %hex::encode(id),
                    "Skipping legacy identity with an out-of-range wallet index",
                );
                out.unreadable = out.unreadable.saturating_add(1);
                continue;
            }
        };

        // Both-or-neither: the legacy `CHECK` guarantees it, so a half-filled
        // link is corruption. Dropping just the link would silently orphan the
        // identity from the wallet that owns its keys, so the row is reported
        // instead — the sentinel stays open and a later build can retry it.
        let wallet = match (wallet, wallet_index) {
            (None, None) => None,
            (Some(seed_hash), Some(index)) => match <[u8; 32]>::try_from(seed_hash.as_slice()) {
                Ok(seed_hash) => Some((seed_hash, index)),
                Err(_) => {
                    tracing::warn!(
                        target = "database::legacy_import",
                        identity = %hex::encode(id),
                        "Skipping legacy identity whose wallet link is not a 32-byte seed hash",
                    );
                    out.unreadable = out.unreadable.saturating_add(1);
                    continue;
                }
            },
            _ => {
                tracing::warn!(
                    target = "database::legacy_import",
                    identity = %hex::encode(id),
                    "Skipping legacy identity with a half-filled wallet link",
                );
                out.unreadable = out.unreadable.saturating_add(1);
                continue;
            }
        };

        let Ok(mut qi) = QualifiedIdentity::from_bytes(&data) else {
            tracing::warn!(
                target = "database::legacy_import",
                identity = %hex::encode(id),
                "Skipping legacy identity whose stored data could not be decoded",
            );
            out.unreadable = out.unreadable.saturating_add(1);
            continue;
        };

        // Neither field is in the bincode blob — the legacy encoder skipped both
        // and kept them in columns. Without this, every imported identity reads
        // back as `Unknown` status on mainnet.
        qi.status = IdentityStatus::from(status);
        qi.network = network;
        // The column is a denormalised copy of the blob's alias (design.md §6);
        // only fall back to it when the blob itself carries none.
        if qi.alias.is_none() {
            qi.alias = alias;
        }

        out.identities.push(LegacyIdentityRow { id, qi, wallet });
    }

    Ok(out)
}
```

<sub>source: ['codex-general']</sub>

thepastaclaw, line 439: **🔴 Blocking: Skip-if-present check and the actual insert key can diverge on a corrupted row**

After a successful decode, nothing verifies that `qi.identity.id()` (the id embedded in the blob) matches the SQLite row's `id` column. The migration's skip-if-present check (`migrate_identities_from_conn`) calls `is_present(&Identifier::from(row.id))`, but `insert_local_qualified_identity` (src/context/identity_db.rs:522) derives the actual storage/vault key from `qualified_identity.identity.id()` — the blob's embedded id, not the id passed in. If a corrupted or hand-edited legacy row has a row `id` that differs from the id encoded in its `data` blob, the presence check can pass (the row's id isn't in the store yet) while the write silently `INSERT OR REPLACE`s a *different*, already-loaded identity under the blob's embedded id — the exact class of overwrite the design doc's skip-if-present rule was built to prevent (design.md "Row already in the k/v store" edge case). Reject the row as unreadable when the two ids disagree, before it ever reaches the skip-if-present check.

```suggestion
        let Ok(mut qi) = QualifiedIdentity::from_bytes(&data) else {
            tracing::warn!(
                target = "database::legacy_import",
                identity = %hex::encode(id),
                "Skipping legacy identity whose stored data could not be decoded",
            );
            out.unreadable = out.unreadable.saturating_add(1);
            continue;
        };

        if qi.identity.id().to_buffer() != id {
            tracing::warn!(
                target = "database::legacy_import",
                row_id = %hex::encode(id),
                blob_id = %qi.identity.id(),
                "Skipping legacy identity whose row id does not match the id encoded in its data blob",
            );
            out.unreadable = out.unreadable.saturating_add(1);
            continue;
        }

        // Neither field is in the bincode blob — the legacy encoder skipped both
        // and kept them in columns. Without this, every imported identity reads
        // back as `Unknown` status on mainnet.
        qi.status = IdentityStatus::from(status);
        qi.network = network;

        out.identities.push(LegacyIdentityRow { id, qi, wallet });
```

<sub>source: ['codex-general']</sub>

src/backend_task/migration/finish_unwire.rs
```diff
@@ -283,12 +305,61 @@ pub async fn run(app_context: &Arc<AppContext>) -> Result<bool, TaskError> {
...
+                    "App-data import failed on the same launch as the identity import; both retry on the next launch",
+                );
+            }
+            return Err(identity_error);
+        }
+    };
     let app_data = app_data?;
-    let moved_data = wallet_moved || app_data.moved_data();
+
+    let moved_data = wallet_moved || app_data.moved_data() || identities.moved_data();
```
thepastaclaw: **🔴 Blocking: An app-data hard failure still masks the identity-unreadable banner**

This very commit's stated goal is to stop a corrupt vote queue from stranding identity keys, and it does fix that: `migrate_identities` (line 323) now runs and persists its imports regardless of the app-data outcome. But `let app_data = app_data?;` (line 342) still runs *before* the `if identities.unreadable > 0 { ... }` check (line 349) that publishes `MigrationState::SucceededWithUnreadableIdentities` and the user-facing "reload this identity" banner. So when `identities` returns `Ok(outcome)` with `outcome.unreadable > 0` *and* `app_data` is a hard `Err` on the same launch (e.g. the same corrupt vote-index blob this commit's own regression test exercises), the function returns the app-data error via `?` and never reaches the identity-unreadable banner. Since the app-data failure is deterministic and recurs on every launch, and the identity sentinel is withheld while `unreadable > 0`, this specific combination permanently hides the identity-specific warning behind the generic app-data retry banner — directly contradicting the commit message's own claim that "the identity failure takes precedence when both fail." The already-imported keys are not lost (they're persisted inside `migrate_identities` before this point), but the user is never told a specific identity needs to be reloaded.

```suggestion
    let identities = match identities {
        Ok(outcome) => outcome,
        Err(identity_error) => {
            if let Err(app_data_error) = &app_data {
                tracing::warn!(
                    target = "migration::finish_unwire",
                    error = ?app_data_error,
                    "App-data import failed on the same launch as the identity import; both retry on the next launch",
                );
            }
            return Err(identity_error);
        }
    };

    // Judge the identity outcome *before* propagating an app-data error: an
    // app-data failure is often deterministic (e.g. a corrupt vote queue) and
    // recurs on every launch, so unwrapping it first would permanently hide
    // the identity-specific "some identities need to be reloaded" banner
    // behind the generic app-data retry banner.
    if identities.unreadable > 0 {
        if let Err(app_data_error) = &app_data {
            tracing::warn!(
                target = "migration::finish_unwire",
                error = ?app_data_error,
                "App-data import also failed on this launch",
            );
        }
        tracing::warn!(
            target = "migration::finish_unwire",
            unreadable = identities.unreadable,
            imported = identities.imported,
            network = ?app_context.network,
            "Some legacy identities could not be decoded; they stay in the previous version's data.db and must be loaded again",
        );
        status.set_state(MigrationState::SucceededWithUnreadableIdentities {
            count: identities.unreadable,
        });
        return Ok(wallet_moved || identities.moved_data());
    }

    let app_data = app_data?;

    let moved_data = wallet_moved || app_data.moved_data() || identities.moved_data();
```

<sub>source: ['codex-general']</sub>

thepastaclaw (changes requested):
## Code Review

Verified all four findings directly against HEAD (ce1205aa). The two data-integrity bugs in `legacy_import.rs` are real and independently confirmable: the alias-fallback gap contradicts this PR's own design doc (`docs/ai-design/2026-07-13-legacy-identity-migration/design.md:196`), and the missing blob-vs-row id check is a genuine skip-if-present bypass since `insert_local_qualified_identity` keys off the blob's embedded id while the pre-check uses the row's id column. The `finish_unwire.rs` banner-suppression bug is also confirmed live in the HEAD commit's own rewritten control flow — `app_data?` still runs before the `identities.unreadable > 0` check, so a deterministic app-data failure permanently masks the identity warning banner, undercutting the very fix this commit claims to make. The missing user-story entry is a real, low-severity documentation gap with a strong direct precedent already in the file.

Source: reviewers codex/general=gpt-5.6-sol (high, completed), sonnet5/general=claude-sonnet-5 (high, completed); verifier sonnet5=claude-sonnet-5 (high, completed); orchestrator openai/gpt-5.6-sol (high, orchestration-only, not reviewer/verifier); experiment sonnet-primary-opus-quarter-sample-20260710, cohort=sonnet_primary, bucket=1.

<details>

```

In `src/database/legacy_import.rs`:
- [BLOCKING] src/database/legacy_import.rs:277-379: Alias-column fallback required by the design doc is never implemented
  The `SELECT` at line 278 omits `alias`, and nothing in the function ever reads it, so an imported identity's alias always comes from whatever is baked into the `data` blob — even when it's `None`. This PR's own design doc (`docs/ai-design/2026-07-13-legacy-identity-migration/design.md:196`) is explicit about the required behavior: "Column is a denormalised copy. Prefer the blob; the column is a fallback if `qi.alias` is `None`." That fallback path is entirely missing from the implementation. Any legacy identity whose blob alias is `None` but whose column carries a value — a realistic v0.9.3 state, since renaming an identity plausibly touched the cheap alias column without re-serializing the whole key-bearing blob — silently imports with no alias, contradicting the documented contract. Read the nullable `alias` column and apply it as a fallback when `qi.alias.is_none()`.
- [BLOCKING] src/database/legacy_import.rs:363-379: Skip-if-present check and the actual insert key can diverge on a corrupted row
  After a successful decode, nothing verifies that `qi.identity.id()` (the id embedded in the blob) matches the SQLite row's `id` column. The migration's skip-if-present check (`migrate_identities_from_conn`) calls `is_present(&Identifier::from(row.id))`, but `insert_local_qualified_identity` (src/context/identity_db.rs:522) derives the actual storage/vault key from `qualified_identity.identity.id()` — the blob's embedded id, not the id passed in. If a corrupted or hand-edited legacy row has a row `id` that differs from the id encoded in its `data` blob, the presence check can pass (the row's id isn't in the store yet) while the write silently `INSERT OR REPLACE`s a *different*, already-loaded identity under the blob's embedded id — the exact class of overwrite the design doc's skip-if-present rule was built to prevent (design.md "Row already in the k/v store" edge case). Reject the row as unreadable when the two ids disagree, before it ever reaches the skip-if-present check.

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:329-344: An app-data hard failure still masks the identity-unreadable banner
  This very commit's stated goal is to stop a corrupt vote queue from stranding identity keys, and it does fix that: `migrate_identities` (line 323) now runs and persists its imports regardless of the app-data outcome. But `let app_data = app_data?;` (line 342) still runs *before* the `if identities.unreadable > 0 { ... }` check (line 349) that publishes `MigrationState::SucceededWithUnreadableIdentities` and the user-facing "reload this identity" banner. So when `identities` returns `Ok(outcome)` with `outcome.unreadable > 0` *and* `app_data` is a hard `Err` on the same launch (e.g. the same corrupt vote-index blob this commit's own regression test exercises), the function returns the app-data error via `?` and never reaches the identity-unreadable banner. Since the app-data failure is deterministic and recurs on every launch, and the identity sentinel is withheld while `unreadable > 0`, this specific combination permanently hides the identity-specific warning behind the generic app-data retry banner — directly contradicting the commit message's own claim that "the identity failure takes precedence when both fail." The already-imported keys are not lost (they're persisted inside `migrate_identities` before this point), but the user is never told a specific identity needs to be reloaded.

In `docs/user-stories.md`:
- [SUGGESTION] docs/user-stories.md:675-683: No user story added for identities/keys preserved across an app upgrade
  This PR introduces a significant user-facing behavior change — legacy v0.9.3 identities (including a masternode's owner/voting keys) are now automatically imported on first launch after upgrade, with a dedicated progress step (`MigrationStep::Identities`) and a warning banner for undecodable rows. CLAUDE.md requires: "When a PR adds or significantly changes user-facing features, check `docs/user-stories.md` ... add one." There's a direct precedent already in the file for this exact shape of feature: `DPN-009: Scheduled votes preserved across an app upgrade [Implemented]` (line 676) and `NET-021: App settings preserved across an app upgrade [Implemented]` (line 1258). No equivalent `IDN-0xx` story exists for identities/keys surviving an upgrade — confirmed by scanning all `IDN-` entries in the file.
```
</details>

src/database/legacy_import.rs
lklimek: Verified against current HEAD (`e2c0970a`): still open. `read_identities`'s `SELECT` (src/database/legacy_import.rs:296-298) still omits `alias`, and there is no fallback assignment when `qi.alias` is `None`. Needs the fix as suggested.

lklimek: Verified against current HEAD (`e2c0970a`): still open. `read_identities` (src/database/legacy_import.rs:382-398) still pushes `LegacyIdentityRow { id, qi, wallet }` without checking `qi.identity.id()` against the row's `id` column, so the skip-if-present precheck and the actual `insert_local_qualified_identity` write key can still diverge on a corrupted row. Needs the fix as suggested.

src/backend_task/migration/finish_unwire.rs
lklimek: Verified against current HEAD (`e2c0970a`): still open. The QA-001 fix reordered the *passes* (identity import now runs before either DET-owned result is unwrapped), but the *outcome check order* is unchanged: `let app_data = app_data?;` (src/backend_task/migration/finish_unwire.rs:367) still runs before the `if identities.unreadable > 0` check (line 374). If `identities` is `Ok` with `unreadable > 0` and `app_data` is a hard `Err` on the same launch, line 367 still short-circuits via `?` and the identity-unreadable banner is never reached. Needs the reorder as suggested (check `identities.unreadable` before unwrapping `app_data`).

src/database/legacy_import.rs
lklimek: Fixed in `b56ebffd`. `read_identities`'s `SELECT` now includes `alias`, applied as a fallback only when `qi.alias.is_none()` — the blob still always wins. Regression test added (`identities_alias_falls_back_to_column_only_when_blob_has_none`), confirmed red against the prior code before the fix.

lklimek: Fixed in `b56ebffd`. `read_identities` now compares `qi.identity.id()` against the row's `id` column right after decode and rejects (counts as unreadable, skips) any row where they disagree. Regression test added (`identities_skip_rows_whose_row_id_and_blob_id_disagree`), confirmed red against the prior code before the fix.

src/backend_task/migration/finish_unwire.rs
lklimek: Fixed in `b56ebffd`. `finish_unwire::run` now checks `identities.unreadable > 0` before unwrapping `app_data` with `?`, so a hard app-data failure on the same launch can no longer mask the identity-unreadable banner. The `app_data` error is logged in that branch and `moved_data` still accounts for `wallet_moved`/`identities.moved_data()` either way. Existing `finish_unwire` orchestration tests still green (108/108 across the relevant test filter).

src/database/legacy_import.rs
```diff
@@ -241,6 +275,160 @@ pub(crate) fn read_scheduled_votes(
...
+        // back as `Unknown` status on mainnet.
+        qi.status = IdentityStatus::from(status);
+        qi.network = network;
+
+        // The blob is the source of truth for the alias; the column is a
+        // denormalised copy. Fall back to it only when the blob carries none, so
+        // a row that has both always keeps the blob's value.
+        if qi.alias.is_none() {
+            qi.alias = alias;
+        }
```
thepastaclaw: **🔴 Blocking: Treat the legacy alias column as authoritative**

The new fallback still loses legitimate v0.9.3 alias edits. In v0.9.3, `set_identity_alias` updated only the SQL `alias` column, while every identity loader decoded the blob and then unconditionally assigned the column value. A rename or alias removal therefore legitimately leaves a stale `Some` value in the blob. Keeping the blob whenever it is populated resurrects the old alias or reverses the user's removal during upgrade. The resolved thread confirms that the current implementation deliberately makes the blob win, but the v0.9.3 setter and loaders establish that the column is authoritative.

```suggestion
        qi.alias = alias;
```

src/backend_task/migration/finish_unwire.rs
```diff
@@ -308,12 +330,82 @@ pub async fn run(app_context: &Arc<AppContext>) -> Result<bool, TaskError> {
...
+            unreadable = identities.unreadable,
+            imported = identities.imported,
+            votes_unreadable,
+            network = ?app_context.network,
+            "Some legacy identities could not be decoded; they stay in the previous version's data.db and must be loaded again",
+        );
+        status.set_state(MigrationState::SucceededWithUnreadableIdentities {
+            count: identities.unreadable,
+        });
+        return Ok(moved_data);
```
thepastaclaw: **🔴 Blocking: Do not report simultaneous app-data and identity failures as success**

When an identity row is unreadable and `migrate_app_data` returns a hard error, this branch logs the app-data error, publishes only `SucceededWithUnreadableIdentities`, and returns `Ok`. Consequently `run_migration_task` never publishes `MigrationState::Failed`, so the user receives no app-data failure or retry action. Both sentinels remain absent, and the unreadable legacy identity is counted again before the modern-store presence check on every launch, allowing the same branch to hide a deterministic vote or top-up failure indefinitely. Preserve the identity-specific warning while also surfacing the hard app-data failure through a structured combined outcome.

src/database/legacy_import.rs
```diff
@@ -241,6 +275,160 @@ pub(crate) fn read_scheduled_votes(
...
+        // that does not fit. Widen the read and convert explicitly: a narrow
+        // `row.get::<u8>` would raise `IntegralValueOutOfRange` through `?` and
+        // take the whole identity read down — every other row-level corruption
+        // here is counted and skipped, and this one must behave the same.
+        let status: i64 = row.get(2)?;
+        let wallet: Option<Vec<u8>> = row.get(3)?;
+        let wallet_index: Option<i64> = row.get(4)?;
+        // Denormalised copy of the blob's own alias. Read as a fallback only —
+        // the blob wins whenever it carries one (see below).
+        let alias: Option<String> = row.get(5)?;
```
thepastaclaw: **🔴 Blocking: Count malformed identity column types per row**

The identity reader applies `?` to every `row.get` before reaching its per-row corruption handling; the latest delta adds the same failure mode for `alias`. SQLite's legacy table is not strict, so a wrong storage class in `alias`, `id`, `data`, `status`, `wallet`, or `wallet_index` produces `InvalidColumnType` and exits `read_identities`, discarding all readable rows accumulated in the pass. No identity then reaches the vault writer, and the hard failure repeats on retry. Decode these columns through a per-row `Result` and increment `unreadable`, matching `read_scheduled_votes` and the PR's stated row-isolation policy.

thepastaclaw (changes requested):

Carried-forward prior findings: c9ab4490eb43 remains valid because the current blob-first alias fallback still loses renamed or cleared aliases under v0.9.3's column-authoritative behavior; prior-4-user-story also remains open. New findings in latest delta: simultaneous unreadable identities and hard app-data errors are reported as success, and malformed SQLite column types can abort the entire identity import. The ID-mismatch and original identity-banner masking findings are fixed; the exact HEAD and targeted tests were verified.

### Review provenance

<details>

```

In `src/database/legacy_import.rs`:
- [BLOCKING] src/database/legacy_import.rs:419-424: Treat the legacy alias column as authoritative
  The new fallback still loses legitimate v0.9.3 alias edits. In v0.9.3, `set_identity_alias` updated only the SQL `alias` column, while every identity loader decoded the blob and then unconditionally assigned the column value. A rename or alias removal therefore legitimately leaves a stale `Some` value in the blob. Keeping the blob whenever it is populated resurrects the old alias or reverses the user's removal during upgrade. The resolved thread confirms that the current implementation deliberately makes the blob win, but the v0.9.3 setter and loaders establish that the column is authoritative.
- [BLOCKING] src/database/legacy_import.rs:307-321: Count malformed identity column types per row
  The identity reader applies `?` to every `row.get` before reaching its per-row corruption handling; the latest delta adds the same failure mode for `alias`. SQLite's legacy table is not strict, so a wrong storage class in `alias`, `id`, `data`, `status`, `wallet`, or `wallet_index` produces `InvalidColumnType` and exits `read_identities`, discarding all readable rows accumulated in the pass. No identity then reaches the vault writer, and the hard failure repeats on retry. Decode these columns through a per-row `Result` and increment `unreadable`, matching `read_scheduled_votes` and the PR's stated row-isolation policy.

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:378-402: Do not report simultaneous app-data and identity failures as success
  When an identity row is unreadable and `migrate_app_data` returns a hard error, this branch logs the app-data error, publishes only `SucceededWithUnreadableIdentities`, and returns `Ok`. Consequently `run_migration_task` never publishes `MigrationState::Failed`, so the user receives no app-data failure or retry action. Both sentinels remain absent, and the unreadable legacy identity is counted again before the modern-store presence check on every launch, allowing the same branch to hide a deterministic vote or top-up failure indefinitely. Preserve the identity-specific warning while also surfacing the hard app-data failure through a structured combined outcome.

In `docs/user-stories.md`:
- [SUGGESTION] docs/user-stories.md:614-624: Add the identity-and-key upgrade preservation user story
  The Identity Operations catalog still ends at IDN-015 without a story for preserving legacy identities and their keys across an upgrade. This PR adds significant user-facing migration behavior, including progress and unreadable-row recovery states, so CLAUDE.md requires a matching story. DPN-009 and NET-021 document the analogous vote and settings migrations.
```
</details>

src/database/legacy_import.rs
lklimek: Good catch — and you were right, so I went and checked the actual `v0.9.3` tag rather than trust either of us. `set_identity_alias` (`src/database/identities.rs`) ran `UPDATE identity SET alias = ? WHERE id = ?` — column only, blob untouched. And all three loaders (`get_local_qualified_identities`, `..._in_wallets`, `get_identity_by_id`) did `let mut identity = QualifiedIdentity::from_bytes(&data); identity.alias = alias;` — an *unconditional* overwrite with the column value. So the column was authoritative and the blob's copy was purely decorative; a rename/removal left the blob stale.

Fixed: the import now assigns the column unconditionally (`qi.alias = alias;`), including a NULL column clearing a stale blob alias. The design doc (`design.md:196`) had it backwards — corrected with the v0.9.3 evidence. The alias test now pins column-always-wins across all three cases (override, fill-empty, NULL-clears). Commit 5f5614c5.

src/backend_task/migration/finish_unwire.rs
lklimek: Fixed. When unreadable identities and a hard app-data failure land on the same launch, the run no longer publishes only `SucceededWithUnreadableIdentities` (which swallowed the app-data failure, gave no retry, and recurred silently every launch). Added `MigrationState::FailedWithUnreadableIdentities { count, error }` — a single retryable Error banner that names *both* problems: reload the identities AND retry the app-data update. The typed error chain is carried through (`app_data_error_chain` reuses a `MigrationFailed` source verbatim, wraps any stray non-migration `TaskError` in a new `AppDataImport` variant so nothing is dropped and `is_backend_not_ready` stays reachable).

Funds stay safe (the drain still runs regardless), and neither DET-owned sentinel is written, so both passes retry next launch. New end-to-end test `both_app_data_and_identity_failures_surface_together` stages exactly this collision and asserts both signals surface. Commit 5f5614c5.

src/database/legacy_import.rs
lklimek: Fixed. The six column reads now go through a per-row `decode_identity_columns` helper (mirroring `decode_scheduled_vote_columns`/`decode_top_up_columns` in the same file), so a wrong SQLite storage class on any of id/data/status/wallet/wallet_index/alias increments `unreadable` and `continue`s to the next row instead of `?`-escaping `read_identities` and discarding every row already accumulated in the batch.

Regression test `identities_skip_malformed_column_types_without_discarding_read_rows` stages an INTEGER-in-`data` row and a BLOB-in-`alias` row after a good row and asserts the good one survives (`unreadable == 2`). Verified RED against the pre-fix code — it panicked with `InvalidColumnType(1, "data", Integer)`, exactly the whole-batch discard — then green after. Commit 5f5614c5.

src/backend_task/migration/finish_unwire.rs
```diff
@@ -308,12 +362,95 @@ pub async fn run(app_context: &Arc<AppContext>) -> Result<bool, TaskError> {
...
+                    unreadable = identities.unreadable,
+                    imported = identities.imported,
+                    votes_unreadable = outcome.votes_unreadable,
+                    network = ?app_context.network,
+                    "Some legacy identities could not be decoded; they stay in the previous version's data.db and must be loaded again",
+                );
+                status.set_state(MigrationState::SucceededWithUnreadableIdentities {
+                    count: identities.unreadable,
+                });
+                return Ok(moved_data);
```
thepastaclaw: **🔴 Blocking: Do not let unreadable identities permanently hide unreadable votes**

When both passes finish with unreadable rows, `migrate_app_data` has already persisted the unreadable-vote warning and its completion sentinel. This branch then publishes only `SucceededWithUnreadableIdentities` and returns before `read_vote_warning` at lines 459-469. On later launches, app-data short-circuits on its sentinel while the corrupt identity is counted again because its sentinel remains absent, so the same early return repeats. Manually loading the identity does not resolve this: unreadable rows are counted before the modern-store presence check. The user therefore never sees the vote warning or rescheduling action and can miss an active vote window. Surface both durable recovery messages in a combined state or otherwise ensure the stored vote warning remains reachable.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest delta fixes all three prior blocking code findings, and the targeted regression tests pass. One blocking lifecycle bug remains: a persistent unreadable identity prevents the durable unreadable-vote warning and recovery action from ever being shown. The required identity-upgrade user story also remains missing.

### Review provenance

<details>

```

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:414-429: Do not let unreadable identities permanently hide unreadable votes
  When both passes finish with unreadable rows, `migrate_app_data` has already persisted the unreadable-vote warning and its completion sentinel. This branch then publishes only `SucceededWithUnreadableIdentities` and returns before `read_vote_warning` at lines 459-469. On later launches, app-data short-circuits on its sentinel while the corrupt identity is counted again because its sentinel remains absent, so the same early return repeats. Manually loading the identity does not resolve this: unreadable rows are counted before the modern-store presence check. The user therefore never sees the vote warning or rescheduling action and can miss an active vote window. Surface both durable recovery messages in a combined state or otherwise ensure the stored vote warning remains reachable.

In `docs/user-stories.md`:
- [SUGGESTION] docs/user-stories.md:614-624: Add the identity-and-key upgrade preservation user story
  The Identity Operations catalog still ends at IDN-015 without a story covering preservation of legacy identities and their keys across an upgrade. This PR introduces substantial user-facing migration and recovery behavior, so CLAUDE.md requires a matching story. DPN-009 and NET-021 document the analogous scheduled-vote and settings migrations.
```
</details>

src/backend_task/migration/finish_unwire.rs
lklimek: Fixed in `4587f254`. Confirmed the finding is real: with an unreadable identity present, `identities.unreadable > 0` recurs on every launch (its sentinel stays unwritten), and that branch returned right after publishing `SucceededWithUnreadableIdentities` — *before* the durable `read_vote_warning` re-publish. Meanwhile the app-data pass short-circuits on its own sentinel from the second launch on and honestly reports zero unreadable votes, so the count could not be rescued from `outcome` counters either. Net effect exactly as described: a user with one corrupt identity row and one corrupt vote row was never told about the vote on any launch, and could miss a live deadline.

Fix: the identity branch now reads the pending vote warning from durable storage (not from `outcome`, which goes quiet once the app-data sentinel is written) and publishes both counts on one terminal state, `MigrationState::SucceededWithUnreadableIdentitiesAndVotes { identities, votes }`, rendered as a single sticky Warning banner naming both remedies (load the identities again; re-schedule the votes on the Scheduled Votes screen). Acknowledging retires only the durable vote half; the identity half then keeps arriving as plain `SucceededWithUnreadableIdentities` until a build with a fixed decoder imports the rows. A k/v read that itself fails is surfaced as the retryable `FailedWithUnreadableIdentities` rather than dropping either signal.

Also added the CLAUDE.md-required user story `IDN-016` (identities and their keys preserved across an app upgrade).

Regression test `unreadable_identities_do_not_hide_the_unreadable_vote_warning` stages exactly this collision across three launches: discovery run surfaces both counts, a later launch proves the vote half survives from the durable record while the app-data sentinel is already written and the identity sentinel is not, and acknowledging retires only the vote half.

Verified under an **isolated** `CARGO_TARGET_DIR` (fresh 2m11s compile, not a cached replay): 52/52 `finish_unwire` lib tests pass, and `cargo clippy --all-features --all-targets -- -D warnings` is clean (exit 0, zero warnings).

src/backend_task/migration/finish_unwire.rs
```diff
@@ -773,6 +962,187 @@ where
...
+    };
+
+    for row in legacy.identities {
+        let id = Identifier::from(row.id);
+
+        // INSERT-OR-REPLACE below, so this check is what keeps a retry from
+        // overwriting an identity the user has edited since it was imported.
+        if is_present(&id).map_err(import_failed)? {
+            outcome.skipped_existing = outcome.skipped_existing.saturating_add(1);
+            continue;
```
thepastaclaw: **🔴 Blocking: Merge legacy keys into partially loaded existing identities**

Presence in the modern store does not prove that every legacy key was preserved. This PR explicitly uses a separate identity sentinel so installations that already ran the wallet drain under a build without this importer still migrate their identities. Before this PR, `MasternodeLoadForm` allowed loading an identity from only its ProTxHash because voting, owner, and payout keys were all optional; `load_identity` then persisted a valid identity with an empty or partial key map. When that reachable installation upgrades, this branch skips the legacy row wholesale even if its blob contains owner, voting, or payout keys missing from the modern record. The skipped row is not counted as unreadable, so `migrate_identities` writes the completion sentinel and those legacy-only keys never reach the vault or get retried. Reconcile missing legacy keys and associations into the existing identity while preserving modern aliases, protection state, and keys instead of treating blob presence as a complete migration.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All five carried-forward findings are fixed at the current HEAD, and the latest delta introduces no new finding. The cumulative PR still has one blocking identity-key preservation defect: an existing modern identity is skipped even when legacy storage contains keys absent from that identity. At exact HEAD 4587f2543135f25bb9a09f4705616f011018e882, the library suite passed with 1673 tests passed and 1 ignored.

### Review provenance

🔴 1 blocking

<details>

```

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1099-1103: Merge legacy keys into partially loaded existing identities
  Presence in the modern store does not prove that every legacy key was preserved. This PR explicitly uses a separate identity sentinel so installations that already ran the wallet drain under a build without this importer still migrate their identities. Before this PR, `MasternodeLoadForm` allowed loading an identity from only its ProTxHash because voting, owner, and payout keys were all optional; `load_identity` then persisted a valid identity with an empty or partial key map. When that reachable installation upgrades, this branch skips the legacy row wholesale even if its blob contains owner, voting, or payout keys missing from the modern record. The skipped row is not counted as unreadable, so `migrate_identities` writes the completion sentinel and those legacy-only keys never reach the vault or get retried. Reconcile missing legacy keys and associations into the existing identity while preserving modern aliases, protection state, and keys instead of treating blob presence as a complete migration.
```
</details>

src/backend_task/migration/finish_unwire.rs
lklimek: Confirmed and fixed in e8b61821. `migrate_identities_from_conn`'s skip-if-present path previously dropped legacy-only owner/voting/payout keys for partially-loaded identities without ever retrying them. Promoted the existing gap-merge pattern (`load_identity.rs`'s `merge_existing_keys_into`) to a shared `QualifiedIdentity::merge_gaps_from` in `model/`, and now reuse it here: the migration fetches the existing modern identity and merges in only the legacy keys/associations it's missing — modern data always wins, re-persisted only when something actually changed (new `reconciled` counter). New regression test `a_present_but_partial_identity_gains_the_legacy_only_keys` covers the fix; full lib suite (1674/1674) and clippy -D warnings both clean.

src/backend_task/migration/finish_unwire.rs
```diff
@@ -773,6 +962,212 @@ where
...
+        // Gap-merge them in (modern always wins) instead of skipping wholesale,
+        // and re-persist only when something changed so a retry never clobbers a
+        // user edit with the stale legacy copy (design.md §7).
+        if let Some(mut modern) = get_existing(&id).map_err(import_failed)? {
+            let before = modern.clone();
+            modern.merge_gaps_from(row.qi);
+            if modern == before {
+                outcome.skipped_existing = outcome.skipped_existing.saturating_add(1);
+            } else {
+                update(&modern).map_err(import_failed)?;
```
thepastaclaw: **🔴 Blocking: Handle protected identities when reconciling legacy keys**

When the legacy row adds a `Clear` or `AlwaysClear` key to an existing identity whose other keys are password-protected, this merge creates a mixed protected/plaintext record. `update_local_qualified_identity` passes that record to `encode_identity_blob_vault_first`, whose protection guard returns `IdentityKeyProtectionDowngrade`. The migration exits before writing its sentinel, saves none of the recovered keys, and repeats the same failure on every launch. The interactive identity-load merge handles this state by verifying the identity password and sealing newly merged keys before persistence; this migration needs an equivalent protected-key path or must defer the reconciliation without losing its ability to retry.

<sub>source: ['codex']</sub>

src/model/qualified_identity/mod.rs
```diff
@@ -506,6 +506,35 @@ impl Signer<IdentityPublicKey> for QualifiedIdentity {
...
+    /// (e.g. adding a voting key without erasing the Owner/Payout keys), and the
+    /// legacy-identity migration reconciling an already-present modern identity
+    /// with keys still held only in the previous version's blob. In both cases
+    /// `self` is the authoritative record and `other` is the supplement.
+    pub fn merge_gaps_from(&mut self, other: QualifiedIdentity) {
+        for (key, value) in other.private_keys.private_keys {
+            self.private_keys.private_keys.entry(key).or_insert(value);
+        }
+        if self.alias.is_none() {
+            self.alias = other.alias;
```
thepastaclaw: **🔴 Blocking: Do not infer migration gaps from field absence**

For migration purposes, a missing value does not prove that the modern identity was only partially loaded. Clearing an alias deliberately persists `None`, and “Remove private key from DET” deliberately removes the corresponding map entry and persists the result. On an installation without the new identity sentinel, or on any retry caused by another unreadable legacy row, this merge fills those intentional absences from the stale legacy blob. That restores a removed alias or makes a private key available in DET again, violating the PR's modern-wins and retry-preserves-edits guarantees. Restrict reconciliation to records known to be partial or persist enough provenance or tombstones to distinguish missing legacy data from intentional removal.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior finding: STILL VALID in the narrower password-protected identity case, where recovered legacy keys cannot be persisted. New latest-delta finding: the absence-based merge can restore aliases and private keys that users explicitly removed from the modern record. Focused tests confirm the individual merge, protection guard, and alias-removal behaviors, but do not cover these failing combinations.

### Review provenance

🔴 2 blocking

<details>

```

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1113-1117: Handle protected identities when reconciling legacy keys
  When the legacy row adds a `Clear` or `AlwaysClear` key to an existing identity whose other keys are password-protected, this merge creates a mixed protected/plaintext record. `update_local_qualified_identity` passes that record to `encode_identity_blob_vault_first`, whose protection guard returns `IdentityKeyProtectionDowngrade`. The migration exits before writing its sentinel, saves none of the recovered keys, and repeats the same failure on every launch. The interactive identity-load merge handles this state by verifying the identity password and sealing newly merged keys before persistence; this migration needs an equivalent protected-key path or must defer the reconciliation without losing its ability to retry.

In `src/model/qualified_identity/mod.rs`:
- [BLOCKING] src/model/qualified_identity/mod.rs:520-525: Do not infer migration gaps from field absence
  For migration purposes, a missing value does not prove that the modern identity was only partially loaded. Clearing an alias deliberately persists `None`, and “Remove private key from DET” deliberately removes the corresponding map entry and persists the result. On an installation without the new identity sentinel, or on any retry caused by another unreadable legacy row, this merge fills those intentional absences from the stale legacy blob. That restores a removed alias or makes a private key available in DET again, violating the PR's modern-wins and retry-preserves-edits guarantees. Restrict reconciliation to records known to be partial or persist enough provenance or tombstones to distinguish missing legacy data from intentional removal.
```
</details>

src/backend_task/migration/finish_unwire.rs
lklimek: Confirmed real and fixed in `b454dba5`. Verified against the code: `encode_identity_blob_vault_first` (identity_db.rs:414-415) returns `IdentityKeyProtectionDowngrade` when an identity has resident plaintext to vault AND any existing key is Tier-2 protected — so merging a legacy `Clear` key into a protected identity made `update_local_qualified_identity` fail, the pass errored before its sentinel, and it recurred every launch. Migration has no identity password, so it genuinely cannot seal a merged key the way the interactive load does.

Fix: reconciliation now runs **only when the modern record holds no private keys at all** (a bare ProTxHash-only load). A protected identity always holds keys, so it never enters the reconcile branch and can never reach the downgrade guard. A keyed-but-partial or protected identity that is genuinely missing a key is instead recovered through the interactive load, which has the password to seal it. This defers without a hard failure (the sentinel still writes for the unrelated rows; nothing loops).

Regression test `a_present_keyed_identity_is_left_untouched_never_reconciled` asserts a keyed identity is skipped (`update` never called). Verified under an isolated `CARGO_TARGET_DIR`: full lib suite 1675/1675, clippy `-D warnings` clean.

src/model/qualified_identity/mod.rs
lklimek: Confirmed real and fixed in `b454dba5`. Verified the premise against the code: "Remove private key from DET" (key_info_screen.rs:485) deletes the map entry, and clearing an alias persists `None` — so field absence genuinely does not prove a partial load, and the gap-merge would have refilled those intentional absences from the stale legacy blob on a pre-sentinel install or a retry, resurrecting a removed alias or a deliberately-removed private key.

Fix: reconciliation is now restricted to records **known to be partial** in the one unambiguous way — the modern private-keys map is *empty* (the ProTxHash-only bare load). For a bare record we take the legacy key set and fill only the missing masternode role associations; alias is never merged in migration at all. Any record that already holds keys is left untouched, so an absent key/alias there is never refilled — removals can't be resurrected.

I also reverted the shared `QualifiedIdentity::merge_gaps_from` extraction this finding was filed against: the gap-merge (fill-from-absence) is safe only with a user actively re-loading, so it goes back to the private `merge_existing_keys_into` in `load_identity` and is no longer reachable from the migration. Migration carries its own bare-record reconcile.

Regression test `a_present_keyed_identity_is_left_untouched_never_reconciled` stages a keyed modern record with a cleared alias + a legacy blob carrying an alias/key/association, and asserts nothing is re-persisted. Isolated verify: full lib suite 1675/1675, clippy `-D warnings` clean.

src/backend_task/migration/finish_unwire.rs
```diff
@@ -773,6 +962,234 @@ where
...
+                    modern.associated_voter_identity = row.qi.associated_voter_identity;
+                }
+                if modern.associated_operator_identity.is_none() {
+                    modern.associated_operator_identity = row.qi.associated_operator_identity;
+                }
+                if modern == before {
+                    outcome.skipped_existing = outcome.skipped_existing.saturating_add(1);
+                } else {
+                    update(&modern).map_err(import_failed)?;
+                    outcome.reconciled = outcome.reconciled.saturating_add(1);
```
thepastaclaw: **🔴 Blocking: Do not treat an empty key map as migration provenance**

An empty map does not uniquely identify a ProTxHash-only load. The key-removal flow can persist an identity after deleting its final resident key, without recording a tombstone. An overwrite can also replace a protected identity with an empty record without deleting its existing protected vault labels. On a pre-sentinel installation or a retry held open by another unreadable row, this branch restores the stale legacy key set. An unprotected identity regains keys deliberately removed from DET; when a restored key matches a retained protected label, `update_local_qualified_identity` instead returns `IdentityKeyProtectionDowngrade`, aborting before the sentinel and repeating on every launch. Use persisted provenance, removal tombstones, or an explicit recovery decision instead of map emptiness.

<sub>source: ['codex']</sub>

src/backend_task/migration/finish_unwire.rs
```diff
@@ -773,6 +962,234 @@ where
...
+                }
+            } else {
+                outcome.skipped_existing = outcome.skipped_existing.saturating_add(1);
```
thepastaclaw: **🔴 Blocking: Do not finalize migration while legacy-only keys remain on keyed identities**

The masternode load form accepts voting, owner, and payout keys independently, so a valid modern record can contain one key while its legacy blob contains additional keys. This branch skips every such record, and `migrate_identities` writes the completion sentinel whenever no row was unreadable, permanently leaving the legacy-only material outside the vault. The normal load form cannot recover it because `RejectIfExists` rejects the stored ProTxHash, while the scoped merge supplies only a voting key. The per-key screen requires the user to possess and re-enter the missing WIF and cannot recover it from `data.db`; adding a missing key directly also fails for a protected identity because it introduces plaintext without the identity password. Keep completion pending or obtain an explicit recovery decision while legacy-only owner, voting, or payout keys remain.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

This head prevents ordinary keyed protected identities from entering reconciliation, but two blocking migration defects remain. Empty key maps can still represent intentional removal or coexist with retained protected vault labels, while keyed-but-partial identities are finalized without recovering legacy-only keys. Full cargo validation was not rerun because only 706 MiB of disk space remained; `git diff --check` passed.

### Review provenance

🔴 2 blocking

<details>

```

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1118-1136: Do not treat an empty key map as migration provenance
  An empty map does not uniquely identify a ProTxHash-only load. The key-removal flow can persist an identity after deleting its final resident key, without recording a tombstone. An overwrite can also replace a protected identity with an empty record without deleting its existing protected vault labels. On a pre-sentinel installation or a retry held open by another unreadable row, this branch restores the stale legacy key set. An unprotected identity regains keys deliberately removed from DET; when a restored key matches a retained protected label, `update_local_qualified_identity` instead returns `IdentityKeyProtectionDowngrade`, aborting before the sentinel and repeating on every launch. Use persisted provenance, removal tombstones, or an explicit recovery decision instead of map emptiness.
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1138-1139: Do not finalize migration while legacy-only keys remain on keyed identities
  The masternode load form accepts voting, owner, and payout keys independently, so a valid modern record can contain one key while its legacy blob contains additional keys. This branch skips every such record, and `migrate_identities` writes the completion sentinel whenever no row was unreadable, permanently leaving the legacy-only material outside the vault. The normal load form cannot recover it because `RejectIfExists` rejects the stored ProTxHash, while the scoped merge supplies only a voting key. The per-key screen requires the user to possess and re-enter the missing WIF and cannot recover it from `data.db`; adding a missing key directly also fails for a protected identity because it introduces plaintext without the identity password. Keep completion pending or obtain an explicit recovery decision while legacy-only owner, voting, or payout keys remain.
```
</details>

src/backend_task/migration/finish_unwire.rs
lklimek: Spot on — map emptiness is not provenance, and I'm not going to pretend otherwise. The empty-map reconcile is gone entirely as of 08242ecd. `migrate_identities_from_conn` is back to plain skip-if-present: an identity already in the store is skipped wholesale and never re-persisted, so no branch reads the stale legacy key set and the `IdentityKeyProtectionDowngrade` path can't fire on a retry. The concern this thread raised is designed out rather than guarded around.

src/backend_task/migration/finish_unwire.rs
lklimek: You've correctly identified the real trade-off: skip-if-present writes the sentinel while legacy-only keys can remain stranded on a partially-loaded record (the voting-only masternode case). After three heuristic attempts — unconditional merge, empty-map gate, bare-only reconcile — each got holed here for the same root reason: field absence can't be told apart from a deliberate removal without provenance the model doesn't carry, and merging plaintext into a protected identity trips the downgrade guard. Keeping completion pending instead would re-fire an unactionable warning on every launch.

So this is a scoping decision, not a dismissal (made with maintainer sign-off): revert to safe skip-if-present and accept the stranding as a **documented known limitation** rather than ship a fourth unsound heuristic. Crucially, no data is lost — the legacy `data.db` is preserved verbatim, so those keys stay fully recoverable. Documented in the design doc §7 ("Known limitation"), and the proper provenance-aware, opt-in recovery flow is tracked as a follow-up: #889. Reverted in 08242ecd.

src/backend_task/migration/finish_unwire.rs
```diff
@@ -773,6 +1125,214 @@ where
...
+        // `MigrationState::Running` — which spans this whole pass.
+        if is_present(&id).map_err(import_failed)? {
+            outcome.skipped_existing = outcome.skipped_existing.saturating_add(1);
+            continue;
+        }
+
+        // The wallet link travels verbatim, never nulled — even when that wallet
+        // did not come across: it is what re-attaches the identity when the
+        // wallet is restored or unlocked later.
+        insert(&row.qi, &row.wallet).map_err(import_failed)?;
```
thepastaclaw, line 1308: **🔴 Blocking: Preserve deletions across retries after partial hard failures**

If a later insert or sentinel-related write fails after earlier identities commit, deleting one before Retry lets the sentinel-less retry restore its stale alias and keys.

<sub>source: ['codex']</sub>

thepastaclaw, line 1308: **🔴 Blocking: Serialize the identity presence check with the replacement write**

Previously started tasks and direct UI writers can interleave between the presence check and INSERT-OR-REPLACE, resurrecting deletions or overwriting modern identity state.

<sub>source: ['codex']</sub>

src/backend_task/migration/mod.rs
```diff
@@ -49,38 +59,52 @@ impl AppContext {
...
                 Ok(BackendTaskSuccessResult::Refresh)
             }
+            MigrationTask::AcknowledgeUnreadableIdentities => {
+                finish_unwire::acknowledge_unreadable_identities(self)?;
+                Ok(BackendTaskSuccessResult::Refresh)
+            }
+            MigrationTask::AcknowledgeUnreadableIdentitiesAndVotes => {
+                finish_unwire::acknowledge_unreadable_votes(self)?;
+                finish_unwire::acknowledge_unreadable_identities(self)?;
+                Ok(BackendTaskSuccessResult::Refresh)
```
thepastaclaw, line 106: **🟡 Suggestion: Publish acknowledgement failures instead of suppressing them**

Acknowledgement errors are suppressed without publishing failure state, and combined acknowledgement can delete the vote warning before identity-warning deletion fails.

<sub>source: ['codex']</sub>

docs/ai-design/2026-07-13-legacy-identity-migration/design.md
```diff
@@ -0,0 +1,547 @@
...
+| **`data IS NULL` or `is_local = 0`** | Skip silently, do not count as failure. | Not user identity data; v0.9.3's own readers ignore these rows. |
+| **`data` fails to decode** | Count `unreadable`, log at `warn` with the identity id, **withhold the sentinel**, publish a terminal warning state. Do **not** fail the pass. | Diverges from the scheduled-vote policy (which writes the sentinel anyway) on purpose: a corrupt vote row is unrecoverable, but an undecodable identity blob may be a *decoder* defect (bincode drift, §4.1) that a later build fixes. Withholding the sentinel keeps the retry door open at the cost of one cheap re-attempt per launch; the skip-if-present rule makes the retry a no-op for everything that already landed. Failing the pass instead would be wrong — it would gate nothing useful and shout at a user who cannot act. |
+| **`id` not 32 bytes / `wallet` not 32 bytes** | Count `unreadable`, same as above. | Corruption. |
```
thepastaclaw, line 264: **🟡 Suggestion: Update the shipped design to the durable-warning sentinel contract**

The shipped design still promises sentinel withholding and automatic retry although current code writes the sentinel and preserves an acknowledgeable durable warning.

<sub>source: ['codex']</sub>

src/backend_task/error.rs
```diff
@@ -37,8 +37,13 @@ pub enum TaskError {
...
     /// particular wallet is not yet registered with it (still loading, or
     /// skipped during load). User-actionable — waiting and retrying resolves it.
-    #[error("This wallet is still loading. Please wait a moment and try again.")]
-    WalletNotLoaded,
+    #[error("The wallet \"{wallet_label}\" is still loading. Please wait a moment and try again.")]
+    WalletNotLoaded {
+        /// Display alias for the affected wallet, or a fallback hex prefix of
+        /// the seed hash when no alias has been set. With several wallets
+        /// loaded, this is the only thing that says *which* one to wait for.
+        wallet_label: String,
```
thepastaclaw, line 45: **🟡 Suggestion: Keep the wallet identity typed inside TaskError**

WalletNotLoaded stores a preformatted String despite the repository contract requiring typed TaskError data formatted through Display.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At head3326548a181a5445ea04b32a08f6b04c39fc2858, two blocking identity-import defects and three suggestions remain;12 priors are fixed, two intentionally deferred, and one outdated.

### Review provenance

🔴 2 blocking | 🟡 3 suggestion(s)

<details>

```

In `src/backend_task/migration/finish_unwire.rs`:
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1294-1308: Preserve deletions across retries after partial hard failures
  If a later insert or sentinel-related write fails after earlier identities commit, deleting one before Retry lets the sentinel-less retry restore its stale alias and keys.
- [BLOCKING] src/backend_task/migration/finish_unwire.rs:1294-1308: Serialize the identity presence check with the replacement write
  Previously started tasks and direct UI writers can interleave between the presence check and INSERT-OR-REPLACE, resurrecting deletions or overwriting modern identity state.

In `src/backend_task/migration/mod.rs`:
- [SUGGESTION] src/backend_task/migration/mod.rs:95-106: Publish acknowledgement failures instead of suppressing them
  Acknowledgement errors are suppressed without publishing failure state, and combined acknowledgement can delete the vote warning before identity-warning deletion fails.

In `docs/ai-design/2026-07-13-legacy-identity-migration/design.md`:
- [SUGGESTION] docs/ai-design/2026-07-13-legacy-identity-migration/design.md:263-264: Update the shipped design to the durable-warning sentinel contract
  The shipped design still promises sentinel withholding and automatic retry although current code writes the sentinel and preserves an acknowledgeable durable warning.

- [SUGGESTION] src/backend_task/error.rs:40-45: Keep the wallet identity typed inside TaskError
  WalletNotLoaded stores a preformatted String despite the repository contract requiring typed TaskError data formatted through Display.
```
</details>

PR #887, opened 2026-07-13
fix: live-QA follow-up batch (masternodes, wallets, fonts)

lklimek:
## Why this PR exists

- **Problem**: A batch of QA follow-ups on top of #885 — masternode-tab usability defects and a withdrawal-key hardening pass.
- **What breaks without it**:
  1. **The masternode status dot was unreadable.** A bare coloured dot plus a status word (e.g. "Pending Creation") sat on every masternode card with no indication of *what* it described. Users read it as Core network or PoSe health; it actually reports only whether the node's **Platform identity** can be found.
  2. Masternodes were listed in arbitrary order, so finding a node in a long list meant scanning it.
- **Blocking relationship**: stacked atop #885 (`feat/legacy-identity-migration`), whose head is merged in.

## What was done

- `fix(masternodes)`: relabel the status dot/text as **"Platform identity: \<status\>"** with a clarifying tooltip ("…does not show Core network or PoSe status").
- `fix(masternodes)`: sort the node grid by display name.
- Withdrawal-key hardening, masternode load-form lifecycle, nav routing, and wallet address-listing fixes (see commit history).

## Withdrawn from this PR

The **migration wallet-password prompt** was implemented here and has been **reverted** (`91dc077e`). A review of the combined diff surfaced four blocking defects, and given the blast radius it is being reworked in a dedicated PR stacked on this one rather than fixed in place:

1. The migration awaited a `tokio::sync::Notify` whose only producer is the egui frame loop, while `mcp/resolve.rs` drives the same migration with **no frame loop** — headless `det-cli`/MCP would hang forever on an install with a password-protected legacy wallet.
2. A forgotten password wedged the app on every launch, and the only exit the UI physically permitted was `remove_wallet` — **deleting the wallet and its seed**.
3. Migration spuriously reported `RegistrationIncomplete` even when the correct password was supplied (double-registration race; reproduced as a real test failure).
4. The "non-dismissible" modal did not actually block input — clicks reached widgets behind it.

The rest of this PR is unaffected by the revert.

## Testing

- `cargo +nightly fmt --all` — clean.
- `cargo clippy --all-features --all-targets -- -D warnings` — clean.
- `cargo test --all-features --workspace` — green on the reverted tree.
- A full multi-agent review (security / project-consistency / adversarial QA) was run against this branch; report drove the revert above.

## Breaking changes

None.

## Checklist

- [x] #885 head merged in
- [x] Migration password prompt reverted; reworked in a follow-up PR
[...]

PR #890, opened 2026-07-14
test(masternode-upgrade): standalone masternode-identity upgrade harness + spec

shumkov:
## What

Adds a documented, reproducible **test harness + spec** for verifying that a
masternode (evonode) identity added as a standalone identity in **DET 0.9.3**
survives an in-place upgrade to a newer DET (validated here against PR #887),
across a `dashmate` v3→v4 network upgrade (protocol 11→12).

Docs/scripts only — no product code changes.

## Spec

See `docs/testing/masternode-identity-upgrade/SPEC.md`. Versions exercised:

| Role | DET | SDK pin |
|------|-----|---------|
| baseline | v0.9.3 (`3268b736`) | platform 2.1.0 (`29f7492e`) — GroveDBProof **V0** |
| upgrade | PR #887 `fix/qa-followups-885` (`8a47745a`) | platform 4.0 (`93b967f9`) — **V1** |

## Results

- ✅ **Migration** — DB v11→v38 clean; evonode identity + owner/voting/payout keys survive.
- ✅ **On-chain liveness** — direct DAPI query shows the identity live on protocol 12 with a real balance.
- ✅ **Key functionality** — new DET's *Sign Message* on the migrated **owner** and **payout (TRANSFER)** keys yields signatures that recover to the right address and are **byte-identical** to `dashd signmessage` (deterministic RFC6979), accepted by `dashd verifymessage`.

## Finding for the authors

A **withdrawal through the new DET can't be exercised on a local network.** Its
SPV-only proof verifier (`SpvProvider`, no core-RPC fallback) needs an
`llmq_test_dip0024` **rotation quorum** (type 103, minSize 4) that a 3-masternode
`dashmate` local net can't form, and regtest exposes no param to shrink it — so
`masternodes_ready()` never becomes true and every proof-verified op fails with
"servers unreachable". DET 0.9.3 was unaffected because it read quorum keys from
its local Core node over RPC. **Open question:** is local `dashmate` a supported
target for the SPV verifier, or should the masternode sync fall back to the
legacy `mnlistdiff` path when no rotation quorum exists? Until then the
key-signing check is the local substitute for an end-to-end withdrawal. Details
in SPEC.md "Known limitation".

## Contents

`docs/testing/masternode-identity-upgrade/`
- `SPEC.md` — short spec: goal, versions, steps, results, finding
- `harness/` — reproducible scripts (network setup, key extraction, funding, v3→v4 upgrade, SPV P2P-port proxy, DAPI liveness query) and `verify_signed_message.py`, a dependency-free secp256k1 verifier for the key-functionality check

Secrets (private keys, wallet DBs, profile snapshots) are **gitignored and
regenerated locally** — nothing sensitive is committed. DET UI steps are manual
today; this lands the deterministic network/keys/verification tooling an
automated test can build on.

## Commits

1. `test(masternode-upgrade): DET 0.9.3 baseline harness + spec`
2. `test(masternode-upgrade): v3->v4 upgrade + new-DET verification (PR #887)`

## Summary by CodeRabbit

* **New Features**
  * Added a reproducible end-to-end test harness for upgrading masternode identities from protocol version 11 to 12.
  * Added tooling to set up a local network, extract/validate masternode keys, fund/withdraw, upgrade in place, and confirm identity liveness via direct queries.
  * Added deterministic signed-message verification and a TCP proxy to support local connectivity checks during the test.
* **Documentation**
  * Documented prerequisites, step-by-step execution, verification, an encrypted-wallet survival scenario, and known limitations.
* **Chores**
  * Added safeguards to keep generated local secrets and database artifacts out of version control.

### DIFF
--- a/docs/testing/masternode-identity-upgrade/SPEC.md
+++ b/docs/testing/masternode-identity-upgrade/SPEC.md
@@ -51,6 +51,27 @@ mainnet protocol upgrade between the two DET versions.
   RFC6979), and (c) is accepted by `dashd verifymessage`. This proves the private
   key material survived migration and still signs correctly.
 
+## Additional scenario: password-encrypted wallet survives the upgrade
+
+Separately verifies that a **password-protected wallet** created in DET 0.9.3
+survives the upgrade and can still be unlocked with the same password.
+
+**What we did.** In DET 0.9.3, created an HD wallet with a strong password
+(0.9.3 encrypts the seed with AES-256-GCM, key derived from the password via
+Argon2). Confirmed at the storage layer — `data.db` `wallet.uses_password = 1`
+for that wallet (vs `0` for a no-password wallet). Snapshotted the profile, then
+ran the new DET on a copy.
+
+**Result: pass.** After the DB migration the wallet appears in the new DET as
+`HD: <name>`, is correctly recognized as encrypted (it starts **locked** with an
+**Unlock** button), and **unlocks with the original password** (the button flips
+to **Lock** and address-derivation controls appear). Because the seed is stored
+with *authenticated* encryption (AES-256-GCM), a successful unlock is conclusive:
+a wrong password or a seed corrupted by migration would fail the GCM auth tag and
+refuse to unlock. So the encrypted seed and the password both survive the
+migration intact. (Address rows don't populate in the UI while SPV is unsynced —
+same dip0024 cause as below — but that's independent of the unlock proof.)
+
 ## Known limitation (candidate finding for the PR)
 
 A **withdrawal through the new DET cannot be exercised on a local network.** The
--- a/docs/testing/masternode-identity-upgrade/SPEC.md
+++ b/docs/testing/masternode-identity-upgrade/SPEC.md
@@ -1,0 +1,82 @@
+# Masternode-identity upgrade test — spec
+
+## Goal
+
+Prove that a masternode (evonode) **identity added as a standalone identity in
+DET 0.9.3** survives an in-place upgrade to a newer DET — the identity, its keys,
+and its on-chain state must all still be usable after the app and the platform
+network are upgraded, the way a real user's install would be.
+
+## Versions under test
+
+| Role | DET | pin |
+|------|-----|-----|
+| baseline (old) | `v0.9.3` — commit `3268b736` | dash-sdk platform rev `29f7492e` (platform 2.1.0, grovedb 3.1) |
+| upgrade (new)  | PR #887 `fix/qa-followups-885` — commit `8a47745a`, base `feat/legacy-identity-migration` | dash-sdk platform rev `93b967f9` (platform 4.0) |
+
+Network: local `dashmate` group, **3 masternodes**. Started on dashmate **v3**
+(protocol 11), then upgraded in place to **v4** (protocol 12) — mirroring a real
+mainnet protocol upgrade between the two DET versions.
+
+## What we did
+
+1. **Baseline on 0.9.3 (v3 network / protocol 11).** Loaded an evonode identity
+   by proTxHash + owner/voting/payout WIFs (for a masternode the platform owner
+   identity id *is* the proTxHash). Created a DET wallet, funded it from the
+   local_seed Core wallet, topped the identity up via asset lock, and **withdrew
+   credits to L1 Dash** — confirmed the asset-unlock landed on the Core chain.
+   Snapshotted the DET profile (`backup-1` post-load, `backup-2` post-withdrawal).
+
+2. **Upgraded the network v3 -> v4 in place.** `dashmate@4 group start` migrated
+   the config and images; validators signalled protocol 12, which activated at
+   the next epoch boundary. From then on Drive emits `GroveDBProof::V1`, which
+   0.9.3's SDK cannot decode (expected — this is why the app must be upgraded).
+
+3. **Ran the new DET (PR #887) on a copy of the 0.9.3 profile.** The DB migrated
+   cleanly (schema v11 -> v38, no data loss) and the evonode identity appeared in
+   the new **Masternodes** section: Active, all three key roles (Voting / Owner /
+   Payout) and the derived voter identity intact.
+
+## Results
+
+- **Migration: pass.** Identity + keys + alias survive the 0.9.3 -> PR-887 DB
+  migration.
+- **On-chain liveness: pass.** Direct DAPI `getIdentity`/`getIdentityBalance`
+  (bypassing DET) shows the identity live on protocol 12 with a real, growing
+  balance — proof the migrated reference points at the genuine on-chain identity.
+- **Key functionality: pass.** DET's *Key Info -> Sign Message* signs a known
+  message with the migrated **owner** and **payout (TRANSFER)** keys; each
+  signature (a) recovers to the key's expected address, (b) is byte-identical to
+  what `dashd signmessage` produces with the genuine private key (deterministic
+  RFC6979), and (c) is accepted by `dashd verifymessage`. This proves the private
+  key material survived migration and still signs correctly.
+
+## Known limitation (candidate finding for the PR)
+
+A **withdrawal through the new DET cannot be exercised on a local network.** The
+new DET verifies platform proofs *trustlessly* via its own SPV-synced masternode
+list (`SpvProvider` in `src/context_provider.rs`; no core-RPC fallback — the old
+`core_backend_mode` was dropped, migration v38 notes "SPV-only now"). That SPV
+masternode sync uses the DIP24 `qrinfo` path, which needs an `llmq_test_dip0024`
+rotation quorum. That quorum is **type 103, size 4, minSize 4** (per dash-spv
+`test_utils/masternode_network.rs`), so it cannot form on a 3-masternode local
+network, and regtest exposes no param to shrink it (only `llmq_test`,
+`llmq_test_instantsend`, `llmq_test_platform` are resizable). Result: the new
+DET's `masternodes_ready()` never becomes true, every DAPI address gets banned
+("servers unreachable"), and no proof-verified op (balance refresh, withdrawal)
+can run locally.
+
+DET 0.9.3 was unaffected because it fetched quorum keys straight from its local
+Core node over RPC (`core.get_quorum_public_key`), which has no dip0024 dependency.
+
+**Open question for the PR authors:** is a standard `dashmate` local network a
+supported target for the new SPV-only verifier, or should the SPV masternode sync
+fall back to the legacy `mnlistdiff` path (or accept trusted core quorum data)
+when no rotation quorum exists? Until then, an end-to-end withdrawal has to be
+tested on devnet/testnet (which have rotation quorums), and the **key-signing
+check above is the local substitute** that proves the migrated keys are usable.
+
+## How to reproduce
+
+See `harness/README.md`. The scripts regenerate the network, keys, and profiles
+locally; secrets (private keys, wallet DBs) are gitignored and never committed.
--- a/docs/testing/masternode-identity-upgrade/SPEC.md
+++ b/docs/testing/masternode-identity-upgrade/SPEC.md
@@ -9,10 +9,11 @@ network are upgraded, the way a real user's install would be.
 
 ## Versions under test
 
-| Role | DET | pin |
-|------|-----|-----|
-| baseline (old) | `v0.9.3` — commit `3268b736` | dash-sdk platform rev `29f7492e` (platform 2.1.0, grovedb 3.1) |
-| upgrade (new)  | PR #887 `fix/qa-followups-885` — commit `8a47745a`, base `feat/legacy-identity-migration` | dash-sdk platform rev `93b967f9` (platform 4.0) |
+- **Baseline (old):** DET `v0.9.3` at commit `3268b736`, using dash-sdk
+  platform rev `29f7492e` (platform 2.1.0, grovedb 3.1).
+- **Upgrade (new):** DET PR #887, branch `fix/qa-followups-885` at commit
+  `8a47745a` and based on `feat/legacy-identity-migration`, using dash-sdk
+  platform rev `93b967f9` (platform 4.0).
 
 Network: local `dashmate` group, **3 masternodes**. Started on dashmate **v3**
 (protocol 11), then upgraded in place to **v4** (protocol 12) — mirroring a real
@@ -33,9 +34,9 @@ mainnet protocol upgrade between the two DET versions.
    0.9.3's SDK cannot decode (expected — this is why the app must be upgraded).
 
 3. **Ran the new DET (PR #887) on a copy of the 0.9.3 profile.** The DB migrated
-   cleanly (schema v11 -> v38, no data loss) and the evonode identity appeared in
-   the new **Masternodes** section: Active, all three key roles (Voting / Owner /
-   Payout) and the derived voter identity intact.
+   cleanly (schema v11 -> v38, no data loss) and the evonode identity appeared
+   in the new **Masternodes** section: Active, all three key roles (Voting /
+   Owner / Payout) and the derived voter identity intact.
 
 ## Results
 
@@ -47,9 +48,10 @@ mainnet protocol upgrade between the two DET versions.
 - **Key functionality: pass.** DET's *Key Info -> Sign Message* signs a known
   message with the migrated **owner** and **payout (TRANSFER)** keys; each
   signature (a) recovers to the key's expected address, (b) is byte-identical to
-  what `dashd signmessage` produces with the genuine private key (deterministic
-  RFC6979), and (c) is accepted by `dashd verifymessage`. This proves the private
-  key material survived migration and still signs correctly.
+  what Dash Core's `signmessage` RPC produces with the genuine private key
+  (deterministic RFC6979), and (c) is accepted by Dash Core's `verifymessage`
+  RPC. This proves the private key material survived migration and still signs
+  correctly.
 
 ## Additional scenario: password-encrypted wallet survives the upgrade
 
--- a/docs/testing/masternode-identity-upgrade/harness/.gitignore
+++ b/docs/testing/masternode-identity-upgrade/harness/.gitignore
@@ -2,9 +2,14 @@
 # and recoverable wallet seeds (the test wallet has no password).
 keys.json
 *.db
+*.db-*
 *.sqlite
 *.sqlite-*
+*-journal
+*-shm
+*-wal
 backups/
+backup-*/
 home-*/
 *.tgz
 network-backup*/
--- a/docs/testing/masternode-identity-upgrade/harness/.gitignore
+++ b/docs/testing/masternode-identity-upgrade/harness/.gitignore
@@ -1,0 +1,10 @@
+# Never commit regenerable local secrets / fixtures — they contain private keys
+# and recoverable wallet seeds (the test wallet has no password).
+keys.json
+*.db
+*.sqlite
+*.sqlite-*
+backups/
+home-*/
+*.tgz
+network-backup*/
--- a/docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
+++ b/docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+# Phase 0 — stand up a local platform network on dashmate v3 (protocol 11).
+#
+# Why v3: v3 Drive (grovedb 4) only emits GroveDBProof::V0, which DET 0.9.3's
+# SDK (platform 2.1, grovedb 3.1) can decode. That keeps the 0.9.3 baseline
+# working with no time pressure. The network is later upgraded to v4 in
+# 04_upgrade_to_v4.sh, which is the actual v3->v4 protocol-11->12 migration.
+#
+# Requires: docker running, npx available. Wipes any existing ~/.dashmate.
+set -e
+
+DASHMATE_V3="${DASHMATE_V3:-dashmate@3.0.2}"   # last stable v3 line (3.1 never released)
+NODES="${NODES:-3}"
+
+echo ">> Removing any existing local network + ~/.dashmate"
+docker ps -a --format '{{.Names}}' | grep '^dashmate_' | xargs -r docker rm -f >/dev/null 2>&1 || true
+docker volume ls --format '{{.Name}}' | grep '^dashmate_' | xargs -r docker volume rm >/dev/null 2>&1 || true
+rm -rf ~/.dashmate
+
+echo ">> dashmate setup local (${DASHMATE_V3}, ${NODES} masternodes)"
+npx -y "$DASHMATE_V3" setup local --node-count="$NODES" --no-debug-logs --miner-interval=1m --verbose
+
+# Fast, steady platform blocks so DKG/sync and manual UI steps don't stall.
+for cfg in local_1 local_2 local_3 local_seed; do
+  npx -y "$DASHMATE_V3" config set --config="$cfg" \
+    platform.drive.tenderdash.consensus.createEmptyBlocksInterval 10s
+done
+
+echo ">> Starting group"
+npx -y "$DASHMATE_V3" group start --verbose
+
+echo ">> Waiting for platform blocks to flow"
+until curl -s --max-time 3 http://127.0.0.1:46657/status \
+      | python3 -c "import json,sys; sys.exit(0 if json.load(sys.stdin)['sync_info']['latest_block_height']!='1' else 1)" 2>/dev/null; do
+  sleep 5
+done
+curl -s http://127.0.0.1:46657/status | python3 -c "import json,sys; d=json.load(sys.stdin); print('platform height', d['sync_info']['latest_block_height'], 'protocol', d['node_info']['protocol_version']['app'])"
+echo ">> v3 network ready (expect protocol 11)."
--- a/docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
+++ b/docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+"""Extract each local masternode's platform-identity keys from the running
+network, so they can be loaded into DET as a standalone evonode identity.
+
+For a masternode/evonode, the platform OWNER identity id == the proTxHash, and
+its keys are the owner / voting / payout address keys. dashmate only persists
+the operator BLS key in config.json; the owner/voting/payout private keys live
+in the local_seed Core wallet ("main"), retrievable via dumpprivkey.
+
+Writes keys.json next to this script. keys.json is gitignored (contains WIF
+private keys) — regenerate it locally, never commit it.
+
+Requires: the network from 01_setup_v3_network.sh running.
+"""
+import json
+import os
+import subprocess
+
+DASHMATE_CONFIG = os.path.expanduser("~/.dashmate/config.json")
+SEED_RPC = "http://127.0.0.1:20302/"
+OUT = os.path.join(os.path.dirname(__file__), "keys.json")
+
+
+def rpc_password():
+    cfg = json.load(open(DASHMATE_CONFIG))
+    return cfg["configs"]["local_seed"]["core"]["rpc"]["users"]["dashmate"]["password"]
+
+
+def call(pw, method, params=None, wallet=None):
+    url = SEED_RPC + (f"wallet/{wallet}" if wallet else "")
+    body = json.dumps({"jsonrpc": "1.0", "id": "h", "method": method, "params": params or []})
+    r = subprocess.run(
+        ["curl", "-s", "--user", f"dashmate:{pw}", "--data-binary", body,
+         "-H", "content-type: text/plain;", url],
+        capture_output=True, text=True,
+    )
+    d = json.loads(r.stdout)
+    if d.get("error"):
+        raise RuntimeError(d["error"])
+    return d["result"]
+
+
+def main():
+    pw = rpc_password()
+    out = {}
+    for p in call(pw, "protx", ["list", "registered", True]):
+        st = p["state"]
+        entry = {
+            "proTxHash": p["proTxHash"],          # == platform owner identity id
+            "type": p.get("type"),
+            "ownerAddress": st["ownerAddress"],
+            "votingAddress": st["votingAddress"],
+            "payoutAddress": st["payoutAddress"],
+            "operatorPubKey": st["pubKeyOperator"],
+        }
+        for role, addr_key in (("ownerPrivkey", "ownerAddress"),
+                               ("votingPrivkey", "votingAddress"),
+                               ("payoutPrivkey", "payoutAddress")):
+            entry[role] = call(pw, "dumpprivkey", [st[addr_key]], wallet="main")
+        out[p["proTxHash"][:8]] = entry
+    json.dump(out, open(OUT, "w"), indent=2)
+    print(f"wrote {OUT} ({len(out)} masternode identities)")
+    for k, v in out.items():
+        print(f"  {k} {v['type']} owner={v['ownerAddress']} payout={v['payoutAddress']}")
+
+
+if __name__ == "__main__":
+    main()
--- a/docs/testing/masternode-identity-upgrade/harness/03_fund_address.py
+++ b/docs/testing/masternode-identity-upgrade/harness/03_fund_address.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+"""Fund a DET wallet receive address from the local_seed Core wallet ("main"),
+then wait for one confirmation. Used in phase 1 to give the DET wallet UTXOs so
+it can top up the identity via asset lock.
+
+Usage: 03_fund_address.py <dash_address> [amount_dash]
+Requires: the network running.
+"""
+import json
+import os
+import subprocess
+import sys
+import time
+
+DASHMATE_CONFIG = os.path.expanduser("~/.dashmate/config.json")
+SEED_RPC = "http://127.0.0.1:20302/"
+
+
+def call(pw, method, params=None, wallet=None):
+    url = SEED_RPC + (f"wallet/{wallet}" if wallet else "")
+    body = json.dumps({"jsonrpc": "1.0", "id": "h", "method": method, "params": params or []})
+    r = subprocess.run(
+        ["curl", "-s", "--user", f"dashmate:{pw}", "--data-binary", body,
+         "-H", "content-type: text/plain;", url],
+        capture_output=True, text=True,
+    )
+    d = json.loads(r.stdout)
+    if d.get("error"):
+        raise RuntimeError(d["error"])
+    return d["result"]
+
+
+def main():
+    if len(sys.argv) < 2:
+        print("usage: 03_fund_address.py <dash_address> [amount_dash]")
+        raise SystemExit(2)
+    address = sys.argv[1]
+    amount = float(sys.argv[2]) if len(sys.argv) > 2 else 10.0
+    pw = json.load(open(DASHMATE_CONFIG))["configs"]["local_seed"]["core"]["rpc"]["users"]["dashmate"]["password"]
+
+    txid = call(pw, "sendtoaddress", [address, amount], wallet="main")
+    print(f"sent {amount} DASH to {address}: {txid}")
+    for _ in range(30):
+        conf = call(pw, "gettransaction", [txid], wallet="main").get("confirmations", 0)
+        if conf >= 1:
+            print(f"confirmed ({conf} conf)")
+            return
+        time.sleep(10)
+    print("WARNING: not confirmed within timeout")
+    raise SystemExit(1)
+
+
+if __name__ == "__main__":
+    main()
--- a/docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
+++ b/docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+# Phase 2 — upgrade the SAME running network from v3 to v4 in place.
+#
+# dashmate v4 migrates the config file (configFormatVersion 3 -> 4) and bumps
+# every image tag to v4 (drive:4, tenderdash 1.6.0, rs-dapi:4). v4 Drive reads
+# the v3 on-disk state (the mainnet-supported upgrade path). v4 validators
+# signal protocol 12, which activates at the next epoch boundary; from then on
+# Drive emits GroveDBProof::V1 and DET 0.9.3 can no longer decode proofs
+# (expected — mirrors a real mainnet upgrade).
+#
+# BACK UP FIRST if you want to be able to return to the v3 state:
+#   tar czf dashmate-home.tgz -C ~ .dashmate
+#   for v in $(docker volume ls --format '{{.Name}}' | grep '^dashmate_'); do
+#     docker run --rm -v "$v":/vol -v "$PWD":/b alpine tar czf "/b/$v.tgz" -C /vol .
+#   done
+set -e
+DASHMATE_V4="${DASHMATE_V4:-dashmate@4.0.0}"
+
+echo ">> Stopping v3 group"
+npx -y dashmate@3.0.2 group stop --force || true
+
+echo ">> Starting group with ${DASHMATE_V4} (migrates config + images)"
+npx -y "$DASHMATE_V4" group start --verbose
+
+echo ">> Waiting for protocol 12 to activate (next epoch boundary)"
+until curl -s --max-time 3 http://127.0.0.1:46657/block \
+      | python3 -c "import json,sys; sys.exit(0 if json.load(sys.stdin)['block']['header']['version']['app']=='12' else 1)" 2>/dev/null; do
+  sleep 30
+  curl -s http://127.0.0.1:46657/block | python3 -c "import json,sys; h=json.load(sys.stdin)['block']['header']; print('height', h['height'], 'active', h['version']['app'], 'proposed', h.get('proposed_protocol_version'))" 2>/dev/null || true
+done
+echo ">> Network is now on protocol 12 (v4). Proofs are V1."
--- a/docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
+++ b/docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
@@ -23,9 +23,16 @@ echo ">> Starting group with ${DASHMATE_V4} (migrates config + images)"
 npx -y "$DASHMATE_V4" group start --verbose
 
 echo ">> Waiting for protocol 12 to activate (next epoch boundary)"
+attempt=1
+max_attempts=60
 until curl -s --max-time 3 http://127.0.0.1:46657/block \
       | python3 -c "import json,sys; sys.exit(0 if json.load(sys.stdin)['block']['header']['version']['app']=='12' else 1)" 2>/dev/null; do
+  if [ "$attempt" -ge "$max_attempts" ]; then
+    echo "ERROR: protocol 12 did not activate after ${max_attempts} attempts (about 30 minutes)." >&2
+    exit 1
+  fi
+  curl -s --max-time 3 http://127.0.0.1:46657/block | python3 -c "import json,sys; h=json.load(sys.stdin)['block']['header']; print('height', h['height'], 'active', h['version']['app'], 'proposed', h.get('proposed_protocol_version'))" 2>/dev/null || true
   sleep 30
-  curl -s http://127.0.0.1:46657/block | python3 -c "import json,sys; h=json.load(sys.stdin)['block']['header']; print('height', h['height'], 'active', h['version']['app'], 'proposed', h.get('proposed_protocol_version'))" 2>/dev/null || true
+  attempt=$((attempt + 1))
 done
 echo ">> Network is now on protocol 12 (v4). Proofs are V1."
--- a/docs/testing/masternode-identity-upgrade/harness/README.md
+++ b/docs/testing/masternode-identity-upgrade/harness/README.md
@@ -1,0 +1,61 @@
+# Harness — masternode-identity upgrade test
+
+Reproducible scripts for the test described in [`../SPEC.md`](../SPEC.md). They
+regenerate everything locally; **no secrets or binaries are committed** (see
+`.gitignore`). The DET UI steps are still manual today — the value here is the
+deterministic network/keys/verification tooling an automated test can build on.
+
+## Prerequisites
+
+- Docker running; `npx` (Node) available.
+- A checkout of `dashpay/platform` (for DAPI `.proto` files) — default path
+  `~/Projects/dashpay/platform`.
+- `grpcurl` (for `query_identity_dapi.sh`).
+- Two DET builds: `v0.9.3` and the PR branch, each run against an **isolated
+  profile** via `HOME` override, e.g.
+  `HOME=/path/to/profile /path/to/dash-evo-tool` — DET derives its data dir from
+  `$HOME` (`~/Library/Application Support/Dash-Evo-Tool` on macOS), so this keeps
+  the test off your real profile. Snapshot the profile dir to back it up.
+
+## Order of operations
+
+| Step | Script / action | Phase |
+|------|-----------------|-------|
+| 1 | `01_setup_v3_network.sh` — dashmate v3 local net, protocol 11 | 0 |
+| 2 | `02_extract_masternode_keys.py` — writes `keys.json` (gitignored) | 0 |
+| 3 | *DET 0.9.3*: Load Identity (Evonode) with a proTxHash + its owner/voting/payout WIFs from `keys.json`; create wallet; fund from Core; top up; withdraw to L1. Snapshot the profile. | 1 |
+| 4 | `04_upgrade_to_v4.sh` — in-place v3->v4, waits for protocol 12 | 2 |
+| 5 | `p2p_proxy.py` — **required for the new DET** (see gotcha) | 3 |
+| 6 | *New DET (PR)*: run on a copy of the 0.9.3 profile; confirm the identity migrated (Masternodes screen). | 3 |
+| 7 | `query_identity_dapi.sh <proTxHash>` — on-chain liveness/balance | 3 |
+| 8 | *New DET*: Masternodes -> identity -> Owner key / Payout key -> Sign Message; copy the signature; verify with `verify_signed_message.py` | 3 |
+
+## Key-functionality check (the local substitute for a withdrawal)
+
+In the new DET, open a migrated key (Owner or Payout), type a known message,
+click **Sign Message**, copy the Base64 signature, then:
+
+```
+python3 verify_signed_message.py <key_address> <base64_signature> "<message>"
+```
+
+Exit 0 / `MATCH` means the migrated private key still produces a valid signature
+for its address. For a byte-for-byte cross-check, `dashd signmessage <address>
+"<message>"` (the address is in the local_seed `main` wallet) yields the
+identical signature — deterministic RFC6979 — proving DET holds the same key.
+
+## Gotchas
+
+- **P2P port proxy (step 5).** The PR's SPV client hardcodes the regtest Core P2P
+  port `19899`, but dashmate's `local_1` Core listens on `20001`. Run
+  `python3 p2p_proxy.py` (bridges `127.0.0.1:19899 -> 20001`) before starting the
+  new DET, or its SPV client can't connect at all.
+- **dip0024 / withdrawal.** See SPEC "Known limitation" — the new DET's SPV
+  verifier needs an `llmq_test_dip0024` rotation quorum that a 3-MN local net
+  can't form, so proof-verified ops (including withdrawal) fail locally with
+  "servers unreachable". The Sign Message check above is the workaround.
+- **Signing a *fresh* message.** Always sign a message you control and verify
+  against *that* message. Recovering a stale signature against the wrong message
+  yields a valid-but-wrong address and looks like a failure.
+- **Text entry on non-US keyboard layouts.** Automated typing tools may fail to
+  map characters; paste via the clipboard instead.
--- a/docs/testing/masternode-identity-upgrade/harness/README.md
+++ b/docs/testing/masternode-identity-upgrade/harness/README.md
@@ -44,6 +44,33 @@ for its address. For a byte-for-byte cross-check, `dashd signmessage <address>
 "<message>"` (the address is in the local_seed `main` wallet) yields the
 identical signature — deterministic RFC6979 — proving DET holds the same key.
 
+## Additional check — encrypted wallet survives the upgrade
+
+Verifies a password-protected wallet created in 0.9.3 still unlocks in the new
+DET (see SPEC "Additional scenario"). No network is needed for the wallet itself;
+run it alongside or independently of the identity flow.
+
+1. **In DET 0.9.3** — Wallets -> **Create Wallet**. Hover the entropy grid, click
+   **Generate**, tick **"I wrote it down"**, give it a **name** and a **strong
+   password** (the strength meter must read at least "Strong"), then **Save
+   Wallet**. Record the name + password.
+2. **Confirm it's encrypted at the storage layer** (profile still on 0.9.3):
+   ```
+   sqlite3 "<profile>/Library/Application Support/Dash-Evo-Tool/data.db" \
+     "SELECT alias, uses_password FROM wallet;"
+   ```
+   The new wallet must show `uses_password = 1` (a no-password wallet shows `0`).
+3. **Snapshot** the profile dir, then **run the new DET on a copy** of it. Choose
+   *Just Explore*, dismiss the sync modal (*Continue in the background*), open
+   **Wallets**, and select the wallet.
+4. **Unlock it** — click **Unlock**, enter the original password, confirm. Pass =
+   the wallet is present and the **Unlock button flips to Lock** (authenticated
+   AES-256-GCM decryption succeeded, so the password + seed survived migration).
+   A wrong password shows an error and the button stays **Unlock**.
+
+Note: address rows may stay empty while SPV is unsynced (the dip0024 limitation
+below) — that does not affect the unlock result.
+
 ## Gotchas
 
 - **P2P port proxy (step 5).** The PR's SPV client hardcodes the regtest Core P2P
--- a/docs/testing/masternode-identity-upgrade/harness/README.md
+++ b/docs/testing/masternode-identity-upgrade/harness/README.md
@@ -14,69 +14,96 @@ deterministic network/keys/verification tooling an automated test can build on.
 - Two DET builds: `v0.9.3` and the PR branch, each run against an **isolated
   profile** via `HOME` override, e.g.
   `HOME=/path/to/profile /path/to/dash-evo-tool` — DET derives its data dir from
-  `$HOME` (`~/Library/Application Support/Dash-Evo-Tool` on macOS), so this keeps
-  the test off your real profile. Snapshot the profile dir to back it up.
+  `$HOME` (`~/Library/Application Support/Dash-Evo-Tool` on macOS), so this
+  keeps the test off your real profile. Snapshot the profile dir to back it up.
 
 ## Order of operations
 
-| Step | Script / action | Phase |
-|------|-----------------|-------|
-| 1 | `01_setup_v3_network.sh` — dashmate v3 local net, protocol 11 | 0 |
-| 2 | `02_extract_masternode_keys.py` — writes `keys.json` (gitignored) | 0 |
-| 3 | *DET 0.9.3*: Load Identity (Evonode) with a proTxHash + its owner/voting/payout WIFs from `keys.json`; create wallet; fund from Core; top up; withdraw to L1. Snapshot the profile. | 1 |
-| 4 | `04_upgrade_to_v4.sh` — in-place v3->v4, waits for protocol 12 | 2 |
-| 5 | `p2p_proxy.py` — **required for the new DET** (see gotcha) | 3 |
-| 6 | *New DET (PR)*: run on a copy of the 0.9.3 profile; confirm the identity migrated (Masternodes screen). | 3 |
-| 7 | `query_identity_dapi.sh <proTxHash>` — on-chain liveness/balance | 3 |
-| 8 | *New DET*: Masternodes -> identity -> Owner key / Payout key -> Sign Message; copy the signature; verify with `verify_signed_message.py` | 3 |
+1. **Phase 0:** Run `01_setup_v3_network.sh` to start a dashmate v3 local
+   network on protocol 11.
+2. **Phase 0:** Run `02_extract_masternode_keys.py` to write the gitignored
+   `keys.json` file.
+3. **Phase 1:** In *DET 0.9.3*, load an Evonode identity using a proTxHash and
+   its owner, voting, and payout WIFs from `keys.json`. Create a wallet and copy
+   its receive address.
+4. **Phase 1:** Run
+   `python3 03_fund_address.py <receive-address> [amount]` to fund the DET wallet
+   from the Core `main` wallet and wait for confirmation.
+5. **Phase 1:** In *DET 0.9.3*, top up the identity, withdraw to L1, and snapshot
+   the profile.
+6. **Phase 2:** Run `04_upgrade_to_v4.sh` to upgrade the network in place from
+   v3 to v4 and wait for protocol 12.
+7. **Phase 3:** Run `p2p_proxy.py`. This is **required for the new DET**; see the
+   gotcha below.
+8. **Phase 3:** Run the *new DET (PR)* on a copy of the 0.9.3 profile and confirm
+   the identity migrated on the Masternodes screen.
+9. **Phase 3:** Run `query_identity_dapi.sh <proTxHash>` to check on-chain
+   liveness and balance.
+10. **Phase 3:** In the *new DET*, open Masternodes -> identity -> Owner key or
+    Payout key -> Sign Message. Copy the signature and verify it with
+    `verify_signed_message.py`.
 
 ## Key-functionality check (the local substitute for a withdrawal)
 
 In the new DET, open a migrated key (Owner or Payout), type a known message,
 click **Sign Message**, copy the Base64 signature, then:
 
-```
+```bash
 python3 verify_signed_message.py <key_address> <base64_signature> "<message>"
 ```
 
 Exit 0 / `MATCH` means the migrated private key still produces a valid signature
-for its address. For a byte-for-byte cross-check, `dashd signmessage <address>
-"<message>"` (the address is in the local_seed `main` wallet) yields the
-identical signature — deterministic RFC6979 — proving DET holds the same key.
+for its address. For a byte-for-byte cross-check against the address in the
+`local_seed` Core `main` wallet, run:
+
+```bash
+npx -y dashmate@4.0.0 core cli \
+  "-rpcwallet=main signmessage <address> \"<message>\"" \
+  --config=local_seed
+```
+
+The command yields the identical signature — deterministic RFC6979 — proving
+DET holds the same key.
 
 ## Additional check — encrypted wallet survives the upgrade
 
 Verifies a password-protected wallet created in 0.9.3 still unlocks in the new
-DET (see SPEC "Additional scenario"). No network is needed for the wallet itself;
-run it alongside or independently of the identity flow.
+DET (see SPEC "Additional scenario"). No network is needed for the wallet
+itself; run it alongside or independently of the identity flow.
 
-1. **In DET 0.9.3** — Wallets -> **Create Wallet**. Hover the entropy grid, click
-   **Generate**, tick **"I wrote it down"**, give it a **name** and a **strong
-   password** (the strength meter must read at least "Strong"), then **Save
-   Wallet**. Record the name + password.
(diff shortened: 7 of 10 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 7279666c)

docs/testing/masternode-identity-upgrade/harness/.gitignore
```diff
@@ -0,0 +1,10 @@
+# Never commit regenerable local secrets / fixtures — they contain private keys
+# and recoverable wallet seeds (the test wallet has no password).
+keys.json
+*.db
+*.sqlite
+*.sqlite-*
+backups/
+home-*/
```
thepastaclaw, line 13: **🔴 Blocking: Ignore documented profile snapshots and SQLite sidecars**

The PR states that wallet databases and profile snapshots are gitignored, but the README permits an arbitrary HOME profile path and the spec names snapshots backup-1 and backup-2. Under the harness, those names do not match backups/ or home-*/, and *.db does not match SQLite sidecars such as data.db-journal, data.db-wal, or data.db-shm. git check-ignore confirms these paths remain trackable, and a focused SQLite reproduction showed a rollback journal retaining prior secret-like page contents; the pinned DET 0.9.3 database also serializes imported identity private keys. Add rules for both sidecars and the documented snapshot names so a user cannot accidentally stage this local key material.

```suggestion
*.db
*.db-*
*.sqlite
*.sqlite-*
backups/
backup-*/
home-*/
```

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/verify_signed_message.py
```diff
@@ -0,0 +1,124 @@
...
+    return r
+
+
+def varint(n):
+    return bytes([n]) if n < 0xFD else b"\xfd" + n.to_bytes(2, "little")
+
+
+def dash_message_hash(message: str) -> bytes:
+    data = b"\x19DarkCoin Signed Message:\n" + varint(len(message)) + message.encode()
+    return hashlib.sha256(hashlib.sha256(data).digest()).digest()
```
thepastaclaw: **🔴 Blocking: Serialize the UTF-8 byte length of signed messages**

Python len(message) counts Unicode code points, while the pinned Rust signed_msg_hash uses str::len(), which is the UTF-8 byte length, before hashing message.as_bytes(). The current verifier therefore computes a different digest for a valid message such as Dash 🚀 and reports MISMATCH for its genuine DET/Core signature. varint() also raises OverflowError at 65,536 bytes even though the DET multiline field has no length cap. Encode once, serialize that byte length with the full CompactSize range, and hash the same bytes.

```suggestion
def varint(n):
    if n < 0xFD:
        return bytes([n])
    if n <= 0xFFFF:
        return b"\xfd" + n.to_bytes(2, "little")
    if n <= 0xFFFFFFFF:
        return b"\xfe" + n.to_bytes(4, "little")
    return b"\xff" + n.to_bytes(8, "little")


def dash_message_hash(message: str) -> bytes:
    encoded_message = message.encode("utf-8")
    data = b"\x19DarkCoin Signed Message:\n" + varint(len(encoded_message)) + encoded_message
    return hashlib.sha256(hashlib.sha256(data).digest()).digest()
```

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,61 @@
...
+click **Sign Message**, copy the Base64 signature, then:
+
+```
+python3 verify_signed_message.py <key_address> <base64_signature> "<message>"
+```
+
+Exit 0 / `MATCH` means the migrated private key still produces a valid signature
+for its address. For a byte-for-byte cross-check, `dashd signmessage <address>
+"<message>"` (the address is in the local_seed `main` wallet) yields the
+identical signature — deterministic RFC6979 — proving DET holds the same key.
```
thepastaclaw: **🔴 Blocking: Replace the unusable dashd cross-check command**

dashd is the node daemon, not the RPC command-line client, so dashd signmessage is not a valid way to invoke the wallet RPC. The prerequisites provide Core through the Dashmate network rather than a local dashd executable, and the extracted keys live in the local_seed main wallet. Document a working Dashmate Core CLI or authenticated JSON-RPC invocation that selects wallet main; otherwise readers cannot reproduce the byte-identical signmessage cross-check claimed here and in the spec.

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo ">> Starting group with ${DASHMATE_V4} (migrates config + images)"
+npx -y "$DASHMATE_V4" group start --verbose
+
+echo ">> Waiting for protocol 12 to activate (next epoch boundary)"
+until curl -s --max-time 3 http://127.0.0.1:46657/block \
+      | python3 -c "import json,sys; sys.exit(0 if json.load(sys.stdin)['block']['header']['version']['app']=='12' else 1)" 2>/dev/null; do
+  sleep 30
+  curl -s http://127.0.0.1:46657/block | python3 -c "import json,sys; h=json.load(sys.stdin)['block']['header']; print('height', h['height'], 'active', h['version']['app'], 'proposed', h.get('proposed_protocol_version'))" 2>/dev/null || true
+done
```
thepastaclaw, line 37: **🟡 Suggestion: Fail when protocol activation never completes**

The activation loop has no deadline or attempt limit. If validators stall or never activate protocol 12, the harness waits forever instead of producing a failed test result; the diagnostic curl can also block indefinitely because it has no maximum time. Add a finite, generously sized deadline with a nonzero exit and apply the same readiness bound to the setup loop in 01_setup_v3_network.sh.

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,61 @@
...
+  the test off your real profile. Snapshot the profile dir to back it up.
+
+## Order of operations
+
+| Step | Script / action | Phase |
+|------|-----------------|-------|
+| 1 | `01_setup_v3_network.sh` — dashmate v3 local net, protocol 11 | 0 |
+| 2 | `02_extract_masternode_keys.py` — writes `keys.json` (gitignored) | 0 |
+| 3 | *DET 0.9.3*: Load Identity (Evonode) with a proTxHash + its owner/voting/payout WIFs from `keys.json`; create wallet; fund from Core; top up; withdraw to L1. Snapshot the profile. | 1 |
+| 4 | `04_upgrade_to_v4.sh` — in-place v3->v4, waits for protocol 12 | 2 |
```
thepastaclaw: **🟡 Suggestion: Include the funding helper in the ordered workflow**

The ordered procedure moves from 02_extract_masternode_keys.py to 04_upgrade_to_v4.sh and tells the reader only to fund from Core. The committed 03_fund_address.py is explicitly the phase-1 helper and includes deterministic sending and confirmation waiting, but the README never names or invokes it. Add 03_fund_address.py <receive-address> [amount] to step 3 so the documented workflow actually uses the supplied funding phase.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At the pinned head, three blocking issues undermine the harness's stated secret-exclusion and verification guarantees: documented profile artifacts can remain trackable, the message verifier hashes non-ASCII messages incorrectly, and the documented Core cross-check invokes the daemon instead of an RPC client. Two additional reproducibility suggestions survive; the macOS xargs claim and the proposed Platform checkout pin were refuted.

### Review provenance

🔴 3 blocking | 🟡 2 suggestion(s)

<details>

```

In `docs/testing/masternode-identity-upgrade/harness/.gitignore`:
- [BLOCKING] docs/testing/masternode-identity-upgrade/harness/.gitignore:4-8: Ignore documented profile snapshots and SQLite sidecars
  The PR states that wallet databases and profile snapshots are gitignored, but the README permits an arbitrary HOME profile path and the spec names snapshots backup-1 and backup-2. Under the harness, those names do not match backups/ or home-*/, and *.db does not match SQLite sidecars such as data.db-journal, data.db-wal, or data.db-shm. git check-ignore confirms these paths remain trackable, and a focused SQLite reproduction showed a rollback journal retaining prior secret-like page contents; the pinned DET 0.9.3 database also serializes imported identity private keys. Add rules for both sidecars and the documented snapshot names so a user cannot accidentally stage this local key material.

In `docs/testing/masternode-identity-upgrade/harness/verify_signed_message.py`:
- [BLOCKING] docs/testing/masternode-identity-upgrade/harness/verify_signed_message.py:61-67: Serialize the UTF-8 byte length of signed messages
  Python len(message) counts Unicode code points, while the pinned Rust signed_msg_hash uses str::len(), which is the UTF-8 byte length, before hashing message.as_bytes(). The current verifier therefore computes a different digest for a valid message such as Dash 🚀 and reports MISMATCH for its genuine DET/Core signature. varint() also raises OverflowError at 65,536 bytes even though the DET multiline field has no length cap. Encode once, serialize that byte length with the full CompactSize range, and hash the same bytes.

In `docs/testing/masternode-identity-upgrade/harness/README.md`:
- [BLOCKING] docs/testing/masternode-identity-upgrade/harness/README.md:43-45: Replace the unusable dashd cross-check command
  dashd is the node daemon, not the RPC command-line client, so dashd signmessage is not a valid way to invoke the wallet RPC. The prerequisites provide Core through the Dashmate network rather than a local dashd executable, and the extracted keys live in the local_seed main wallet. Document a working Dashmate Core CLI or authenticated JSON-RPC invocation that selects wallet main; otherwise readers cannot reproduce the byte-identical signmessage cross-check claimed here and in the spec.
- [SUGGESTION] docs/testing/masternode-identity-upgrade/harness/README.md:24-27: Include the funding helper in the ordered workflow
  The ordered procedure moves from 02_extract_masternode_keys.py to 04_upgrade_to_v4.sh and tells the reader only to fund from Core. The committed 03_fund_address.py is explicitly the phase-1 helper and includes deterministic sending and confirmation waiting, but the README never names or invokes it. Add 03_fund_address.py <receive-address> [amount] to step 3 so the documented workflow actually uses the supplied funding phase.

In `docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh`:
- [SUGGESTION] docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh:25-30: Fail when protocol activation never completes
  The activation loop has no deadline or attempt limit. If validators stall or never activate protocol 12, the harness waits forever instead of producing a failed test result; the diagnostic curl can also block indefinitely because it has no maximum time. Add a finite, generously sized deadline with a nonzero exit and apply the same readiness bound to the setup loop in 01_setup_v3_network.sh.
```
</details>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,88 @@
...
+1. **In DET 0.9.3** — Wallets -> **Create Wallet**. Hover the entropy grid, click
+   **Generate**, tick **"I wrote it down"**, give it a **name** and a **strong
+   password** (the strength meter must read at least "Strong"), then **Save
+   Wallet**. Record the name + password.
+2. **Confirm it's encrypted at the storage layer** (profile still on 0.9.3):
+   ```
+   sqlite3 "<profile>/Library/Application Support/Dash-Evo-Tool/data.db" \
+     "SELECT alias, uses_password FROM wallet;"
+   ```
+   The new wallet must show `uses_password = 1` (a no-password wallet shows `0`).
```
thepastaclaw: **🟡 Suggestion: Use the platform-specific profile path for the encryption check**

The new sqlite3 command always queries <profile>/Library/Application Support/Dash-Evo-Tool/data.db, which is only the macOS location. With the documented HOME-isolated workflow on native Linux, ProjectDirs resolves the database under <profile>/.config/dash-evo-tool/data.db. The command therefore fails or checks the wrong file on Linux; document platform-specific paths or define the database path separately.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Full-range revalidation confirms that all five carried-forward prior findings remain valid: three are blocking and two are suggestions. The latest delta introduces one additional valid documentation suggestion because its encrypted-wallet database check uses only the macOS profile path. The surviving blockers require a preliminary REQUEST_CHANGES review.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

<details>

```

In `docs/testing/masternode-identity-upgrade/harness/README.md`:
- [SUGGESTION] docs/testing/masternode-identity-upgrade/harness/README.md:57-62: Use the platform-specific profile path for the encryption check
  The new sqlite3 command always queries <profile>/Library/Application Support/Dash-Evo-Tool/data.db, which is only the macOS location. With the documented HOME-isolated workflow on native Linux, ProjectDirs resolves the database under <profile>/.config/dash-evo-tool/data.db. The command therefore fails or checks the wrong file on Linux; document platform-specific paths or define the database path separately.
```
</details>

docs/testing/masternode-identity-upgrade/harness/.gitignore
Claudius-Maginificent: Fixed in `7279666c` — added `*-journal`, `*-shm`, `*-wal` sidecar patterns and `backup-*/`/`*.db-*` snapshot patterns.

docs/testing/masternode-identity-upgrade/harness/verify_signed_message.py
Claudius-Maginificent: Fixed in `7279666c` — same root cause as CodeRabbit's finding on this line: the varint prefix now encodes the UTF-8 byte length via proper CompactSize (1/3/5/9-byte forms), fixing both the char-vs-byte-count bug and the overflow above 64KB.

docs/testing/masternode-identity-upgrade/harness/README.md
Claudius-Maginificent: Fixed in `7279666c` — replaced the invalid `dashd signmessage` example with a working Dashmate Core CLI invocation (`npx dashmate core cli "-rpcwallet=main signmessage ..."`).

docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
Claudius-Maginificent: Fixed in `7279666c` — the protocol-activation wait loop is now bounded to 60 attempts and fails loudly with a clear timeout error instead of looping forever.

docs/testing/masternode-identity-upgrade/harness/README.md
Claudius-Maginificent: Fixed in `7279666c` — added the missing `03_fund_address.py` funding step to the numbered walkthrough.

docs/testing/masternode-identity-upgrade/harness/README.md
Claudius-Maginificent: Fixed in `7279666c` — documented both the macOS and Linux database paths for the encryption-check command instead of hardcoding macOS only.

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+# working with no time pressure. The network is later upgraded to v4 in
+# 04_upgrade_to_v4.sh, which is the actual v3->v4 protocol-11->12 migration.
+#
+# Requires: docker running, npx available. Wipes any existing ~/.dashmate.
+set -e
+
+DASHMATE_V3="${DASHMATE_V3:-dashmate@3.0.2}"   # last stable v3 line (3.1 never released)
+NODES="${NODES:-3}"
+
+echo ">> Removing any existing local network + ~/.dashmate"
```
Claudius-Maginificent, line 15: 🟡 **Unconditional `rm -rf ~/.dashmate` plus force-removal of every `dashmate_*` container and volume — no confirmation, no scoping, no backup** *(merge-blocking)*

The documented first step of the runbook destroys state with no prompt and no dry-run. The `grep '^dashmate_'` glob is **not scoped to the local network** — it matches every dashmate container and volume on the host, including a testnet or mainnet masternode the developer happens to run on the same machine. Docker volumes for such a node hold the Core wallet; removing them is irreversible and, for a real masternode, means loss of key material and access to collateral. The `|| true` on the wipe lines also means `set -e` will not halt a partial failure.

The comment on line 9 warns "Wipes any existing ~/.dashmate", but a comment is not a guardrail.

**Recommendation:** Gate the wipe behind an explicit opt-in — require `FORCE_WIPE=1` or an interactive y/N confirmation that lists exactly which containers and volumes will be destroyed *before* touching anything. Scope the grep to the local-network prefix actually created by `dashmate setup local` rather than the broad `^dashmate_`, so a co-resident testnet/mainnet node cannot be caught. Better still, honour a `DASHMATE_HOME`/`--home` override so the harness never points at the developer's default dashmate home. Drop the `|| true` on the wipe so a partial failure stops the run.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+for cfg in local_1 local_2 local_3 local_seed; do
+  npx -y "$DASHMATE_V3" config set --config="$cfg" \
+    platform.drive.tenderdash.consensus.createEmptyBlocksInterval 10s
+done
+
+echo ">> Starting group"
+npx -y "$DASHMATE_V3" group start --verbose
+
+echo ">> Waiting for platform blocks to flow"
+until curl -s --max-time 3 http://127.0.0.1:46657/status \
```
Claudius-Maginificent, line 33: 🟡 **Readiness check exits the wait loop at height 0, declaring the network ready before it produced a block — and the loop is unbounded**

Two coupled problems in this wait loop.

**(1) Correctness.** `until` terminates when the condition returns 0, and the predicate returns 0 for *any* height that is not the literal string `'1'`. Tenderdash reports `latest_block_height` as `'0'` in the window after it starts serving `/status` but before the first block. `'0' != '1'` is true, so the loop exits immediately and the script prints `>> v3 network ready` against a chain with **zero blocks**. The intent is clearly "wait until height has advanced past 1" — a `>` comparison on an integer, not a `!=` on a string. Real consequence: step 2 (`02_extract_masternode_keys.py`) then fails for reasons unrelated to its actual cause.

**(2) Robustness.** The loop has no iteration cap, no max-wait, and no escape hatch — if dashmate/docker never reaches the expected height (image pull failure, port clash, OOM), it hangs forever. It also prints nothing at all while waiting. Note `04_upgrade_to_v4.sh` was bounded to 60 attempts in `7279666c`; this loop was not, so the two are now inconsistent — and `03_fund_address.py` already models the right pattern with `range(30)` + `raise SystemExit(1)`.

**Recommendation:**
```sh
sys.exit(0 if int(json.load(sys.stdin)['sync_info']['latest_block_height']) > 1 else 1)
```
and bound the loop with a max retry count + explicit non-zero exit and diagnostic on timeout, matching what `04` now does. Additionally, *assert* protocol 11 before baseline work rather than printing an "expect protocol 11" message.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+DASHMATE_V3="${DASHMATE_V3:-dashmate@3.0.2}"   # last stable v3 line (3.1 never released)
+NODES="${NODES:-3}"
+
+echo ">> Removing any existing local network + ~/.dashmate"
+docker ps -a --format '{{.Names}}' | grep '^dashmate_' | xargs -r docker rm -f >/dev/null 2>&1 || true
+docker volume ls --format '{{.Name}}' | grep '^dashmate_' | xargs -r docker volume rm >/dev/null 2>&1 || true
+rm -rf ~/.dashmate
+
+echo ">> dashmate setup local (${DASHMATE_V3}, ${NODES} masternodes)"
+npx -y "$DASHMATE_V3" setup local --node-count="$NODES" --no-debug-logs --miner-interval=1m --verbose
```
Claudius-Maginificent, line 21: 🟡 **`npx -y` installs and executes dashmate plus its full transitive tree with lifecycle scripts enabled**

Six separate `npx -y dashmate@<ver>` invocations across `01` and `04` auto-download and execute an npm package with `-y` suppressing the install confirmation, no `--ignore-scripts`, no lockfile, and no integrity pinning. The top-level version pin (`dashmate@3.0.2`, `dashmate@4.0.0`) is the right instinct and materially better than a range — but the pin does **not** extend to the transitive dependency tree, which resolves fresh on every run and executes `preinstall`/`install`/`postinstall` hooks with the invoking user's full permissions.

Research performed and stated plainly: OSV.dev returned `{}` for dashmate 3.0.2, 4.0.0 and all versions — **no known advisories**; the GitHub Global Advisory Database is likewise clean for the npm package. This is a supply-chain *exposure* finding, not a report of a live compromise.

The exposure matters here specifically because the same host is expected to hold masternode owner/voting/payout keys.

**Recommendation:** Pass `--ignore-scripts` to the npx invocations if dashmate tolerates it, or install dashmate once from a committed lockfile / pinned container image rather than resolving the tree on every run. At minimum, document in README "Prerequisites" that these scripts execute third-party npm lifecycle code and should be run on a scratch machine or container, not on a host holding mainnet keys.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
```diff
@@ -0,0 +1,38 @@
...
+# Phase 2 — upgrade the SAME running network from v3 to v4 in place.
+#
+# dashmate v4 migrates the config file (configFormatVersion 3 -> 4) and bumps
+# every image tag to v4 (drive:4, tenderdash 1.6.0, rs-dapi:4). v4 Drive reads
+# the v3 on-disk state (the mainnet-supported upgrade path). v4 validators
+# signal protocol 12, which activates at the next epoch boundary; from then on
+# Drive emits GroveDBProof::V1 and DET 0.9.3 can no longer decode proofs
+# (expected — mirrors a real mainnet upgrade).
+#
+# BACK UP FIRST if you want to be able to return to the v3 state:
```
Claudius-Maginificent, line 11: 🟡 **Backup recipe writes a secret-bearing `~/.dashmate` tarball into `$PWD`, and the root `.gitignore` does not cover it**

The header comment instructs the operator to run `tar czf dashmate-home.tgz -C ~ .dashmate` and `docker run ... -v "$PWD":/b alpine tar czf "/b/$v.tgz" ...`. Both write into the *current working directory*, and the comment never states which directory that must be.

A `.dashmate` archive is high-value: it contains the Core wallet (owner/voting/payout private keys), the dashmate RPC passwords, and operator BLS keys.

**Verified:** the harness-local `.gitignore` covers `*.tgz`, but that file only governs `docs/testing/masternode-identity-upgrade/harness/`. The repository root `.gitignore` has **no** `*.tgz`, `keys.json`, or `*.db` entry (confirmed against this branch's HEAD). An operator who runs the recipe from the repo root — the natural cwd — produces an untracked-but-stageable secret archive.

**Recommendation:** Make the recipe path-explicit and self-protecting rather than relying on the operator's cwd: write backups to an absolute path outside the repo (e.g. `${BACKUP_DIR:-$HOME/dashmate-backups}`) and `mkdir -p` it in the script. Failing that, add `*.tgz`, `keys.json`, and `*.db` to the **root** `.gitignore` so the rule holds regardless of cwd. Belt and braces: do both.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

Claudius-Maginificent (commented):
## 🧐 Grumpy review — 4-way pass (security + project-consistency + QA + Codex Sol)

Reviewed at `7279666c`. Four independent reviewers produced **36 raw findings**, consolidated to **26** after merging 10 duplicate clusters; **20 comments** are posted here after dropping what `7279666c` already fixed (see below). **Zero CRITICAL, zero HIGH** — that distribution is honest rather than generous: this PR adds no product code, touches no `.rs` file, and is unreachable from CI (`tests.yml`/`clippy.yml` filter on `**/*.rs` and friends; nothing under `docs/**` matches) and from every packaging path. The blast radius of everything here is the operator who runs the harness.

**Advisory only** — commenting, not requesting changes.

### ✅ What is genuinely good (verified, not assumed)

- **No secrets committed** anywhere in the tree or in any commit — the only 64-hex hits are the public secp256k1 domain constants. The `.gitignore` was written *before* the secrets exist.
- **The SPEC's load-bearing architectural claims hold.** `core_backend_mode` dropped, migration v38's "SPV-only now" note, `SpvProvider` in `src/context_provider.rs`, `REGTEST_P2P_PORT = 19899`, DB v11→v38 — all independently spot-checked against the current `origin/v1.0-dev` tip.
- **The crypto core survived adversarial testing.** An independent re-implementation of ECDSA signing on top of the script's own primitives round-trips correctly; wrong-message and bit-flipped inputs are correctly rejected; hash160/Base58Check matches a known Bitcoin test vector.
- **The dip0024 limitation is diagnosed to the quorum-parameter level** and escalated as an open question rather than buried. That is the right instinct.
- **`03_fund_address.py` hardcodes its RPC endpoint** and cannot be pointed at mainnet even by a careless operator.
- **All shell scripts are shellcheck-clean.**
- **Supply-chain check came back clean:** OSV.dev and the GitHub Global Advisory Database report no known advisories for dashmate 3.0.2 / 4.0.0 / any version.

### 🟡 The MEDIUM items worth your attention

1. **[`01_setup_v3_network.sh:15`] Unguarded destructive first step.** The documented step 1 force-removes *every* `dashmate_*` container and volume on the host and runs `rm -rf ~/.dashmate` with no prompt, no dry-run, and no scoping to the local network. A developer with a co-resident testnet or mainnet dashmate node loses its Docker volumes — which hold the Core wallet — merely by starting this test.
2. **[`01_setup_v3_network.sh:33`] Real correctness bug — readiness check passes at height 0.** `!='1'` on a *string* returns true for Tenderdash's `'0'`, so the loop exits and prints ">> v3 network ready" against a chain with zero blocks, making step 2 fail for reasons unrelated to its actual cause. The same loop is also unbounded (unlike `04`, which `7279666c` bounded — the two are now inconsistent).
3. **[`SPEC.md:41`] `test(...)` title, zero tests.** 0 `.rs` files, nothing CI-enforced, and the scripts *print* rather than *assert* the behaviours the Results section records as "pass". The PR body is honest that UI steps are manual; the gap is between that honesty and the title plus the unqualified pass verdicts.
4. **[`SPEC.md:1`] Undeclared `docs/testing/` docs root + branch is 2 commits behind `v1.0-dev`** (`mergeable_state: "blocked"`). Genuinely mitigating: the convention that governs placement (`docs/gui-testing/`) landed in #860 — one of the two commits this branch is behind — so it was invisible from this base.
5. **[`04_upgrade_to_v4.sh:11`] Backup recipe writes a secret-bearing `.dashmate` tarball into `$PWD`**, and the **root** `.gitignore` has no `*.tgz`/`keys.json`/`*.db` entry (the harness-local one does not govern the repo root).
6. **[`01_setup_v3_network.sh:21`] `npx -y` with lifecycle scripts enabled** on a host expected to hold masternode keys — exposure, not a live compromise.
7. **[`README.md:14`] macOS-only in practice, never says so**, and omits `DASH_EVO_DATA_DIR` for the new DET.

The remaining LOW findings are robustness/idiom/DRY nits, posted inline.

### 🗑️ Dropped as already fixed in `7279666c` (and duplicating existing threads)

Credit where due — the fix commit landed after this review ran and resolved three real findings, so they are **not** re-posted:

- **The message-hash byte-length bug** (`dash_message_hash` varinted the *character* count, not the UTF-8 byte length — a confident false negative on any non-ASCII message). Now correctly encodes-then-measures, and `varint()` was upgraded to full CompactSize, which also fixes the `OverflowError` above 64 KB. This was our headline correctness bug; nice fix.
- **`dashd signmessage` naming the wrong executable** — now a working `npx dashmate core cli "-rpcwallet=main signmessage ..."` invocation, with SPEC.md updated to say "Dash Core's `signmessage` RPC".
- **README never naming `03_fund_address.py`** — now step 4 of the walkthrough.

Also **partially** addressed and re-scoped rather than dropped: the unbounded-poll-loop finding (`04` bounded, `01` still not) and the macOS-path finding (the `sqlite3` check now lists both paths; the Prerequisites rationale and `DASH_EVO_DATA_DIR` remain).

### Bottom line

A genuinely useful, honestly-written harness. The concept is sound and the fixes are small — nothing here is about the idea, and the crypto that matters most is correct. The two I would not merge without are the unguarded `rm -rf` and the height-0 readiness check.

docs/testing/masternode-identity-upgrade/harness/04_upgrade_to_v4.sh
```diff
@@ -0,0 +1,38 @@
...
+#   tar czf dashmate-home.tgz -C ~ .dashmate
+#   for v in $(docker volume ls --format '{{.Name}}' | grep '^dashmate_'); do
+#     docker run --rm -v "$v":/vol -v "$PWD":/b alpine tar czf "/b/$v.tgz" -C /vol .
+#   done
+set -e
+DASHMATE_V4="${DASHMATE_V4:-dashmate@4.0.0}"
+
+echo ">> Stopping v3 group"
+npx -y dashmate@3.0.2 group stop --force || true
```
Claudius-Maginificent, line 20: 🟢 **`04` hardcodes `dashmate@3.0.2` while `01` parameterizes the same value via `$DASHMATE_V3`**

`01` exposes `DASHMATE_V3="${DASHMATE_V3:-dashmate@3.0.2}"` specifically so an operator can pin a different v3 patch release, and uses that variable for every dashmate@3 invocation. `04` exposes `DASHMATE_V4` but then stops the v3 group with a literal `npx -y dashmate@3.0.2 group stop --force || true` — it does not read `$DASHMATE_V3` at all.

Override `DASHMATE_V3` when setting up (the documented use case for the variable) and the teardown silently runs a *different* dashmate version against the network the other one created — npx will happily fetch 3.0.2 and operate on a config it did not write.

**Recommendation:** Use `"${DASHMATE_V3:-dashmate@3.0.2}"` in `04`'s stop command, matching `01`'s pattern — or hoist both version pins into a shared sourced `versions.sh` so the two scripts cannot disagree about which network they are talking to.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
```diff
@@ -0,0 +1,68 @@
...
+
+def rpc_password():
+    cfg = json.load(open(DASHMATE_CONFIG))
+    return cfg["configs"]["local_seed"]["core"]["rpc"]["users"]["dashmate"]["password"]
+
+
+def call(pw, method, params=None, wallet=None):
+    url = SEED_RPC + (f"wallet/{wallet}" if wallet else "")
+    body = json.dumps({"jsonrpc": "1.0", "id": "h", "method": method, "params": params or []})
+    r = subprocess.run(
```
Claudius-Maginificent, line 32: 🟢 **`call()` leaks the RPC password into the host process list, ignores curl's exit status, and sets no timeouts**

Grouping three defects in this one helper (duplicated verbatim into `03_fund_address.py`, see separate note):

**(1) Credential in argv.** The dashmate RPC password is handed to curl as `--user dashmate:{pw}`. Process arguments are world-readable on Linux via `/proc/<pid>/cmdline` and `ps aux`, so any unprivileged local process can read it for the window each curl runs — and `02` issues one curl per masternode per key role while `03` polls up to 30 times, so the window is repeated, not momentary. Impact on a dev box is genuinely low (the credential guards a localhost regtest node) and is not being inflated — it is flagged as a credential-handling pattern that should not be copied forward.

**(2) Exit status ignored.** `subprocess.run(...)` is called without `check=True` and `r.returncode`/`r.stderr` are never inspected. When curl fails — daemon down, wrong port, TLS refusal, i.e. the *most likely* failure modes in a harness whose whole job is standing networks up and down — `r.stdout` is empty and the next line raises a bare `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`. That traceback sends the operator debugging JSON instead of connectivity.

**(3) No timeouts.** No `--connect-timeout`/`--max-time`, so a stalled node blocks the helper indefinitely.

**Recommendation:** Drop the subprocess entirely and use `urllib.request` from the stdlib with an `Authorization: Basic` header and an explicit timeout — no new dependency, keeps the credential out of argv, gives real exceptions, removes the fork-per-call cost, and is consistent with `verify_signed_message.py`'s zero-dependency ethos. Add bounded retries.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
```diff
@@ -0,0 +1,68 @@
...
+            "votingAddress": st["votingAddress"],
+            "payoutAddress": st["payoutAddress"],
+            "operatorPubKey": st["pubKeyOperator"],
+        }
+        for role, addr_key in (("ownerPrivkey", "ownerAddress"),
+                               ("votingPrivkey", "votingAddress"),
+                               ("payoutPrivkey", "payoutAddress")):
+            entry[role] = call(pw, "dumpprivkey", [st[addr_key]], wallet="main")
+        out[p["proTxHash"][:8]] = entry
+    json.dump(out, open(OUT, "w"), indent=2)
```
Claudius-Maginificent, line 61: 🟢 **`keys.json` written with the default umask, leaving WIF private keys world-readable inside the repo worktree**

`json.dump(out, open(OUT, "w"), indent=2)` creates `keys.json` with the process default umask — 0644 on a typical Linux/macOS developer account, i.e. readable by **every local user**. The contents are exactly what the module docstring says: owner, voting, and payout WIF private keys for every masternode, plus the proTxHash identifying the platform identity.

Compounding this, the file is written into the repository worktree and stays there after the run. `.gitignore` prevents normal Git *discovery* — it does not protect against other local users, backup daemons, editor indexers, or `git add -f`. This conflicts with least-privilege secret handling (OWASP ASVS V13.3.1).

**Recommendation:** Create the file with restrictive permissions from the outset rather than chmod-ing after the fact (which leaves a race):
```python
fd = os.open(OUT, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
with os.fdopen(fd, "w") as f:
    json.dump(out, f, indent=2)
```
Better still, write it outside the repository into a mode-0700 temp directory, document its lifetime, and remove it after the manual import step.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+import subprocess
+
+DASHMATE_CONFIG = os.path.expanduser("~/.dashmate/config.json")
+SEED_RPC = "http://127.0.0.1:20302/"
+OUT = os.path.join(os.path.dirname(__file__), "keys.json")
+
+
+def rpc_password():
+    cfg = json.load(open(DASHMATE_CONFIG))
```
Claudius-Maginificent, line 25: 🟢 **Idiom nits: files opened without context managers; bare `except Exception: pass`; magic 8-char proTxHash key with no collision guard**

Three small slips, grouped:

**(1) Unclosed file handles** — `json.load(open(DASHMATE_CONFIG))` (:25) and `json.dump(out, open(OUT, "w"), indent=2)` (:61) rely on CPython refcounting to close the handle. The `json.dump` case is the one that bites: on any non-refcounting interpreter, `keys.json` can be left **truncated** — and the whole run depends on that file.

**(2) Silent proxy failures** — `p2p_proxy.py` swallows every exception with bare `except Exception: pass` in `pipe()` and `handle()` (:14-20, :26), so a proxy that has silently stopped relaying is indistinguishable from one that is idle. Unhelpful for the one component the README calls "required for the new DET".

**(3) Truncated dict key** — `out[p["proTxHash"][:8]] = entry` (:60) uses an unexplained 8-hex-char (32-bit) truncation as a dict key with no collision guard; a collision silently drops a masternode's keys.

**Recommendation:** Use `with open(...) as f:` for both reads and the write; log the exception in the proxy's handlers (even a one-line `print(..., flush=True)`) instead of `pass`; key the dict on the full `proTxHash` (or assert no collision before writing) and truncate only for display in the print loop.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/03_fund_address.py
```diff
@@ -0,0 +1,54 @@
...
+import os
+import subprocess
+import sys
+import time
+
+DASHMATE_CONFIG = os.path.expanduser("~/.dashmate/config.json")
+SEED_RPC = "http://127.0.0.1:20302/"
+
+
+def call(pw, method, params=None, wallet=None):
```
Claudius-Maginificent, line 19: 🟢 **The entire Core-RPC `call()` helper and its config lookup are duplicated verbatim across `02` and `03`**

`DASHMATE_CONFIG`, `SEED_RPC`, and the 12-line `call(pw, method, params, wallet)` body are byte-identical in both files.

The password lookup is duplicated too, but **inconsistently**: `02` factors it into a named `rpc_password()` helper (02:24-26) while `03` inlines the same six-level subscript chain on one 100-column line (03:39). Two copies of an RPC client is where the third copy comes from — and the divergence between the named and inlined variants shows the drift started before the code even landed.

The magic `20302` is likewise pinned in two places here and a third in `docs/local-network.md`.

**Recommendation:** Extract a `dashmate_rpc.py` module next to the scripts exposing `rpc_password()` and `call()`, and import it from both. That also gives the error handling, timeouts, and the `dashmate config get` accessor recommended elsewhere in this review exactly one place to live.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/verify_signed_message.py
```diff
@@ -0,0 +1,137 @@
...
+    n = int.from_bytes(d, "big")
+    alphabet = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+    out = ""
+    while n > 0:
+        n, r = divmod(n, 58)
+        out = alphabet[r] + out
+    return "1" * (len(d) - len(d.lstrip(b"\x00"))) + out
+
+
+def recover_address(sig_b64: str, message: str, version: int) -> str:
```
Claudius-Maginificent, line 98: 🟢 **Hand-rolled secp256k1 recovery performs no input or curve validation; degenerate signatures crash with an unhandled traceback instead of the documented MISMATCH contract**

`recover_address()` implements ECDSA public-key recovery from scratch and validates nothing beyond the 65-byte length. Specifically: `r` and `s` are never range-checked against `[1, N-1]`; `header` is never checked to be in `27..34` (a header of 200 is silently accepted); `x = r + (recid >> 1) * N` is never checked to be `< P`; there is no low-S/malleability check; and critically the square root `y = pow(v, (P+1)//4, P)` — the Tonelli-Shanks shortcut valid only for `P ≡ 3 (mod 4)` — is never verified to satisfy `y² == x³ + 7`.

**Confirmed empirically:** 98 of the first 199 candidate `x` values are quadratic non-residues for which that shortcut returns a bogus `y`, after which `point_add`/`point_mul` happily operate on a point that is not on the curve and return a plausible-looking wrong address.

To be clear about what survived testing: the crypto **core is correct** — an independent re-implementation of ECDSA signing on the script's own primitives round-trips, wrong-message and bit-flipped inputs are correctly rejected, and hash160/Base58Check matches a known Bitcoin test vector. This finding is about malformed-input handling, not about the happy path.

**Recommendation:** Add the cheap guards that make failures self-describing: assert `1 <= r,s < N`; assert `27 <= header <= 34`; assert `x < P`; and after computing `y`, assert `(y*y - (x**3 + 7)) % P == 0`, raising a distinct "signature is malformed" error rather than proceeding onto a non-curve point. Print a clean `MISMATCH (malformed signature)` and exit 1 instead of letting exceptions reach a bare traceback — so a harness/input fault stays distinguishable from a real key-migration failure. The zero-dependency constraint is understandable and worth keeping.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/query_identity_dapi.sh
```diff
@@ -0,0 +1,20 @@
...
+PROTX="$1"
+PLATFORM="${2:-$HOME/Projects/dashpay/platform}"
+PROTO="$PLATFORM/packages/dapi-grpc/protos/platform/v0/platform.proto"
+GATEWAY="${GATEWAY:-127.0.0.1:2443}"   # dashmate local DAPI gateway (self-signed TLS)
+
+if [ -z "$PROTX" ]; then echo "usage: $0 <proTxHash_hex> [platform_repo_path]"; exit 2; fi
+ID_B64=$(python3 -c "import base64,sys; print(base64.b64encode(bytes.fromhex(sys.argv[1])).decode())" "$PROTX")
+
+echo ">> getIdentityBalance for $PROTX"
+grpcurl -insecure -max-time 15 -import-path "$PLATFORM/packages/dapi-grpc/protos" -proto "$PROTO" \
```
Claudius-Maginificent, line 18: 🟢 **`grpcurl -insecure` disables TLS verification while the gateway host is environment-overridable**

`-insecure` disables server certificate verification entirely. Against the documented default `GATEWAY=127.0.0.1:2443` — dashmate's local gateway with a self-signed cert — that is reasonable and necessary, and the inline comment says so.

The issue is that `GATEWAY` is deliberately env-overridable (`GATEWAY="${GATEWAY:-127.0.0.1:2443}"`) while `-insecure` is **hardcoded**. The moment an operator follows the SPEC's own advice that "an end-to-end withdrawal has to be tested on devnet/testnet" and re-points `GATEWAY` at a remote host, they get an unverified, silently MITM-able TLS channel with no warning.

Contrast `03_fund_address.py`, which hardcodes its RPC endpoint and therefore *cannot* be pointed at a remote host by a careless operator — that script gets this right.

**Recommendation:** Couple the flag to the target instead of hardcoding it: apply `-insecure` only when `GATEWAY` resolves to a loopback address, and require an explicit `INSECURE=1` opt-in otherwise. A two-line guard that errors out on a non-loopback `GATEWAY` without that opt-in closes this cleanly.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/SPEC.md
```diff
@@ -0,0 +1,105 @@
+# Masternode-identity upgrade test — spec
```
Claudius-Maginificent, line 1: 🟡 **`docs/testing/` invents a third docs location that matches no documented convention — and the branch is behind `v1.0-dev`, where the convention that governs it landed** *(merge-blocking)*

Two findings with one shared remediation.

**(1) Undeclared docs root.** The repo documents exactly two homes for this artifact. At this branch's merge-base (`87ba5b71`), CLAUDE.md already read: *"docs/ai-design should contain architecture, technical design and **manual testing scenarios** files, grouped in subdirectories prefixed with ISO-formatted date."* On current `origin/v1.0-dev`, CLAUDE.md keeps that rule **and** adds `docs/gui-testing/` as the home for reusable, non-date-grouped GUI testing practice. This PR creates `docs/testing/masternode-identity-upgrade/` — a third top-level docs tree that is (a) not under `docs/ai-design/`, (b) not ISO-date-prefixed, (c) not `docs/gui-testing/`, and (d) not mentioned anywhere in CLAUDE.md.

**(2) Branch is 2 commits behind `origin/v1.0-dev`, and GitHub reports `mergeable_state: "blocked"`.** Verified via git and the GitHub API, not the diff alone: `git merge-base HEAD origin/v1.0-dev` = `87ba5b71`; `git log 87ba5b71..origin/v1.0-dev --oneline` shows exactly 2 commits ahead — `1397b434 feat: rewrite Dash Evo Tool onto the new platform-wallet (#860)` and `3b32bed2 ci: replace weekly build check with weekly draft pre-release (#896)`. This is not cosmetic: **#860 is the very commit that introduced `docs/gui-testing/`**, so the convention that should govern this PR's placement is invisible from its base. In fairness, that is genuinely mitigating — the authors could not have read a convention that did not exist on their base.

**Recommendation:** Rebase onto current `origin/v1.0-dev` (docs-only diff — should be a no-op rebase) and confirm `mergeable_state` clears. Then re-decide placement with current conventions in view: move to `docs/ai-design/2026-07-XX-masternode-identity-upgrade/` if the SPEC is a dated point-in-time record, or put the manual scenario under `docs/gui-testing/scenarios/` if the intent is standing reusable practice. If `docs/testing/` is genuinely meant to be a new fourth category, declare it in CLAUDE.md's Documentation section **in this same PR** — an undeclared docs root is drift by definition.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/SPEC.md
```diff
@@ -0,0 +1,105 @@
...
+   the config and images; validators signalled protocol 12, which activated at
+   the next epoch boundary. From then on Drive emits `GroveDBProof::V1`, which
+   0.9.3's SDK cannot decode (expected — this is why the app must be upgraded).
+
+3. **Ran the new DET (PR #887) on a copy of the 0.9.3 profile.** The DB migrated
+   cleanly (schema v11 -> v38, no data loss) and the evonode identity appeared
+   in the new **Masternodes** section: Active, all three key roles (Voting /
+   Owner / Payout) and the derived voter identity intact.
+
+## Results
```
Claudius-Maginificent, line 41: 🟡 **PR is titled `test(masternode-upgrade)` but adds zero automated, asserting, or CI-enforced tests; the recorded pass claims are a frozen manual snapshot**

Confirmed via `git diff origin/v1.0-dev...HEAD --name-only`: 10 files changed, all under `docs/testing/masternode-identity-upgrade/`, **0 `.rs` files**. No `#[test]`, no `#[ignore]`-gated harness, nothing registered under `tests/backend-e2e/` — the project's own convention for network-dependent tests per CLAUDE.md and `tests/backend-e2e/README.md`. Nothing under `docs/**` matches the path filters in `tests.yml`/`clippy.yml`, so none of this is reachable from CI or any packaging path.

Beyond the absence of CI enforcement, the committed scripts do not **assert** the behaviours this Results section records as passes. No script inspects the migrated alias/key roles, compares pre/post state, or performs a post-upgrade DET operation. The single DAPI helper *prints* a balance; it does not check one. So "Migration: pass" is a claim about one operator's one run at one commit, not a reproducible verdict — and nothing will tell the next person if it regresses.

Stated fairly: the PR body is honest that "DET UI steps are manual today", and the harness genuinely is the deterministic groundwork an automated test could build on. The gap is between that honest framing and the `test(...)` title plus the unqualified "pass" verdicts here.

**Recommendation:** Either (a) reframe as docs rather than `test(...)` — this is a QA report plus a reusable manual runbook — or (b) wire the deterministic, non-GUI parts (key extraction, DAPI liveness query, signature verification) into an actual `#[ignore]`-gated backend-e2e test per `tests/backend-e2e/README.md`. Either way: record a pre-upgrade manifest (proTxHash, alias, role→address mapping, balance, schema version) and add an orchestrated post-upgrade check that **asserts** those values rather than printing them. Mark the GUI-driven parts explicitly as manual and re-run-required.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/SPEC.md
```diff
@@ -0,0 +1,105 @@
...
+
+- **Migration: pass.** Identity + keys + alias survive the 0.9.3 -> PR-887 DB
+  migration.
+- **On-chain liveness: pass.** Direct DAPI `getIdentity`/`getIdentityBalance`
```
Claudius-Maginificent, line 45: 🟢 **SPEC claims `getIdentity`/`getIdentityBalance` were used, but the script only ever calls `getIdentityBalance`**

This Results bullet states: *"Direct DAPI `getIdentity`/`getIdentityBalance` (bypassing DET) shows the identity live on protocol 12 with a real, growing balance — proof the migrated reference points at the genuine on-chain identity."*

`query_identity_dapi.sh` contains exactly **one** grpcurl invocation — `Platform/getIdentityBalance` (:20). There is no `getIdentity` call anywhere in the harness.

The distinction is not pedantic. `getIdentity` is what returns the identity's *key material and revision* — i.e. the direct on-chain corroboration that the migrated keys match the chain — whereas `getIdentityBalance` proves only that credits exist at that id. This bullet leans on the result as "proof the migrated reference points at the genuine on-chain identity", which is precisely the claim a balance query alone does **not** support.

**Recommendation:** Pick one and make them agree — either add the `getIdentity` grpcurl call to `query_identity_dapi.sh` (preferred; it is four lines and materially strengthens the evidence), or correct this bullet to say `getIdentityBalance` only and soften the "genuine on-chain identity" claim to what a balance query actually supports.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,115 @@
...
+`.gitignore`). The DET UI steps are still manual today — the value here is the
+deterministic network/keys/verification tooling an automated test can build on.
+
+## Prerequisites
+
+- Docker running; `npx` (Node) available.
+- A checkout of `dashpay/platform` (for DAPI `.proto` files) — default path
+  `~/Projects/dashpay/platform`.
+- `grpcurl` (for `query_identity_dapi.sh`).
+- Two DET builds: `v0.9.3` and the PR branch, each run against an **isolated
```
Claudius-Maginificent, line 14: 🟡 **Harness is macOS-only in practice but never says so, and omits `DASH_EVO_DATA_DIR` for the new DET**

*(Partially addressed in `7279666c` — the encrypted-wallet `sqlite3` check now documents both the macOS and Linux paths. Two coupled issues remain.)*

**(1) The macOS assumption is still unstated here.** Lines 14-18 tell the operator DET "derives its data dir from `$HOME` (`~/Library/Application Support/Dash-Evo-Tool` on macOS)" — the only path given in Prerequisites. This repo targets Linux (x86_64/aarch64), Windows **and** macOS, and CI is Linux; on Linux the path is `~/.config/dash-evo-tool` (`src/app_dir.rs`). Prerequisites list Docker and npx but never state "macOS only", so a Linux developer reads a `$HOME`-override rationale written against a path their build does not use.

**(2) `DASH_EVO_DATA_DIR` is never mentioned.** The `HOME`-override advice is **correct and justified for the 0.9.3 baseline** — that build predates `DASH_EVO_DATA_DIR`, so `HOME` really is the only lever. But the *new* DET supports `DASH_EVO_DATA_DIR` directly, which is what `docs/gui-testing/` safety rule 1 calls for, and it is the more precise instrument (isolates the data dir without relocating everything else keyed off `$HOME`). The reasoning behind the current advice is sound; it just is not written down — and undocumented good judgement gets refactored away by the next person.

**Recommendation:** State the macOS assumption in Prerequisites, or parameterize the profile path (`PROFILE_DB="${PROFILE_DB:-...}"`) and give the Linux equivalent alongside. Separately, note *why* the `HOME` override is required for 0.9.3, and use `DASH_EVO_DATA_DIR` for the new-DET steps.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,115 @@
...
+  `$HOME` (`~/Library/Application Support/Dash-Evo-Tool` on macOS), so this
+  keeps the test off your real profile. Snapshot the profile dir to back it up.
+
+## Order of operations
+
+1. **Phase 0:** Run `01_setup_v3_network.sh` to start a dashmate v3 local
+   network on protocol 11.
+2. **Phase 0:** Run `02_extract_masternode_keys.py` to write the gitignored
+   `keys.json` file.
+3. **Phase 1:** In *DET 0.9.3*, load an Evonode identity using a proTxHash and
```
Claudius-Maginificent, line 26: 🟢 **Manual DET UI steps re-invent the `docs/gui-testing/` tier instead of using its `TEMPLATE.md` and safety rules**

Steps 3, 8 and 10 of the walkthrough, and the whole "Additional check — encrypted wallet survives the upgrade" section, are manual driving of the compiled binary through a real display. That is precisely the tier `docs/gui-testing/` owns: it ships a canonical `scenarios/TEMPLATE.md` with structured sections (Verifies / Tier justification / Prerequisites / Setup / Procedure / Safety constraints / Expected outcome / Known gotchas) plus seven "Non-negotiable safety rules". This PR re-derives that structure ad hoc and re-derives the isolation rule in its own words.

**Mitigating context, stated plainly:** `docs/gui-testing/` landed on `v1.0-dev` via PR #860 (commit `1397b434`), which is one of the two commits this branch is behind. The authors could not have used a template that did not exist on their base — this is drift created by the branch being stale, not by carelessness.

**Recommendation:** Rebase onto current `v1.0-dev`, read `docs/gui-testing/README.md`, and re-express the manual UI portions as a scenario file from `scenarios/TEMPLATE.md`, keeping the genuinely novel dashmate/network/key tooling as scripts the scenario links to. Do not maintain two scenario formats.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,115 @@
+# Harness — masternode-identity upgrade test
+
+Reproducible scripts for the test described in [`../SPEC.md`](../SPEC.md). They
+regenerate everything locally; **no secrets or binaries are committed** (see
+`.gitignore`). The DET UI steps are still manual today — the value here is the
+deterministic network/keys/verification tooling an automated test can build on.
+
+## Prerequisites
```
Claudius-Maginificent, line 8: 🟢 **Harness README duplicates `docs/local-network.md` and diverges on how to obtain the RPC password**

`docs/local-network.md` already documents dashmate local-network setup, the `20302` seed RPC port, and the sanctioned way to read the RPC password:
```sh
dashmate config get core.rpc.users.dashmate.password --config=local_seed
```

The harness restates the prerequisites and instead parses `~/.dashmate/config.json` directly (`cfg['configs']['local_seed']['core']['rpc']['users']['dashmate']['password']`), hardcodes `SEED_RPC = 'http://127.0.0.1:20302/'`, and never references `docs/local-network.md`.

Two documented ways to obtain the same value is one too many — and the `config.json` path is an **internal layout that dashmate is free to change between v3 and v4**, which is a real hazard given this harness's entire purpose is spanning a v3→v4 upgrade.

**Recommendation:** Link `docs/local-network.md` from this README rather than restating prerequisites, and prefer the supported `dashmate config get` accessor over reaching into `config.json` internals. At minimum, note that the `config.json` shape is version-coupled.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/README.md
```diff
@@ -0,0 +1,115 @@
...
+   its owner, voting, and payout WIFs from `keys.json`. Create a wallet and copy
+   its receive address.
+4. **Phase 1:** Run
+   `python3 03_fund_address.py <receive-address> [amount]` to fund the DET wallet
+   from the Core `main` wallet and wait for confirmation.
+5. **Phase 1:** In *DET 0.9.3*, top up the identity, withdraw to L1, and snapshot
+   the profile.
+6. **Phase 2:** Run `04_upgrade_to_v4.sh` to upgrade the network in place from
+   v3 to v4 and wait for protocol 12.
+7. **Phase 3:** Run `p2p_proxy.py`. This is **required for the new DET**; see the
```
Claudius-Maginificent, line 36: 🟢 **The `NN_` step-numbering convention is applied to only 4 of the documented steps**

The walkthrough numbers 10 steps. Scripts `01`–`04` carry the `NN_` prefix; but `p2p_proxy.py` (step 7), `query_identity_dapi.sh` (step 9), and `verify_signed_message.py` (step 10) are unprefixed.

The convention self-evidently exists — it is just abandoned halfway, so an `ls` of the directory no longer tells you the running order and the README becomes the only source of sequence.

**Recommendation:** Either prefix the remaining ordered scripts (`05_p2p_proxy.py`, `07_query_identity_dapi.sh`, `08_verify_signed_message.py`) or drop the prefixes entirely and let the README own sequencing. Half a convention is worse than either whole one.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/p2p_proxy.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
Claudius-Maginificent, line 1: 🟢 **`p2p_proxy.py` permanently encodes a workaround for a hardcoded DET port with no tracking issue**

The gotcha claim checks out — `const REGTEST_P2P_PORT: u16 = 19899;` is confirmed at `src/wallet_backend/mod.rs:2528` on `origin/v1.0-dev`, so DET genuinely cannot reach dashmate's `local_1` Core on 20001. Credit for diagnosing it precisely.

But the PR's response is to commit a permanent TCP proxy and a README note, while the *other* limitation it found (dip0024) gets a prominent SPEC "Open question for the PR authors". A hardcoded regtest P2P port that makes the product unusable against a stock dashmate local network is a product defect of the same family and deserves the same visibility. As written, the workaround will quietly outlive the bug and nobody will know to delete it.

**Recommendation:** File a GitHub issue for the hardcoded `REGTEST_P2P_PORT` (making it configurable) and reference it (`#NNNN`) from this docstring and the README gotcha, so the proxy has a defined death date.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/.gitignore
```diff
@@ -0,0 +1,15 @@
+# Never commit regenerable local secrets / fixtures — they contain private keys
```
Claudius-Maginificent, line 1: 🟢 **`.gitignore` comment "(the test wallet has no password)" went stale within this same PR**

The header reads: *"Never commit regenerable local secrets / fixtures — they contain private keys and recoverable wallet seeds (the test wallet has no password)."*

A later commit on this branch adds the "Additional scenario: password-encrypted wallet survives the upgrade" flow, whose entire point is a wallet created **with** a strong password. So the parenthetical is now wrong for one of the two wallets the harness produces, within the same PR that wrote it.

Minor, but it is precisely the cross-artifact drift that compounds: the rationale for ignoring `*.db` is now half-stated.

**Recommendation:** Reword to state the invariant rather than a per-wallet fact — e.g. *"these files contain private keys and wallet seeds; a no-password wallet's seed is recoverable outright, and a password-protected one is still an offline-attackable target."* The ignore rule is right either way; only the justification needs to stop being version-specific.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (security-engineer-smythe + project-reviewer-adams + qa-engineer-marvin + Codex Sol)</sub>

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+# Requires: docker running, npx available. Wipes any existing ~/.dashmate.
+set -e
+
+DASHMATE_V3="${DASHMATE_V3:-dashmate@3.0.2}"   # last stable v3 line (3.1 never released)
+NODES="${NODES:-3}"
+
+echo ">> Removing any existing local network + ~/.dashmate"
+docker ps -a --format '{{.Names}}' | grep '^dashmate_' | xargs -r docker rm -f >/dev/null 2>&1 || true
+docker volume ls --format '{{.Name}}' | grep '^dashmate_' | xargs -r docker volume rm >/dev/null 2>&1 || true
+rm -rf ~/.dashmate
```
thepastaclaw, line 18: **🔴 Blocking: Require explicit consent before deleting Dashmate state**

The first documented workflow step force-removes every container and volume whose name begins with `dashmate_`, then recursively deletes the user's entire default `~/.dashmate` directory. These operations are not scoped to the network created by this harness, so unrelated Dashmate profiles, wallets, configuration, keys, and node data on the same host can be irreversibly destroyed without confirmation. Require an explicit destructive opt-in and run the harness against an isolated Dashmate home or configuration before deleting any state.

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+  npx -y "$DASHMATE_V3" config set --config="$cfg" \
+    platform.drive.tenderdash.consensus.createEmptyBlocksInterval 10s
+done
+
+echo ">> Starting group"
+npx -y "$DASHMATE_V3" group start --verbose
+
+echo ">> Waiting for platform blocks to flow"
+until curl -s --max-time 3 http://127.0.0.1:46657/status \
+      | python3 -c "import json,sys; sys.exit(0 if json.load(sys.stdin)['sync_info']['latest_block_height']!='1' else 1)" 2>/dev/null; do
```
thepastaclaw, line 34: **🔴 Blocking: Do not treat block height zero as network readiness**

The `until` command succeeds for every height except the literal string `"1"`. Tenderdash can serve `/status` with `latest_block_height` equal to `"0"` before producing its first block, so the script can immediately announce that the v3 network is ready and advance to key extraction against an unready network. Parse the height numerically and require it to advance beyond 1.

```suggestion
until curl -s --max-time 3 http://127.0.0.1:46657/status \
      | python3 -c "import json,sys; sys.exit(0 if int(json.load(sys.stdin)['sync_info']['latest_block_height']) > 1 else 1)" 2>/dev/null; do
```

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh
```diff
@@ -0,0 +1,38 @@
...
+  sleep 5
+done
```
thepastaclaw, line 36: **🟡 Suggestion: Bound the v3 setup readiness wait**

The prior timeout finding explicitly requested a finite bound for both the protocol-activation loop and this setup-readiness loop. The latest commit bounds `04_upgrade_to_v4.sh`, but this loop still waits forever if Dashmate, Docker, or the status endpoint never becomes usable. Add a maximum attempt count, a diagnostic timeout message, and a nonzero exit, matching the bounded pattern in the upgrade script.

<sub>source: ['codex']</sub>

docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py
thepastaclaw, line 61: **🟡 Suggestion: Create the private-key file with restricted permissions**

`open(OUT, "w")` creates `keys.json` using the process umask; with the common `022` setting, verification produced mode `0644`. The file contains every masternode's owner, voting, and payout WIF private keys, so it should not be readable by other local users. `.gitignore` prevents ordinary Git discovery but does not enforce filesystem access controls. Create the file as mode `0600` and explicitly restrict an existing file before writing it.

```suggestion
    fd = os.open(OUT, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
    os.fchmod(fd, 0o600)
    with os.fdopen(fd, "w") as output_file:
        json.dump(out, output_file, indent=2)
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Exact-head verification confirms two in-scope blocking defects in initial network setup: the script destructively wipes broad Dashmate state without explicit consent, and its readiness predicate accepts block height zero. Two suggestions also remain: the setup wait is unbounded, as identified by the partially resolved prior finding, and the extracted WIF file is created with permissive default filesystem permissions. Five of the six indexed prior findings are fully resolved.

Source: Codex reviewer `gpt-5.6-sol`; verifier `gpt-5.6-sol`.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 2 blocking | 🟡 2 suggestion(s)

<details>

```

In `docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh`:
- [BLOCKING] docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh:15-18: Require explicit consent before deleting Dashmate state
  The first documented workflow step force-removes every container and volume whose name begins with `dashmate_`, then recursively deletes the user's entire default `~/.dashmate` directory. These operations are not scoped to the network created by this harness, so unrelated Dashmate profiles, wallets, configuration, keys, and node data on the same host can be irreversibly destroyed without confirmation. Require an explicit destructive opt-in and run the harness against an isolated Dashmate home or configuration before deleting any state.
- [BLOCKING] docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh:33-34: Do not treat block height zero as network readiness
  The `until` command succeeds for every height except the literal string `"1"`. Tenderdash can serve `/status` with `latest_block_height` equal to `"0"` before producing its first block, so the script can immediately announce that the v3 network is ready and advance to key extraction against an unready network. Parse the height numerically and require it to advance beyond 1.
- [SUGGESTION] docs/testing/masternode-identity-upgrade/harness/01_setup_v3_network.sh:33-36: Bound the v3 setup readiness wait
  The prior timeout finding explicitly requested a finite bound for both the protocol-activation loop and this setup-readiness loop. The latest commit bounds `04_upgrade_to_v4.sh`, but this loop still waits forever if Dashmate, Docker, or the status endpoint never becomes usable. Add a maximum attempt count, a diagnostic timeout message, and a nonzero exit, matching the bounded pattern in the upgrade script.

In `docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py`:
- [SUGGESTION] docs/testing/masternode-identity-upgrade/harness/02_extract_masternode_keys.py:61: Create the private-key file with restricted permissions
  `open(OUT, "w")` creates `keys.json` using the process umask; with the common `022` setting, verification produced mode `0644`. The file contains every masternode's owner, voting, and payout WIF private keys, so it should not be readable by other local users. `.gitignore` prevents ordinary Git discovery but does not enforce filesystem access controls. Create the file as mode `0600` and explicitly restrict an existing file before writing it.
```
</details>

lklimek:
outdated, replaced by #983

PR #893, opened 2026-07-15
fix(dashpay): preserve paid-action guards across identity switch and paginate contact info

Claudius-Maginificent:
## Why this PR exists

- **Problem**: Two DashPay defects remain open in `ContactsState` and `create_or_update_contact_info`. A paid-action double-submit guard is discarded whenever the user switches identity, wallet, or network — even while the guarded backend task is still running. Separately, the owner-scoped `contactInfo` lookup reads only the first page of results despite a comment claiming it reads all of them, so metadata selection is made against an incomplete set.

- **What breaks without it**:
  1. *Paying twice for one action.* Start a paid contact action (its backend task is still in flight) → switch identity, wallet, or network → `reset_for_identity_change` wipes every double-submit guard, including the live one → the action can be submitted again while the first is still running, and the user pays for it twice.
  2. *Silent contact-data loss, or a duplicate document.* An identity with 100+ established contacts has some `contactInfo` documents beyond the first page. Editing a contact whose record sits past that page → the lookup never sees the existing document → the code either writes a fresh one (duplicating that contact's record) or proceeds without the existing field values, silently dropping the contact's accepted-accounts allow-list, nickname, and note.

- **Blocking relationship**: Stacked atop #894, which was squash-merged into this base and carried the rest of this PR's original scope with it. Everything else previously described here — the wallet cold-boot lockout, migration-gate coverage, shielded chokepoint, DashPay preservation defaults, and the passphrase transition-frame fixes — is already merged and is **no longer part of this PR**. What remains is the single unmerged commit below.

## What was done

- **Preserve live guards across a context change** — `ContactsState::reset_for_identity_change` keeps a paid-action guard whose backend task is genuinely still running, instead of wiping the whole guard set on an identity/wallet/network switch. View-only state still clears. Guards for abandoned confirmations are released, and timed-out entries are pruned, so nothing leaks and expired guards still self-heal.
- **Paginate owner-scoped `contactInfo`** — `create_or_update_contact_info` fetches every page before selecting update/create metadata, so the choice is made against the complete set rather than the first 100 documents.

## Testing

Verification is running against the rebased tree; results will be posted here once complete. The pre-existing suite (`cargo test --all-features --workspace`, `clippy --all-features --all-targets -D warnings`, `cargo +nightly fmt --all`) is the gate, plus the contacts/dashpay tests covering guard lifecycle and pagination.

## Breaking changes

None.

### DIFF
--- a/src/backend_task/dashpay/contact_info.rs
+++ b/src/backend_task/dashpay/contact_info.rs
@@ -19,12 +19,13 @@ use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
 use dash_sdk::dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
 use dash_sdk::dpp::identity::{KeyType, Purpose, SecurityLevel};
 use dash_sdk::dpp::platform_value::{Bytes32, Value};
-use dash_sdk::drive::query::{WhereClause, WhereOperator};
+use dash_sdk::drive::query::{OrderClause, WhereClause, WhereOperator};
 use dash_sdk::platform::documents::transitions::DocumentCreateTransitionBuilder;
 use dash_sdk::platform::proto::get_documents_request::get_documents_request_v0::Start;
 use dash_sdk::platform::{Document, DocumentQuery, FetchMany, Identifier};
 use dash_sdk::query_types::Documents;
 use std::collections::{BTreeMap, HashSet};
+use std::future::Future;
 use std::sync::Arc;
 use zeroize::Zeroizing;
 
@@ -389,6 +390,31 @@ fn next_contact_info_page(docs: &Documents) -> Option<Start> {
         .map(|last_id| Start::StartAfter(last_id.to_buffer().to_vec()))
 }
 
+async fn fetch_all_contact_info_documents<F, Fut>(
+    mut query: DocumentQuery,
+    mut fetch_page: F,
+) -> Result<Documents, TaskError>
+where
+    F: FnMut(DocumentQuery) -> Fut,
+    Fut: Future<Output = Result<Documents, TaskError>>,
+{
+    query.limit = CONTACT_INFO_PAGE_SIZE as u32;
+    let mut documents = Documents::new();
+
+    loop {
+        let page = fetch_page(query.clone()).await?;
+        let next_page = next_contact_info_page(&page);
+        documents.extend(page);
+
+        match next_page {
+            Some(start) => query.start = Some(start),
+            None => break,
+        }
+    }
+
+    Ok(documents)
+}
+
 async fn lookup_contact_info(
     app_context: &Arc<AppContext>,
     sdk: &Sdk,
@@ -408,58 +434,56 @@ async fn lookup_contact_info(
         operator: WhereOperator::Equal,
         value: Value::Identifier(identity_id.to_buffer()),
     });
-    query.limit = CONTACT_INFO_PAGE_SIZE as u32;
+    query = query.with_order_by(OrderClause {
+        field: "$updatedAt".to_string(),
+        ascending: true,
+    });
+    let existing_docs = fetch_all_contact_info_documents(query, |page_query| async move {
+        Document::fetch_many(sdk, page_query)
+            .await
+            .map_err(TaskError::from)
+    })
+    .await?;
+    let mut existing = None;
     let mut next_derivation_index = 0u32;
 
-    loop {
-        let existing_docs = Document::fetch_many(sdk, query.clone()).await?;
-        let next_page = next_contact_info_page(&existing_docs);
-
-        for doc in existing_docs.into_values().flatten() {
-            let props = doc.properties();
-            let Some(derivation_index) = props
-                .get("derivationEncryptionKeyIndex")
-                .and_then(|value| value.to_integer::<u32>().ok())
-            else {
-                continue;
-            };
-            next_derivation_index = next_derivation_index.max(derivation_index.saturating_add(1));
-            if props
-                .get("rootEncryptionKeyIndex")
-                .and_then(|value| value.to_integer::<u32>().ok())
-                .is_none()
-            {
-                continue;
-            }
-
-            let (enc_user_id_key, private_data_key) =
-                derive_contact_info_keys(app_context, identity, derivation_index).await?;
-            let Some(Value::Bytes(enc_user_id)) = props.get("encToUserId") else {
-                continue;
-            };
-            if decrypt_to_user_id(enc_user_id, &enc_user_id_key).ok()
-                == Some(contact_user_id.to_buffer())
-            {
-                return Ok(ContactInfoLookup {
-                    existing: Some(ExistingContactInfo {
-                        document: doc,
-                        derivation_index,
-                        enc_user_id_key,
-                        private_data_key,
-                    }),
-                    next_derivation_index,
-                });
-            }
+    for doc in existing_docs.into_values().flatten() {
+        let props = doc.properties();
+        let Some(derivation_index) = props
+            .get("derivationEncryptionKeyIndex")
+            .and_then(|value| value.to_integer::<u32>().ok())
+        else {
+            continue;
+        };
+        next_derivation_index = next_derivation_index.max(derivation_index.saturating_add(1));
+        if props
+            .get("rootEncryptionKeyIndex")
+            .and_then(|value| value.to_integer::<u32>().ok())
+            .is_none()
+        {
+            continue;
         }
 
-        match next_page {
-            Some(start) => query.start = Some(start),
-            None => break,
+        let (enc_user_id_key, private_data_key) =
+            derive_contact_info_keys(app_context, identity, derivation_index).await?;
+        let Some(Value::Bytes(enc_user_id)) = props.get("encToUserId") else {
+            continue;
+        };
+        if existing.is_none()
+            && decrypt_to_user_id(enc_user_id, &enc_user_id_key).ok()
+                == Some(contact_user_id.to_buffer())
+        {
+            existing = Some(ExistingContactInfo {
+                document: doc,
+                derivation_index,
+                enc_user_id_key,
+                private_data_key,
+            });
         }
     }
 
     Ok(ContactInfoLookup {
-        existing: None,
+        existing,
         next_derivation_index,
     })
 }
@@ -689,6 +713,10 @@ pub async fn create_or_update_contact_info(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use dash_sdk::dpp::system_data_contracts::{SystemDataContract, load_system_data_contract};
+    use dash_sdk::dpp::version::PlatformVersion;
+    use std::collections::VecDeque;
+    use std::future::ready;
 
     const KEY: [u8; 32] = [7u8; 32];
     const OTHER_KEY: [u8; 32] = [9u8; 32];
@@ -749,6 +777,43 @@ mod tests {
             updated_at_core_block_height: None,
             transferred_at_core_block_height: None,
         })
+    }
+
+    #[tokio::test]
+    async fn contact_info_fetch_walks_every_page() {
+        let contract =
+            load_system_data_contract(SystemDataContract::Dashpay, PlatformVersion::latest())
+                .expect("DashPay contract fixture");
+        let query = DocumentQuery::new(contract, "contactInfo").expect("contactInfo query");
+        let first_page_last_id = id(100);
+        let first_page = (1..=100)
+            .map(|byte| (id(byte), None))
+            .collect::<Documents>();
+        let second_page = [(id(101), None)].into_iter().collect::<Documents>();
+        let mut pages = VecDeque::from([first_page, second_page]);
+        let mut calls = 0;
+
+        let documents = fetch_all_contact_info_documents(query, |query| {
+            calls += 1;
+            match calls {
+                1 => assert!(query.start.is_none()),
+                2 => assert!(matches!(
+                    query.start,
+                    Some(Start::StartAfter(ref cursor))
+                        if cursor == &first_page_last_id.to_buffer().to_vec()
+                )),
+                _ => panic!("the short second page must end pagination"),
+            }
+            ready(Ok::<_, TaskError>(
+                pages.pop_front().expect("a page for every fetch"),
+            ))
+        })
+        .await
+        .expect("all pages fetched");
+
+        assert_eq!(calls, 2);
+        assert_eq!(documents.len(), 101);
+        assert_eq!(documents.keys().last().copied(), Some(id(101)));
     }
 
     #[test]
--- a/src/backend_task/dashpay/contact_info.rs
+++ b/src/backend_task/dashpay/contact_info.rs
@@ -381,6 +381,8 @@ struct ContactInfoLookup {
     next_derivation_index: u32,
 }
 
+type ContactInfoKeys = (Zeroizing<[u8; 32]>, Zeroizing<[u8; 32]>);
+
 const CONTACT_INFO_PAGE_SIZE: usize = 100;
 
 fn next_contact_info_page(docs: &Documents) -> Option<Start> {
@@ -415,6 +417,88 @@ where
     Ok(documents)
 }
 
+fn contact_info_derivation_index(document: &Document) -> Option<u32> {
+    document
+        .properties()
+        .get("derivationEncryptionKeyIndex")
+        .and_then(|value| value.to_integer::<u32>().ok())
+}
+
+fn has_contact_info_key_material(document: &Document) -> bool {
+    contact_info_derivation_index(document).is_some()
+        && document
+            .properties()
+            .get("rootEncryptionKeyIndex")
+            .and_then(|value| value.to_integer::<u32>().ok())
+            .is_some()
+        && matches!(
+            document.properties().get("encToUserId"),
+            Some(Value::Bytes(_))
+        )
+}
+
+fn find_existing_contact_info<F>(
+    documents: Vec<Document>,
+    contact_user_id: Identifier,
+    mut derive_keys: F,
+) -> Option<ExistingContactInfo>
+where
+    F: FnMut(u32) -> Result<ContactInfoKeys, TaskError>,
+{
+    for document in documents {
+        let Some(derivation_index) = contact_info_derivation_index(&document) else {
+            continue;
+        };
+        if !has_contact_info_key_material(&document) {
+            continue;
+        }
+
+        let Ok((enc_user_id_key, private_data_key)) = derive_keys(derivation_index) else {
+            continue;
+        };
+        let Some(Value::Bytes(enc_user_id)) = document.properties().get("encToUserId") else {
+            continue;
+        };
+        if decrypt_to_user_id(enc_user_id, &enc_user_id_key).ok()
+            == Some(contact_user_id.to_buffer())
+        {
+            return Some(ExistingContactInfo {
+                document,
+                derivation_index,
+                enc_user_id_key,
+                private_data_key,
+            });
+        }
+    }
+
+    None
+}
+
+async fn scan_contact_info_documents<F, Fut>(
+    documents: Vec<Document>,
+    contact_user_id: Identifier,
+    with_key_session: F,
+) -> Result<ContactInfoLookup, TaskError>
+where
+    F: FnOnce(Vec<Document>, Identifier) -> Fut,
+    Fut: Future<Output = Result<Option<ExistingContactInfo>, TaskError>>,
+{
+    let next_derivation_index = documents
+        .iter()
+        .filter_map(contact_info_derivation_index)
+        .fold(0u32, |next, index| next.max(index.saturating_add(1)));
+    let existing = if documents.iter().any(has_contact_info_key_material) {
+        with_key_session(documents, contact_user_id).await?
+    } else {
+        None
+    };
+
+    Ok(ContactInfoLookup {
+        existing,
+        next_derivation_index,
+    })
+}
+
 async fn lookup_contact_info(
     app_context: &Arc<AppContext>,
     sdk: &Sdk,
@@ -444,48 +528,43 @@ async fn lookup_contact_info(
             .map_err(TaskError::from)
     })
     .await?;
-    let mut existing = None;
-    let mut next_derivation_index = 0u32;
+    let documents = existing_docs.into_values().flatten().collect();
+    scan_contact_info_documents(
+        documents,
+        contact_user_id,
+        |documents, contact_user_id| async move {
+            let seed_hash = identity
+                .dashpay_wallet_seed_hash()
+                .ok_or(TaskError::ContactWalletSeedUnavailable)?;
+            let network = identity.network;
 
-    for doc in existing_docs.into_values().flatten() {
-        let props = doc.properties();
-        let Some(derivation_index) = props
-            .get("derivationEncryptionKeyIndex")
-            .and_then(|value| value.to_integer::<u32>().ok())
-        else {
-            continue;
-        };
-        next_derivation_index = next_derivation_index.max(derivation_index.saturating_add(1));
-        if props
-            .get("rootEncryptionKeyIndex")
-            .and_then(|value| value.to_integer::<u32>().ok())
-            .is_none()
-        {
-            continue;
-        }
-
-        let (enc_user_id_key, private_data_key) =
-            derive_contact_info_keys(app_context, identity, derivation_index).await?;
-        let Some(Value::Bytes(enc_user_id)) = props.get("encToUserId") else {
-            continue;
-        };
-        if existing.is_none()
-            && decrypt_to_user_id(enc_user_id, &enc_user_id_key).ok()
-                == Some(contact_user_id.to_buffer())
-        {
-            existing = Some(ExistingContactInfo {
-                document: doc,
-                derivation_index,
-                enc_user_id_key,
-                private_data_key,
-            });
-        }
-    }
-
-    Ok(ContactInfoLookup {
-        existing,
-        next_derivation_index,
-    })
+            app_context
+                .wallet_backend()?
+                .secret_access()
+                .with_secret_session(
+                    &crate::wallet_backend::SecretScope::HdSeed { seed_hash },
+                    async |session| {
+                        let plaintext = session.plaintext();
+                        let seed = plaintext
+                            .expose_hd_seed()
+                            .ok_or(TaskError::ContactWalletSeedUnavailable)?;
+                        Ok(find_existing_contact_info(
+                            documents,
+                            contact_user_id,
+                            |derivation_index| {
+                                crate::wallet_backend::derive_contact_info_encryption_keys(
+                                    seed,
+                                    network,
+                                    derivation_index,
+                                )
+                            },
+                        ))
+                    },
+                )
+                .await
+        },
+    )
+    .await
 }
 
 pub(super) async fn contact_info_is_hidden(
@@ -715,6 +794,7 @@ mod tests {
     use super::*;
     use dash_sdk::dpp::system_data_contracts::{SystemDataContract, load_system_data_contract};
     use dash_sdk::dpp::version::PlatformVersion;
+    use std::cell::Cell;
     use std::collections::VecDeque;
     use std::future::ready;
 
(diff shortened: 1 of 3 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 2c69239d)

src/app.rs
```diff
@@ -1758,13 +1768,11 @@ impl App for AppState {
...
-        // the block to suppress its focus management so the prompt keeps the keyboard.
-        ProgressOverlay::render_global(ctx, self.active_secret_prompt.is_some());
+        // A blocking progress overlay remains active underneath a secret prompt,
+        // but renders no dimmer, card, or focus trap until the prompt resolves.
+        // Every passphrase prompt — cancellable or not — supplies its own
+        // outside-window input barrier in its place (`passphrase_modal`).
+        ProgressOverlay::render_global(ctx, self.has_blocking_secret_prompt());
 
         // Render any just-in-time passphrase prompt on top of the screen.
         self.render_secret_prompt(ctx);
```
thepastaclaw, line 2084: **🔴 Blocking: Install passphrase input barriers before rendering the screen**

When a queued secret prompt becomes active, `claim_overlay_input()` deliberately yields, the underlying screen processes the frame's input, and only afterward does `render_secret_prompt()` install the modal sink. Migration password prompts follow the same sequence because `is_prompting()` yields before `update_password_prompt()` renders at line 1801. In egui 0.35, `set_modal_layer()` records the modal for the current frame while widget interaction checks the modal layer carried from the previous frame, so a sink installed after an underlying button has returned `clicked()` cannot revoke that action. Preinstall the prompt's modal barrier before invoking the visible screen; the current kittest primes the modal for a frame before clicking and therefore does not cover this transition frame.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Three in-scope blocking issues remain at the exact head. Unsupported shielded tasks initialize wallet storage before their capability check, the migration gate loses DashPay request correlation, and newly activated passphrase prompts install their input barrier too late to prevent first-frame click-through.

### Review provenance

🔴 3 blocking

<details>

```

In `src/backend_task/mod.rs`:
- [BLOCKING] src/backend_task/mod.rs:587-599: Reject unavailable shielded tasks before wallet initialization
  A dispatched `ShieldedTask` reaches `ensure_wallet_backend()` before `run_shielded_task()` checks `FeatureGate::ShieldedOperations`. On a cold context, initialization opens and hydrates wallet storage, then `bootstrap_loaded_wallets()` resolves every open wallet seed and can register wallets, bind Orchard keys, and register contact accounts. The capability is unavailable on every current protocol version, so unsupported requests can still cross the wallet and secret boundary or return a terminal storage-open error before the intended refusal. Check shielded availability before this generic initialization path.
- [BLOCKING] src/backend_task/mod.rs:607-613: Preserve DashPay request identity at the migration gate
  The migration gate returns a bare `WalletStorageNotReady` before `run_dashpay_task()` can attach the request ID for Accept, Reject, or Cancel. The legacy contact-request screen therefore cannot match the failure and leaves its request state and loading flag set. The Identity Hub instead clears every paid-action guard for this error, but backend tasks run concurrently and migration publishes `Running` asynchronously: request A can still be executing when request B is rejected by this gate, causing B's error to release A's guard and permit a duplicate paid submission. Wrap the gate error with the rejected DashPay action's request ID so each screen retires only that action.

In `src/app.rs`:
- [BLOCKING] src/app.rs:1752-1778: Install passphrase input barriers before rendering the screen
  When a queued secret prompt becomes active, `claim_overlay_input()` deliberately yields, the underlying screen processes the frame's input, and only afterward does `render_secret_prompt()` install the modal sink. Migration password prompts follow the same sequence because `is_prompting()` yields before `update_password_prompt()` renders at line 1801. In egui 0.35, `set_modal_layer()` records the modal for the current frame while widget interaction checks the modal layer carried from the previous frame, so a sink installed after an underlying button has returned `clicked()` cannot revoke that action. Preinstall the prompt's modal barrier before invoking the visible screen; the current kittest primes the modal for a frame before clicking and therefore does not cover this transition frame.
```
</details>

src/app.rs
Claudius-Maginificent: Fixed and independently verified — twice.

- **`7de49aa9`**: `AppState::update()` now detects the inactive→active transition on prompt activation (`prompt_blocking && !prompt_was_blocking`) and neutralizes that frame's pointer click via a new `passphrase_modal::drop_activation_frame_pointer_click(ctx)` helper, in addition to (not instead of) installing the modal sink — matches your diagnosis of the egui 0.35 `set_modal_layer`/`Focus::end_pass` one-frame lag exactly.
- **`3591ea8c`**: the first two pinning tests (`transition_frame_click_leaks_through_a_newly_activated_prompt` / `..._migration_prompt`) called the extracted helper directly rather than driving the real update loop, so an independent QA pass caught that mutating the actual `app.rs` call site left the suite green. Closed with two new tests (`appstate_jit_prompt_activation_drops_transition_frame_click`, `appstate_migration_prompt_activation_drops_transition_frame_click`) that build a real `AppState` via `Harness::builder().build_eframe(...)` and drive it through `harness.step()`. Mutation-killed against the real production call site, independently reproduced by a second agent from a clean isolated build.

On severity: independently re-derived as **LOW** (risk 0.35 / impact 0.25 / scope 0.40, OWASP-normalized) rather than Blocking — real and worth fixing (which we did), but a single stray click on a narrow transition frame, not a fund-loss or data-integrity path. Happy to discuss if you weigh it differently.

Now on `origin/integration/review-remediation` @ `3591ea8c`.

Claudius-Maginificent:
@thepastaclaw both blocking findings from this review are fixed in `3e69b2fd` (now on `origin/integration/review-remediation` @ `3591ea8c`):

**Shielded pre-check ordering (`src/backend_task/mod.rs:587-599`)**: `run_backend_task` now checks `FeatureGate::ShieldedOperations::is_available()` for any `BackendTask::ShieldedTask` as its very first statement, before the shared `ensure_wallet_backend()` bootstrap runs. Your read of `bootstrap_loaded_wallets()`'s reach (seed exposure, upstream registration, Orchard key binding, contact-account registration) was accurate — this makes the gate's own "before any wallet/backend access" claim literally true instead of aspirational.

**DashPay gate request-identity scoping (`src/backend_task/mod.rs:607-613`, `src/ui/identity/hub_screen.rs:576-587`)**: added `dashpay_request_id(&BackendTask) -> Option<Identifier>`, matching `AcceptContactRequest`/`RejectContactRequest`/`CancelContactRequest`. The migration gate now wraps a rejected DashPay contact action in `TaskError::DashPayContactRequestActionFailed { request_id, source }` instead of returning a bare `WalletStorageNotReady`, and `release_request_guard_for_error` was narrowed to release *only* the matching request's guard — the old blanket `WalletStorageNotReady => clear_in_flight()` arm was deleted entirely (confirmed: zero remaining callers).

On severity: both independently re-derived as **LOW**, not Blocking:
- Shielded gate — `risk 0.29 / impact 0.06 / scope 0.20`. No shielded operation could actually execute (the in-handler gate still refused before any fund movement); UI shielded buttons are separately gated; reachable only via MCP/direct dispatch; seed material never left the sanctioned secret-seam chokepoint or the caller.
- DashPay guard — `risk 0.21 / impact 0.07 / scope 0.20`. The re-entry window is real, but all three contact actions re-check freshly-fetched Platform state before acting (not the clicked row), and DashPay contact-request documents carry a Platform-enforced uniqueness constraint — a re-dispatched action in the TOCTOU window gets a clean rejection or no-op, not a silent duplicate. Also a net narrowing of pre-existing base behavior (the prior blanket-clear already accepted "worst case: a re-enabled row" as its tradeoff).

Both fixes were independently mutation-tested (revert → RED with the exact regression symptom; restore → GREEN) by an agent that didn't write them, plus a full `cargo test --all-features --workspace` / `clippy -D warnings` / `fmt --check` gate. Full write-up in the PR description.

thepastaclaw:
Thanks — I see the shielded and DashPay remediation in `3e69b2fd`, plus the transition-frame production-wiring regression coverage at head `3591ea8c`. I’ve prioritized an independent exact-head re-review, with this thread context included as evidence.

src/ui/state/contacts_view.rs
```diff
@@ -82,6 +92,11 @@ impl ContactsState {
         self.hidden.clear();
         self.show_hidden = false;
         self.search.clear();
+    }
+
+    /// Reset identity-scoped data, including guards that belong to the old identity.
+    pub fn reset_for_identity_change(&mut self) {
+        self.reset();
         self.in_flight.clear();
```
thepastaclaw: **🔴 Blocking: Identity switching clears guards for paid actions that are still running**

`reset_for_identity_change()` clears every paid-action guard, but switching identity, wallet, or network does not cancel tasks already spawned by `AppState`. A user can start Accept, Decline, or Cancel for identity A, switch away and back while that task is running, reload the pending request, and dispatch it again because `begin_request()` no longer sees the original guard. This defeats the PR's per-request double-submit protection and can submit two paid state transitions. Retain guards independently of the selected view and key them by network, acting identity, and request ID; release only the guard named by its matching result, error, or timeout.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Current head fixes all three previously indexed blockers. Two in-scope DashPay blockers remain: identity switching can release a paid-action guard while its task is still running, and contact-info preservation examines only the first 100 documents. Diff and formatting checks passed; focused test execution remains unavailable because the required Tenderdash download was declined.

### Review provenance

🔴 2 blocking

<details>

```

In `src/ui/state/contacts_view.rs`:
- [BLOCKING] src/ui/state/contacts_view.rs:97-100: Identity switching clears guards for paid actions that are still running
  `reset_for_identity_change()` clears every paid-action guard, but switching identity, wallet, or network does not cancel tasks already spawned by `AppState`. A user can start Accept, Decline, or Cancel for identity A, switch away and back while that task is running, reload the pending request, and dispatch it again because `begin_request()` no longer sees the original guard. This defeats the PR's per-request double-submit protection and can submit two paid state transitions. Retain guards independently of the selected view and key them by network, acting identity, and request ID; release only the guard named by its matching result, error, or timeout.

In `src/backend_task/dashpay/contact_info.rs`:
- [BLOCKING] src/backend_task/dashpay/contact_info.rs:400-407: Preserve silently fails for contact records beyond the first page
  The owner-scoped lookup stops after 100 documents and does not paginate. `found_existing_doc` controls both the `Preserve` input at lines 473-474 and the replace-versus-create branch at line 528, so a target outside that page is treated as absent: preserved fields are resolved from empty data and the code attempts to create another document. The contract's unique owner/root/derivation index may reject that create when the truncated high-water mark collides; if the selected coordinate is free, a second document with empty nickname, note, and accepted-account data can be created. Either outcome violates this PR's contact-metadata preservation guarantee. Fetch every page in deterministic order and compute the derivation high-water mark across all owned documents before concluding that the target is absent.
```
</details>

src/ui/state/contacts_view.rs
Claudius-Maginificent: Fixed and independently verified.

`ContactsState::reset_for_identity_change` no longer wipes `in_flight` wholesale — a guard for a request whose backend task is still genuinely running now survives an identity/wallet/network switch. What still gets cleared: the view-only caches (incoming/outgoing/contacts/search) via the existing `reset()`, expired guards (past `REQUEST_IN_FLIGHT_TIMEOUT`) are physically pruned, and an *abandoned* contact-info overwrite-confirmation dialog releases only its own guard (same release logic already used when a user manually cancels that dialog) — genuinely-running actions stay guarded either way.

The pre-existing test that pinned the old (buggy) clear-everything behavior as intended was revised to pin the corrected behavior instead: request-2's guard survives the identity-change reset while the view state still clears and a pre-seeded expired guard is pruned.

On the guard-key design: your report didn't prescribe a specific scheme, but the simplest fix (drop the wholesale clear, keep the guard keyed by bare `request_id`) is sound rather than needing identity/network scoping — contact-request document IDs are `SHA256d(contract_id ‖ owner_id ‖ document_type ‖ 32 bytes of fresh entropy)` (`generate_document_id_v0` in rs-dpp), so a collision across two identities' unrelated requests would require breaking SHA256d or an owner_id match plus an entropy repeat. Independently traced to source, not assumed.

Independently re-derived severity: **LOW** (risk 0.61 / impact 0.10 / scope 0.20, OWASP-normalized) — reachable via ordinary navigation (no special skill needed), but the realistic worst case is a resubmitted paid action that Platform's document-revision concurrency check likely rejects rather than a true double state-change, and the blast radius is one narrow interaction path.

Fixed in commit `27605357` on `integration/review-remediation` — verification complete, pending push to origin.

Claudius-Maginificent:
@thepastaclaw fixed and independently verified.

`create_or_update_contact_info`'s owner-scoped `contactInfo` lookup now paginates fully via `Start::StartAfter`, ordered `$updatedAt ASC` on the existing `ownerIdAndUpdatedAt` index, before concluding a target contact's document is absent — both the existing-document search and the derivation-index high-water mark now cover every page the identity owns, not just the first 100. The misleading "Get all contact info documents" comment on the old single-page fetch is corrected along with it. Added a deterministic 100+1-document pagination test (fake multi-page fetch seam) proving the loop actually walks a second page rather than asserting it in prose.

Independently re-derived severity: **LOW** (risk 0.51 / impact 0.15 / scope 0.18, OWASP-normalized) — real and directly relevant given this PR's own contact-metadata preservation guarantee, but gated behind an identity owning 100+ established `contactInfo` documents, an uncommon precondition at current adoption.

Fixed in the same commit as the identity-switch guard finding above: `27605357` on `integration/review-remediation` — verification complete (mutation-tested by an agent that didn't write the fix, full `fmt`/`clippy`/`cargo test --all-features --workspace` gate clean), pending push to origin.

src/ui/dashpay/contact_requests.rs
```diff
@@ -151,6 +157,10 @@ impl ContactRequests {
             self.outgoing_requests.clear();
             self.has_fetched_requests = false;
             self.pending_profile_fetches.clear();
+            self.request_in_flight = None;
+            self.request_task_in_flight = None;
+            self.overwrite_confirmation_dialog = None;
+            self.loading = false;
```
thepastaclaw, line 163: **🔴 Blocking: Identity switching re-enables an unfinished paid request**

`set_selected_identity()` clears `request_in_flight` and `request_task_in_flight`, but changing identities does not cancel the backend Accept or Reject task already spawned by `AppState`. Switching away and back reloads the still-pending request with active buttons, so it can dispatch a second paid state transition while the first remains in progress. Preserve live guards independently of the selected identity and release them only for their matching task outcome or a deliberate timeout.

<sub>source: ['codex']</sub>

src/ui/dashpay/contact_requests.rs
```diff
@@ -879,7 +953,23 @@ impl ScreenLike for ContactRequests {
...
+                self.request_in_flight.as_ref() == Some(request_id)
+            }
+            _ => false,
+        };
+        if completed_request {
+            self.request_in_flight = None;
+            self.request_task_in_flight = None;
+        }
+        if self.request_in_flight.is_none() {
+            self.loading = false;
```
thepastaclaw, line 989: **🔴 Blocking: Navigating away can strand the paid-action guard**

This guard is released only when this persistent Contacts root screen receives the matching result. `AppState` sends ordinary task results exclusively to `visible_screen_mut()`, so navigating to My Profile, another root screen, or a detail screen before completion delivers the result elsewhere. Returning to Requests leaves `request_in_flight` set and `loading` true, and this implementation has no timeout; the spinner remains until an identity switch or restart. Move completion tracking to a task-scoped owner that survives navigation, or route correlated outcomes back to the originating screen.

<sub>source: ['codex']</sub>

src/ui/dashpay/contact_details.rs
```diff
@@ -620,11 +663,15 @@ impl ScreenLike for ContactDetailsScreen {
...
                 }
             }
-            BackendTaskSuccessResult::DashPayContactInfoUpdated(contact_id) => {
-                if contact_id == self.contact_id {
+            BackendTaskSuccessResult::DashPayContactInfoUpdated {
+                identity,
+                contact_id,
+            } => {
+                if identity == self.identity.identity.id() && contact_id == self.contact_id {
+                    self.commit_pending_update();
```
thepastaclaw, line 678: **🔴 Blocking: Navigating away drops the local contact-info commit**

The PR moved the local nickname, note, and hidden-state write into `commit_pending_update()`, which runs only when this screen receives the successful Platform result. The loading view still exposes Back and global navigation, so the screen can be popped before completion; because results go only to the currently visible screen, the confirmed Platform write then never updates the WalletBackend private-info sidecar. Reopening the detail or using an offline view reads the stale sidecar until a network refresh happens. Commit the sidecar in the backend success path or another task-scoped owner that survives navigation.

<sub>source: ['codex']</sub>

src/backend_task/dashpay/contact_requests.rs
```diff
@@ -848,38 +852,50 @@ pub async fn reject_contact_request(
...
+///
+/// The marker is the only thing that retires the row — Platform keeps the
+/// `contactRequest` document forever — so a failed write must surface rather
+/// than be reported as a completed rejection.
+fn complete_rejection(
+    request_id: Identifier,
+    mark_declined: impl FnOnce() -> Result<(), TaskError>,
+) -> Result<BackendTaskSuccessResult, TaskError> {
+    mark_declined()?;
     Ok(BackendTaskSuccessResult::DashPayContactRequestRejected(
```
thepastaclaw, line 1027: **🟡 Suggestion: Sidecar failure loses the paid action's partial-success state**

`create_or_update_contact_info()` has already confirmed the paid Platform replacement before `mark_declined()` can fail. The resulting request-scoped error releases the UI guard, but without the local marker the immutable request appears again on reload; retrying performs another paid replacement before retrying the same marker. Cancellation has the equivalent split-brain path at lines 964-973. Represent this as a partial-completion outcome and retry only the idempotent marker through a durable marker-only task instead of reopening the paid action.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At exact head 27605357859d94d796efd7a1659c3923513d3f15, both carried-forward prior findings are fixed, so no prior blockers remain. The latest remediation delta introduces no new finding, but the cumulative PR range still contains three blocking legacy DashPay lifecycle and persistence defects plus one partial-success handling suggestion. Formatting and cumulative/latest diff checks passed; focused tests could not run because approval to download the pinned Tenderdash source was declined.

### Review provenance

<details>

```

In `src/ui/dashpay/contact_requests.rs`:
- [BLOCKING] src/ui/dashpay/contact_requests.rs:160-163: Identity switching re-enables an unfinished paid request
  `set_selected_identity()` clears `request_in_flight` and `request_task_in_flight`, but changing identities does not cancel the backend Accept or Reject task already spawned by `AppState`. Switching away and back reloads the still-pending request with active buttons, so it can dispatch a second paid state transition while the first remains in progress. Preserve live guards independently of the selected identity and release them only for their matching task outcome or a deliberate timeout.
- [BLOCKING] src/ui/dashpay/contact_requests.rs:955-971: Navigating away can strand the paid-action guard
  This guard is released only when this persistent Contacts root screen receives the matching result. `AppState` sends ordinary task results exclusively to `visible_screen_mut()`, so navigating to My Profile, another root screen, or a detail screen before completion delivers the result elsewhere. Returning to Requests leaves `request_in_flight` set and `loading` true, and this implementation has no timeout; the spinner remains until an identity switch or restart. Move completion tracking to a task-scoped owner that survives navigation, or route correlated outcomes back to the originating screen.

In `src/ui/dashpay/contact_details.rs`:
- [BLOCKING] src/ui/dashpay/contact_details.rs:666-671: Navigating away drops the local contact-info commit
  The PR moved the local nickname, note, and hidden-state write into `commit_pending_update()`, which runs only when this screen receives the successful Platform result. The loading view still exposes Back and global navigation, so the screen can be popped before completion; because results go only to the currently visible screen, the confirmed Platform write then never updates the WalletBackend private-info sidecar. Reopening the detail or using an offline view reads the stale sidecar until a network refresh happens. Commit the sidecar in the backend success path or another task-scoped owner that survives navigation.

In `src/backend_task/dashpay/contact_requests.rs`:
- [SUGGESTION] src/backend_task/dashpay/contact_requests.rs:898-899: Sidecar failure loses the paid action's partial-success state
  `create_or_update_contact_info()` has already confirmed the paid Platform replacement before `mark_declined()` can fail. The resulting request-scoped error releases the UI guard, but without the local marker the immutable request appears again on reload; retrying performs another paid replacement before retrying the same marker. Cancellation has the equivalent split-brain path at lines 964-973. Represent this as a partial-completion outcome and retry only the idempotent marker through a durable marker-only task instead of reopening the paid action.
```
</details>

src/backend_task/dashpay/contact_info.rs
```diff
@@ -408,58 +434,56 @@ async fn lookup_contact_info(
...
+            && decrypt_to_user_id(enc_user_id, &enc_user_id_key).ok()
+                == Some(contact_user_id.to_buffer())
+        {
+            existing = Some(ExistingContactInfo {
+                document: doc,
+                derivation_index,
+                enc_user_id_key,
+                private_data_key,
+            });
         }
```
thepastaclaw: **🔴 Blocking: The full scan repeatedly unlocks the wallet and can fail after a match**

The new full-owner scan invokes `derive_contact_info_keys()` for every valid document, including every document after `existing` has been populated. Each call enters `SecretAccess::with_secret`; with the default `RememberPolicy::None`, a protected wallet is decrypted only for that call, so an identity with 101 contactInfo documents can require 101 password submissions for one save, decline, cancel, or unhide. Continuing after a match also propagates unrelated derivation failures: the contract permits any non-negative integer, but hardened BIP32 rejects indices at or above 2^31, so a later document can abort an already-successful lookup. Compute the public high-water mark without key access, hold one secret session while deriving candidate keys, and stop deriving once the target is found.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At exact head f9d641f53b5f361fb1db7519ff9076e8420ca40d, two carried-forward blockers remain: hidden legacy Requests screens miss correlated completions, and popped contact-detail screens lose their local sidecar commit. The latest delta adds one blocker in the full contact-info scan; repeated protected-wallet prompts and failures from later invalid derivation indices are both valid consequences of the same loop and are combined below. The shared backend claim and durable action phases fix the other two prior findings.

### Review provenance

🔴 3 blocking

<details>

```

In `src/backend_task/dashpay/contact_info.rs`:
- [BLOCKING] src/backend_task/dashpay/contact_info.rs:467-482: The full scan repeatedly unlocks the wallet and can fail after a match
  The new full-owner scan invokes `derive_contact_info_keys()` for every valid document, including every document after `existing` has been populated. Each call enters `SecretAccess::with_secret`; with the default `RememberPolicy::None`, a protected wallet is decrypted only for that call, so an identity with 101 contactInfo documents can require 101 password submissions for one save, decline, cancel, or unhide. Continuing after a match also propagates unrelated derivation failures: the contract permits any non-negative integer, but hardened BIP32 rejects indices at or above 2^31, so a later document can abort an already-successful lookup. Compute the public high-water mark without key access, hold one secret session while deriving candidate keys, and stop deriving once the target is found.
```
</details>

src/backend_task/dashpay/contact_requests.rs
thepastaclaw: Resolved in this update — _Sidecar failure loses the paid action's partial-success state_ no longer present.

src/ui/dashpay/contact_requests.rs
thepastaclaw: Resolved in this update — _Identity switching re-enables an unfinished paid request_ no longer present.

src/ui/dashpay/contact_requests.rs
lklimek: Independently verified: **still live**, but out of this PR's scope — not fixed here.

This PR was rebased down to a single commit (`f9d641f5`) touching only `src/backend_task/dashpay/contact_info.rs`, `src/ui/identity/hub_screen.rs`, and `src/ui/state/contacts_view.rs`. `src/ui/dashpay/contact_requests.rs` is no longer in the diff — it reached the base via #894's squash-merge, so this is pre-existing base code rather than something this PR introduces or touches.

Confirming your diagnosis against current code rather than closing it as merely out-of-scope:

- `0fda1f67` added `route_contact_request_result_to_hidden_hub` / `route_contact_request_error_to_hidden_hub` (`app.rs:1625-1647`), but that forwarding is **Hub-specific** — it only targets `main_screens[RootScreenIdentityHub]`, never `Screen::DashPayScreen`. The catch-all fallback (`app.rs:1926-1931`) still routes strictly to `visible_screen_mut()`.
- `ContactRequests::display_task_result` / `display_task_error` (`contact_requests.rs:930-990`) therefore still clear `request_in_flight` / `loading` only when the result lands on that exact instance, and this widget has no timeout.
- The path is genuinely reachable, not dead code: Identity Hub → Settings → "Change photo" (`settings.rs:250-258`) → Contacts tab (`dashpay_subscreen_chooser_panel.rs:29`) → `RootScreenDashPayContacts`. `left_panel.rs:19-25` documents that the Dashpay entries are hidden from nav but deliberately kept reachable.

Worth noting the modern Hub analog this PR *does* touch is engineered correctly for exactly this: `ContactsState::reset_for_identity_change` retains live guards and prunes only past `REQUEST_IN_FLIGHT_TIMEOUT` (`contacts_view.rs:18-20, 81-99`). The legacy widget has neither routing coverage nor a timeout.

Independently re-derived severity: **LOW** (risk 0.4 / impact 0.3 / scope 0.15) rather than Blocking. It is reachable in a few clicks, but the realistic worst case is a stuck spinner and disabled buttons on one de-emphasised legacy screen, self-healing on identity switch or restart — no double-pay (the backend `try_claim_contact_request_action` chokepoint blocks that independently of any UI flag) and no data loss.

Tracked as a follow-up against the legacy widget rather than fixed here, to keep this PR at the one commit that is actually its own.

src/ui/dashpay/contact_details.rs
lklimek: Independently verified: the mechanism you describe is **still present in the code**, but the screen containing it is currently **unreachable**, and the file is out of this PR's scope — not fixed here.

Scope: this PR was rebased to a single commit (`f9d641f5`) touching only `src/backend_task/dashpay/contact_info.rs`, `src/ui/identity/hub_screen.rs`, and `src/ui/state/contacts_view.rs`. `src/ui/dashpay/contact_details.rs` reached the base via #894's squash-merge.

The bug itself is exactly as you describe, unchanged:

- `commit_pending_update()` (`contact_details.rs:208-247`) remains the sole writer of the WalletBackend private-info sidecar (`persist_contact_private_info`, `ui/dashpay/mod.rs:33-48`), called only from `display_task_result()` on `DashPayContactInfoUpdated` (`:673-684`).
- The backend task performs only the Platform document write and returns `DashPayContactInfoUpdated` — no sidecar call anywhere in `create_or_update_contact_info` (`contact_info.rs:523-711`), including after this PR's rewrite of it.
- "Back" is rendered unconditionally (`contact_details.rs:267-268`), not gated on `self.loading`.
- I checked `IdentityHubScreen::display_task_result`'s `DashPayContactInfoUpdated` arm (`hub_screen.rs:559-575`) as a possible alternate commit path: it only acts when `pending_contact_info_tasks.remove(&key).is_some()`, never calls `persist_contact_private_info`, and is a no-op for a key it did not register. There is no backend-side or Hub-side fallback commit.

What changes the priority: a repo-wide search for `DashPayContactDetails` finds `ScreenType::DashPayContactDetails` defined and pattern-matched only inside `src/ui/mod.rs`'s own enum/match machinery — **zero call sites construct it**, in production UI, tests, or MCP/CLI. Contact clicks route to `DashPayContactProfileViewer` instead (`contacts_list.rs:836-837`, `contacts.rs:520-526`). So the screen compiles and is exhaustively matched, but nothing in the shipped app can navigate to it.

Severity as shipped: **INFO** (risk ≈ 0, scope ≈ 0 — unreachable). As latent code it is a real landmine: restoring navigation to this screen, or reusing the screen-scoped-commit pattern, reactivates the bug with no other code change. Note `contact_profile_viewer.rs:144` independently calls `persist_contact_private_info` — worth a look under the same lens, though outside what I verified here.

Tracked as a follow-up (fix the pattern, or delete the unreachable screen) rather than fixed in this PR.

src/backend_task/dashpay/contact_info.rs
lklimek: Confirmed and fixed in `2c69239d`. Both halves were real, and this one is squarely this PR's own regression — paginating the scan is what made the per-document unlock unbounded, where the old single-page read capped it at 100. Good catch.

Restructured exactly along the lines you prescribed:

- **Public high-water mark, no key access.** `next_derivation_index` is now computed in a first pass from document properties alone.
- **One secret session.** The second pass derives candidate keys inside a single `SecretAccess::with_secret_session` (the existing seam API — no new secret path, and the seed stays borrowed inside the seam with derived keys still `Zeroizing`). The session callback is `FnOnce`, so the one-session invariant is enforced by the type rather than by a comment.
- **Stop on match.** The scan returns at the first matching document; nothing is derived past the target.
- **A single document no longer aborts the scan.** Derivation failures — including the `>= 2^31` hardened-BIP32 case you identify — are skipped rather than propagated, so a match found earlier survives.

One judgement call worth surfacing, since you didn't prescribe it: **an underivable index still advances the high-water mark.** Index `2^31` yields `next_derivation_index == 2^31 + 1` even though key derivation skips it — the document occupies that coordinate on Platform whether or not a key can be derived for it, so treating it as free would risk colliding with the contract's unique owner/root/derivation index. Indices that are missing, wrong-typed, negative, or outside `u32` do *not* advance it, since neither `ContactInfoLookup` nor the derivation API can represent that coordinate. `saturating_add` guards the `u32::MAX` edge.

Verification — three regression tests, each **confirmed failing before the fix** (red run exited 101 with all three failing), then green:

- `contact_info_scan_uses_one_secret_session_for_many_documents` — 101 candidates → 101 derivations, exactly **1** session (was 101)
- `contact_info_scan_stops_deriving_after_target_match`
- `contact_info_scan_skips_late_hardened_index_and_keeps_high_water_mark`

All three confirmed by exact-name grep in the test log rather than inferred from a green exit. Full module suite 21 passed; `clippy --all-features --all-targets -D warnings` clean; `cargo +nightly fmt --all` clean. Both `lookup_contact_info` consumers (hidden-state read and create/update) keep their existing `ContactInfoLookup` / `UnreadableContactInfoPolicy` behaviour, and the full pagination still runs before the scan.

src/backend_task/dashpay/contact_info.rs
```diff
@@ -408,60 +518,53 @@ async fn lookup_contact_info(
...
-    }
-
-    Ok(ContactInfoLookup {
-        existing: None,
-        next_derivation_index,
+    query = query.with_order_by(OrderClause {
+        field: "$updatedAt".to_string(),
+        ascending: true,
+    });
+    let existing_docs = fetch_all_contact_info_documents(query, |page_query| async move {
```
thepastaclaw, line 525: **🔴 Blocking: Mutable pagination order can skip contact records**

The scan orders documents by `$updatedAt`, then continues with the last document ID from each page. Platform resolves that ID to the document's current contents before constructing the next index query. If another contact-info operation replaces the boundary document between requests, its updated timestamp moves it later in the index, so the next page starts from the new position and skips documents between the old and new positions. Operations for different contacts or request IDs are not serialized per owner. An omitted target is therefore treated as absent, causing creation with incomplete allocation metadata; this can create a duplicate contact record or fail on an index collision. Use a stable owner/key index for pagination or serialize the complete scan-and-write operation per owner.

<sub>source: ['codex']</sub>

src/backend_task/dashpay/contact_info.rs
```diff
@@ -389,6 +392,113 @@ fn next_contact_info_page(docs: &Documents) -> Option<Start> {
...
+    with_key_session: F,
+) -> Result<ContactInfoLookup, TaskError>
+where
+    F: FnOnce(Vec<Document>, Identifier) -> Fut,
+    Fut: Future<Output = Result<Option<ExistingContactInfo>, TaskError>>,
+{
+    let next_derivation_index = documents
+        .iter()
+        .filter_map(contact_info_derivation_index)
+        .fold(0u32, |next, index| next.max(index.saturating_add(1)));
```
thepastaclaw, line 489: **🔴 Blocking: Skipped invalid index still poisons new-contact allocation**

The DashPay contract accepts every non-negative integer derivation index, while hardened BIP32 accepts only values below 2^31. The lookup now skips a document whose index cannot be derived, but this calculation still advances `next_derivation_index` beyond that document. When no existing document matches the requested contact, `create_or_update_contact_info` passes the resulting underivable value to key derivation and fails. One contract-valid high index therefore prevents this identity from creating contact information for every new contact. Track occupied derivable indices and choose an unused value below 2^31 instead of advancing beyond an index the key API cannot represent.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At exact head 2c69239d596c261243c6530b1a9b396e41c77770, focused tests could not compile because the build requires an unavailable Tenderdash download. Carried-forward prior findings: the identity-switch and partial-success defects are fixed; the legacy spinner defect remains real but is intentionally deferred outside the current three-file PR diff; the Contact Details defect is outdated because that screen is unreachable. New findings in the latest delta: the mutable pagination cursor and underivable high-water allocation are blocking defects.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (completed)

🔴 2 blocking

<details>

```

In `src/backend_task/dashpay/contact_info.rs`:
- [BLOCKING] src/backend_task/dashpay/contact_info.rs:521-525: Mutable pagination order can skip contact records
  The scan orders documents by `$updatedAt`, then continues with the last document ID from each page. Platform resolves that ID to the document's current contents before constructing the next index query. If another contact-info operation replaces the boundary document between requests, its updated timestamp moves it later in the index, so the next page starts from the new position and skips documents between the old and new positions. Operations for different contacts or request IDs are not serialized per owner. An omitted target is therefore treated as absent, causing creation with incomplete allocation metadata; this can create a duplicate contact record or fail on an index collision. Use a stable owner/key index for pagination or serialize the complete scan-and-write operation per owner.
- [BLOCKING] src/backend_task/dashpay/contact_info.rs:486-489: Skipped invalid index still poisons new-contact allocation
  The DashPay contract accepts every non-negative integer derivation index, while hardened BIP32 accepts only values below 2^31. The lookup now skips a document whose index cannot be derived, but this calculation still advances `next_derivation_index` beyond that document. When no existing document matches the requested contact, `create_or_update_contact_info` passes the resulting underivable value to key derivation and fails. One contract-valid high index therefore prevents this identity from creating contact information for every new contact. Track occupied derivable indices and choose an unused value below 2^31 instead of advancing beyond an index the key API cannot represent.
```
</details>

src/ui/dashpay/contact_requests.rs
thepastaclaw: Resolved in this update — _Navigating away can strand the paid-action guard_ no longer present.

src/ui/dashpay/contact_details.rs
thepastaclaw: Resolved in this update — _Navigating away drops the local contact-info commit_ no longer present.

PR #894, opened 2026-07-15
fix: rebase SND-003 onto PR893 + QA-campaign fix batches

lklimek:
## Why this PR exists
- **Problem**: The SND-003 "Receive (QR) is inert" fix needed rebasing onto PR892, and a full desktop-GUI walkthrough of the 175 user stories surfaced a batch of real defects (a crash, silent hangs, dead buttons, a wallet self-lockout, a missing delete confirmation) plus security gaps in local data deletion.
- **What breaks without it**: Updating a data contract crashes the app; Load-Identity / Load-Masternode / Fetch-Documents / Add-Token hang silently with no feedback; transfer-credits and several token buttons are dead; a locked wallet becomes permanently stuck; single-key wallets delete with no confirmation; and "Clear Database" can report success while identity private keys (including masternode voting/owner/payout keys) survive on disk.

## What was done
- **SND-003**: the Receive (QR) dialog no longer dismisses itself on the click that opens it.
- **QA fixes**: DOC-002 crash → typed error; IDN-002 / MN-001 / DOC-004 / TOK-003 silent hangs → timeouts + typed errors; IDN-006 / TOK-005 / TOK-011 / TOK-018 dead buttons; WAL-005/006/007 (rename, unlock self-lockout, single-key removal confirmation); SND-005 fee + total shown before send; IDN-008/009/013a / DPN-008 navigation; MCP-001 cross-process wallet visibility; NEW-002/003/004.
- **Security**: SEC-001/002 — "Clear Database" now wipes identity private keys and surfaces partial failures fail-closed (was silent / warn-only). `clear_identity_vault_keys` propagates instead of swallowing read/decode/delete errors, `IdentityKeyView::delete_all` attempts every key rather than short-circuiting, and DashPay sidecar/overlay delete failures are accumulated instead of warn-only.
- **New controls**: NEW-006 developer-only "Wipe Platform Data" — on the **Network Chooser, in the "Database Maintenance" section beside "Clear {Network} Database"**, alongside its sibling destructive controls (devnet only, type-WIPE-to-confirm). NEW-007 "View Profile" for contacts.
- **Hardening**: database-version migration safety gate; contact-request idempotency; platform dependency bump pulling the AssetLockProof rehydration fix; unified test env-var lock (removes a flaky cross-module test race); per-instance modal guard id so two popups on one screen can no longer dismiss each other.

## Test plan
- [x] `cargo build` — clean
- [x] `cargo clippy --all-features --all-targets -- -D warnings` — clean (verified via a forced, non-cached recompile)
- [x] `cargo test --all-features --workspace` — 2142 passed, **0 failed** (68 network-gated backend-e2e ignored, as expected)
- [x] Suite verified green under **both** parallel and serial execution — the previously flaky cross-module `DASH_EVO_DATA_DIR` test race is fixed in this PR
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] Desktop-GUI walkthrough across the 175-story catalog drove the QA fix batches; an annotated status report tracks every finding as Fixed/Open
- [x] **NEW-006 exercised live in the GUI** (Developer role, fresh data dir): the control renders in "Database Maintenance" directly beneath "Clear Devnet Database"; the type-WIPE dialog holds (empty → confirm disabled, lowercase `wipe` → still disabled, Enter does not bypass, `WIPE` → enabled); confirming dispatches without panic. Gate verified live by round-trip with the role held constant — Devnet: present → Testnet: **absent** → Devnet: present.
- [ ] **NEW-007 ("View Profile") is not yet exercised live in the GUI**, nor is NEW-006's removal from the Identity Hub → Settings tab. Both screens require a loaded identity (and, for NEW-007, established DashPay contacts). The QA campaign's dataset cannot currently be opened by this branch: another checkout migrated the shared wallet storage in `~/.config/dash-evo-tool` to schema 4, and this branch's `platform-wallet` pin supports max 3, so it fails closed with `SchemaVersionUnsupported { found: 4, max_supported: 3 }`. NEW-006's removal from the Settings tab is nonetheless certain statically — the button, its dialog field, and its helpers no longer exist in `settings.rs`, so no code path can render it. The security, modal-guard and test-lock fixes remain covered by unit + integration tests only.

## Breaking changes
- None.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 0e79a89b)
_Canonical validated blockers: 9_

src/mcp/resolve.rs
```diff
@@ -118,34 +118,44 @@ pub(crate) fn wallet_arc(
...
-/// Poll until the cold-start storage migration is no longer running.
+/// Wire the wallet backend so persisted wallets are available in memory.
+///
+/// Unlike [`ensure_spv_synced`], this does not start SPV or wait for chain sync.
+pub(crate) async fn ensure_wallets_hydrated(ctx: &Arc<AppContext>) -> Result<(), McpToolError> {
+    let (tx, _) = tokio::sync::mpsc::channel::<crate::app::TaskResult>(32);
+    let sender = crate::utils::egui_mpsc::SenderAsync::new(tx, egui::Context::default());
+    ctx.ensure_wallet_backend(sender)
+        .await
+        .map_err(McpToolError::TaskFailed)
```
thepastaclaw: **🔴 Blocking: Run headless migration before resolving hydrated wallets**

`ensure_wallets_hydrated` only constructs the backend, whose initial hydration reads the modern sidecars. The migration tests explicitly establish that these sidecars are empty before `FinishUnwire` imports legacy `data.db` and rehydrates the context. Wallet-facing tools call this helper and then resolve the wallet before `ensure_spv_synced`, which is the path that dispatches `FinishUnwire`; Platform-only tools never reach that path, and `core_wallets_list` reads the empty map directly. Consequently legacy wallets appear missing or produce an empty list, and protected installations never receive the instruction to complete migration in the desktop app. The pre-resolution readiness path must run or join `FinishUnwire` and use its post-migration hydration before exposing `ctx.wallets`.

<sub>source: ['codex']</sub>

src/backend_task/dashpay/contact_requests.rs
```diff
@@ -848,38 +852,50 @@ pub async fn reject_contact_request(
...
-    }
+    // DashPay has no on-chain "rejected" flag, so a DET-local marker retires the
+    // row. `DashpayView::contact_requests` reads it by counterparty id under the
+    // acting identity's scope, and it is incoming-only: it must not silence a
+    // request we later send to that same person.
+    complete_rejection(request_id, || {
+        app_context
+            .wallet_backend()?
+            .dashpay_mark_declined(&owner_id, &from_identity_id)
+    })
```
thepastaclaw: **🔴 Blocking: Do not rebroadcast a paid action when only its local marker failed**

The hidden `contactInfo` document has already been successfully broadcast before `dashpay_mark_declined` runs. If the local write fails, the error releases the request guard and explicitly tells the user to try again; retrying calls `create_or_update_contact_info` again, which bumps the document revision and purchases another replacement before retrying the marker. `cancel_flow` has the same ordering after its hide broadcast. Track the successful network commit separately so retries reconcile only the local marker instead of broadcasting and paying for another transition.

<sub>source: ['codex']</sub>

src/ui/state/contacts_view.rs
```diff
@@ -210,25 +225,26 @@ impl ContactsState {
...
+        true
     }
 
     /// Whether an action for this request is already running. Drives the row's
     /// disabled state, so the user sees why the buttons do not respond.
     pub fn is_in_flight(&self, request_id: &Identifier) -> bool {
-        self.in_flight.contains(request_id)
+        self.in_flight.get(request_id).is_some_and(|started| {
+            Instant::now().saturating_duration_since(*started) < REQUEST_IN_FLIGHT_TIMEOUT
+        })
```
thepastaclaw, line 241: **🔴 Blocking: Do not expire a paid-action guard without backend completion**

After five minutes, `is_in_flight` reports the request as idle and `begin_request` replaces its timestamp, allowing a second Accept, Decline, or Cancel. The backend action has no matching five-minute outer timeout, cancellation, or completion reconciliation; it performs several sequential SDK fetch and transition awaits, so the original task can still commit after the UI dispatches the duplicate. Release the guard only from an authoritative result, or add an actual backend timeout/cancellation protocol that reconciles whether the transition committed before allowing another paid action.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Exact head 82a96aefb33a45cb59066ddeda914a04bd0d35e3 was verified against the full PR range. Four in-scope blocking defects remain: legacy wallets are resolved before headless migration, DashPay retries can purchase duplicate transitions, paid-action guards can expire while work remains active, and scheduled votes can age out while migration waits for wallet passwords.

### Review provenance

🔴 4 blocking

<details>

```

In `src/mcp/resolve.rs`:
- [BLOCKING] src/mcp/resolve.rs:124-129: Run headless migration before resolving hydrated wallets
  `ensure_wallets_hydrated` only constructs the backend, whose initial hydration reads the modern sidecars. The migration tests explicitly establish that these sidecars are empty before `FinishUnwire` imports legacy `data.db` and rehydrates the context. Wallet-facing tools call this helper and then resolve the wallet before `ensure_spv_synced`, which is the path that dispatches `FinishUnwire`; Platform-only tools never reach that path, and `core_wallets_list` reads the empty map directly. Consequently legacy wallets appear missing or produce an empty list, and protected installations never receive the instruction to complete migration in the desktop app. The pre-resolution readiness path must run or join `FinishUnwire` and use its post-migration hydration before exposing `ctx.wallets`.

In `src/backend_task/dashpay/contact_requests.rs`:
- [BLOCKING] src/backend_task/dashpay/contact_requests.rs:868-872: Do not rebroadcast a paid action when only its local marker failed
  The hidden `contactInfo` document has already been successfully broadcast before `dashpay_mark_declined` runs. If the local write fails, the error releases the request guard and explicitly tells the user to try again; retrying calls `create_or_update_contact_info` again, which bumps the document revision and purchases another replacement before retrying the marker. `cancel_flow` has the same ordering after its hide broadcast. Track the successful network commit separately so retries reconcile only the local marker instead of broadcasting and paying for another transition.

In `src/ui/state/contacts_view.rs`:
- [BLOCKING] src/ui/state/contacts_view.rs:239-242: Do not expire a paid-action guard without backend completion
  After five minutes, `is_in_flight` reports the request as idle and `begin_request` replaces its timestamp, allowing a second Accept, Decline, or Cancel. The backend action has no matching five-minute outer timeout, cancellation, or completion reconciliation; it performs several sequential SDK fetch and transition awaits, so the original task can still commit after the UI dispatches the duplicate. Release the guard only from an authoritative result, or add an actual backend timeout/cancellation protocol that reconciles whether the transition committed before allowing another paid action.

In `src/app.rs`:
- [BLOCKING] src/app.rs:1794-1798: Pause scheduled-vote sweeps until migration imports identities
  `FinishUnwire` imports scheduled votes first, then can remain in `AwaitingWalletPasswords` during wallet registration, and imports legacy identities only afterward. The periodic `CastDueScheduledVotes` task is dispatched unconditionally and `ContestedResourceTask` is not covered by the migration gate. While migration waits, the sweep finds the imported vote but no corresponding modern voting identity and silently skips it. Because eligibility ends 120 seconds after the scheduled time, a password prompt left open beyond that window makes the vote permanently ineligible for automatic casting. Defer these sweeps while migration is in progress or preserve the vote's eligibility until its identity import completes.
```
</details>

Claudius-Maginificent:
I reviewed commit `82a96aef` across the 90-file diff. CI is green, and the dependency pin audit found no new advisory or package-surface regression.

The pending inline review contains 1 HIGH, 3 MEDIUM, and 1 LOW findings that are not covered by the existing review. The merge blocker is:

- **Clear Database can return success without deleting persisted wallet data** when backend wiring is pending or failed.

I also found cross-task UI state corruption, a permanently stranded token-refresh guard after a true hang, and a migration-modal transition-frame click-through. I omitted two independently confirmed findings—MCP migration ordering and the paid DashPay partial-commit retry—because `thepastaclaw` posted equivalent inline comments while this review was being prepared.

Please also refresh the PR body: it still says later QA batches will land later and its test counts predate much of the current diff, including the migration and destructive-storage changes.

src/context/wallet_lifecycle/spv.rs
```diff
@@ -21,14 +21,13 @@ impl AppContext {
...
         // "delete all local data" promise holds — wallets must NOT rehydrate
         // on next launch and encrypted seeds must NOT persist. Clear the
         // persisted state (seed-envelope vault, wallet-meta + single-key
         // sidecars, shielded notes, session cache) BEFORE the in-memory maps
-        // below, so a mid-failure crash cannot strand a recoverable seed. The
+        // below, so a mid-failure crash cannot strand current state. The
+        // pre-update database remains a read-only recovery artifact. The
         // upstream (watch-only) persistor rows have no seed and are removed
         // asynchronously off the main thread. Best-effort when the backend is
         // not wired yet — there is no such state in that case.
```
Claudius-Maginificent: **HIGH — the wipe can report success while persistent wallets remain.** `wallet_backend()` can be unavailable while startup wiring is pending or has failed even though vault/sidecar files from a prior run exist. In that case every persistent-store deletion is skipped, the in-memory maps are still cleared, and this method returns `Ok(())`; the UI reports success and the wallets can rehydrate later. Please make current-store availability a precondition (or erase the known stores independently), return a typed error when deletion cannot complete, and add an unwired/failed-backend test. The new regression test always wires the backend first.

src/ui/contracts_documents/contracts_documents_screen.rs
```diff
@@ -541,13 +554,13 @@ impl ScreenLike for DocumentQueryScreen {
...
     }
 
-    fn display_message(&mut self, message: &str, message_type: MessageType) {
+    fn display_message(&mut self, _message: &str, message_type: MessageType) {
         // Banner display is handled globally by AppState; this is only for side-effects.
-        if message.contains("Error fetching documents")
-            && matches!(message_type, MessageType::Error | MessageType::Warning)
+        if self
+            .document_query_status
+            .fail_if_fetch_in_flight(message_type)
```
Claudius-Maginificent: **MEDIUM — this cannot tell which task failed.** `AppState` forwards every unhandled backend error to the currently visible screen using only a `MessageType`. If a document query is running while any concurrent task errors, this marks the query failed and clears its banner although its future is still active. The same new pattern appears in token claims and token-balance refresh, where it can re-enable duplicate work. Please correlate terminal state with the originating operation/request (or route typed results to their owner), and replace the helper-only tests with an AppState-level test showing unrelated failure B cannot terminate operation A.

src/backend_task/tokens/query_my_token_balances.rs
```diff
@@ -34,14 +49,29 @@ impl AppContext {
...
-        self.refresh_upstream_token_balances(watch_sets, &sender)
-            .await?;
+        let refresh_guard = self.begin_token_balance_refresh()?;
+        let context = Arc::clone(self);
+        await_managed_network_request_with_timeout(
+            self.subtasks.clone(),
+            "token_balance_refresh_reaper",
+            NETWORK_REQUEST_TIMEOUT,
+            async move {
+                let _refresh_guard = refresh_guard;
```
Claudius-Maginificent, line 59: **MEDIUM — a true hang strands this guard forever.** The guard is owned by the managed request and resets only when that future ends. On timeout, the helper moves the `JoinHandle` into an unbounded reaper; if the upstream future never completes, every later refresh returns `TokenBalanceRefreshInProgress` for the rest of the process even though the timeout message tells the user to refresh the Tokens screen. Please provide a cancellation-safe/bounded recovery path (or accurately require restart while the latch remains held) and test a never-completing future; the current test only completes shortly after timeout.

src/app.rs
```diff
@@ -1743,6 +1807,22 @@ impl App for AppState {
...
+        // its first visible frame never shares a pointer sink or focus trap.
+        self.activate_secret_prompt(ctx);
+
+        // On the frame a passphrase prompt first becomes active — a just-in-time
+        // unlock promoted above, or the migration password prompt — egui has
+        // already resolved this frame's click against the previous, prompt-less
+        // frame, before the modal installs its input sink. Drop that one pending
+        // click so it cannot fall through to the screen beneath; the sink covers
+        // every later frame.
+        let prompt_blocking = self.has_blocking_secret_prompt();
```
Claudius-Maginificent: **MEDIUM — the migration prompt can become active after this snapshot.** Migration state is published asynchronously, while `migration.update_banner()` reconciles/renders its password prompt only after the visible screen has consumed input. If state changes between this line and that later reconciliation, the underlying screen returns an action, the prompt appears, and the action is still dispatched; the next frame's pointer drop is too late. Please reconcile migration prompting before this snapshot and use one state snapshot for input ownership and rendering (or defer a newly detected prompt to the next frame), with a mid-frame transition test.

docs/user-stories.md
```diff
@@ -1346,7 +1358,8 @@ As a user, while a long operation that is unsafe to interrupt is running (broadc
 
 - A full-window dimming overlay with an indeterminate spinner and an optional "Step N of M" counter and description appears while the operation runs, and lowers automatically when it finishes (success or error).
 - All interaction beneath the block is suppressed: pointer clicks hit a sink, and keyboard/text input is claimed at frame start so nothing reaches a focused field beneath (FR-8 / QA-001). The block is never dismissable by Esc, Enter, Space, or Tab.
-- The block yields to a passphrase prompt: when a secret prompt is shown above the overlay it keeps the keyboard (Enter/Esc/Tab) so the user can still authenticate or cancel (SEC-004).
+- The block yields completely to a passphrase prompt: it remains active but paints no dimmer, pointer sink, card, or focus trap until the prompt resolves, so the user can type and use every prompt action (SEC-004).
```
Claudius-Maginificent: **LOW — `SEC-004` is a transient review finding ID.** Repository policy keeps ephemeral audit IDs out of committed specifications because they lose meaning outside the originating review artifact. Please remove it or replace it with a durable requirement, issue, or design-document reference.

src/mcp/resolve.rs
lklimek: Verified against current code (82a96aef): confirmed real. `ensure_wallets_hydrated` only wires the backend; `FinishUnwire` is only dispatched from `ensure_spv_synced`. Several tool call sites (`core_wallets_list`, and multiple identity/shielded tools) call only the former, so legacy wallets pending migration read as absent. Tracked for a dedicated fix — not resolving yet.

src/backend_task/dashpay/contact_requests.rs
lklimek: Verified against current code (82a96aef): confirmed real. The `contactInfo` broadcast in `create_or_update_contact_info` happens before `complete_rejection`'s local marker write; a local-write failure surfaces a retry-worthy error with no record that the paid broadcast already succeeded, so a retry re-broadcasts and re-pays. Tracked for a dedicated fix — not resolving yet.

src/ui/state/contacts_view.rs
lklimek: Verified against current code (82a96aef): confirmed real. `is_in_flight` is purely time-based (5-minute `REQUEST_IN_FLIGHT_TIMEOUT`) with no link to actual backend task completion, so a slow-but-still-running paid action past that window would allow a duplicate dispatch. Tracked for a dedicated fix — not resolving yet.

Claudius-Maginificent (commented):
I found five actionable inline issues not covered by the existing review (1 HIGH, 3 MEDIUM, 1 LOW). Two independently confirmed findings were omitted as duplicates. This review is intentionally pending so you can inspect and publish it manually.

src/context/wallet_lifecycle/spv.rs
lklimek: Verified against current code (82a96aef): confirmed real. `clear_network_database` treats an unwired backend as "nothing to clear" (`if let Ok(backend) = self.wallet_backend()` silently skips the whole persisted-secret-store deletion block), but that assumption is only true on a genuinely fresh install — a prior run's seed-envelope vault and sidecars can already exist on disk while the current session's backend is still wiring or failed to wire. The function still returns `Ok(())` in that case, so the UI reports a successful wipe while secret-bearing files remain. Tracked for a dedicated fix — not resolving yet.

src/ui/contracts_documents/contracts_documents_screen.rs
lklimek: Verified against current code (82a96aef): confirmed real, and worth being honest about — this is a genuine trade-off from the string-match fix, not a new independent bug. The old code matched on the exact literal error text, which was fragile but happened to correlate (crudely) to the one call path that produced that exact string. The new `fail_if_fetch_in_flight`/`stop_on_failure` checks only structurally on MessageType while `WaitingForResult`/`Refreshing`, with no task/request-id correlation, so any concurrent unrelated Error/Warning can now trip it. Same shape in token claims and token-balance refresh, as you note. Tracked for a dedicated fix (proper per-operation result routing) — not resolving yet.

src/backend_task/tokens/query_my_token_balances.rs
lklimek: Verified against current code (82a96aef): confirmed real. `_refresh_guard` is only dropped when the spawned request future itself completes; on timeout the JoinHandle moves to an unbounded reaper, but if the upstream call never returns at all (a true hang, not just a slow one) the guard is held forever and every subsequent refresh attempt returns `TokenBalanceRefreshInProgress` for the rest of the process, while the message tells the user to just wait/retry. Tracked for a dedicated fix (bounded recovery, or an honest "restart required" message) — not resolving yet.

src/app.rs
lklimek: Plausible and consistent with the surrounding code's own documented sensitivity to frame-precise input/prompt races (see the comments a few lines above on the "one-frame interactive gap" and pointer-sink timing) — I did not do a full cross-frame trace to independently reproduce this one, so treating it as tracked-but-not-independently-confirmed rather than fully verified. Not resolving.

docs/user-stories.md
lklimek: Verified: confirmed present at docs/user-stories.md:1361 ("...every prompt action (SEC-004)"), a bare transient review-ID reference with no resolvable context in a durable spec doc. Easy fix — tracked, not resolving yet.

src/backend_task/tokens/query_my_token_balances.rs
```diff
@@ -34,14 +49,29 @@ impl AppContext {
...
+                context.refresh_upstream_token_balances(watch_sets).await
```
thepastaclaw, line 60: **🔴 Blocking: Do not strand the token refresh single-flight guard**

The guard that clears `token_balance_refresh_in_flight` is moved into the spawned network request. `await_managed_network_request_with_timeout` returns a timeout after 90 seconds but deliberately leaves that request running under a reaper. If the underlying request remains hung, the guard is never dropped and every later refresh returns `TokenBalanceRefreshInProgress` for the rest of the session. Add bounded cancellation or a completion protocol that can safely release and reconcile the single-flight state after timeout.

<sub>source: ['codex']</sub>

src/app.rs
```diff
@@ -1727,10 +1811,26 @@ impl App for AppState {
         // screen learns which votes are in progress / cast via
         // `display_task_result`, so a slow or failing query never stalls a frame.
         let now = Instant::now();
-        if now.duration_since(self.last_scheduled_vote_check) > Duration::from_secs(60) {
+        let network = active_context.network;
+        if !migration_allows_scheduled_vote_sweep(migration_state.as_ref()) {
+            self.scheduled_vote_sweep_deferred_since_ms
+                .entry(network)
+                .or_insert_with(unix_time_ms);
```
thepastaclaw, line 2015: **🔴 Blocking: Do not treat terminal no-op migration state as permanently in progress**

`migration_allows_scheduled_vote_sweep` excludes `Idle`, but `FinishUnwire::terminal_state(false)` deliberately returns `Idle` when the completion sentinel already exists or no legacy data was moved. Acknowledging unreadable-data warnings also returns to `Idle`. The reconciler's per-network dispatch guard prevents another migration run, while this branch records a deferred sweep every frame and never resumes voting. Track whether the cold-start run completed instead of interpreting every `Idle` state as pre-dispatch.

<sub>source: ['codex']</sub>

src/app.rs
```diff
@@ -1727,10 +1811,26 @@ impl App for AppState {
...
+        } else if let Some(preserve_eligibility_since_ms) =
+            self.scheduled_vote_sweep_deferred_since_ms.remove(&network)
+        {
             self.last_scheduled_vote_check = now;
             self.handle_backend_task(BackendTask::ContestedResourceTask(
-                ContestedResourceTask::CastDueScheduledVotes,
+                ContestedResourceTask::CastDueScheduledVotes {
+                    preserve_eligibility_since_ms: Some(preserve_eligibility_since_ms),
+                },
```
thepastaclaw, line 2023: **🔴 Blocking: Retain recovery eligibility until the deferred sweep succeeds**

The preserved migration cutoff is removed before the asynchronous recovery sweep reads the database, loads identities, or casts any vote. Task-level failures do not restore it, and per-vote network failures are caught while the sweep returns outer success. Every later periodic sweep receives `None`, so votes that became stale during migration are permanently filtered out despite being logged as left for the next sweep. Keep the cutoff until all eligible recovery work reaches a durable terminal outcome.

<sub>source: ['codex']</sub>

src/app.rs
```diff
@@ -1929,6 +2047,63 @@ impl App for AppState {
...
+            },
+        );
+
+        assert!(
+            !MigrationReconciler::is_prompting(&frame_state),
+            "a transition published mid-frame must wait for the next frame",
+        );
+        assert!(
+            MigrationReconciler::is_prompting(&status.state()),
+            "the next frame snapshot must observe the prompt",
```
thepastaclaw, line 2265: **🟡 Suggestion: Exercise AppState snapshot wiring instead of ArcSwap ownership**

This test proves only that an `Arc` returned by `MigrationStatus::state` retains the previous allocation after a replacement is published. It never drives `AppState::update` or verifies that input claiming, prompt activation, overlay rendering, banner reconciliation, and scheduled-vote gating all consume the same frame snapshot. A direct `migration_status().state()` read could be reintroduced in any one consumer without failing this test. Add an AppState-level regression that publishes during a frame and checks those consumers remain consistent until the next frame.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Exact head ffa7ea0b4582afe3fddef648e4473ff00abed911 has seven in-scope blocking defects: three unresolved prior issues, a stranded token-refresh guard, and three scheduled-vote regressions. The original migration/password scheduled-vote race is fixed, but no-op completion, retry eligibility, and nested vote-result handling remain incorrect. Focused tests could not run because fetching the pinned Platform revision required network access that was not approved.

### Review provenance

🔴 7 blocking | 🟡 1 suggestion(s)

_3 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `src/backend_task/tokens/query_my_token_balances.rs`:
- [BLOCKING] src/backend_task/tokens/query_my_token_balances.rs:58-60: Do not strand the token refresh single-flight guard
  The guard that clears `token_balance_refresh_in_flight` is moved into the spawned network request. `await_managed_network_request_with_timeout` returns a timeout after 90 seconds but deliberately leaves that request running under a reaper. If the underlying request remains hung, the guard is never dropped and every later refresh returns `TokenBalanceRefreshInProgress` for the rest of the session. Add bounded cancellation or a completion protocol that can safely release and reconcile the single-flight state after timeout.

In `src/app.rs`:
- [BLOCKING] src/app.rs:1815-1818: Do not treat terminal no-op migration state as permanently in progress
  `migration_allows_scheduled_vote_sweep` excludes `Idle`, but `FinishUnwire::terminal_state(false)` deliberately returns `Idle` when the completion sentinel already exists or no legacy data was moved. Acknowledging unreadable-data warnings also returns to `Idle`. The reconciler's per-network dispatch guard prevents another migration run, while this branch records a deferred sweep every frame and never resumes voting. Track whether the cold-start run completed instead of interpreting every `Idle` state as pre-dispatch.
- [BLOCKING] src/app.rs:1819-1826: Retain recovery eligibility until the deferred sweep succeeds
  The preserved migration cutoff is removed before the asynchronous recovery sweep reads the database, loads identities, or casts any vote. Task-level failures do not restore it, and per-vote network failures are caught while the sweep returns outer success. Every later periodic sweep receives `None`, so votes that became stale during migration are permanently filtered out despite being logged as left for the next sweep. Keep the cutoff until all eligible recovery work reaches a durable terminal outcome.
- [SUGGESTION] src/app.rs:2052-2068: Exercise AppState snapshot wiring instead of ArcSwap ownership
  This test proves only that an `Arc` returned by `MigrationStatus::state` retains the previous allocation after a replacement is published. It never drives `AppState::update` or verifies that input claiming, prompt activation, overlay rendering, banner reconciliation, and scheduled-vote gating all consume the same frame snapshot. A direct `migration_status().state()` read could be reintroduced in any one consumer without failing this test. Add an AppState-level regression that publishes during a frame and checks those consumers remain consistent until the next frame.

In `src/backend_task/contested_names/mod.rs`:
- [BLOCKING] src/backend_task/contested_names/mod.rs:242-246: Do not mark a rejected recovered vote as executed
  `vote_on_dpns_name` returns `Ok(DPNSVoteResults(...))` even when the Platform broadcast failed; the failure is stored inside the result vector. This sweep treats every outer `Ok` as success and emits `CastScheduledVote`. `AppState` then marks the database row executed and shows a success banner, permanently suppressing retries for a rejected vote. Inspect the nested result for the scheduled voter and emit success only when its Platform result is `Ok`.
```
</details>

src/model/data_migration.rs
```diff
@@ -0,0 +1,27 @@
+/// Oldest released DET `data.db` version the direct storage update supports.
+pub(crate) const MIN_DIRECT_MIGRATION_VERSION: i64 = 11;
+
+/// Newest known pre-unwire DET data version the compatibility readers support.
+///
+/// Deliberately kept a few versions above `DEFAULT_DB_VERSION` (38) as headroom,
+/// so data written by a slightly newer build (39, 40) still migrates rather than
+/// failing closed; only a version above this ceiling is rejected as too new.
+pub(crate) const MAX_DIRECT_MIGRATION_VERSION: i64 = 40;
```
thepastaclaw, line 9: **🔴 Blocking: Reject database versions newer than the understood schema**

The current schema and migration ladder end at version 38, but this gate accepts versions 39 and 40 despite having no definitions for either layout. A newer build can change SQL columns or serialized blob formats under either version, after which this older binary may skip undecodable records and write completion sentinels. Set the ceiling to the newest schema whose exact layout is understood, currently DEFAULT_DB_VERSION.

<sub>source: ['codex']</sub>

src/context/wallet_lifecycle/spv.rs
```diff
@@ -90,19 +103,15 @@ impl AppContext {
...
-                    }
-                });
-        }
+        let backend = Arc::clone(&backend);
+        self.subtasks
+            .spawn_sync("shielded_coordinator_clear", async move {
+                if let Err(error) = backend.clear_shielded().await {
+                    tracing::warn!(%error, "Shielded coordinator reset failed during clear");
+                }
+            });
```
thepastaclaw, line 117: **🔴 Blocking: Wait for and propagate every database-clear operation**

Upstream wallet removal and shielded-store clearing are detached, while DashPay listing and deletion failures are only logged. The method can return Ok before durable cleanup finishes or after persisted contacts, wallet state, or shielded state failed to clear, and the UI immediately reports that all local data was removed. Await the asynchronous operations and add every synchronous or asynchronous failure to the authoritative WalletDataClearIncomplete result.

<sub>source: ['codex']</sub>

src/app.rs
```diff
@@ -80,6 +81,23 @@ pub const MIGRATION_IDENTITIES_ACK_ACTION_ID: &str = "migration:ack:unreadable_i
...
 
+fn migration_allows_scheduled_vote_sweep(state: &MigrationState) -> bool {
+    matches!(
+        state,
+        MigrationState::Success
+            | MigrationState::SucceededWithUnreadableVotes { .. }
+            | MigrationState::SucceededWithUnreadableIdentities { .. }
+            | MigrationState::SucceededWithUnreadableIdentitiesAndVotes { .. }
+    )
+}
```
thepastaclaw, line 92: **🔴 Blocking: Treat a completed no-op migration as terminal**

migration_allows_scheduled_vote_sweep excludes Idle, while finish_unwire deliberately leaves a successful no-op run in Idle. MigrationReconciler dispatches only once per network, so an ordinary launch with no data to move remains Idle permanently and defers scheduled votes on every frame. Represent completed-with-no-changes as a distinct terminal state or track cold-start completion separately from MigrationState::Idle.

<sub>source: ['codex']</sub>

src/backend_task/contested_names/mod.rs
```diff
@@ -232,9 +241,9 @@ impl AppContext {
             {
                 Ok(_) => {
                     let _ = sender
-                        .send(TaskResult::Success(Box::new(
+                        .send(TaskResult::unattributed_success(
                             BackendTaskSuccessResult::CastScheduledVote(vote),
-                        )))
+                        ))
                         .await;
```
thepastaclaw, line 247: **🔴 Blocking: Do not mark nested vote failures as successful casts**

vote_on_dpns_name returns an outer Ok(DPNSVoteResults) even when Platform rejected the submission, storing that failure inside the nested result vector. This branch treats every outer Ok as success and emits CastScheduledVote, after which AppState marks the database row executed and shows a success message. Inspect the nested result and emit CastScheduledVote only when the scheduled voter's submission succeeded.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward revalidation confirms three scheduled-vote blockers remain, while four prior issues are fixed, token-refresh guard retention is intentional, and the AppState snapshot test remains a suggestion. The latest delta adds six blockers involving future database versions, incomplete data wiping, and uncorrelated document results. With nine confirmed blockers total, this PR requires changes.

### Review provenance

🔴 9 blocking

_4 additional finding(s) omitted (not in diff)._

<details>

```

In `src/model/data_migration.rs`:
- [BLOCKING] src/model/data_migration.rs:6-9: Reject database versions newer than the understood schema
  The current schema and migration ladder end at version 38, but this gate accepts versions 39 and 40 despite having no definitions for either layout. A newer build can change SQL columns or serialized blob formats under either version, after which this older binary may skip undecodable records and write completion sentinels. Set the ceiling to the newest schema whose exact layout is understood, currently DEFAULT_DB_VERSION.

- [BLOCKING] src/wallet_backend/mod.rs:1194-1214: Use fallible enumeration for the authoritative wallet wipe
  forget_all_wallets_local relies on WalletMetaView::list and SingleKeyView::list_persisted to discover every secret that must be removed. Both APIs log and suppress listing, key-parsing, and blob-decoding failures, returning an empty or incomplete vector. Clear Database can therefore report success while seeds or imported private keys remain on disk. Add fallible enumeration for this destructive path and propagate every listing or decoding failure into ClearAllOutcome.failures.

- [BLOCKING] src/context/identity_db.rs:994-1011: Do not purge identity metadata when its private keys were not deleted
  clear_identity_vault_keys silently returns when the stored identity cannot be read or decoded and only logs delete_all failures. Both Clear Database and Wipe Platform Data then purge the identity scope and enumeration index, leaving private keys in the secret store while deleting the metadata needed to identify and retry those deletions. Return a typed error and preserve the identity record and index until all associated vault rows are removed successfully.

In `src/context/wallet_lifecycle/spv.rs`:
- [BLOCKING] src/context/wallet_lifecycle/spv.rs:41-114: Wait for and propagate every database-clear operation
  Upstream wallet removal and shielded-store clearing are detached, while DashPay listing and deletion failures are only logged. The method can return Ok before durable cleanup finishes or after persisted contacts, wallet state, or shielded state failed to clear, and the UI immediately reports that all local data was removed. Await the asynchronous operations and add every synchronous or asynchronous failure to the authoritative WalletDataClearIncomplete result.

In `src/ui/contracts_documents/contracts_documents_screen.rs`:
- [BLOCKING] src/ui/contracts_documents/contracts_documents_screen.rs:582-604: Correlate successful document-query results before updating the screen
  Failures are checked against pending_fetch_context, but successful Documents and PageDocuments results still use the context-blind callback. Fetch Documents remains enabled while a request is pending, and document results can also arrive from other screens. An older or unrelated completion can therefore clear the active request and replace its documents and pagination state. Override display_backend_task_result and mutate state only when the supplied context matches pending_fetch_context.

In `src/ui/tokens/view_token_claims_screen.rs`:
- [BLOCKING] src/ui/tokens/view_token_claims_screen.rs:80-99: Prevent stale document results from replacing token claims
  The claims screen correlates failures but accepts every successful Documents result routed to the visible screen. Both Refresh and Fetch claims remain available during Fetching, so multiple requests can overlap, and results from unrelated document queries can also reach this callback. Correlate successful results to the pending operation; because repeated claims queries have identical contexts, also disable duplicate dispatches while fetching or add a unique operation identifier.

In `src/app.rs`:
- [BLOCKING] src/app.rs:84-92: Treat a completed no-op migration as terminal
  migration_allows_scheduled_vote_sweep excludes Idle, while finish_unwire deliberately leaves a successful no-op run in Idle. MigrationReconciler dispatches only once per network, so an ordinary launch with no data to move remains Idle permanently and defers scheduled votes on every frame. Represent completed-with-no-changes as a distinct terminal state or track cold-start completion separately from MigrationState::Idle.

In `src/backend_task/contested_names/mod.rs`:
- [BLOCKING] src/backend_task/contested_names/mod.rs:231-247: Do not mark nested vote failures as successful casts
  vote_on_dpns_name returns an outer Ok(DPNSVoteResults) even when Platform rejected the submission, storing that failure inside the nested result vector. This branch treats every outer Ok as success and emits CastScheduledVote, after which AppState marks the database row executed and shows a success message. Inspect the nested result and emit CastScheduledVote only when the scheduled voter's submission succeeded.
```
</details>

src/mcp/resolve.rs
thepastaclaw: Resolved in `68a2090` — _Run headless migration before resolving hydrated wallets_ no longer present.

src/backend_task/dashpay/contact_requests.rs
thepastaclaw: Resolved in `0fda1f6` — _Do not rebroadcast a paid action when only its local marker failed_ no longer present.

src/app.rs
thepastaclaw: Resolved in `0fda1f6` — _Do not treat terminal no-op migration state as permanently in progress_ no longer present.

src/app.rs
thepastaclaw: Resolved in `0fda1f6` — _Exercise AppState snapshot wiring instead of ArcSwap ownership_ no longer present.

src/ui/state/contacts_view.rs
thepastaclaw: Resolved in `0fda1f6` — _Do not expire a paid-action guard without backend completion_ no longer present.

src/backend_task/tokens/query_my_token_balances.rs
thepastaclaw: Resolved in `b633a52` — _Do not strand the token refresh single-flight guard_ no longer present.

src/app.rs
thepastaclaw: Resolved in `0fda1f6` — _Retain recovery eligibility until the deferred sweep succeeds_ no longer present.

src/ui/identities/add_existing_identity_screen.rs
```diff
@@ -1114,8 +1114,11 @@ impl ScreenLike for AddExistingIdentityScreen {
             egui::CentralPanel::default()
                 .frame(egui::Frame::NONE)
                 .show(ui, |ui| {
-                    let mut popup =
-                        InfoPopup::new("Load Identity Information", &show_pop_up_info_text);
+                    let mut popup = InfoPopup::new(
+                        egui::Id::new("load_identity_info_popup"),
+                        "Load Identity Information",
+                        &show_pop_up_info_text,
```
thepastaclaw, line 1120: **🔴 Blocking: Give each dynamic help popup its own guard ID**

This screen multiplexes at least five help messages through one `InfoPopup` ID. Because the popup does not block background input, clicking a different help icon while one popup is open replaces the message before rendering. The guard sees the same ID on consecutive passes, treats that opening click as an ordinary outside click, and immediately closes the second popup. Include the message or another stable help-site discriminator in the ID; the same dynamic-single-ID pattern appears in `AddNewIdentityScreen` and `ContactProfileViewerScreen`.

```suggestion
let mut popup = InfoPopup::new(
    egui::Id::new("load_identity_info_popup").with(&show_pop_up_info_text),
    "Load Identity Information",
    &show_pop_up_info_text,
);
```

<sub>source: ['codex']</sub>

src/ui/helpers.rs
```diff
@@ -15,6 +36,40 @@ pub fn clicked_outside_window(ctx: &egui::Context, window_rect: egui::Rect) -> b
...
+pub(crate) fn clicked_outside_window_after_open_by_id(
+    ctx: &egui::Context,
+    window_rect: egui::Rect,
+    id: egui::Id,
+) -> bool {
+    let this_pass = ctx.cumulative_pass_nr();
+    let last_pass: Option<u64> = ctx.data(|d| d.get_temp(id));
+    ctx.data_mut(|d| d.insert_temp(id, this_pass));
+    let is_opening_frame = last_pass != Some(this_pass.wrapping_sub(1));
+    !is_opening_frame && clicked_outside_window(ctx, window_rect)
```
thepastaclaw, line 70: **🟡 Suggestion: Re-arm a modal guard after an outside-click dismissal**

The helper records the current pass before reporting an outside-click dismissal. If the same popup ID is reopened on the immediately following pass, the stored pass still looks continuous, so the reopening click closes the popup again. Remove the temporary pass entry when dismissal is reported so a new rendering episode is recognized even without a full non-rendering pass between episodes.

```suggestion
let this_pass = ctx.cumulative_pass_nr();
let last_pass: Option<u64> = ctx.data(|d| d.get_temp(id));
ctx.data_mut(|d| d.insert_temp(id, this_pass));
let is_opening_frame = last_pass != Some(this_pass.wrapping_sub(1));
if is_opening_frame {
    return false;
}
let clicked = clicked_outside_window(ctx, window_rect);
if clicked {
    ctx.data_mut(|d| d.remove::<u64>(id));
}
clicked
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All nine carried-forward blocker candidates and the prior test-coverage suggestion remain valid; the three Clear Database blockers are consolidated into one comment because they share the same false-success outcome. Latest-delta review adds one blocker for dynamic help popups sharing a guard ID and one suggestion for immediate same-popup reopening. With eight consolidated blockers and two suggestions, this PR requires changes.

### Review provenance

🔴 8 blocking | 🟡 1 suggestion(s)

<details>

```

In `src/ui/contracts_documents/contracts_documents_screen.rs`:
- [BLOCKING] src/ui/contracts_documents/contracts_documents_screen.rs:582-604: Correlate successful document-query results before updating the screen
  Failures are checked against `pending_fetch_context`, but successful `Documents` and `PageDocuments` results still use the context-blind callback. The fetch control remains available while a request is pending, and results are routed to whichever screen is visible. An older or unrelated completion can therefore clear the active request and replace its documents and pagination state. Override `display_backend_task_result` and mutate state only when the supplied context matches the pending query.

In `src/ui/tokens/view_token_claims_screen.rs`:
- [BLOCKING] src/ui/tokens/view_token_claims_screen.rs:80-99: Prevent stale document results from replacing token claims
  The claims screen correlates failures but accepts every successful `Documents` result routed to the visible screen. Refresh and Fetch claims can dispatch overlapping requests, and an unrelated document query can also complete while this screen is visible. Correlate successful results with the expected claims-query context and prevent duplicate identical requests while a fetch is active, or add a unique operation identifier.

In `src/ui/identities/add_existing_identity_screen.rs`:
- [BLOCKING] src/ui/identities/add_existing_identity_screen.rs:1117-1120: Give each dynamic help popup its own guard ID
  This screen multiplexes at least five help messages through one `InfoPopup` ID. Because the popup does not block background input, clicking a different help icon while one popup is open replaces the message before rendering. The guard sees the same ID on consecutive passes, treats that opening click as an ordinary outside click, and immediately closes the second popup. Include the message or another stable help-site discriminator in the ID; the same dynamic-single-ID pattern appears in `AddNewIdentityScreen` and `ContactProfileViewerScreen`.

In `src/ui/helpers.rs`:
- [SUGGESTION] src/ui/helpers.rs:66-70: Re-arm a modal guard after an outside-click dismissal
  The helper records the current pass before reporting an outside-click dismissal. If the same popup ID is reopened on the immediately following pass, the stored pass still looks continuous, so the reopening click closes the popup again. Remove the temporary pass entry when dismissal is reported so a new rendering episode is recognized even without a full non-rendering pass between episodes.
```
</details>

docs/ai-design/2026-07-14-pr892-user-story-qa/scenarios/WAL.md
```diff
@@ -0,0 +1,850 @@
...
+`SND.md` under SND-003. Filed there since SND-003 ("Receive Dash with QR code") is the
+story it maps to; noting the cross-reference here since it was discovered during WAL testing.
+
+---
+
+## WAL-002: Import wallet via mnemonic — PASS
+
+Steps:
+1. Used the Create Wallet entropy-grid flow to generate a fresh 12-word mnemonic (`sail
+   eager shrug goose primary position under shuffle swarm occur fall diet`), noted the
```
thepastaclaw, line 117: **🔴 Blocking: Remove committed wallet recovery material**

These lines contain a checksum-valid 12-word BIP39 mnemonic, and line 140 contains a checksum-valid compressed testnet WIF private key. Deleting the test wallets or recording that they currently have no balance does not revoke these credentials: anyone with repository access can derive their addresses and spend funds sent there later, and removing the text in a later commit leaves it recoverable from history. Remove the literals, rewrite the affected history, and treat both credentials as compromised.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Nine in-scope blocking issues remain at head 0e79a89bf4263c087d4455c8410a9ae470b93d93, including a newly committed, cryptographically valid mnemonic and testnet private key. All ten prior findings remain valid; the two modal-guard findings are consolidated into one canonical blocker, leaving nine blockers and one suggestion within the comment budget.

### Review provenance

🔴 9 blocking

_7 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `docs/ai-design/2026-07-14-pr892-user-story-qa/scenarios/WAL.md`:
- [BLOCKING] docs/ai-design/2026-07-14-pr892-user-story-qa/scenarios/WAL.md:116-117: Remove committed wallet recovery material
  These lines contain a checksum-valid 12-word BIP39 mnemonic, and line 140 contains a checksum-valid compressed testnet WIF private key. Deleting the test wallets or recording that they currently have no balance does not revoke these credentials: anyone with repository access can derive their addresses and spend funds sent there later, and removing the text in a later commit leaves it recoverable from history. Remove the literals, rewrite the affected history, and treat both credentials as compromised.

In `src/ui/contracts_documents/contracts_documents_screen.rs`:
- [BLOCKING] src/ui/contracts_documents/contracts_documents_screen.rs:582-604: Correlate successful document-query results before updating the screen
  Failures are correlated against `pending_fetch_context`, but successful `Documents` and `PageDocuments` results still use the legacy context-blind callback. The fetch control remains available while a request is pending, and results are routed asynchronously to whichever screen is visible. An older or unrelated completion can therefore clear the active request and replace its documents and pagination state. Override `display_backend_task_result` and mutate state only when its context matches the pending query.

In `src/ui/tokens/view_token_claims_screen.rs`:
- [BLOCKING] src/ui/tokens/view_token_claims_screen.rs:80-99: Prevent stale document results from replacing token claims
  The claims screen compares task context only for failures and accepts every successful `Documents` result routed to it. Both Refresh and Fetch claims can dispatch while work is active, and unrelated document queries can also complete while this screen is visible. A stale completion can overwrite the claims and reset `fetch_status`; successful results need the same context correlation as failures, with duplicate identical requests suppressed or assigned unique operation identifiers.
```
</details>

PR #896, opened 2026-07-15
ci: replace weekly build check with weekly draft pre-release

Claudius-Maginificent:
## Why this PR exists
- **Problem**: The weekly build workflow (`weekly-build.yml`) only builds the app and posts pass/fail to Slack with per-artifact download links (GitHub login required). There's no easy way for the team or testers to grab a recent build without digging through Actions run artifacts.
- **What breaks without it**: Anyone wanting a recent build has to find a green weekly Actions run, open it, and download artifacts individually — no changelog-style landing page, no versioning, no single link to share.
- **Blocking relationship**: none — this PR's own diff is CI-only and touches no application code. Originally stacked atop #894 purely for branch-basing purposes; once #894 squash-merged and its branch was deleted, GitHub auto-retargeted this PR onto `docs/platform-wallet-migration-design` (#860), inflating the visible diff to 133 unrelated files. Since none of this PR's 9 commits touch anything #860 does, and `v1.0-dev` was already a full ancestor of this branch, the branch was rebuilt by cherry-picking the 9 CI-only commits directly onto `origin/v1.0-dev` and force-pushed — collapsing the diff back to the real 4 files with a clean, conflict-free `MERGEABLE` state.

## What was done
Replaces `weekly-build.yml` with a workflow that cuts a **draft** GitHub pre-release every Saturday from `v1.0-dev`, versioned `v<core>-weekly.<YYYYMMDD>` (core read live from `Cargo.toml`), and posts a Slack notification linking to the release page instead of per-artifact links.

- **New `gate` job**: skips the run when there are no new commits since the last weekly tag, or when the latest completed `tests.yml`/`clippy.yml` run on the triggering ref is red (a hard block, not overridable by `force` or a manual `version`). Supports `workflow_dispatch` inputs `force` (re-cut/update today's draft in place) and `version` (custom semver, regex-validated to guard against shell/tag injection). The CI-red check keys off `github.ref_name` — the same ref checkout/tag/build already use by default — instead of a hardcoded branch name, so a manual dispatch from another ref can no longer read one branch's CI status while gating/tagging a different branch's history, and dispatching from an intentional custom branch keeps working (an earlier version of this fix hardcoded `v1.0-dev` directly, which broke custom-branch dispatch and would've gone stale the moment the active dev branch rolled to v1.1-dev/v2.0-dev — reworked after review).
- **New `tag` job**: deletes any still-drafted prior weekly release (+ its tag) before pushing this week's tag, so at most one weekly draft ever exists. Pushes with the default `GITHUB_TOKEN` only — this is load-bearing: GitHub suppresses new workflow runs triggered by the default token, which is what keeps `release.yml`'s `push: tags: v*` trigger dormant for this tag. A PAT here would fire a duplicate non-draft prerelease. Tags `HEAD` (not `$GITHUB_SHA`, which stays bound to whatever ref originally triggered the run) and, on a force re-cut of an already-tagged day, moves the tag forward to the current commit instead of leaving it stranded on the old one — but only after verifying the existing tag is actually one of *this workflow's own* still-draft releases (a still-draft GH release carrying our weekly-version marker). Without that check, `force=true` combined with a `version` matching an already-existing official release (e.g. `0.9.3`) would have silently rewritten — and a subsequent build failure could have let `cleanup` delete — a real, already-published release tag. Dedup and stale-draft cleanup both now use that same draft+marker signal instead of tag-name pattern matching, so custom (non-canonical) weekly versions are tracked identically to auto-derived ones.
- **New `cleanup` job**: restores the invariant "a weekly tag exists ⟺ a complete release exists" — deletes the orphaned tag/draft whenever tagging succeeded but the platform build didn't. Its fallback tag-delete no longer swallows API failures with `|| true` — a genuine deletion failure now fails the job instead of `cleanup` reporting success while an orphaned tag silently remains.
- **`release.yml`**: additive-only `draft` and `version_override` inputs (both default to today's exact behavior when omitted). A new shared composite action, `.github/actions/apply-version-override`, rewrites `Cargo.toml`/`Cargo.lock` before each build job so the weekly version (+ short commit sha as semver build-metadata) bakes into the binary — `det.log`'s existing startup log line then carries both automatically, no source change needed. Each build job's checkout now first checks via the GitHub API whether `inputs.tag` actually exists as a ref, and pins to it only when it does — real verification instead of trusting a context value. (An earlier version of this gated on `github.event_name == 'workflow_call'`, which never actually engages: a reusable workflow inherits the *caller's* original event name — schedule/workflow_dispatch — it never becomes `'workflow_call'`. Caught by review, replaced with the API check.) Direct `workflow_dispatch` of `release.yml` keeps checking out `github.ref` unchanged, since that path's `tag` input often names a release still to be *created*, not an existing ref yet.
- **`flatpak.yml`**: same `version_override` input, plus a fix for a real edge case — undrafting a weekly release later fires `release: published`, which re-triggers this workflow's existing attach-to-release step. Without the fix it would rebuild from the *committed* `Cargo.toml` (still `1.0.0-dev`), clobbering the correctly-versioned bundle already on the release page. `finalize` now persists the full version as a machine-readable marker in the release notes at creation time, and `flatpak.yml` recovers it from there on `release: published` — this covers *any* accepted version shape, not just the auto-derived weekly one, so a custom `version` dispatch input survives undrafting too (falls back to the old tag-pattern reconstruction only for releases created before this marker existed).
- **`finalize` job**: uploads the Flatpak bundles onto the draft release page (its own `release:`-triggered attach path never fires for a `workflow_call`-created draft) and fetches the release URL for Slack.
- **`notify` job**: now depends on `tag`/`cleanup` in addition to `release`/`flatpak`/`finalize`, and branches on their actual results — it previously could claim success whenever `should_release=true` regardless of whether the build actually finished, and separately could race `cleanup` and announce "orphaned tag swept away" before cleanup ran or even after it failed.
- **Security**: `apply-version-override` and `gate`'s `Decide` step both interpolated `workflow_dispatch` inputs directly into `run:` script bodies — GitHub's documented [script-injection](https://docs.github.com/en/actions/concepts/security/script-injections) class, where a crafted input value breaks out of the shell assignment and runs arbitrary commands with the job's `GITHUB_TOKEN`. Both now pass inputs through `env:` instead, so values arrive as data, never as re-parsed script text.
- **Permissions**: `permissions: {}` at the workflow level; every job grants only what it needs (see the job table in the design doc).

## Testing
- All files (`weekly-build.yml`, `release.yml`, `flatpak.yml`, `apply-version-override/action.yml`) pass `actionlint` clean on all new/changed content (pre-existing findings on unrelated lines of `release.yml` are stale action-version/shellcheck style nits, unrelated to this change). Zero CRLF regressions.
- **Live end-to-end dispatch verified green**: run [29445812074](https://github.com/dashpay/dash-evo-tool/actions/runs/29445812074) completed successfully top to bottom — `gate`, `tag`, all 5 platform-matrix legs, both Flatpak legs, `Create GitHub Release`, `Attach Flatpak bundles`, and `Slack notification` all succeeded; `cleanup` correctly no-op'd. Produced a real draft release (`v1.0.0-weekly.20260715`) with all 7 expected assets (5 platform installers + 2 Flatpak bundles). The earlier `startup_failure` noted in this section previously was a propagation-timing issue on first dispatch after push, not a workflow defect — resolved on retry.
- The `SHOULD_RELEASE=true`-but-build-failed notify/cleanup gap, the script-injection findings, the force-recut tag/version mismatch, the `TARGET_BRANCH` redesign, and the flatpak custom-version recovery were caught by automated review (Copilot + `thepastaclaw`) and one human review comment after that green run, verified against the code, and fixed across follow-up commits.
- **Second live end-to-end dispatch verified green** on top of the full fix stack: run [29488701117](https://github.com/dashpay/dash-evo-tool/actions/runs/29488701117), commit `0f7a9b42`. Every job succeeded (`gate`, `tag`, all 5 platform legs, both Flatpak legs, `Create GitHub Release`, `Attach Flatpak bundles`, `Slack notification`); `cleanup` correctly skipped (nothing to clean up). Produced `v1.0.0-weekly.20260716` with all 7 expected assets, and its release body contains exactly the expected marker (`<!-- weekly-version: 1.0.0-weekly.20260716+0f7a9b423 -->`), confirming the flatpak version-recovery persistence works in real conditions, not just in isolated logic review. The prior day's draft (`v1.0.0-weekly.20260715`) was correctly deleted by `tag`'s stale-draft cleanup, leaving only one weekly draft at a time as designed.
- A further review round (`thepastaclaw`) on commit `0f7a9b42` found 5 more findings, most notably a real one: `force=true` + a `version` matching an existing official release could have hijacked/destroyed that release's tag. All 5 fixed in `95f0238a`, verified line-by-line against the exact intended diff before commit (not just "tests pass"), given the severity of the tag-hijack fix specifically.
- **A live dispatch (on the pre-retarget base) caught a bug in that very fix**: the tag-hijack ownership guard used `GET /repos/{owner}/{repo}/releases/tags/{tag}`, which is documented as returning only *published* releases and empirically 404s for any draft regardless of token — confirmed by testing it directly against our own known-good draft (`ghsudo gh api .../releases/tags/v1.0.0-weekly.20260716` → 404) alongside the paginated list endpoint, which correctly found it. This meant the guard rejected every force re-cut, not just malicious ones — a functional regression in the very fix meant to prevent tag hijacking. Fixed by switching to the same paginated `/releases` list + `tag_name` filter already used by the dedup logic.
- **That fix itself then failed a second live dispatch** with the same symptom, for an unrelated reason: `gh api --jq --arg tag "$TAG" '...'` is invalid `gh` CLI syntax (`gh api` doesn't accept extra jq flags after `--jq`), so the lookup errored and was silently swallowed, leaving the guard permanently fail-closed. Root-caused by testing the exact committed syntax live and comparing it against working alternatives. Fixed by interpolating `$TAG` directly into a double-quoted `--jq` expression (the working form for `gh api`'s built-in jq processor; `$TAG` is already regex-validated upstream, so this is safe).
- **A third live dispatch, run [29501651417](https://github.com/dashpay/dash-evo-tool/actions/runs/29501651417), verified the corrected guard end-to-end**: `gate` and `tag` succeeded — the guard correctly recognized the existing `v1.0.0-weekly.20260716` draft as its own and force-moved the tag to the new commit — followed by all 5 platform legs, both Flatpak legs, `Create GitHub Release`, `Attach Flatpak bundles`, and `Slack notification`, all green; `cleanup` correctly skipped.
- **A further review round (`thepastaclaw` ×6, `coderabbitai` ×1) on commit `dbb771eb` found 7 more findings**, all verified against current code before fixing: 3 fresh script-injection instances (`release.yml`'s 3 checkout-ref-resolver steps interpolating `inputs.tag`/`github.ref` directly into the script; `weekly-build.yml`'s gate step doing the same with `github.ref_name`; `flatpak.yml` doing the same with `inputs.version_override`/`github.event.release.tag_name`) — all missed when the earlier injection fixes landed in different locations; the checkout-ref resolver also silently treated any `gh api` failure (auth/rate-limit/network) the same as a confirmed 404, now fixed to fail closed on anything but a real 404 (live-tested against a real tag, a missing tag, and a simulated 401 before shipping); the no-new-commits gate only considered draft releases, letting an undrafted weekly release drop out of the comparison and cause a duplicate re-cut with zero new commits; and `tag`'s stale-draft deletion ran before the force-recut ownership check, so a bad `version` input could destroy the legitimate current draft before being correctly refused — split into its own "Verify force-recut ownership" step that now runs first. `flatpak.yml`'s `gh release view` also silently swallowed API/auth failures identically to "no marker found"; suppression removed. All 7 fixed in `03468c4d`.
- **A fourth live dispatch, run [29516790796](https://github.com/dashpay/dash-evo-tool/actions/runs/29516790796), verified the full fix stack including the reordered ownership check**: confirmed via per-step API inspection that `Check out code` → `Verify force-recut ownership` → `Delete stale weekly drafts` → `Push weekly tag` executed in that exact order and all succeeded (today's tag already existed, so this genuinely exercised the force-recut/ownership path, not a no-op). Full pipeline green end to end: `gate`, `tag`, all 5 platform legs, both Flatpak legs, release creation, asset attach, Slack notification; `cleanup` correctly skipped.

## Breaking changes
None to the real tag-push release flow — both new `release.yml`/`flatpak.yml` inputs default to today's exact behavior when omitted.

## Follow-ups
- **Retention policy, once weekly releases go fully automatic (not draft-gated).** Today's `tag` job already deletes the *previous* weekly release before cutting a new one, but only while it's still a draft — anything a human undrafts/publishes is intentionally kept forever. The end goal for this pipeline is full auto-publish (no manual approval step), at which point every weekly build becomes permanent by default unless we add pruning.
  - Checked GitHub's actual constraints before assuming a size problem: 2 GiB max per release asset, up to 1000 assets per release, and **release-asset storage/bandwidth is explicitly uncapped** — not counted against the git repo size guidance (<1GB ideal / <5GB recommended), which only concerns git objects. So this is not a storage-quota problem.
  - The real reason to add a retention window is signal-to-noise, same as any project with a rolling nightly/weekly channel: an unbounded Releases page eventually buries the build people actually want, and list/autocomplete tooling gets noisier over time.
  - Recommended default: rolling window (e.g. keep the last 8-12 weeks), pruning older ones the same way `tag` already prunes stale drafts today, just widened from "drafts only" to "all weekly releases." A "promote to keep forever" escape hatch (e.g. a label/marker a human can set) is a reasonable addition if specific historical builds ever need permanent pinning.
  - Not designed or scoped yet — flagging so it isn't lost before the auto-publish switch flips.

## Checklist
- [x] `actionlint` clean on all new/changed workflow files
- [x] Diff reviewed against design spec, additive-only to `release.yml`/`flatpak.yml`
- [x] Live end-to-end dispatch verified green (run 29445812074)
- [x] Script-injection findings (Copilot + `thepastaclaw`) fixed and verified against current code
- [x] `TARGET_BRANCH` hardcoding reworked to key off `github.ref_name` (human review feedback)
- [x] Flatpak custom-version recovery fixed (`thepastaclaw`)
- [x] Fresh live end-to-end dispatch verified green on top of all fixes in this round (run 29488701117)
- [x] Force-recut tag-hijack guard + reusable-workflow ref-pinning fix + 3 more findings (`thepastaclaw`) fixed and verified against current code
- [x] Force-recut ownership guard fixed again — the by-tag API endpoint used in the first fix structurally excludes drafts (confirmed via live failure + direct API comparison); switched to the same paginated list-endpoint approach already proven in the dedup logic
- [x] Retargeted base to `v1.0-dev` — rebuilt the branch via cherry-pick onto `origin/v1.0-dev` and force-pushed (see Blocking relationship above)
- [x] Live end-to-end dispatch verified green on top of the corrected tag-hijack guard, on the retargeted base (run 29501651417)
- [x] 7 further findings (`thepastaclaw` ×6, `coderabbitai` ×1) fixed and verified against current code — 3 more script-injection instances, fail-closed checkout-ref resolution, no-new-commits gate covering published releases, and the force-recut ownership check reordered before stale-draft deletion
- [x] Live end-to-end dispatch verified green on top of the full fix stack, including per-step confirmation the reordered ownership check actually ran first (run 29516790796)

## Attribution


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 03468c4d)

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,241 @@
...
+          # (it's a safety net, not a convenience toggle). Treated as "latest
+          # *completed* run of each CI workflow on v1.0-dev" rather than
+          # HEAD-sha check-runs, because tests.yml/clippy.yml are path-filtered
+          # and a docs-only HEAD would otherwise have zero check-runs and read
+          # as falsely green. A workflow that has never run is treated as
+          # absent, not red.
+          red=false
+          for wf in tests.yml clippy.yml; do
+            concl=$(gh run list --workflow "$wf" --branch v1.0-dev --status completed \
+                      --limit 1 --json conclusion --jq '.[0].conclusion // ""')
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed still open. `gate`'s CI-red check is hardcoded to `--branch v1.0-dev` (line 71), but every job's `actions/checkout` defaults to `github.ref` — the ref that actually triggered the run. On a `workflow_dispatch` from a non-`v1.0-dev` ref, this reads CI status from one branch while building another, exactly as described.

Recommend a single `TARGET_BRANCH: v1.0-dev` env (or `GITHUB_REF_NAME`) used consistently by both the CI check and the checkout steps. Not addressed in this PR yet — leaving unresolved.

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,241 @@
...
+      - name: Push weekly tag
+        run: |
+          set -euo pipefail
+          if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+            echo "Tag $TAG already exists (force re-cut) — reusing."
+            exit 0
+          fi
+          # See the SECURITY NOTE at the top of this file: GITHUB_TOKEN only.
+          git tag -a "$TAG" -m "Weekly draft pre-release $TAG" "$GITHUB_SHA"
+          git push origin "refs/tags/$TAG"
```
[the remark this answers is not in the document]
Claudius-Maginificent: Checked this against `actions/checkout@v4`'s actual source (`ref-helper.ts`): with `fetch-depth: 0` (set here, lines 143-145), the action calls `getRefSpecForAllHistory()`, which unconditionally includes `+refs/tags/*:refs/tags/*` in the fetch refspec — independent of the `fetch-tags` input, which only matters when `fetch-depth > 0`. So this checkout already pulls every remote tag before the `git rev-parse -q --verify refs/tags/$TAG` check on line 163.

I don't believe there's a bug here — leaving unresolved for a maintainer call rather than auto-resolving, but no code change is planned on my end.

.github/actions/apply-version-override/action.yml
```diff
@@ -0,0 +1,55 @@
...
+  steps:
+    - shell: bash
+      run: |
+        set -euo pipefail
+        v="${{ inputs.version }}"
+        if [ -z "$v" ]; then
+          echo "No version override supplied — leaving Cargo.toml/Cargo.lock untouched."
+          exit 0
+        fi
+        python3 - "$v" <<'PY'
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed still open. `apply-version-override/action.yml` regex-substitutes `inputs.version` directly into `Cargo.toml`/`Cargo.lock` (lines 32-33, 43-48) with no validation, and `release.yml`'s `workflow_dispatch.version_override` (lines 21, 35) accepts an unvalidated free-form string. A value containing `"` or a newline could break out of the TOML string and inject arbitrary manifest content.

`weekly-build.yml` is safe in practice — `gate` enforces a strict semver pattern before setting `version_override` (line 90) — but the direct `release.yml` manual-dispatch path isn't guarded. Recommend moving the same semver regex check into `apply-version-override` itself so every caller is protected, not just this one. Not addressed in this PR yet — leaving unresolved.

.github/actions/apply-version-override/action.yml
```diff
@@ -0,0 +1,55 @@
...
+        import re, sys, pathlib
+
+        v = sys.argv[1]
+
+        ct = pathlib.Path("Cargo.toml")
+        ct_text = ct.read_text()
+        new_ct_text, n = re.subn(
+            r'(?m)^version\s*=\s*"[^"]*"', f'version = "{v}"', ct_text, count=1
```
thepastaclaw: **🔴 Blocking: Validate and safely pass the version override before using it**

`release.yml` exposes `version_override` through `workflow_dispatch`, and this action substitutes that value directly into `v="${{ inputs.version }}"` before Bash runs. A value containing `"; command; #` terminates the assignment and executes arbitrary commands with the release job's credentials. Passing the value through `env` prevents shell parsing, but the action must also reject values that are not valid Rust semver before inserting them into quoted TOML strings. GitHub documents this expression-substitution injection pattern in [Script injections](https://docs.github.com/en/actions/concepts/security/script-injections).

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,48 +1,276 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Decide
+        id: decide
+        run: |
+          set -euo pipefail
+          FORCE="${{ inputs.force || false }}"
+          MANUAL_VERSION="${{ inputs.version }}"
```
thepastaclaw: **🔴 Blocking: Do not interpolate dispatch inputs into the gate script**

`inputs.version` is substituted into the generated Bash program before the semver check executes, so a quoted assignment followed by a command runs before validation. Move both dispatch expressions into the step's `env` mapping and read `FORCE` and `MANUAL_VERSION` from the environment.

```suggestion
- name: Decide
  id: decide
  env:
    FORCE: ${{ inputs.force || false }}
    MANUAL_VERSION: ${{ inputs.version }}
  run: |
    set -euo pipefail
```

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,48 +1,276 @@
...
+          # (it's a safety net, not a convenience toggle). Treated as "latest
+          # *completed* run of each CI workflow on v1.0-dev" rather than
+          # HEAD-sha check-runs, because tests.yml/clippy.yml are path-filtered
+          # and a docs-only HEAD would otherwise have zero check-runs and read
+          # as falsely green. A workflow that has never run is treated as
+          # absent, not red.
+          red=false
+          for wf in tests.yml clippy.yml; do
+            concl=$(gh run list --workflow "$wf" --branch v1.0-dev --status completed \
+                      --limit 1 --json conclusion --jq '.[0].conclusion // ""')
```
thepastaclaw: **🔴 Blocking: Gate the same ref that the workflow builds**

A manual dispatch can select any ref, and checkout plus the reusable build workflows operate on that triggering ref. The gate nevertheless reads the latest completed CI runs from hard-coded `v1.0-dev`, allowing a different commit to be released using another branch's green status. Restrict release-producing dispatches to `v1.0-dev`, or use one target ref consistently for checkout, CI queries, tagging, and builds.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,48 +1,276 @@
...
+            echo "Removing stale weekly draft $old"
+            gh release delete "$old" --yes --cleanup-tag
+          done
+
+      - name: Push weekly tag
+        run: |
+          set -euo pipefail
+          if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+            echo "Tag $TAG already exists (force re-cut) — reusing."
+            exit 0
```
thepastaclaw, line 197: **🔴 Blocking: Force re-cuts build a commit that the release tag does not identify**

When `force=true` is dispatched again on the same UTC date after the branch advances, the gate derives the same tag but a new version containing the new short SHA. This block leaves the existing tag on its old commit while the reusable workflows check out and build the new triggering SHA. The assets and embedded version therefore describe new source, but the release tag and GitHub source archives identify the old source. A force re-cut must either move the tag deliberately or rebuild the existing tagged commit with its original version.

<sub>source: ['codex']</sub>

.github/workflows/flatpak.yml
```diff
@@ -37,6 +43,34 @@ jobs:
...
+              # Undrafting a weekly release later fires this `release: published`
+              # trigger, which would otherwise rebuild from the *committed*
+              # Cargo.toml (still e.g. 1.0.0-dev) and clobber the correctly
+              # weekly-versioned bundle already on the release page. Reconstruct
+              # the identical version instead: checkout on a `release` event
+              # checks out the tag's own commit, so `--short=9` here reproduces
+              # the exact sha the weekly workflow baked in (must match the
+              # `--short=9` width used in weekly-build.yml's gate job).
+              ov="${tag#v}+$(git rev-parse --short=9 HEAD)"
+            fi
```
thepastaclaw: **🔴 Blocking: Preserve custom weekly versions when published drafts rebuild Flatpak**

The weekly dispatch accepts arbitrary semver values and initially passes the exact value into the Flatpak build. Publishing the draft later triggers this workflow without `workflow_call` inputs, but the recovery logic only handles canonical `vX.Y.Z-weekly.YYYYMMDD` tags. A custom value such as `1.2.0-rc.1` is rebuilt from committed version `1.0.0-dev` and overwrites the correctly versioned Flatpak assets. A custom weekly-shaped value with build metadata is also reconstructed with the tag commit SHA instead of its original metadata. Persist and recover the original override, or constrain accepted versions to a deterministically reconstructable format.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -55,82 +283,16 @@ jobs:
...
-
-            if [ -n "$DOWNLOAD_SECTION" ]; then
-              TEXT="${TEXT}${NL}Despite the chaos, my magnificence preserved these artifacts:${DOWNLOAD_SECTION}"
-            fi
+          if [ "$CI_RED" = "true" ]; then
+            TEXT=":warning: *I, Claudius the Magnificent, decline to bless a broken week.* CI on \`v1.0-dev\` is red — no weekly draft was cut. Mend the pipeline: <${RUN_URL}|the run>."
+          elif [ "$SHOULD_RELEASE" = "true" ] && [ "$FINALIZE_RESULT" = "success" ]; then
+            TEXT=":sparkles: *I, Claudius the Magnificent, have sealed another weekly draft for \`${REPO}\`.* Version \`${VERSION}\` awaits your royal assent — undraft it here: <${RELEASE_URL}|the release page>."
+          elif [ "$SHOULD_RELEASE" = "true" ]; then
+            TEXT=":rotating_light: <!channel> *Even I, Claudius the Magnificent, cannot forge a release from a broken build.* Version \`${VERSION}\` for \`${REPO}\` was tagged, yet the platform build faltered — I have swept the orphaned tag away so next week begins unblemished. Inspect the wreckage: <${RUN_URL}|the run>."
```
thepastaclaw: **🔴 Blocking: Report tag and cleanup outcomes only after they complete**

`notify` does not depend on `tag` or `cleanup`. If tagging fails, downstream jobs are skipped while `SHOULD_RELEASE` remains true, so this branch falsely says the version was tagged and removed. If a build fails after a successful tag, `notify` and `cleanup` can run concurrently, allowing Slack to claim cleanup before it finishes or even when it fails. Add both jobs to the notification dependencies and derive the message from their actual results.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The exact range ef3d5c43bc51b11f1cdd2ca6d11a1d0dfc280458..a4865d024d5e9037d1f0ec86a3a383082b08d642 contains six confirmed blockers: two command-injection paths, cross-ref CI gating, inconsistent force re-cuts, incorrect Flatpak version recovery, and premature failure notifications. The draft-reuse, missing-tags, required-action-input, and workflow-dispatch-default concerns are not actionable against the current implementations.

### Review provenance
- Codex reviewers: `unknown model` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 6 blocking

<details>

```

In `.github/actions/apply-version-override/action.yml`:
- [BLOCKING] .github/actions/apply-version-override/action.yml:20-33: Validate and safely pass the version override before using it
  `release.yml` exposes `version_override` through `workflow_dispatch`, and this action substitutes that value directly into `v="${{ inputs.version }}"` before Bash runs. A value containing `"; command; #` terminates the assignment and executes arbitrary commands with the release job's credentials. Passing the value through `env` prevents shell parsing, but the action must also reject values that are not valid Rust semver before inserting them into quoted TOML strings. GitHub documents this expression-substitution injection pattern in [Script injections](https://docs.github.com/en/actions/concepts/security/script-injections).

In `.github/workflows/weekly-build.yml`:
- [BLOCKING] .github/workflows/weekly-build.yml:55-60: Do not interpolate dispatch inputs into the gate script
  `inputs.version` is substituted into the generated Bash program before the semver check executes, so a quoted assignment followed by a command runs before validation. Move both dispatch expressions into the step's `env` mapping and read `FORCE` and `MANUAL_VERSION` from the environment.
- [BLOCKING] .github/workflows/weekly-build.yml:71-72: Gate the same ref that the workflow builds
  A manual dispatch can select any ref, and checkout plus the reusable build workflows operate on that triggering ref. The gate nevertheless reads the latest completed CI runs from hard-coded `v1.0-dev`, allowing a different commit to be released using another branch's green status. Restrict release-producing dispatches to `v1.0-dev`, or use one target ref consistently for checkout, CI queries, tagging, and builds.
- [BLOCKING] .github/workflows/weekly-build.yml:163-165: Force re-cuts build a commit that the release tag does not identify
  When `force=true` is dispatched again on the same UTC date after the branch advances, the gate derives the same tag but a new version containing the new short SHA. This block leaves the existing tag on its old commit while the reusable workflows check out and build the new triggering SHA. The assets and embedded version therefore describe new source, but the release tag and GitHub source archives identify the old source. A force re-cut must either move the tag deliberately or rebuild the existing tagged commit with its original version.
- [BLOCKING] .github/workflows/weekly-build.yml:289-292: Report tag and cleanup outcomes only after they complete
  `notify` does not depend on `tag` or `cleanup`. If tagging fails, downstream jobs are skipped while `SHOULD_RELEASE` remains true, so this branch falsely says the version was tagged and removed. If a build fails after a successful tag, `notify` and `cleanup` can run concurrently, allowing Slack to claim cleanup before it finishes or even when it fails. Add both jobs to the notification dependencies and derive the message from their actual results.

In `.github/workflows/flatpak.yml`:
- [BLOCKING] .github/workflows/flatpak.yml:51-64: Preserve custom weekly versions when published drafts rebuild Flatpak
  The weekly dispatch accepts arbitrary semver values and initially passes the exact value into the Flatpak build. Publishing the draft later triggers this workflow without `workflow_call` inputs, but the recovery logic only handles canonical `vX.Y.Z-weekly.YYYYMMDD` tags. A custom value such as `1.2.0-rc.1` is rebuilt from committed version `1.0.0-dev` and overwrites the correctly versioned Flatpak assets. A custom weekly-shaped value with build metadata is also reconstructed with the tag commit SHA instead of its original metadata. Persist and recover the original override, or constrain accepted versions to a deterministically reconstructable format.
```
</details>

.github/workflows/flatpak.yml
Claudius-Maginificent: Confirmed still open — and my mistake: this one didn't make it into the last consolidation pass (I fixed the other 5 blocking findings from this review plus 2 Copilot nits, but missed this one entirely).

`flatpak.yml`'s "Resolve version override" step only reconstructs a version on `release: published` when the tag matches the canonical `vX.Y.Z-weekly.YYYYMMDD` shape (line ~54). `weekly-build.yml`'s own `version` dispatch input accepts arbitrary semver, not just that shape, so a custom-versioned weekly build's Flatpak bundle would get silently rebuilt from the committed `Cargo.toml` version on undraft, clobbering the correctly-versioned asset already on the release page — exactly as described.

Leaving unresolved. Will scope and fix in a follow-up pass — either persist the original override somewhere recoverable (e.g. as release metadata) or constrain accepted custom versions to something deterministically reconstructable.

.github/workflows/weekly-build.yml
```diff
@@ -29,6 +29,9 @@ concurrency:
   group: weekly-build
   cancel-in-progress: false # weekly cadence: never interrupt a half-cut release
 
+env:
+  TARGET_BRANCH: v1.0-dev
```
lklimek: Not sure this is correct. First of all, we default to whatever is the "main" branch at the time - it can be v1.0-dev, v1.1-dev, v2.0-dev - it's a moving target. Also we want to be able to build from some other, custom branch  - in this case, we should just take that branch into consideration, not v1.0-dev.

Claudius-Maginificent: Fair catch, and you're right — I picked the wrong one of the two options the original comment offered. `TARGET_BRANCH=v1.0-dev` hardcoded two problems: it goes stale the moment the active dev branch rolls to v1.1-dev/v2.0-dev, and it silently drops the ability to intentionally dispatch from a custom branch (checkout was forced onto `v1.0-dev` regardless of what ref was actually selected).

Redesigning to the other option instead: drop the `TARGET_BRANCH` env and the explicit checkout `ref:` overrides entirely (reverting to `actions/checkout`'s default — whatever ref triggered the run), and change the CI-red check's hardcoded `--branch v1.0-dev` to `--branch "${{ github.ref_name }}"`. That makes the CI check, checkout, tagging, and build all consistently key off whatever ref actually triggered the run — no hardcoded branch name anywhere, and dispatching from a custom branch works exactly as before, just with a consistent (not mismatched) CI check. For the `schedule` trigger specifically, `github.ref_name` resolves to the repo's actual configured default branch, so as long as that setting gets flipped when the project rolls to a new dev branch, this workflow tracks it automatically with zero further edits.

Fixing now alongside the flatpak custom-version-recovery finding (thepastaclaw, below). Will ping once pushed.

Claudius-Maginificent: Pushed in `0f7a9b42`: `TARGET_BRANCH` and the explicit checkout `ref:` overrides are gone (checkout back to default — whatever ref triggered the run), and the CI-red check now uses `--branch "${{ github.ref_name }}"` instead of a hardcoded branch name. Checkout, tagging, build, and the CI check all consistently track the same ref again, with nothing hardcoded — a schedule run picks up the repo's actual default branch automatically, and a custom-branch dispatch works as it did before.

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,290 @@
...
+
+      - name: Push weekly tag
+        run: |
+          set -euo pipefail
+          if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+            echo "Tag $TAG already exists — force re-cut: moving it to the current commit."
+            git config user.name "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git tag -f -a "$TAG" -m "Weekly draft pre-release $TAG" HEAD
+            git push --force origin "refs/tags/$TAG"
```
thepastaclaw, line 196: **🔴 Blocking: Force re-cuts can rewrite official release tags**

The manual version accepts any semver, including existing official versions such as `0.9.3`, and `force` bypasses the existing-tag gate. The tag job then force-moves that official tag to the dispatch branch's HEAD. If the build subsequently fails, cleanup can also delete the associated official release and tag through `gh release delete --cleanup-tag`. Restrict force re-cuts to tags owned by this weekly workflow before allowing an existing ref to be moved or deleted.

<sub>source: ['codex']</sub>

.github/workflows/release.yml
```diff
@@ -50,6 +70,19 @@ jobs:
...
       - name: Check out code
         uses: actions/checkout@v4
+        with:
+          # Only pin to inputs.tag for the workflow_call path (weekly-build.yml,
+          # the only caller — it always passes an already-pushed tag). Direct
+          # workflow_dispatch's `tag` input is a NEW tag name softprops/action-gh-release
+          # creates at the end of this job, not an existing ref — checking it
+          # out here would fail with "ref not found" for that path, so it must
+          # keep falling back to github.ref (the dispatch-selected branch).
+          ref: ${{ (github.event_name == 'workflow_call' && inputs.tag) || github.ref }}
```
thepastaclaw: **🔴 Blocking: The reusable workflow never selects the weekly tag**

A reusable workflow inherits the caller's GitHub event context, so `github.event_name` remains `schedule` or `workflow_dispatch`; it does not become `workflow_call`. This condition is therefore false for weekly invocations at lines 80, 213, and 542, and checkout uses the caller's branch ref instead of `inputs.tag`. A commit arriving after the tag job, or a later job rerun, can consequently build a branch tip that no longer matches the release tag.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,290 @@
...
+          if [ "$FORCE" != "true" ]; then
+            if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+              {
+                echo "should_release=false"
+                echo "reason=already_released"
+              } >>"$GITHUB_OUTPUT"
+              exit 0
+            fi
+            LAST=$(git tag -l 'v*-weekly.*' --sort=-creatordate | head -n1)
+            if [ -n "$LAST" ] && [ "$(git rev-list --count "${LAST}..HEAD")" -eq 0 ]; then
```
thepastaclaw: **🔴 Blocking: Custom-version releases bypass weekly lifecycle tracking**

The custom `version` input is documented as replacing the auto-derived weekly version, but deduplication only searches `v*-weekly.*`. A custom release such as `v1.2.0-rc.1` cut after the last canonical weekly tag is therefore ignored, so the next schedule can cut another draft without any commits since the custom release. The stale-draft scan at lines 151-153 uses an even narrower canonical-weekly regex, leaving that custom draft behind as well. Track weekly-owned releases independently of their user-selected version shape, for example through the persisted marker.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,290 @@
...
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      TAG: ${{ needs.gate.outputs.tag }}
+    steps:
+      - name: Delete orphaned weekly tag
+        run: |
+          set -euo pipefail
+          if gh release view "$TAG" -R "$GITHUB_REPOSITORY" >/dev/null 2>&1; then
+            gh release delete "$TAG" --yes --cleanup-tag -R "$GITHUB_REPOSITORY"
+          else
+            gh api --method DELETE "repos/$GITHUB_REPOSITORY/git/refs/tags/$TAG" || true
```
thepastaclaw: **🔴 Blocking: Cleanup reports success when tag deletion fails**

The fallback tag deletion suppresses every API failure with `|| true`, causing the cleanup job to finish successfully even when permissions or a transient API error leave the orphaned tag in place. The notification then selects its successful-cleanup branch and explicitly claims the tag was swept away.

```suggestion
gh api --method DELETE "repos/$GITHUB_REPOSITORY/git/refs/tags/$TAG"
```

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -55,82 +295,20 @@ jobs:
...
-              done <<< "$PASSED"
-            fi
-
-            TEXT=":rotating_light: *Even I, Claudius the Magnificent, cannot overcome every limitation of your rudimentary build infrastructure.* Here's what broke on \`${REPO}\`:${NL}${NL}${BLOCKS}"
-
-            if [ -n "$DOWNLOAD_SECTION" ]; then
-              TEXT="${TEXT}${NL}Despite the chaos, my magnificence preserved these artifacts:${DOWNLOAD_SECTION}"
-            fi
+          if [ "$CI_RED" = "true" ]; then
+            TEXT=":warning: *I, Claudius the Magnificent, decline to bless a broken week.* CI on \`v1.0-dev\` is red — no weekly draft was cut. Mend the pipeline: <${RUN_URL}|the run>."
```
thepastaclaw: **🔴 Blocking: CI-red notifications identify the wrong branch**

The gate now checks CI for `github.ref_name` so custom-branch dispatches work, but the corresponding Slack message still states that `v1.0-dev` is red. A failed custom-branch dispatch therefore directs responders to investigate the wrong branch, contradicting the custom-ref behavior added by this head commit.

```suggestion
TEXT=":warning: *I, Claudius the Magnificent, decline to bless a broken week.* CI on \`${GITHUB_REF_NAME}\` is red — no weekly draft was cut. Mend the pipeline: <${RUN_URL}|the run>."
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All six prior findings are fixed at the exact head. Five new in-scope blockers remain: forced custom versions can rewrite official release tags, reusable builds do not actually check out the weekly tag, custom-version drafts bypass lifecycle tracking, failed cleanup is reported as successful, and CI-red notifications identify the wrong branch.

### Review provenance

🔴 5 blocking

<details>

```

In `.github/workflows/weekly-build.yml`:
- [BLOCKING] .github/workflows/weekly-build.yml:164-169: Force re-cuts can rewrite official release tags
  The manual version accepts any semver, including existing official versions such as `0.9.3`, and `force` bypasses the existing-tag gate. The tag job then force-moves that official tag to the dispatch branch's HEAD. If the build subsequently fails, cleanup can also delete the associated official release and tag through `gh release delete --cleanup-tag`. Restrict force re-cuts to tags owned by this weekly workflow before allowing an existing ref to be moved or deleted.
- [BLOCKING] .github/workflows/weekly-build.yml:113-114: Custom-version releases bypass weekly lifecycle tracking
  The custom `version` input is documented as replacing the auto-derived weekly version, but deduplication only searches `v*-weekly.*`. A custom release such as `v1.2.0-rc.1` cut after the last canonical weekly tag is therefore ignored, so the next schedule can cut another draft without any commits since the custom release. The stale-draft scan at lines 151-153 uses an even narrower canonical-weekly regex, leaving that custom draft behind as well. Track weekly-owned releases independently of their user-selected version shape, for example through the persisted marker.
- [BLOCKING] .github/workflows/weekly-build.yml:265: Cleanup reports success when tag deletion fails
  The fallback tag deletion suppresses every API failure with `|| true`, causing the cleanup job to finish successfully even when permissions or a transient API error leave the orphaned tag in place. The notification then selects its successful-cleanup branch and explicitly claims the tag was swept away.
- [BLOCKING] .github/workflows/weekly-build.yml:300: CI-red notifications identify the wrong branch
  The gate now checks CI for `github.ref_name` so custom-branch dispatches work, but the corresponding Slack message still states that `v1.0-dev` is red. A failed custom-branch dispatch therefore directs responders to investigate the wrong branch, contradicting the custom-ref behavior added by this head commit.

In `.github/workflows/release.yml`:
- [BLOCKING] .github/workflows/release.yml:80: The reusable workflow never selects the weekly tag
  A reusable workflow inherits the caller's GitHub event context, so `github.event_name` remains `schedule` or `workflow_dispatch`; it does not become `workflow_call`. This condition is therefore false for weekly invocations at lines 80, 213, and 542, and checkout uses the caller's branch ref instead of `inputs.tag`. A commit arriving after the tag job, or a later job rerun, can consequently build a branch tip that no longer matches the release tag.
```
</details>

.github/workflows/release.yml
```diff
@@ -48,8 +68,39 @@ jobs:
...
+        id: resolve_ref
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          tag="${{ inputs.tag }}"
+          if [ -n "$tag" ] && gh api "repos/$GITHUB_REPOSITORY/git/refs/tags/$tag" >/dev/null 2>&1; then
+            echo "ref=$tag" >> "$GITHUB_OUTPUT"
+          else
+            echo "ref=${{ github.ref }}" >> "$GITHUB_OUTPUT"
```
thepastaclaw: **🔴 Blocking: Keep checkout refs out of the generated shell program**

GitHub substitutes `inputs.tag` and `github.ref` into the Bash source before execution. A dispatch input such as `v1"; command; #` escapes the assignment, and valid Git ref names can also contain quotes and shell metacharacters. This executes in all three build jobs with release-write and OIDC permissions before checkout. Pass both values through `env`, validate the requested tag as a Git ref, and make the same change in the resolver blocks around lines 228 and 575.

<sub>source: ['codex']</sub>

.github/workflows/release.yml
```diff
@@ -48,8 +68,39 @@ jobs:
...
+          fi
```
thepastaclaw: **🔴 Blocking: Fail closed when checkout-tag verification cannot complete**

Every failure from `gh api` enters the same fallback used for a confirmed missing tag. A 404 is expected when a direct dispatch intends to create a new tag, but authentication failures, rate limits, and transient network or API failures also select `github.ref`. The weekly caller has already pushed the requested tag, so this fallback can make different platform jobs build a newer branch commit while publishing under the older tag. Distinguish a confirmed missing ref from operational errors and require successful tag resolution for callers that pass an existing tag.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,305 @@
...
+          # (it's a safety net, not a convenience toggle). Treated as "latest
+          # *completed* run of each CI workflow on v1.0-dev" rather than
+          # HEAD-sha check-runs, because tests.yml/clippy.yml are path-filtered
+          # and a docs-only HEAD would otherwise have zero check-runs and read
+          # as falsely green. A workflow that has never run is treated as
+          # absent, not red.
+          red=false
+          for wf in tests.yml clippy.yml; do
+            concl=$(gh run list --workflow "$wf" --branch "${{ github.ref_name }}" --status completed \
+                      --limit 1 --json conclusion --jq '.[0].conclusion // ""')
```
thepastaclaw: **🔴 Blocking: Pass the selected branch to the CI query through the environment**

`github.ref_name` is inserted directly into Bash inside a quoted argument. Git permits branch names containing quotes, semicolons, dollar signs, and `#`, so a dispatched custom branch can terminate the argument and execute commands in the gate step. That command can also forge `GITHUB_OUTPUT` values that control the privileged tag and release jobs. Use the already-safe environment mechanism for the branch name and reference the shell variable in `gh run list`.

<sub>source: ['codex']</sub>

.github/workflows/flatpak.yml
```diff
@@ -37,6 +43,37 @@ jobs:
...
+      - name: Resolve version override
+        id: vsel
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          ov="${{ inputs.version_override }}"
+          if [ -z "$ov" ] && [ "${{ github.event_name }}" = "release" ]; then
+            tag="${{ github.event.release.tag_name }}"
```
thepastaclaw: **🔴 Blocking: Treat Flatpak version and release-tag values as shell data**

The release tag is substituted directly into `tag="..."` before Bash parses the step, and valid Git tags may contain quote and command-separator characters. Publishing a release for such a tag can therefore execute commands in this `contents: write` job before the tag is queried or any recovered version reaches the validating composite action. Pass the release tag and reusable-workflow input through `env` rather than embedding either expression in the script body.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,305 @@
...
+            if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+              {
+                echo "should_release=false"
+                echo "reason=already_released"
+              } >>"$GITHUB_OUTPUT"
+              exit 0
+            fi
+            LAST=$(gh api "repos/$GITHUB_REPOSITORY/releases" --paginate \
+              --jq '[.[] | select(.draft == true and ((.body // "") | test("<!-- weekly-version:")))] | sort_by(.created_at) | reverse | .[0].tag_name // empty')
+            if [ -n "$LAST" ] && git rev-parse -q --verify "refs/tags/$LAST" >/dev/null && [ "$(git rev-list --count "${LAST}..HEAD")" -eq 0 ]; then
```
thepastaclaw, line 116: **🔴 Blocking: Published weekly releases bypass the no-new-commits gate**

The last-release query only considers marker-bearing releases while they are drafts. Publishing the draft preserves its weekly-version marker but removes it from this query. On the next scheduled date, the derived tag is different, so the workflow creates another build even when HEAD has no commits after the published weekly tag. Use all marker-bearing releases for the no-new-commits comparison and retain the `draft == true` restriction only for stale-draft deletion.

<sub>source: ['codex']</sub>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,305 @@
...
+        run: |
+          set -euo pipefail
+          mapfile -t olds < <(gh api "repos/$GITHUB_REPOSITORY/releases" --paginate \
+            --jq '.[] | select(.draft == true and ((.body // "") | test("<!-- weekly-version:"))) | .tag_name')
+          for old in "${olds[@]}"; do
+            [ -z "$old" ] && continue
+            [ "$old" = "$TAG" ] && continue # never touch today's tag (force re-cut)
+            echo "Removing stale weekly draft $old"
+            gh release delete "$old" --yes --cleanup-tag
+          done
```
thepastaclaw, line 184: **🔴 Blocking: Validate the force re-cut target before deleting stale drafts**

The tag job deletes every other marker-bearing draft and its tag before checking whether the requested target tag is owned by this workflow. For example, `force=true` with `version=0.9.3` deletes the current valid weekly draft and then fails when the next step recognizes `v0.9.3` as an official tag. Preflight the target tag's ownership before performing any stale-release deletion.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Most prior release-workflow issues are fixed, but six in-scope blockers remain. The workflows still contain three expression-to-shell injection paths, can silently build the triggering branch when tag verification fails, ignore published weekly releases during deduplication, and delete valid stale drafts before rejecting an invalid force re-cut.

### Review provenance

🔴 6 blocking

<details>

```

In `.github/workflows/release.yml`:
- [BLOCKING] .github/workflows/release.yml:77-81: Keep checkout refs out of the generated shell program
  GitHub substitutes `inputs.tag` and `github.ref` into the Bash source before execution. A dispatch input such as `v1"; command; #` escapes the assignment, and valid Git ref names can also contain quotes and shell metacharacters. This executes in all three build jobs with release-write and OIDC permissions before checkout. Pass both values through `env`, validate the requested tag as a Git ref, and make the same change in the resolver blocks around lines 228 and 575.
- [BLOCKING] .github/workflows/release.yml:78-82: Fail closed when checkout-tag verification cannot complete
  Every failure from `gh api` enters the same fallback used for a confirmed missing tag. A 404 is expected when a direct dispatch intends to create a new tag, but authentication failures, rate limits, and transient network or API failures also select `github.ref`. The weekly caller has already pushed the requested tag, so this fallback can make different platform jobs build a newer branch commit while publishing under the older tag. Distinguish a confirmed missing ref from operational errors and require successful tag resolution for callers that pass an existing tag.

In `.github/workflows/weekly-build.yml`:
- [BLOCKING] .github/workflows/weekly-build.yml:72-73: Pass the selected branch to the CI query through the environment
  `github.ref_name` is inserted directly into Bash inside a quoted argument. Git permits branch names containing quotes, semicolons, dollar signs, and `#`, so a dispatched custom branch can terminate the argument and execute commands in the gate step. That command can also forge `GITHUB_OUTPUT` values that control the privileged tag and release jobs. Use the already-safe environment mechanism for the branch name and reference the shell variable in `gh run list`.
- [BLOCKING] .github/workflows/weekly-build.yml:113-115: Published weekly releases bypass the no-new-commits gate
  The last-release query only considers marker-bearing releases while they are drafts. Publishing the draft preserves its weekly-version marker but removes it from this query. On the next scheduled date, the derived tag is different, so the workflow creates another build even when HEAD has no commits after the published weekly tag. Use all marker-bearing releases for the no-new-commits comparison and retain the `draft == true` restriction only for stale-draft deletion.
- [BLOCKING] .github/workflows/weekly-build.yml:149-159: Validate the force re-cut target before deleting stale drafts
  The tag job deletes every other marker-bearing draft and its tag before checking whether the requested target tag is owned by this workflow. For example, `force=true` with `version=0.9.3` deletes the current valid weekly draft and then fails when the next step recognizes `v0.9.3` as an official tag. Preflight the target tag's ownership before performing any stale-release deletion.

In `.github/workflows/flatpak.yml`:
- [BLOCKING] .github/workflows/flatpak.yml:53-55: Treat Flatpak version and release-tag values as shell data
  The release tag is substituted directly into `tag="..."` before Bash parses the step, and valid Git tags may contain quote and command-separator characters. Publishing a release for such a tag can therefore execute commands in this `contents: write` job before the tag is queried or any recovered version reaches the validating composite action. Pass the release tag and reusable-workflow input through `env` rather than embedding either expression in the script body.
```
</details>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,307 @@
...
+            # could let `cleanup` delete — a real, already-published release
+            # tag. Only proceed if a GH release for this tag exists, is still
+            # a draft, and carries our own weekly-version marker in its body
+            # (the same marker `finalize` writes) — i.e. it's provably one of
+            # THIS workflow's own not-yet-published drafts, never anything
+            # else.
+            # shellcheck disable=SC2016
+            owned=$(gh api "repos/$GITHUB_REPOSITORY/releases" --paginate \
+              --jq --arg tag "$TAG" '.[] | select(.tag_name == $tag and .draft == true and ((.body // "") | test("<!-- weekly-version:"))) | "true"' \
+              2>/dev/null | head -n1 || echo "")
```
thepastaclaw: **🔴 Blocking: The new ownership lookup cannot run**

`gh api` accepts one expression after `--jq`; it does not forward standalone jq options such as `--arg`. With the runner-compatible gh 2.96.0, this exact command exits before any API request with `accepts 1 arg(s), received 4`. The discarded stderr and `|| echo ""` then leave `owned` empty, so every legitimate force re-cut of an existing weekly draft is rejected. This happens after the preceding step may already have deleted other weekly drafts.

```suggestion
owned=$(
  gh api "repos/$GITHUB_REPOSITORY/releases" --paginate --slurp |
    jq -r --arg tag "$TAG" 'any(.[][]; .tag_name == $tag and .draft == true and ((.body // "") | test("<!-- weekly-version:")))'
)
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The exact rebased head remains non-mergeable: all seven reported blockers are confirmed in the current PR range. Six earlier blockers remain, and the newest ownership lookup adds a regression because `gh api` rejects its unsupported `--arg` arguments before making an API request. Eleven older findings are fixed.

### Review provenance

🔴 7 blocking

<details>

```

In `.github/workflows/weekly-build.yml`:
- [BLOCKING] .github/workflows/weekly-build.yml:175-177: The new ownership lookup cannot run
  `gh api` accepts one expression after `--jq`; it does not forward standalone jq options such as `--arg`. With the runner-compatible gh 2.96.0, this exact command exits before any API request with `accepts 1 arg(s), received 4`. The discarded stderr and `|| echo ""` then leave `owned` empty, so every legitimate force re-cut of an existing weekly draft is rejected. This happens after the preceding step may already have deleted other weekly drafts.
```
</details>

.github/workflows/weekly-build.yml
```diff
@@ -1,50 +1,306 @@
...
+              echo "reason=ci_red"
+            } >>"$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "ci_red=false" >>"$GITHUB_OUTPUT"
+
+          # Version / tag derivation
+          if [ -n "$MANUAL_VERSION" ]; then
+            # Fail-closed on malformed human input — it flows into `git tag`.
+            if ! printf '%s' "$MANUAL_VERSION" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'; then
```
thepastaclaw, line 92: **🔴 Blocking: Reject multiline version inputs before writing job outputs**

`grep -Eq` applies the anchored expression to individual input lines, so a multiline dispatch value passes whenever any line is valid semver. For example, `1.2.3\nshould_release=false\ntag=v0.9.3` passes this check; lines 124-129 then write the embedded lines into `GITHUB_OUTPUT`, where they are parsed as additional outputs and can override values controlling the privileged tag and release jobs. The identical validator in `.github/actions/apply-version-override/action.yml:28` also accepts multiline data, so both checks should validate the complete string rather than any matching line.

```suggestion
if [[ ! "$MANUAL_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest delta fixes prior `38758cb8b6ff` and introduces no new delta-specific blocker. Six carried-forward current-head priors remain blocking. Cumulative current-PR inspection newly confirms one additional blocker: multiline custom-version input can inject gate outputs.

### Review provenance

🔴 7 blocking

<details>

```

In `.github/workflows/weekly-build.yml`:
- [BLOCKING] .github/workflows/weekly-build.yml:91: Reject multiline version inputs before writing job outputs
  `grep -Eq` applies the anchored expression to individual input lines, so a multiline dispatch value passes whenever any line is valid semver. For example, `1.2.3\nshould_release=false\ntag=v0.9.3` passes this check; lines 124-129 then write the embedded lines into `GITHUB_OUTPUT`, where they are parsed as additional outputs and can override values controlling the privileged tag and release jobs. The identical validator in `.github/actions/apply-version-override/action.yml:28` also accepts multiline data, so both checks should validate the complete string rather than any matching line.
```
</details>

.github/workflows/weekly-build.yml
thepastaclaw: Resolved in `dbb771e` — _The new ownership lookup cannot run_ no longer present.

thepastaclaw (changes requested):

At exact head 03468c4da889de3d396a68032f8d7ae08226c6f4, all seven findings indexed from d99e2303 are fixed. The blocker carried forward from the later dbb771eb review remains valid because both whole-version validators are still line-oriented and the weekly gate still writes the accepted multiline value into job outputs; no separate new finding was identified in d99e2303..03468c4d.

### Review provenance

🔴 1 blocking

.github/workflows/weekly-build.yml
thepastaclaw: Resolved in this update — _Pass the selected branch to the CI query through the environment_ no longer present.

.github/workflows/weekly-build.yml
thepastaclaw: Resolved in this update — _Validate the force re-cut target before deleting stale drafts_ no longer present.

.github/workflows/release.yml
thepastaclaw: Resolved in this update — _Keep checkout refs out of the generated shell program_ no longer present.

.github/workflows/release.yml
thepastaclaw: Resolved in this update — _Fail closed when checkout-tag verification cannot complete_ no longer present.

.github/workflows/flatpak.yml
thepastaclaw: Resolved in this update — _Treat Flatpak version and release-tag values as shell data_ no longer present.

.github/workflows/weekly-build.yml
Claudius-Maginificent: Fixed in `0f7a9b42` — `TARGET_BRANCH` was removed entirely. The CI-red check now keys off `github.ref_name` (the ref that actually triggered the run) and checkout uses the default `github.ref`, so it always tracks whatever branch you dispatch from — no more hardcoded `v1.0-dev`, no moving-target problem, and custom-branch dispatch works correctly.

One residual nit: the file's top-of-file comment (line 3) still reads "every Saturday from v1.0-dev" — accurate for the schedule trigger (fires from the default branch) but could be read as implying the branch is hardcoded everywhere. Happy to tighten that wording if you'd like, otherwise leaving open for you to close.

PR #897, opened 2026-07-16
fix(platform): keep cause-less transition results unconfirmed

lklimek:
## Why this PR exists

- **Problem**: Dash Platform can accept a state-transition broadcast while the
  separate result wait fails. The SDK represents a cause-less wait failure as
  `StateTransitionBroadcastError`, and DET currently reports every such error as
  a Platform rejection.
- **What breaks without it**: A mainnet identity withdrawal was delivered after
  its broadcast succeeded, but DET displayed “The platform rejected this
  request” and invited a retry. Retrying an action that may already have
  completed can duplicate a funds-moving or paid operation.
- **Blocking relationship**: Stacked on #894. The generic Platform-side
  retry/error-contract fix is tracked in dashpay/platform#4137.

## What was done

- Added `TaskError::PlatformResultUnconfirmed` with a user-facing message that
  says the request was submitted but could not be confirmed and directs the user
  to check whether it completed before retrying.
- Classify `StateTransitionBroadcastError { cause: None, .. }` as unconfirmed.
- Keep structured `cause: Some(consensus_error)` errors on the existing
  consensus-specific or `PlatformRejected` paths.
- Removed the message-only duplicate-key fallback for cause-less broadcast
  envelopes. Without a structured consensus cause, message text alone does not
  prove rejection.
- Added regression coverage for the observed code-14
  `Tenderdash is not available` response and for a cause-less message that looks
  like a duplicate-key rejection.
- Updated the changelog.

## Testing

- Regression tests confirmed failing before the production change.
- `cargo +nightly fmt --all -- --check`
- `cargo test --lib backend_task::error::tests::` — 98 passed
- `cargo clippy --lib --all-features -- -D warnings`

## Breaking changes

None.

## Checklist

- [x] The branch is based on the current PR #894 head.
- [x] User-visible behavior is covered by regression tests.
- [x] Formatting and targeted lint/test gates pass.
- [x] No dependencies or persisted data formats changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,6 +180,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **Submitted Platform actions are no longer reported as rejected when only
+  confirmation failed**: if a state transition was broadcast but its result
+  could not be confirmed, the app now tells you to check whether it completed
+  before trying again instead of showing an unsafe rejection-and-retry message.
+
 - **Shielded actions now say when they are unavailable instead of failing
   obscurely**: if shielding, sending, or withdrawing shielded funds is not
   available on your network yet, the app says so and points you at a regular
--- a/src/backend_task/error.rs
+++ b/src/backend_task/error.rs
@@ -1095,6 +1095,15 @@ pub enum TaskError {
         source_error: Box<SdkError>,
     },
 
+    /// Platform accepted the request, but its result could not be confirmed.
+    #[error(
+        "Your request was submitted, but its result could not be confirmed. Check whether it completed before trying again."
+    )]
+    PlatformResultUnconfirmed {
+        #[source]
+        source_error: Box<SdkError>,
+    },
+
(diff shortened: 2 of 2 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit d278dcc7)

thepastaclaw (approved):

This is a small, well-scoped fix that correctly closes a safety gap: a state transition broadcast that succeeds but whose result-confirmation fails with no structured consensus cause was previously misclassified in ways that could invite an unsafe retry. The new TaskError::PlatformResultUnconfirmed variant, its guard ordering in the SdkError->TaskError match, and the removal of the fragile message-string duplicate-key heuristic are all correctly implemented and covered by new tests for both the removed heuristic path and an unrelated cause-less code (Unavailable). No blocking issues found by either agent, and independent verification of the diff, the match-arm guard, and the shielded_broadcast_error dead-code claim all check out.

### Review provenance
- Verifier: `claude-sonnet-5` — final-verifier
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — general (completed)

PR #898, opened 2026-07-16
ci: allow manual test/clippy runs, cover rust-toolchain.toml, drop dead push branch

lklimek:
## Why this PR exists

- **Problem**: Three narrow gaps in the `tests.yml` / `clippy.yml` triggers. A draft PR cannot run CI at all. A toolchain bump runs neither tests nor clippy. And the push trigger points at a branch that does not exist.

- **What breaks without it**:
  1. *A draft PR cannot be tested.* Both jobs are gated on `github.event.pull_request.draft != true`, and neither workflow declared `workflow_dispatch` — so a draft had **zero** CI and **no** manual override. The only way to get a run was to mark the PR ready for review. Since PRs here are opened as drafts by convention, "push it and let CI check" silently did nothing.
  2. *A toolchain bump is never tested.* `rust-toolchain.toml` pins the compiler for every build, but matched no path filter. Edit it alone — changing the toolchain the whole workspace compiles against — and neither tests nor clippy fire.
  3. *Pushes to the release branch ran nothing.* The push trigger listed `main`, which does not exist in this repo (the branches are `master` and `v1.0-dev`), so it matched nothing at all.

- **Blocking relationship**: None. Targets `v1.0-dev` directly.

## What was done

- **Added `workflow_dispatch`** to both workflows, so a draft PR can be exercised from the Actions tab or via `gh workflow run tests.yml --ref <branch>`. The existing draft guard already permits this: for a `workflow_dispatch` event `github.event.pull_request` is null, so `draft != true` evaluates true. A manual run tests the **branch head**, not the PR merge commit, so it does not prove the merged result is green.
- **Added `rust-toolchain.toml`** to the path filters. `**/*.rs` already covers `build.rs` and every crate's sources, so the filters now fire on any Rust code change and on no documentation change — which is the intent.
- **Reduced the push branches to `v*-dev`.** `main` does not exist; `master` is release-only and needs no push-triggered run.
- **Shared the path lists between the `push` and `pull_request` triggers via a YAML anchor**, so the two lists cannot silently drift apart. GitHub Actions has supported anchors since [2025-09-18](https://github.blog/changelog/2025-09-18-actions-yaml-anchors-and-non-public-workflow-templates/), and path filters are a documented use case. (Merge keys remain unsupported; none are used here.)

The `paths` filters are deliberately **kept** — documentation-only changes should not spend CI minutes.

## Testing

Both files parse, and the triggers resolve as intended:

| Workflow | Triggers | push branches | `rust-toolchain.toml` filtered | anchor resolves push == PR | `workflow_dispatch` |
|---|---|---|---|---|---|
| `tests.yml` | push, pull_request, workflow_dispatch | `v*-dev` | yes | yes | yes |
| `clippy.yml` | push, pull_request, workflow_dispatch | `v*-dev` | yes | yes | yes |

This PR edits both workflow files, and each filter lists its own workflow file, so both gates run against it here.

## Breaking changes

None. Behaviour only widens: `workflow_dispatch` is additive, `rust-toolchain.toml` adds a trigger path, and the dropped push branches (`main`, `master`) triggered nothing useful — `main` does not exist, and `master` receives no development pushes.

## Summary by CodeRabbit

* **Chores**
  * Updated automated checks to run only for development branch pushes and relevant code changes.
  * Added manual options to run linting and tests on demand.
  * Improved workflow coverage for changes to CI configuration and toolchain settings.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -50,12 +50,19 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-cargo-registry-
 
+      # No `toolchain` or `components` input: the action installs the channel and
+      # the components rust-toolchain.toml pins, so a bump there is linted by the
+      # compiler it introduces. Naming 1.92 here duplicated that pin and would
+      # silently keep linting 1.92 after the file moved on.
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: 1.92
-          components: clippy, rustfmt
-          override: true
+          # This action defaults RUSTFLAGS to "-D warnings". The Clippy step
+          # already passes `-D warnings` itself, and setting it globally would
+          # also deny warnings for the fmt check's build — keep it unset.
+          rustflags: ""
+          # Cargo caching is handled by the actions/cache step above.
+          cache: false
 
       - name: Install system dependencies
         run: |
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,23 +1,26 @@
 name: Clippy
 
+# Path filters below fire on any change that can affect a lint or fmt result,
+# and on nothing that cannot. `**/*.rs` covers build.rs and every crate's
+# sources; documentation changes deliberately match nothing.
 on:
   push:
     branches:
-      - master
       - "v*-dev"
-    paths:
-      - "**/*.rs"
+    paths: &code_paths
+      - "**/*.rs" # includes build.rs
       - "**/Cargo.toml"
       - "Cargo.lock"
       - ".cargo/config.toml"
+      - "rust-toolchain.toml"
       - ".github/workflows/clippy.yml"
-  # No `paths` filter: every non-draft pull request runs clippy and the fmt
-  # check, so a green PR always means they actually ran.
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-  # Allows a manual run against a branch — the only way to exercise a draft PR,
-  # whose `pull_request` runs are suppressed by the draft guard below. Note a
-  # manual run tests the branch head, not the PR merge commit.
+    paths: *code_paths
+  # Allows a manual run against a branch. Draft PRs are suppressed by the
+  # `draft != true` guard below and otherwise have no way to run CI at all; a
+  # manual run covers them. Note it tests the branch head, not the PR merge
+  # commit, so it does not prove the merged result is green.
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -3,7 +3,7 @@ name: Clippy
 on:
   push:
     branches:
-      - main
+      - master
       - "v*-dev"
     paths:
       - "**/*.rs"
@@ -11,14 +11,14 @@ on:
       - "Cargo.lock"
       - ".cargo/config.toml"
       - ".github/workflows/clippy.yml"
+  # No `paths` filter: every non-draft pull request runs clippy and the fmt
+  # check, so a green PR always means they actually ran.
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    paths:
-      - "**/*.rs"
-      - "**/Cargo.toml"
-      - "Cargo.lock"
-      - ".cargo/config.toml"
-      - ".github/workflows/clippy.yml"
+  # Allows a manual run against a branch — the only way to exercise a draft PR,
+  # whose `pull_request` runs are suppressed by the draft guard below. Note a
+  # manual run tests the branch head, not the PR merge commit.
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,7 @@ name: Tests
 on:
   push:
     branches:
-      - main
+      - master
       - "v*-dev"
     paths:
       - "**/*.rs"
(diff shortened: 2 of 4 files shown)


.github/workflows/clippy.yml
```diff
@@ -3,22 +3,22 @@ name: Clippy
 on:
   push:
     branches:
-      - main
+      - master
```
lklimek: no master here

thepastaclaw:
✅ Final review complete — no blockers (commit 73965871)

.github/workflows/tests.yml
```diff
@@ -1,26 +1,28 @@
...
-      - main
       - "v*-dev"
-    paths:
-      - "**/*.rs"
+    paths: &code_paths
+      - "**/*.rs" # includes build.rs
       - "**/Cargo.toml"
       - "Cargo.lock"
       - ".cargo/config.toml"
+      - "rust-toolchain.toml"
```
thepastaclaw, line 15: **🔴 Blocking: Toolchain-file changes still are not tested**

Adding `rust-toolchain.toml` to the path filter makes both workflows run when the file changes, but neither job uses its selected compiler. The test job sets `toolchain: stable` with `override: true` at lines 54-58, while `.github/workflows/clippy.yml` sets `toolchain: 1.92` with `override: true` at lines 53-58. These directory overrides determine which toolchain runs instead of `rust-toolchain.toml`; the test job may coincide with the file only when its version happens to be the current stable release, and clippy remains fixed at 1.92. A toolchain bump can therefore pass without testing or linting the version being introduced. Derive both action inputs from `rust-toolchain.toml`, or otherwise install and run the file's exact channel.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The manual dispatch and shared path filters are structurally sound. However, the new toolchain-file trigger does not test the selected toolchain because both jobs replace it with explicit rustup overrides, leaving a stated goal of this PR unmet.

### Review provenance

🔴 1 blocking

<details>

```

In `.github/workflows/tests.yml`:
- [BLOCKING] .github/workflows/tests.yml:15: Toolchain-file changes still are not tested
  Adding `rust-toolchain.toml` to the path filter makes both workflows run when the file changes, but neither job uses its selected compiler. The test job sets `toolchain: stable` with `override: true` at lines 54-58, while `.github/workflows/clippy.yml` sets `toolchain: 1.92` with `override: true` at lines 53-58. These directory overrides determine which toolchain runs instead of `rust-toolchain.toml`; the test job may coincide with the file only when its version happens to be the current stable release, and clippy remains fixed at 1.92. A toolchain bump can therefore pass without testing or linting the version being introduced. Derive both action inputs from `rust-toolchain.toml`, or otherwise install and run the file's exact channel.
```
</details>

.github/workflows/tests.yml
lklimek: Confirmed and fixed in `73965871` — the finding was valid, and verifying it surfaced something worse than a hollow trigger.

`rust-toolchain.toml` pins `channel = "1.92"`, but the test job named `stable` with `override: true`. A rustup directory override takes precedence over the toolchain file, so the suite has **never run on the pinned toolchain at all** — not merely "coincides when 1.92 happens to be stable". Clippy named `1.92` by hand, matching only because the number was duplicated; the next bump would have kept linting the old compiler.

Fix: both jobs now install via `actions-rust-lang/setup-rust-toolchain@v1` with **no** `toolchain` input, so the channel and components come straight from `rust-toolchain.toml`. Clippy's `components` moved into that file. It is now the single source of truth, and because a change to it triggers both workflows, a toolchain bump is tested and linted by the compiler it introduces.

One note on the suggested remedy, which is why this took a different route than proposed: **`dtolnay/rust-toolchain` would not have fixed this.** It takes its toolchain from its `@ref` (`@stable` installs stable) and does not read `rust-toolchain.toml` — so adopting it would have reproduced the same defect with different syntax. `actions-rust-lang/setup-rust-toolchain` is the one that reads the file when no toolchain is given.

Two of that action's defaults are pinned rather than inherited, both of which would have been silent behaviour changes:

- `rustflags` defaults to `-D warnings` — that would newly fail the **test** job on any warning. Clippy owns that gate and its own step already passes the flag, so it is set to `""`.
- `cache` defaults to `true` (Swatinem/rust-cache), which would duplicate the existing `actions/cache` step; set to `false`.

Also drops `actions-rs/toolchain` from both jobs — it is archived.

.github/workflows/clippy.yml
```diff
@@ -1,24 +1,27 @@
...
       - "v*-dev"
-    paths:
-      - "**/*.rs"
+    paths: &code_paths
+      - "**/*.rs" # includes build.rs
       - "**/Cargo.toml"
       - "Cargo.lock"
       - ".cargo/config.toml"
+      - "rust-toolchain.toml"
       - ".github/workflows/clippy.yml"
```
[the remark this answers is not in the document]
lklimek: Confirmed and fixed in `73965871`. Same root cause as the `tests.yml` thread — replying here for completeness.

Clippy's hard-coded `1.92` matched `rust-toolchain.toml` only because the version was duplicated by hand; with `override: true` the rustup directory override wins over the file regardless, so a bump to the file would have kept linting 1.92 exactly as you describe.

Both jobs now install via `actions-rust-lang/setup-rust-toolchain@v1` with no `toolchain` input, taking channel and components from `rust-toolchain.toml`. Clippy's `components = ["clippy", "rustfmt"]` moved into that file, so the version is no longer duplicated anywhere.

On the specific suggestion: **`dtolnay/rust-toolchain@stable` would not have worked here** — it takes its toolchain from the `@ref`, so `@stable` installs stable and ignores `rust-toolchain.toml` entirely, reproducing the bug. (It is used elsewhere in this repo, as you note — but in `flatpak.yml` / `release.yml`, where tracking stable is the intent.) `actions-rust-lang/setup-rust-toolchain` reads the toolchain file when no toolchain input is given, which is what this needs.

thepastaclaw (approved):

This PR adds CI for the project across three commits: enabling tests/clippy on non-draft PRs, adding workflow_dispatch plus path-filter coverage for the toolchain file and dropping the dead `main` push branch, and finally replacing the hardcoded/overridden toolchain install with actions-rust-lang/setup-rust-toolchain reading the pin directly from rust-toolchain.toml. No new findings in the latest delta. The one carried-forward prior finding (toolchain-file changes not actually tested) is fixed at this head: both tests.yml and clippy.yml install via setup-rust-toolchain@v1 with no toolchain/override input, so they run whatever rust-toolchain.toml pins, and that file is in both workflows' path filters, so a bump both triggers and is tested by the toolchain it introduces.

### Review provenance
- Verifier: `claude-sonnet-5` — verifier
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — general (failed)

PR #900, opened 2026-07-16
fix(wallets): show the fee on outgoing transaction history

lklimek:
## Why this PR exists

- **Problem**: The Transaction History "Fee" column renders `-` for every transaction, always. The column is correctly wired to `tx.fee`, but discovered `TransactionRecord`s are always built with `fee: None`, so there is never a value to show. Separately, `estimate_platform_fee` takes no output count, so the advanced Platform→Platform preview underestimates multi-output transfers.

- **What breaks without it**:
  1. *The fee column is dead.* Send any amount from a wallet, open Transaction History, look at "Fee": it shows `-`. It shows `-` for every row, forever — a user cannot see what any transaction cost them, and the column looks broken rather than empty-by-design.
  2. *Multi-output platform transfers are underestimated.* The advanced Platform→Platform preview already computes its output count but has no way to pass it to `estimate_platform_fee`, which prices a single output. A transfer with several outputs is quoted below its real cost.

- **Blocking relationship**: None — targets #860's branch. This commit was **orphaned**: it sat on an abandoned branch while #894 was squash-merged, and the squash absorbed everything around it but not this. It looked merged because the base does contain an SND-005 commit — but that one is the *pre-send fee estimate*, a different defect. The history-column fix was never merged.

## What was done

- **Derive the fee for outgoing transactions** (`wallet_backend/snapshot.rs`). `transaction_fee(record)` computes `sum(known inputs) - sum(outputs)` for ordinary wallet-created outgoing transactions, which already carry every input and output value — so no new persistence is needed. It preserves an already-recorded fee, requires complete and correctly indexed input details, uses checked sums and checked subtraction, and returns `None` on incomplete data, overflow, or underflow, leaving the column at `-` rather than showing a wrong number.
- **Generalize `estimate_platform_fee` to take an output count** (`model/fee_estimation.rs`). Simple callers pass `1`; the advanced Platform→Platform path passes the count it already had.

Reapplied onto the post-#894 base, which independently landed a newer SND-005 pre-send estimate. **That newer base version is kept.** This change drops the orphaned commit's older duplicate send-screen renderer, and its fixed-action shielded-route preview — the latter deliberately, because a fixed two-action estimate would misreport the fee when backend note selection determines the real action count.

## Testing

- `outgoing_transaction_fee_is_derived_from_known_inputs` — new regression test; asserts the derived fee is `Some(300)`. Confirmed **by name** in the run log, not inferred from a green exit.
- 42 fee-related tests pass, including `platform_fee_accounts_for_every_output` covering the output-count fix.
- `cargo clippy --all-features --all-targets -- -D warnings` — clean, which is also the gate proving no dead code was left behind by the dropped shielded-route path.
- `cargo +nightly fmt --all` — clean.

Net diff is 3 files, +91/−8.

## Breaking changes

None. The fee column populates where it previously showed `-`; a transaction whose inputs are not fully known still shows `-`. `estimate_platform_fee`'s new output-count argument is internal.

### DIFF
--- a/src/ui/wallets/send_screen.rs
+++ b/src/ui/wallets/send_screen.rs
@@ -359,7 +359,7 @@ impl WalletSendScreen {
 
         let usable_count = sorted_addresses.len().min(MAX_PLATFORM_INPUTS);
         if usable_count == 0 {
-            return estimate_platform_fee(fee_estimator, 1);
+            return estimate_platform_fee(fee_estimator, 1, 1);
         }
 
         let dest_kind = self.validated_destination.as_ref().map(|v| v.kind());
@@ -383,7 +383,7 @@ impl WalletSendScreen {
             }
         }
 
-        estimate_platform_fee(fee_estimator, usable_count)
+        estimate_platform_fee(fee_estimator, usable_count, 1)
     }
 
     /// Clear the AddressInput widget so it picks up the new network on next frame.
@@ -2830,6 +2830,7 @@ impl WalletSendScreen {
                 Some(estimate_platform_fee(
                     &self.app_context.fee_estimator(),
                     num_inputs,
+                    num_outputs,
                 ))
             }
             _ => None,
--- a/src/wallet_backend/snapshot.rs
+++ b/src/wallet_backend/snapshot.rs
@@ -143,12 +143,39 @@ pub(super) fn map_transaction_record(record: &TransactionRecord) -> WalletTransa
         height: record.height(),
         block_hash: block_info.map(|bi| bi.block_hash()),
         net_amount: record.net_amount,
-        fee: record.fee,
+        fee: transaction_fee(record),
         label: Some(record.label.clone()).filter(|s| !s.is_empty()),
         // Per-wallet history — every record involves our addresses.
         is_ours: true,
         status: status_from_context(&record.context),
     }
+}
+
+fn transaction_fee(record: &TransactionRecord) -> Option<u64> {
+    if record.fee.is_some() {
+        return record.fee;
+    }
+    if record.transaction.input.is_empty()
+        || record.input_details.len() != record.transaction.input.len()
+        || record
+            .input_details
+            .iter()
+            .enumerate()
+            .any(|(index, detail)| detail.index as usize != index)
+    {
+        return None;
+    }
+
+    let input_total = record
+        .input_details
+        .iter()
+        .try_fold(0u64, |total, input| total.checked_add(input.value))?;
+    let output_total = record
+        .transaction
+        .output
+        .iter()
+        .try_fold(0u64, |total, output| total.checked_add(output.value))?;
+    input_total.checked_sub(output_total)
 }
 
 /// The `(transaction, [(outpoint, txout, address)])` payload the asset-lock and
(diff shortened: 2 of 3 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit ca8d7081)

src/ui/wallets/send_screen.rs
```diff
@@ -2830,6 +2830,7 @@ impl WalletSendScreen {
                 Some(estimate_platform_fee(
                     &self.app_context.fee_estimator(),
                     num_inputs,
+                    num_outputs,
                 ))
             }
```
thepastaclaw, line 2835: **🔴 Blocking: Count the Platform outputs actually submitted**

This branch prices every row with a non-empty address, but `send_advanced_platform_to_platform` skips zero-credit rows and inserts outputs into a `BTreeMap`, which coalesces duplicate destinations. Two rows for the same Platform address therefore submit one transition output but are priced as two, and a populated zero-credit row is priced despite being omitted. Because the newly generalized fee calculation charges per output, derive the count from the same distinct, positive-credit destinations used by the send path.

```suggestion
AdvancedSourceType::Platform if has_platform_out && !has_core_out => {
    let num_inputs = self
        .platform_inputs
        .iter()
        .filter(|i| !i.amount.trim().is_empty())
        .count()
        .max(1);
    let num_outputs = self
        .advanced_outputs
        .iter()
        .filter_map(|output| {
            let destination =
                PlatformAddress::from_bech32m_string(output.address.trim()).ok()?;
            let credits = Self::parse_amount_to_credits(&output.amount).ok()?;
            (credits > 0).then_some(destination)
        })
        .collect::<std::collections::BTreeSet<_>>()
        .len();
    if num_outputs == 0 {
        return None;
    }
    Some(estimate_platform_fee(
        &self.app_context.fee_estimator(),
        num_inputs,
        num_outputs,
    ))
}
```

<sub>source: ['codex']</sub>

src/wallet_backend/snapshot.rs
```diff
@@ -851,6 +878,44 @@ mod tests {
...
+            vec![OutputDetail {
+                index: 0,
+                role: OutputRole::Sent,
+                address: Some(destination),
+                value: 9_700,
+            }],
+            -10_000,
+        );
+
+        assert_eq!(map_transaction_record(&record).fee, Some(300));
```
thepastaclaw, line 916: **🟡 Suggestion: Cover the fee derivation's fail-closed behavior**

The new test covers only a single-input happy path. The helper's user-visible correctness also depends on preserving an existing upstream fee and returning `None` for incomplete or misindexed input metadata and when outputs exceed inputs. Add regression cases for those branches so a future simplification cannot display a fabricated transaction fee.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The outgoing transaction-history fee derivation is conservative and uses checked arithmetic, but its fail-closed behavior lacks regression coverage beyond the happy path. The Platform-to-Platform preview counts populated form rows instead of the distinct positive-credit outputs submitted to the backend, so valid duplicate or zero-credit rows produce an incorrect fee estimate.

### Review provenance

<details>

```

In `src/ui/wallets/send_screen.rs`:
- [BLOCKING] src/ui/wallets/send_screen.rs:2823-2835: Count the Platform outputs actually submitted
  This branch prices every row with a non-empty address, but `send_advanced_platform_to_platform` skips zero-credit rows and inserts outputs into a `BTreeMap`, which coalesces duplicate destinations. Two rows for the same Platform address therefore submit one transition output but are priced as two, and a populated zero-credit row is priced despite being omitted. Because the newly generalized fee calculation charges per output, derive the count from the same distinct, positive-credit destinations used by the send path.

In `src/wallet_backend/snapshot.rs`:
- [SUGGESTION] src/wallet_backend/snapshot.rs:882-916: Cover the fee derivation's fail-closed behavior
  The new test covers only a single-input happy path. The helper's user-visible correctness also depends on preserving an existing upstream fee and returning `None` for incomplete or misindexed input metadata and when outputs exceed inputs. Add regression cases for those branches so a future simplification cannot display a fabricated transaction fee.
```
</details>

PR #905, opened 2026-07-17
fix(shutdown): stop wallet backend on GUI close and clear stale shutdown-channel log

Claudius-Maginificent:
## Why this PR exists

- **Problem**: On GUI close, DET never stops the wallet backend (SPV client / platform-wallet coordinators) before tearing down — it's simply abandoned at process exit. Separately, a stale oneshot-channel poll logs a misleading warning on every clean shutdown.
- **What breaks without it**: Launch DET, pick "Just Explore", let SPV start syncing, close the window via a real title-bar/WM close. Every single clean close logs `WARN dash_evo_tool::app: Shutdown channel closed unexpectedly (possible panic)` even though nothing panicked — confusing when triaging a real shutdown issue. More substantively, the wallet backend's SPV client/coordinators are never explicitly asked to shut down through the app's own shutdown path, relying instead on whatever happens at process death rather than a deterministic teardown.
- **Blocking relationship**: none.

## What was done

- `src/app.rs`: GUI close now awaits `WalletBackend::shutdown()` for every initialized network context (including one that finishes initializing mid-shutdown via an in-flight network switch) before the viewport actually closes, bounded by the existing shutdown timeout. The `shutdown_receiver` oneshot is consumed exactly once (cleared once resolved) so a later frame can't re-poll an already-drained channel and log a false "possible panic" warning.
- `src/backend_task/mod.rs`: closes a related race where an in-flight network-switch task could create/publish a new backend context after shutdown had already begun.
- A `shutdown_started` latch is set on the very first shutdown attempt (graceful or force-close) and never cleared, so `on_exit()`'s force-close guard blocks fallback re-entry on every branch — clean, panicked, or deadline-exceeded — not only after a clean finish. This prevents a second close signal from racing wallet-backend teardown against a still-running first attempt.
- `forget_all_secrets()` is now also called again after the bounded wallet-shutdown wait resolves, including the timeout path, so a backend task that re-caches a secret mid-wait is still cleared before exit.
- `SwitchNetwork` now registers its wallet-backend context before its slow network call resolves, racing that call against the shutdown cancellation token — an in-flight network switch is discoverable to the shutdown teardown pass instead of being invisible to it.
- Incidental cleanup: removed the now-dead blocking `TaskManager::shutdown()`, pruned completed blocking-task tracking entries instead of letting them grow unbounded for the app session's lifetime, and centralized the shutdown budget constant instead of duplicating its arithmetic.
- **Out of scope**: the upstream `platform-wallet` coordinator-thread join race (discarded `JoinHandle`s, `quiesce()` not waiting for actual thread exit) is tracked separately at dashpay/platform#3954 — this PR makes DET correctly *ask* the wallet backend to shut down with a bounded wait, but can't fully close that upstream race until the fix there lands.

## Testing

- `cargo fmt --all -- --check` — clean
- `cargo test app:: --all-features` — 35 passed, 0 failed
- `cargo test backend_task:: --all-features` — 374 passed, 0 failed
- `cargo test utils::tasks:: --all-features` — 5 passed, 0 failed
- Manual GUI verification against the exact commit this fix is based on (`3b32bed2`, the commit `v1.0.0-weekly.20260717` was built from): closed via a real ICCCM `WM_DELETE_WINDOW` request (not a forced X11 destroy) with a fresh isolated data directory. Confirmed via logs that both the crash-safety and the log-line regression are resolved — no panics, and the shutdown sequence now shows SPV/wallet-backend components tearing down in order (`SyncCoordinator` → per-task shutdown → peer network manager → storage flush) before the viewport closes.

## Breaking changes

None.

## Checklist

- [x] Builds clean
- [x] Clippy clean (`-D warnings`)
- [x] Formatted (`cargo fmt --all`)
- [x] Manually verified against a real window-close event
- [x] Automated regression tests added for the shutdown-race fixes (`second_exit_after_async_attempt_does_not_repeat_wallet_teardown`, `wallet_shutdown_clears_secrets_again_after_teardown_timeout`, `switch_network_registers_wired_backend_before_cancellation_teardown`)

## Attribution

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved coordinated app shutdown with a hard deadline, including safe handling of late/deduplicated network context registration.
  * Ensured SPV startup wiring occurs before cancellation teardown, with correct wallet secret clearing when shutdown is in progress or times out.
  * Updated viewport close behavior to reflect shutdown completion, with a timed fallback when needed.
* **Reliability**
  * Hardened task shutdown using an admission barrier that blocks new work and reports degraded teardown on timeout.
  * Added shutdown-aware MCP dispatch behavior (in-flight completes; new requests are rejected) with expanded test coverage.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -81,6 +81,22 @@ pub const MIGRATION_IDENTITIES_ACK_ACTION_ID: &str = "migration:ack:unreadable_i
 pub const MIGRATION_UNREADABLE_ACK_ACTION_ID: &str =
     "migration:ack:unreadable_identities_and_votes";
 
+const WALLET_BACKEND_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(30);
+const SHUTDOWN_DEADLINE_MARGIN: Duration = Duration::from_secs(5);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ShutdownOutcome {
+    Complete,
+    TaskManagerFailed,
+    WalletBackendTimedOut,
+}
+
+fn shutdown_hard_deadline() -> Duration {
+    crate::utils::tasks::SHUTDOWN_TIMEOUT
+        + WALLET_BACKEND_SHUTDOWN_TIMEOUT
+        + SHUTDOWN_DEADLINE_MARGIN
+}
+
 fn migration_allows_scheduled_vote_sweep(state: &MigrationState) -> bool {
     matches!(
         state,
@@ -684,10 +700,12 @@ pub struct AppState {
     migration: MigrationReconciler,
     /// Async shutdown receiver. `Some` until a graceful shutdown reaches a
     /// terminal state; the viewport is closed once the receiver resolves.
-    shutdown_receiver: Option<tokio::sync::oneshot::Receiver<()>>,
+    shutdown_receiver: Option<tokio::sync::oneshot::Receiver<ShutdownOutcome>>,
     /// Timestamp when the async shutdown was initiated, used as a hard deadline
     /// to force-close the viewport if the shutdown task stalls.
     shutdown_started: Option<std::time::Instant>,
+    /// True once every managed shutdown stage reached a terminal outcome.
+    shutdown_finished: bool,
     /// Platform-level accessibility (AccessKit) activation reconciler.
     accessibility: AccessibilityActivator,
     /// Shared MCP context -- follows network switches via `ArcSwap`.
@@ -1347,6 +1365,7 @@ impl AppState {
             migration: MigrationReconciler::new(),
             shutdown_receiver: None,
             shutdown_started: None,
+            shutdown_finished: false,
             accessibility: AccessibilityActivator::new(accessibility_enforced),
             #[cfg(feature = "mcp")]
             mcp_app_context,
@@ -1828,25 +1847,77 @@ impl AppState {
         }
     }
 
-    fn collect_created_context(contexts: &mut Vec<Arc<AppContext>>, task_result: TaskResult) {
-        if let TaskResult::Success { result, .. } = task_result
-            && let BackendTaskSuccessResult::NetworkContextCreated { context, .. } = *result
+    fn push_unique_context(contexts: &mut Vec<Arc<AppContext>>, context: Arc<AppContext>) {
+        if !contexts
+            .iter()
+            .any(|existing| Arc::ptr_eq(existing, &context))
         {
             contexts.push(context);
         }
     }
 
-    fn start_async_shutdown(&mut self) -> tokio::sync::oneshot::Receiver<()> {
-        self.subtasks.cancellation_token.cancel();
-        let mut contexts = self.network_contexts.values().cloned().collect::<Vec<_>>();
-        let subtasks = Arc::clone(&self.subtasks);
+    fn collect_created_context(contexts: &mut Vec<Arc<AppContext>>, task_result: TaskResult) {
+        if let TaskResult::Success { result, .. } = task_result
+            && let BackendTaskSuccessResult::NetworkContextCreated { context, .. } = *result
+        {
+            Self::push_unique_context(contexts, context);
+        }
+    }
+
+    async fn shutdown_wallet_backends(contexts: Vec<Arc<AppContext>>) -> ShutdownOutcome {
+        let mut wallet_backends = Vec::new();
+        for context in contexts {
+            if let Ok(backend) = context.wallet_backend()
+                && !wallet_backends
+                    .iter()
+                    .any(|existing| Arc::ptr_eq(existing, &backend))
+            {
+                wallet_backends.push(backend);
+            }
+        }
+
+        for backend in &wallet_backends {
+            backend.forget_all_secrets();
+        }
+
+        let shutdowns = futures::future::join_all(
+            wallet_backends
+                .iter()
+                .map(|wallet_backend| wallet_backend.shutdown()),
+        );
+        if tokio::time::timeout(WALLET_BACKEND_SHUTDOWN_TIMEOUT, shutdowns)
+            .await
+            .is_err()
+        {
+            tracing::warn!(
+                timeout_secs = WALLET_BACKEND_SHUTDOWN_TIMEOUT.as_secs(),
+                "Wallet backend shutdown timed out; closing with degraded teardown"
+            );
+            ShutdownOutcome::WalletBackendTimedOut
+        } else {
+            ShutdownOutcome::Complete
+        }
+    }
+
+    fn initial_shutdown_contexts(&self) -> Vec<Arc<AppContext>> {
+        let mut contexts = Vec::new();
+        for context in self.network_contexts.values() {
+            Self::push_unique_context(&mut contexts, Arc::clone(context));
+        }
+        contexts
+    }
+
+    fn start_async_shutdown(&mut self) -> tokio::sync::oneshot::Receiver<ShutdownOutcome> {
+        let mut contexts = self.initial_shutdown_contexts();
+        let mut task_shutdown = self.subtasks.shutdown_async();
         let (_empty_sender, empty_receiver) = tokiompsc::channel(1);
         let mut task_result_receiver =
             std::mem::replace(&mut self.task_result_receiver, empty_receiver);
+        #[cfg(feature = "mcp")]
+        let mcp_app_context = self.mcp_app_context.clone();
         let (tx, rx) = tokio::sync::oneshot::channel();
 
         tokio::spawn(async move {
-            let mut task_shutdown = subtasks.shutdown_async();
             let task_shutdown_completed = loop {
                 tokio::select! {
                     result = &mut task_shutdown => break result.is_ok(),
@@ -1863,23 +1934,69 @@ impl AppState {
                 Self::collect_created_context(&mut contexts, task_result);
             }
 
-            let wallet_backends = contexts
-                .iter()
-                .filter_map(|context| context.wallet_backend().ok())
-                .collect::<Vec<_>>();
-            futures::future::join_all(
-                wallet_backends
-                    .iter()
-                    .map(|wallet_backend| wallet_backend.shutdown()),
-            )
-            .await;
-
-            if task_shutdown_completed {
-                let _ = tx.send(());
+            #[cfg(feature = "mcp")]
+            if let Some(mcp_app_context) = mcp_app_context {
+                Self::push_unique_context(&mut contexts, mcp_app_context.load_full());
             }
+
+            let wallet_outcome = Self::shutdown_wallet_backends(contexts).await;
+            let outcome = if task_shutdown_completed {
+                wallet_outcome
+            } else {
+                ShutdownOutcome::TaskManagerFailed
+            };
+            let _ = tx.send(outcome);
         });
 
         rx
+    }
+
+    fn run_blocking_shutdown_fallback(&mut self) {
+        const POLL_INTERVAL: Duration = Duration::from_millis(10);
+
+        let mut contexts = self.initial_shutdown_contexts();
+        let mut task_shutdown = self.subtasks.shutdown_async();
+        let task_shutdown_completed = loop {
+            while let Ok(task_result) = self.task_result_receiver.try_recv() {
+                Self::collect_created_context(&mut contexts, task_result);
+            }
+
+            match task_shutdown.try_recv() {
+                Ok(()) => break true,
+                Err(tokio::sync::oneshot::error::TryRecvError::Closed) => break false,
+                Err(tokio::sync::oneshot::error::TryRecvError::Empty) => {
+                    std::thread::sleep(POLL_INTERVAL);
+                }
+            }
+        };
+
+        while let Ok(task_result) = self.task_result_receiver.try_recv() {
+            Self::collect_created_context(&mut contexts, task_result);
+        }
+
+        #[cfg(feature = "mcp")]
+        if let Some(mcp_app_context) = &self.mcp_app_context {
+            Self::push_unique_context(&mut contexts, mcp_app_context.load_full());
+        }
+
+        let (tx, rx) = std::sync::mpsc::sync_channel(1);
+        tokio::spawn(async move {
+            let outcome = Self::shutdown_wallet_backends(contexts).await;
+            let _ = tx.send(outcome);
+        });
+
+        match rx.recv_timeout(WALLET_BACKEND_SHUTDOWN_TIMEOUT + SHUTDOWN_DEADLINE_MARGIN) {
+            Ok(ShutdownOutcome::Complete) if task_shutdown_completed => {}
+            Ok(outcome) => tracing::warn!(
+                ?outcome,
+                task_shutdown_completed,
+                "Blocking shutdown fallback completed with degraded teardown"
+            ),
+            Err(error) => tracing::warn!(
+                ?error,
+                "Blocking wallet backend shutdown did not report a terminal outcome"
+            ),
+        }
     }
 }
 
@@ -1895,8 +2012,19 @@ impl App for AppState {
             // Shutdown already in progress — check if it's done.
             let should_close = match self.shutdown_receiver.as_mut() {
                 Some(rx) => match rx.try_recv() {
-                    Ok(()) => {
-                        tracing::debug!("Async shutdown finished, closing viewport");
+                    Ok(outcome) => {
+                        self.shutdown_finished = true;
+                        match outcome {
+                            ShutdownOutcome::Complete => {
+                                tracing::debug!("Async shutdown finished, closing viewport");
+                            }
+                            ShutdownOutcome::TaskManagerFailed => tracing::warn!(
+                                "Task shutdown failed; closing with degraded teardown"
+                            ),
+                            ShutdownOutcome::WalletBackendTimedOut => tracing::warn!(
+                                "Wallet backend shutdown exceeded its deadline; closing with degraded teardown"
+                            ),
+                        }
                         true
                     }
                     Err(tokio::sync::oneshot::error::TryRecvError::Closed) => {
@@ -1907,8 +2035,7 @@ impl App for AppState {
                     Err(tokio::sync::oneshot::error::TryRecvError::Empty) => {
                         // Still waiting — check hard deadline to prevent infinite loop.
                         if let Some(started) = self.shutdown_started {
-                            let grace = crate::utils::tasks::SHUTDOWN_TIMEOUT
-                                + std::time::Duration::from_secs(5);
+                            let grace = shutdown_hard_deadline();
                             if started.elapsed() > grace {
                                 tracing::warn!(
                                     "Shutdown hard deadline exceeded, force-closing viewport"
@@ -2489,16 +2616,12 @@ impl App for AppState {
         // observers (TouchBar, etc.) while views are still alive.
         crate::platform::order_out_all_windows();
 
-        // The start marker persists after the completion receiver is consumed.
-        // Only force-closes that bypass update() need the blocking fallback.
-        if self.shutdown_started.is_some() {
-            tracing::debug!("on_exit: async shutdown was initiated, skipping blocking fallback");
+        if self.shutdown_finished {
+            tracing::debug!("on_exit: async shutdown finished, skipping blocking fallback");
             return;
         }
         tracing::debug!("on_exit: fallback blocking shutdown");
-        if let Err(e) = self.subtasks.shutdown() {
-            tracing::error!("Error during task shutdown: {}", e);
-        }
+        self.run_blocking_shutdown_fallback();
         tracing::debug!("App shutdown complete");
     }
 }
@@ -2893,5 +3016,58 @@ mod spv_overlay_tests {
                 );
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod shutdown_tests {
+    use super::*;
+
+    #[test]
+    fn collect_created_context_only_keeps_network_context_results() {
+        let temp_dir = tempfile::tempdir().expect("temp dir");
+        let context = crate::context::test_support::test_app_context(temp_dir.path());
+        let mut contexts = Vec::new();
+
+        AppState::collect_created_context(
+            &mut contexts,
+            TaskResult::unattributed_success(BackendTaskSuccessResult::NetworkContextCreated {
+                network: Network::Testnet,
+                context: Arc::clone(&context),
+                spv_started: false,
+            }),
+        );
+        assert_eq!(contexts.len(), 1);
+        assert!(Arc::ptr_eq(&contexts[0], &context));
+
+        AppState::collect_created_context(
+            &mut contexts,
+            TaskResult::unattributed_success(BackendTaskSuccessResult::NetworkContextCreated {
+                network: Network::Testnet,
+                context: Arc::clone(&context),
+                spv_started: false,
+            }),
+        );
+        assert_eq!(contexts.len(), 1);
+
+        AppState::collect_created_context(
+            &mut contexts,
+            TaskResult::unattributed_success(BackendTaskSuccessResult::None),
+        );
+        assert_eq!(contexts.len(), 1);
+
+        AppState::collect_created_context(
+            &mut contexts,
+            TaskResult::unattributed_error(TaskError::NoIdentitiesFound),
+        );
+        assert_eq!(contexts.len(), 1);
+    }
+
+    #[test]
+    fn viewport_deadline_covers_task_and_wallet_shutdown_budgets() {
+        assert!(
+            shutdown_hard_deadline()
+                >= crate::utils::tasks::SHUTDOWN_TIMEOUT + WALLET_BACKEND_SHUTDOWN_TIMEOUT
+        );
     }
 }
--- a/src/app.rs
+++ b/src/app.rs
@@ -1590,7 +1590,7 @@ impl AppState {
         reason: BackendInitReason,
         start_spv: bool,
     ) {
-        subtasks.spawn_sync(reason.task_name(), async move {
+        let _ = subtasks.spawn_sync(reason.task_name(), async move {
             if start_spv {
                 let already_running = app_ctx
                     .wallet_backend()
@@ -1613,7 +1613,7 @@ impl AppState {
         config: crate::mcp::McpConfig,
     ) {
         let cancel = subtasks.cancellation_token.clone();
-        subtasks.spawn_sync("mcp-server", async move {
+        let _ = subtasks.spawn_sync("mcp-server", async move {
             if let Err(error) = crate::mcp::start_http_server(app_context, config, cancel).await {
                 tracing::error!(%error, "MCP server failed");
             }
@@ -1637,7 +1637,7 @@ impl AppState {
         let watcher_context = context.clone();
         let app_context = self.current_app_context().clone();
         let handle = tokio::runtime::Handle::current();
-        self.subtasks.spawn_blocking_sync(
+        let _ = self.subtasks.spawn_blocking_sync(
             "backend_task_join_watcher",
             move || {
                 handle.block_on(async move {
@@ -1672,7 +1672,7 @@ impl AppState {
         let app_context = self.current_app_context().clone();
         let handle = tokio::runtime::Handle::current();
 
-        self.subtasks.spawn_blocking_sync(
+        let _ = self.subtasks.spawn_blocking_sync(
             "backend_tasks_join_watcher",
             move || {
                 handle.block_on(async move {
@@ -2125,6 +2125,22 @@ impl AppState {
         Self::shutdown_wallet_backends(contexts).await
     }
 
+    async fn finish_shutdown_after_tasks<F>(
+        task_shutdown_outcome: Option<TaskShutdownOutcome>,
+        wallet_shutdown: F,
+    ) -> ShutdownOutcome
+    where
+        F: std::future::Future<Output = ShutdownOutcome>,
+    {
+        match task_shutdown_outcome {
+            Some(TaskShutdownOutcome::Complete) => wallet_shutdown.await,
+            Some(TaskShutdownOutcome::BackendTasksTimedOut) => {
+                ShutdownOutcome::BackendTasksTimedOut
+            }
+            None => ShutdownOutcome::TaskManagerFailed,
+        }
+    }
+
     fn start_async_shutdown(&mut self) -> tokio::sync::oneshot::Receiver<ShutdownOutcome> {
         let mut contexts = self.initial_shutdown_contexts();
         let mut task_shutdown = self.subtasks.shutdown_async();
@@ -2148,20 +2164,16 @@ impl AppState {
                 }
             };
 
-            let wallet_outcome = Self::finish_wallet_shutdown(
-                contexts,
-                task_result_receiver,
-                #[cfg(feature = "mcp")]
-                mcp_app_context,
+            let outcome = Self::finish_shutdown_after_tasks(
+                task_shutdown_outcome,
+                Self::finish_wallet_shutdown(
+                    contexts,
+                    task_result_receiver,
+                    #[cfg(feature = "mcp")]
+                    mcp_app_context,
+                ),
             )
             .await;
-            let outcome = match task_shutdown_outcome {
-                Some(TaskShutdownOutcome::Complete) => wallet_outcome,
-                Some(TaskShutdownOutcome::BackendTasksTimedOut) => {
-                    ShutdownOutcome::BackendTasksTimedOut
-                }
-                None => ShutdownOutcome::TaskManagerFailed,
-            };
             let _ = tx.send(outcome);
         });
 
@@ -2194,11 +2206,14 @@ impl AppState {
 
         let (tx, rx) = std::sync::mpsc::sync_channel(1);
         tokio::spawn(async move {
-            let outcome = Self::finish_wallet_shutdown(
-                contexts,
-                task_result_receiver,
-                #[cfg(feature = "mcp")]
-                mcp_app_context,
+            let outcome = Self::finish_shutdown_after_tasks(
+                task_shutdown_outcome,
+                Self::finish_wallet_shutdown(
+                    contexts,
+                    task_result_receiver,
+                    #[cfg(feature = "mcp")]
+                    mcp_app_context,
+                ),
             )
             .await;
             let _ = tx.send(outcome);
@@ -2849,7 +2864,7 @@ impl App for AppState {
                     // only the winner spawns the async teardown. No banner is
                     // needed for a user-initiated stop.
                     if app_ctx.connection_status().begin_spv_stop() {
-                        self.subtasks.spawn_sync("spv_manual_stop", async move {
+                        let _ = self.subtasks.spawn_sync("spv_manual_stop", async move {
                             app_ctx.stop_spv().await;
                         });
                     }
@@ -3443,13 +3458,18 @@ mod shutdown_tests {
 
         let (started_tx, started_rx) = std::sync::mpsc::sync_channel(1);
         let (release_tx, release_rx) = std::sync::mpsc::sync_channel(1);
-        app.subtasks.spawn_blocking_sync(
-            "slow-shutdown-regression-task",
-            move || {
-                started_tx.send(()).expect("report slow task start");
-                release_rx.recv().expect("wait for slow task release");
-            },
-            |_| async {},
+        assert!(
+            app.subtasks
+                .spawn_blocking_sync(
+                    "slow-shutdown-regression-task",
+                    move || {
+                        started_tx.send(()).expect("report slow task start");
+                        release_rx.recv().expect("wait for slow task release");
+                    },
+                    |_| async {},
+                )
+                .is_ok(),
+            "slow shutdown task is accepted"
         );
         started_rx.recv().expect("slow task started");
 
@@ -3527,6 +3547,57 @@ mod shutdown_tests {
             !backend.secret_cached.load(Ordering::Acquire),
             "the final clear still runs when backend shutdown exceeds its timeout"
         );
+    }
+
+    #[tokio::test]
+    async fn backend_task_timeout_skips_wallet_teardown() {
+        let teardown_calls = Arc::new(AtomicUsize::new(0));
+        let calls = Arc::clone(&teardown_calls);
+
+        let outcome = AppState::finish_shutdown_after_tasks(
+            Some(TaskShutdownOutcome::BackendTasksTimedOut),
+            async move {
+                calls.fetch_add(1, Ordering::Relaxed);
+                ShutdownOutcome::Complete
+            },
+        )
+        .await;
+
+        assert_eq!(outcome, ShutdownOutcome::BackendTasksTimedOut);
+        assert_eq!(teardown_calls.load(Ordering::Relaxed), 0);
+    }
+
+    #[tokio::test]
+    async fn task_manager_failure_skips_wallet_teardown() {
+        let teardown_calls = Arc::new(AtomicUsize::new(0));
+        let calls = Arc::clone(&teardown_calls);
+
+        let outcome = AppState::finish_shutdown_after_tasks(None, async move {
+            calls.fetch_add(1, Ordering::Relaxed);
+            ShutdownOutcome::Complete
+        })
+        .await;
+
+        assert_eq!(outcome, ShutdownOutcome::TaskManagerFailed);
+        assert_eq!(teardown_calls.load(Ordering::Relaxed), 0);
+    }
+
+    #[tokio::test]
+    async fn complete_task_shutdown_runs_wallet_teardown_once() {
+        let teardown_calls = Arc::new(AtomicUsize::new(0));
+        let calls = Arc::clone(&teardown_calls);
+
+        let outcome = AppState::finish_shutdown_after_tasks(
+            Some(TaskShutdownOutcome::Complete),
+            async move {
+                calls.fetch_add(1, Ordering::Relaxed);
+                ShutdownOutcome::Complete
+            },
+        )
+        .await;
+
+        assert_eq!(outcome, ShutdownOutcome::Complete);
+        assert_eq!(teardown_calls.load(Ordering::Relaxed), 1);
     }
 
     #[test]
--- a/src/app.rs
+++ b/src/app.rs
@@ -682,8 +682,8 @@ pub struct AppState {
     spv_block: SpvBlockReconciler,
     /// Data-migration banner + cold-start `FinishUnwire` dispatch reconciler.
     migration: MigrationReconciler,
-    /// Async shutdown receiver. `Some` while a graceful shutdown is in progress;
-    /// the viewport is closed once the receiver resolves.
+    /// Async shutdown receiver. `Some` until a graceful shutdown reaches a
+    /// terminal state; the viewport is closed once the receiver resolves.
     shutdown_receiver: Option<tokio::sync::oneshot::Receiver<()>>,
     /// Timestamp when the async shutdown was initiated, used as a hard deadline
     /// to force-close the viewport if the shutdown task stalls.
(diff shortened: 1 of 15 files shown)


thepastaclaw:
🕓 Ready for review — 38 ahead in queue (commit 7603be6c)
_Queue position: 39/39_

src/app.rs
```diff
@@ -1827,6 +1827,60 @@ impl AppState {
...
+        }
+    }
+
+    fn start_async_shutdown(&mut self) -> tokio::sync::oneshot::Receiver<()> {
+        self.subtasks.cancellation_token.cancel();
+        let mut contexts = self.network_contexts.values().cloned().collect::<Vec<_>>();
+        let subtasks = Arc::clone(&self.subtasks);
+        let (_empty_sender, empty_receiver) = tokiompsc::channel(1);
+        let mut task_result_receiver =
+            std::mem::replace(&mut self.task_result_receiver, empty_receiver);
```
thepastaclaw: **🔴 Blocking: Include the active MCP context in wallet-backend shutdown**

With the `mcp` feature enabled, the `network_switch` tool publishes its newly created context directly to `mcp_app_context` through `DashMcpService::swap_context`. That context is neither inserted into `network_contexts` nor returned through the GUI task-result receiver, so an MCP-initiated switch completed before GUI exit leaves the active context absent from this shutdown snapshot. The tool shuts down the outgoing backend before swapping, which means the omitted MCP context is precisely the backend still running at exit. Clone the shared ArcSwap before spawning shutdown, load its final value after the tracked MCP server has drained, and add it to the context set with pointer-based deduplication.

<sub>source: ['codex']</sub>

src/app.rs
```diff
@@ -1827,6 +1827,60 @@ impl AppState {
...
+                tokio::select! {
+                    result = &mut task_shutdown => break result.is_ok(),
+                    task_result = task_result_receiver.recv() => {
+                        let Some(task_result) = task_result else {
+                            break task_shutdown.await.is_ok();
+                        };
+                        Self::collect_created_context(&mut contexts, task_result);
+                    }
+                }
+            };
```
thepastaclaw, line 2185: **🔴 Blocking: Task shutdown is not a registration barrier**

Waiting for `subtasks.shutdown_async()` does not guarantee that every previously submitted task has been joined or aborted. `TaskManager::spawn_sync` records the name synchronously but launches an untracked wrapper that acquires the JoinSet mutex later; if shutdown acquires and drains that mutex first, the wrapper can register its future only after `shutdown_async()` resolves. A delayed `spawn_backend_init` then runs without checking the cancellation token and can create or start a wallet backend after this code has collected and shut down the known backends. Make task submission synchronously register or reserve the task before returning, or introduce a closed state that atomically rejects registration once shutdown begins.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The new shutdown flow correctly consumes the terminal receiver and handles GUI network-switch results, but it does not yet guarantee that every wallet backend is stopped. Embedded MCP can own an active context outside the GUI registry, and TaskManager can accept a delayed task registration after shutdown has already drained its JoinSet. Both gaps can recreate the nondeterministic teardown this PR is intended to fix.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 2 blocking

<details>

```

In `src/app.rs`:
- [BLOCKING] src/app.rs:1840-1845: Include the active MCP context in wallet-backend shutdown
  With the `mcp` feature enabled, the `network_switch` tool publishes its newly created context directly to `mcp_app_context` through `DashMcpService::swap_context`. That context is neither inserted into `network_contexts` nor returned through the GUI task-result receiver, so an MCP-initiated switch completed before GUI exit leaves the active context absent from this shutdown snapshot. The tool shuts down the outgoing backend before swapping, which means the omitted MCP context is precisely the backend still running at exit. Clone the shared ArcSwap before spawning shutdown, load its final value after the tracked MCP server has drained, and add it to the context set with pointer-based deduplication.
- [BLOCKING] src/app.rs:1848-1860: Task shutdown is not a registration barrier
  Waiting for `subtasks.shutdown_async()` does not guarantee that every previously submitted task has been joined or aborted. `TaskManager::spawn_sync` records the name synchronously but launches an untracked wrapper that acquires the JoinSet mutex later; if shutdown acquires and drains that mutex first, the wrapper can register its future only after `shutdown_async()` resolves. A delayed `spawn_backend_init` then runs without checking the cancellation token and can create or start a wallet backend after this code has collected and shut down the known backends. Make task submission synchronously register or reserve the task before returning, or introduce a closed state that atomically rejects registration once shutdown begins.
```
</details>

src/app.rs
```diff
@@ -2431,11 +2489,9 @@ impl App for AppState {
...
-        // in update(). Skip the blocking fallback to avoid double-shutdown.
-        // The blocking path only runs when the window was force-closed without
-        // going through update() (e.g., OS-level kill, alt-F4 on some platforms).
-        if self.shutdown_receiver.is_some() {
+        // The start marker persists after the completion receiver is consumed.
+        // Only force-closes that bypass update() need the blocking fallback.
+        if self.shutdown_started.is_some() {
             tracing::debug!("on_exit: async shutdown was initiated, skipping blocking fallback");
             return;
         }
```
lklimek: 🔴 **Blocking: `on_exit`'s force-close fallback — this code's own documented path — still never calls `WalletBackend::shutdown()`**

This new comment states explicitly that *"Only force-closes that bypass `update()` need the blocking fallback"* — i.e. the authors acknowledge the fallback at the lines just below is a real, reachable path (the pre-existing comment left in place a few lines up names "OS-level kill, alt-F4 on some platforms").

But tracing `TaskManager::shutdown()` (`src/utils/tasks.rs:76-109`): it calls `shutdown_inner` exactly like the async path, and neither it nor anything else on this fallback branch ever iterates `network_contexts` or calls `wallet_backend.shutdown()`. That call exists **only** in the new `start_async_shutdown` — the `update()`-mediated path.

So for the exact scenario this PR's own comment names as still reachable, the wallet backend's SPV client and coordinators are torn down exactly as they were before this fix: **not at all, just abandoned at process exit** — the precise problem the PR body describes as what it fixes. The 'Out of scope' section names only the upstream `platform-wallet` coordinator-thread join race (dashpay/platform#3954); it does not carve out this fallback, so this is a real gap between stated scope (*"stop wallet backend on GUI close"*, unqualified) and actual coverage (graceful close only).

**Recommendation:** either extend the fallback to also drain `network_contexts`' wallet backends (bounded — it's already a blocking context), or scope the title/description explicitly to "the graceful close path only" and file a tracked follow-up for the force-close path.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
```diff
@@ -1837,37 +1891,41 @@ impl App for AppState {
...
+                    }
+                    Err(tokio::sync::oneshot::error::TryRecvError::Empty) => {
+                        // Still waiting — check hard deadline to prevent infinite loop.
+                        if let Some(started) = self.shutdown_started {
+                            let grace = crate::utils::tasks::SHUTDOWN_TIMEOUT
+                                + std::time::Duration::from_secs(5);
+                            if started.elapsed() > grace {
+                                tracing::warn!(
+                                    "Shutdown hard deadline exceeded, force-closing viewport"
+                                );
```
lklimek, line 2296: 🟡 **The viewport's hard deadline can expire before wallet-backend shutdown finishes**

The new sequence waits for `TaskManager` shutdown *before* it starts `WalletBackend::shutdown()`, but this hard deadline is still only `SHUTDOWN_TIMEOUT` (10s) + 5s — a budget that doesn't cover the sequence's own worst case:

- `TaskManager` may consume the full 10s on its own.
- `WalletBackend::shutdown()` then calls the pinned `platform-wallet` `SpvRuntime::stop()`, whose run-loop join alone can wait **15s** before aborting — and only *then* drains the coordinators.

The viewport can therefore force-close while the awaited shutdown future is still running. `on_exit()` sees `shutdown_started` and skips its fallback, and `main` drops the Tokio runtime once `run_native` returns — so the remaining teardown future can simply be aborted. The common path awaits the backend; the slow path does not, which again misses the PR's promise that initialized backends stop before the viewport closes.

**Recommendation:** give each shutdown stage an explicit budget and set the deadline to cover the complete worst-case sequence (task shutdown + SPV stop + coordinator drain), or wrap the whole sequence in one reported timeout whose force-close is an explicit *degraded* outcome. A deterministic test where task shutdown consumes its full budget while backend shutdown stays pending would pin this down — asserting the viewport is not declared cleanly stopped early.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
```diff
@@ -1827,6 +1827,60 @@ impl AppState {
...
+            let wallet_backends = contexts
+                .iter()
+                .filter_map(|context| context.wallet_backend().ok())
+                .collect::<Vec<_>>();
+            futures::future::join_all(
+                wallet_backends
+                    .iter()
+                    .map(|wallet_backend| wallet_backend.shutdown()),
+            )
+            .await;
```
lklimek: 🔵 **Graceful shutdown doesn't eagerly clear remembered wallet secrets**

After task shutdown, this code awaits `WalletBackend::shutdown()`, which stops SPV and the `platform-wallet` coordinators but does **not** call `WalletBackend::forget_all_secrets()`. A plaintext seed or key retained under `RememberPolicy::UntilAppClose` therefore stays in `SecretAccess`'s `Zeroizing` session map for the whole (potentially long — see the deadline comment above) shutdown, and is cleared only when the final `AppState`/`SecretAccess` owners drop.

To be clear about severity: no new code bypasses `wallet_backend/secret_seam.rs`, and normal Rust drop still provides a final zeroization backstop — so this is a hardening gap, not a leak. But the *explicit eager teardown clear* that `WalletBackend::forget_all_secrets()` and the JIT-secret design document is missing, and shutdown is exactly when it's meant to run.

**Recommendation:** once managed wallet-using tasks are conclusively quiesced, call `forget_all_secrets()` on every collected backend before awaiting its async shutdown, and apply the same eager clear to any blocking/fallback exit path. Testable with a remembered session entry: assert it's gone as soon as shutdown enters its backend-teardown phase.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
```diff
@@ -1827,6 +1827,60 @@ impl AppState {
...
         }
     }
+
+    fn collect_created_context(contexts: &mut Vec<Arc<AppContext>>, task_result: TaskResult) {
+        if let TaskResult::Success { result, .. } = task_result
+            && let BackendTaskSuccessResult::NetworkContextCreated { context, .. } = *result
+        {
+            contexts.push(context);
+        }
+    }
```
lklimek: 🔵 **This new pure helper is trivially testable, so the checklist's blanket "tautological" opt-out doesn't cover it**

The PR checklist opts out of tests entirely: *"exercising the receiver-lifecycle behavior requires full `AppState`/eframe integration; a Tokio-oneshot-only unit test would be tautological."* That justification is **accurate for the `start_async_shutdown` orchestration** — no argument there.

But `collect_created_context(&mut Vec<Arc<AppContext>>, TaskResult)` is a pure, free-standing filter+push with no `AppState`, eframe, or tokio dependency. Constructing a `TaskResult::Success { result: Box::new(BackendTaskSuccessResult::NetworkContextCreated { .. }) }` vs. a `Success` carrying another variant vs. a `TaskResult::Error { .. }`, then asserting on `contexts.len()`/contents, is a ~5-line non-tautological test that would catch a future refactor breaking the match guard — precisely the kind of silent breakage that would quietly drop a context from the shutdown sweep.

**Recommendation:** add a `#[cfg(test)]` unit test covering the three cases: `Success` + `NetworkContextCreated` (pushes), `Success` + other variant (no-op), `Error` (no-op).

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

lklimek (commented):
## Automated grumpy-review — PR #905

Reviewed at `fdeeacd2` (unchanged since review; no findings dropped as already-fixed).

### The headline: this PR does not yet fully deliver its stated purpose

The PR body states the problem without qualification — *"On GUI close, DET never stops the wallet backend ... it's simply abandoned at process exit"* — and claims that *"GUI close now awaits `WalletBackend::shutdown()` for every initialized network context."* Two independent reviewers traced the actual call chains rather than trusting the description, and **that claim does not hold in three verified cases**:

1. **MCP-created contexts are structurally invisible to the shutdown sweep** — not a race, but true under *every* timing once an MCP client has called `network_switch`. This independently corroborates @thepastaclaw's already-open thread and is left there rather than duplicated here.
2. **The `TaskManager` registration barrier is not atomic**, so a wallet-touching task can wire and start a backend *after* the sweep has already collected and shut down every known backend — the exact class of race the PR claims to have closed, via call sites the `SwitchNetwork` guard doesn't cover. Also left on @thepastaclaw's existing thread.
3. **`on_exit`'s force-close fallback still never calls `WalletBackend::shutdown()`** — posted inline below. The PR's own new comment names this path as reachable, and the 'Out of scope' section does not carve it out.

Because the 'Out of scope' section names only the upstream `platform-wallet` coordinator-thread join race (dashpay/platform#3954), the remaining gaps read as claimed-closed when they are not. Either close them, or explicitly narrow the title/description and file tracked follow-ups.

### What's genuinely good here

- The **common GUI-close path is now correct**: existing contexts are snapshotted up front, and an in-flight GUI-driven `SwitchNetwork` context is correctly observed via the shared channel before the shutdown task settles.
- The **stale-log fix is sound** — `shutdown_receiver` is consumed via `.take()` exactly once, so a later frame hits the new `None => true` arm instead of re-triggering the false "possible panic" warning. A real sender drop/panic still reaches the `Closed` warning, so the cleanup doesn't hide genuine failures.
- `WalletBackend::shutdown()` is documented idempotent, so the deliberate double-invocation between the `SwitchNetwork` check and the aggregate sweep is **safe by design** — that overlap is not a bug.
- The sweep genuinely *awaits* each future rather than merely signalling stop.
- Call-tree walk found **no stale callers** of any changed/new signature.
- `cargo clippy --all-features --all-targets -- -D warnings` (clean) and `cargo test --lib app::` (30/30) were **independently reproduced**, not taken from the PR's self-report.
- The upstream dashpay/platform#3954 limitation is accurately disclosed.

### Findings dropped as duplicates

Two blocking findings (the MCP-context blind spot and the `TaskManager` registration race) duplicate open @thepastaclaw threads and are **not** re-posted as new comments — each has a reply on its existing thread adding only what's genuinely new (root-cause mechanism, a second uncovered call site, and the secret-session lifetime angle).

**Suggested action: request changes** before merge — the fixes are narrow and well-scoped; no redesign needed.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
lklimek: **Independently corroborated by a second reviewer — this holds up, and it is worse than a race.**

Traced the full call chain rather than taking this comment at face value, and it confirms exactly as written. Adding three details not yet on this thread:

**Root-cause mechanism.** The reason the new context never reaches the GUI receiver is `mcp/dispatch.rs::dispatch_task`, which builds a brand-new *throwaway* `mpsc::channel::<TaskResult>(32)` (dispatch.rs:27) local to that single call and returns the result directly as a function value. So the resulting `NetworkContextCreated { context, .. }` never touches `AppState.task_result_sender`/`task_result_receiver` **at all** — it is not a matter of timing or ordering on the shared channel.

**No pull-back path exists.** Confirmed `current_app_context()` (app.rs:1393-1402) reads exclusively from `self.network_contexts.get(&self.chosen_network)`, never from `mcp_app_context`; and the only code that moves state between them is the **one-way** `finalize_network_switch` (app.rs:1629-1634), GUI → MCP. There is no code path that pulls an MCP-driven switch back into `network_contexts`.

**Consequence is unconditional, not a race.** With the `mcp` feature enabled (CLAUDE.md documents this as an embedded, first-class GUI feature, not experimental) and after at least one `network_switch` call, the resulting backend is never a candidate for the `wallet_backends.filter_map(...)` sweep under **any** timing.

**One impact not yet mentioned here:** beyond the SPV run loop and coordinators surviving, this also leaves **decrypted per-wallet secret session state alive in memory** past the point the user believes the app has closed — secrets JIT-loaded into that context's session cache are never dropped, since nothing ever asks that backend to tear down.

Agreed on the fix as described (clone the ArcSwap, load its final value after the tracked MCP server has drained, pointer-based dedupe). The alternative worth weighing: route MCP-driven `SwitchNetwork` dispatch through the same shared `task_result_sender`, which keeps `network_contexts` authoritative for every context the process creates instead of adding a second place shutdown must remember to look.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
lklimek: **Independently corroborated — and there is a second uncovered call site beyond `spawn_backend_init`.**

Two reviewers converged on this same root cause from different directions, which is worth recording: `spawn_sync` (tasks.rs:32-42) records the name in `active_names` synchronously but defers JoinSet insertion to a nested `tokio::spawn(spawn_subtask(...))`, and `spawn_subtask` (tasks.rs:196-209) only registers once it acquires `subtasks.lock().await` — while `shutdown_inner` holds that same Mutex for the **entire** `join_next()` drain loop.

**The additive part:** this thread names `spawn_backend_init` (app.rs:1429, the boot/manual-connect path). The same race also reaches the **generic wallet-touching-task bootstrap** at `backend_task/mod.rs:783-795` — `self.ensure_wallet_backend(sender.clone()).await`, the lazy-wire path used by identity/wallet/DashPay/token/core/shielded tasks, i.e. *every* wallet-touching task family except `SwitchNetwork`. That call site has **no** cancellation-token check before or after it runs — notably unlike the guard this very PR adds for `SwitchNetwork` at backend_task/mod.rs:919-923:

```rust
if self.subtasks.cancellation_token.is_cancelled()
    && let Ok(backend) = new_ctx.wallet_backend() { backend.shutdown().await; }
```

So the PR demonstrably knows this guard is needed — it just applies it to exactly one of the call sites that needs it. Two independent call sites confirm the underlying `spawn_sync`/`shutdown_inner` registration race is a **structural gap in the barrier this PR relies on**, not an isolated edge case, which strengthens the case for your option (b): fix it in `TaskManager` rather than sprinkling per-call-site checks that the next wallet-touching task will forget.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/app.rs
Claudius-Maginificent: on_exit's force-close fallback now drains tasks/results, clears secrets, and shuts down wallet backends instead of skipping straight past them. Fixed in `80e27aa1`.

src/app.rs
Claudius-Maginificent: Added tests for the context filtering/deduplication, the registration barrier, and the shutdown-deadline behavior — the checklist's blanket test opt-out didn't actually cover this helper. Fixed in `80e27aa1`.

src/app.rs
Claudius-Maginificent: Remembered wallet secrets are now eagerly cleared before backend shutdown on both the normal and force-close paths. Fixed in `80e27aa1`.

src/app.rs
Claudius-Maginificent: Wallet teardown now has an explicit 30-second budget within the 45-second viewport deadline, with an explicit degraded outcome if it's exceeded rather than racing the deadline silently. Fixed in `80e27aa1`.

src/utils/tasks.rs
```diff
@@ -187,29 +202,42 @@ async fn shutdown_inner(
     }
 
     // Abort all remaining tasks
-    subtasks.lock().await.shutdown().await;
-
-    completed.load(std::sync::atomic::Ordering::Relaxed)
-}
+    tasks.shutdown().await;
```
thepastaclaw, line 384: **🔴 Blocking: Aborting join watchers does not stop backend work**

After SHUTDOWN_TIMEOUT expires, shutdown_inner aborts the remaining JoinSet entries and allows shutdown_async to report completion. The GUI entries named backend_task_join_watcher only await JoinHandles created by spawn_blocking in src/app.rs; once those blocking operations have started, aborting or dropping their JoinHandles cannot stop them. AppState then advances to shutdown_wallet_backends, so a slow wallet-touching run_backend_task operation can continue using, wiring, or writing through the same backend while its secrets are cleared and its storage is being shut down. Track actual completion of the blocking operations and wait for them to reach a terminal state, or add cooperative per-operation cancellation that reaches a terminal state before wallet teardown begins.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest delta fixes both prior blockers: shutdown now captures the final MCP context after managed task shutdown, and TaskManager closes registration atomically before taking its JoinSet. One cumulative blocking gap remains: timing out and aborting the tracked join watchers does not stop their already-running spawn_blocking backend operations, so wallet work can continue concurrently with wallet-backend teardown.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 1 blocking

<details>

```

In `src/utils/tasks.rs`:
- [BLOCKING] src/utils/tasks.rs:204-205: Aborting join watchers does not stop backend work
  After SHUTDOWN_TIMEOUT expires, shutdown_inner aborts the remaining JoinSet entries and allows shutdown_async to report completion. The GUI entries named backend_task_join_watcher only await JoinHandles created by spawn_blocking in src/app.rs; once those blocking operations have started, aborting or dropping their JoinHandles cannot stop them. AppState then advances to shutdown_wallet_backends, so a slow wallet-touching run_backend_task operation can continue using, wiring, or writing through the same backend while its secrets are cleared and its storage is being shut down. Track actual completion of the blocking operations and wait for them to reach a terminal state, or add cooperative per-operation cancellation that reaches a terminal state before wallet teardown begins.
```
</details>

src/app.rs
Claudius-Maginificent: Fixed in `80e27aa1` (fix(shutdown): make wallet teardown race-safe). The force-close fallback now routes through `run_blocking_shutdown_fallback()`, which builds the same `contexts` list from `network_contexts` and calls the shared `shutdown_wallet_backends(contexts)` used by the graceful path — so `forget_all_secrets()` + `wallet_backend.shutdown()` now run on both paths, not just the async one. `TaskManager::shutdown()` (the sync method this comment named) is now dead code; both shutdown paths route through `shutdown_async()` instead.

Independently re-verified against current HEAD, not just taken from the commit message.

src/app.rs
Claudius-Maginificent: Fixed in `80e27aa1`. The hard deadline is now `SHUTDOWN_TIMEOUT` (10s) + `WALLET_BACKEND_SHUTDOWN_TIMEOUT` (30s) + `SHUTDOWN_DEADLINE_MARGIN` (5s) = 45s, and `shutdown_wallet_backends` wraps each backend's shutdown future in `tokio::time::timeout(WALLET_BACKEND_SHUTDOWN_TIMEOUT, …)`. That covers the worst case cited here (10s task shutdown + up to 15s SPV-stop join) with margin to spare. `viewport_deadline_covers_task_and_wallet_shutdown_budgets` passes.

Independently re-verified against current HEAD, not just taken from the commit message.

src/app.rs
Claudius-Maginificent: Fixed in `80e27aa1`. `shutdown_wallet_backends` now calls `backend.forget_all_secrets()` for every backend before the `join_all` of `.shutdown()` calls, on both the async and blocking-fallback paths — it's a shared function, so no path skips it.

Independently re-verified against current HEAD, not just taken from the commit message.

src/app.rs
Claudius-Maginificent: Fixed in `80e27aa1`. `shutdown_tests::collect_created_context_only_keeps_network_context_results` is now a `#[cfg(test)]` unit test with no AppState/eframe/tokio dependency, covering exactly the three cases requested: Success+NetworkContextCreated (pushes, dedupes via `Arc::ptr_eq`), Success+other variant (no-op), Error (no-op). Test passes.

Independently re-verified against current HEAD, not just taken from the commit message.

src/utils/tasks.rs
Claudius-Maginificent: Independently re-verified against current HEAD (`38483722`) — still open. `shutdown_inner`'s abort-on-timeout only tears down the `backend_task_join_watcher` wrapper tasks (which just `.await` the underlying `spawn_blocking` `JoinHandle`); dropping the watcher's own future doesn't call `.abort()` on the inner `spawn_blocking` handle, so an already-running `run_backend_task` blocking op keeps executing, unobserved, while `shutdown_wallet_backends` proceeds to clear secrets and tear down storage.

`80e27aa1` (the most recent shutdown commit) only fixed the task-registration barrier race from the earlier review round — it doesn't touch this. Call-tree walk: both callers of `shutdown_inner` (`start_async_shutdown`, `run_blocking_shutdown_fallback`) proceed to `shutdown_wallet_backends` unconditionally afterward, regardless of whether the aborted watchers' underlying blocking work actually finished.

Recommendation stands: track real completion of the blocking ops, or thread per-operation cancellation through `run_backend_task` (per the code's own T7 TODO), before wallet teardown begins.

thepastaclaw (changes requested):

The new-push delta merges v1.0-dev changes and does not modify the shutdown machinery, so the remaining blocker is still present at the exact head. The PR fixes the task-registration barrier and final MCP-context collection, but after the task timeout it can still begin wallet-backend teardown while an already-running blocking backend operation continues. The three focused shutdown tests pass, but none exercises this timeout and detached-operation behavior.

### Review provenance

🔴 1 blocking

src/app.rs
```diff
@@ -1827,6 +1871,195 @@ impl AppState {
...
+                task_result_receiver,
+                #[cfg(feature = "mcp")]
+                mcp_app_context,
+            )
+            .await;
+            let outcome = match task_shutdown_outcome {
+                Some(TaskShutdownOutcome::Complete) => wallet_outcome,
+                Some(TaskShutdownOutcome::BackendTasksTimedOut) => {
+                    ShutdownOutcome::BackendTasksTimedOut
+                }
```
thepastaclaw: **🔴 Blocking: Backend-task timeout still permits concurrent wallet teardown**

`TaskShutdownOutcome::BackendTasksTimedOut` means at least one non-cancellable `spawn_blocking` closure remains active after the two bounded waits, but this code calls `finish_wallet_shutdown()` before inspecting that outcome. Those closures run wallet-touching backend operations that can still be signing, updating storage, or using coordinators, while wallet shutdown clears their secrets and calls `WalletBackend::shutdown()` on the same resources. The blocking fallback repeats this ordering at lines 2039-2055. A network switch also stores its new backend before `bootstrap_loaded_wallets()` finishes, but does not publish `NetworkContextRegistered` until that bootstrap returns, so a timed-out switch can remain absent from the collected teardown contexts. Do not start wallet teardown after `BackendTasksTimedOut`; either wait for actual terminal completion, or make every backend operation cooperatively cancel and acknowledge termination before teardown. If the bounded timeout must force process exit, skip shared wallet teardown on that degraded branch.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest commits fix the ordinary watcher-abort race, MCP-context collection, and task-registration barrier. One blocking race remains: after blocking work times out, both shutdown paths begin wallet teardown while non-cancellable backend work is still running. This is a preliminary Codex-only checkpoint; no Sonnet coverage is claimed.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 1 blocking

<details>

```

In `src/app.rs`:
- [BLOCKING] src/app.rs:1993-2004: Backend-task timeout still permits concurrent wallet teardown
  `TaskShutdownOutcome::BackendTasksTimedOut` means at least one non-cancellable `spawn_blocking` closure remains active after the two bounded waits, but this code calls `finish_wallet_shutdown()` before inspecting that outcome. Those closures run wallet-touching backend operations that can still be signing, updating storage, or using coordinators, while wallet shutdown clears their secrets and calls `WalletBackend::shutdown()` on the same resources. The blocking fallback repeats this ordering at lines 2039-2055. A network switch also stores its new backend before `bootstrap_loaded_wallets()` finishes, but does not publish `NetworkContextRegistered` until that bootstrap returns, so a timed-out switch can remain absent from the collected teardown contexts. Do not start wallet teardown after `BackendTasksTimedOut`; either wait for actual terminal completion, or make every backend operation cooperatively cancel and acknowledge termination before teardown. If the bounded timeout must force process exit, skip shared wallet teardown on that degraded branch.
```
</details>

src/utils/tasks.rs
```diff
@@ -34,124 +103,174 @@ impl TaskManager {
         F: std::future::Future<Output = ()> + Send + 'static,
         F::Output: Send + 'static,
     {
-        if let Ok(mut names) = self.active_names.lock() {
-            names.push(name);
+        let mut state = self.task_state.lock().unwrap_or_else(|e| e.into_inner());
+        if !state.accepting {
+            tracing::debug!(task = name, "Rejected task registration during shutdown");
+            return;
```
thepastaclaw: **🔴 Blocking: Rejected descendant registration detaches admitted work**

When shutdown closes registration, `spawn_sync()` drops the supplied future. `await_managed_network_request_with_timeout()` transfers a timed-out request's `JoinHandle` into such a reaper future at `src/backend_task/mod.rs:97`; if registration is rejected, dropping that handle detaches the still-running request. The admitted outer backend task can then finish and let `TaskManager` report `Complete` while its request continues using the wallet backend. Registration rejection must return ownership so the caller can abort or await the request, or descendants created by admitted work must remain tracked through shutdown.

<sub>source: ['codex']</sub>

src/utils/tasks.rs
```diff
@@ -34,124 +103,174 @@ impl TaskManager {
...
+        state.tasks.spawn(async move {
+            let _active_task = ActiveTaskGuard::new(active_names, name);
+            let result = completion
+                .await
+                .lock()
+                .unwrap_or_else(|error| error.into_inner())
+                .take();
+            if let Some(result) = result {
+                on_join(result).await;
+            }
```
thepastaclaw, line 236: **🟡 Suggestion: Shutdown can discard late blocking-task panic reports**

The blocking closure's completion and its `on_join` callback have different lifetimes. After the ordinary-task timeout, `tasks.shutdown()` aborts the observer that owns `on_join`; if the blocking closure then finishes during the blocking-task phase, `shutdown_blocking_tasks()` treats the shared completion as ready but never takes its stored `JoinError`. A backend closure that panics during this interval can therefore produce a `Complete` shutdown outcome without reaching `forward_backend_task_join_error` or the task-result channel. Keep error observation attached to the tracked completion, or make the blocking shutdown phase consume the result and execute the callback.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR adds bounded wallet-backend teardown, but three shutdown-barrier gaps still allow wallet resources to be cleared or shut down while backend work remains active. The revised blocking-task tracker also loses late panic reporting when its ordinary observer is aborted, so the PR requires changes before merge.

### Review provenance

<details>

```

In `src/utils/tasks.rs`:
- [BLOCKING] src/utils/tasks.rs:107-109: Rejected descendant registration detaches admitted work
  When shutdown closes registration, `spawn_sync()` drops the supplied future. `await_managed_network_request_with_timeout()` transfers a timed-out request's `JoinHandle` into such a reaper future at `src/backend_task/mod.rs:97`; if registration is rejected, dropping that handle detaches the still-running request. The admitted outer backend task can then finish and let `TaskManager` report `Complete` while its request continues using the wallet backend. Registration rejection must return ownership so the caller can abort or await the request, or descendants created by admitted work must remain tracked through shutdown.
- [SUGGESTION] src/utils/tasks.rs:158-167: Shutdown can discard late blocking-task panic reports
  The blocking closure's completion and its `on_join` callback have different lifetimes. After the ordinary-task timeout, `tasks.shutdown()` aborts the observer that owns `on_join`; if the blocking closure then finishes during the blocking-task phase, `shutdown_blocking_tasks()` treats the shared completion as ready but never takes its stored `JoinError`. A backend closure that panics during this interval can therefore produce a `Complete` shutdown outcome without reaching `forward_backend_task_join_error` or the task-result channel. Keep error observation attached to the tracked completion, or make the blocking shutdown phase consume the result and execute the callback.

In `src/mcp/dispatch.rs`:
- [BLOCKING] src/mcp/dispatch.rs:25-32: MCP backend operations remain outside the shutdown barrier
  Embedded MCP requests execute backend operations through a raw `tokio::task::spawn_blocking`, while `TaskManager` tracks only the enclosing HTTP server future. Cancellation or the ordinary-task timeout can drop the request's join handle without stopping its non-cancellable closure, after which the PR's new shutdown path loads the MCP context and tears down its wallet backend. MCP payments, withdrawals, shielded operations, and other backend tasks can therefore overlap wallet teardown. Register MCP dispatch work with the real-completion barrier used by GUI backend tasks, or add an equivalent in-flight request barrier that completes before wallet teardown.
```
</details>

Claudius-Maginificent:
Pushed fixes for the 3 blocking shutdown-race findings, plus the diagnostics-loss nit:

- Wallet teardown (`finish_wallet_shutdown`) now gated on `TaskShutdownOutcome::Complete` in both the async and blocking shutdown paths.
- `spawn_sync`'s timeout-reaper registration no longer silently detaches a still-running request from `TaskManager` tracking.
- MCP dispatch now routes through the tracked `spawn_blocking_sync` instead of a raw `spawn_blocking`, so it's included in shutdown draining.
- Late `JoinError`s from aborted blocking tasks are now consumed/logged instead of silently dropped.

Commit: `38dbb789`

src/utils/tasks.rs
```diff
@@ -1,160 +1,340 @@
...
+        let active_names = Arc::clone(&self.active_names);
+        state.tasks.spawn(async move {
+            let _active_task = ActiveTaskGuard::new(active_names, name);
+            let result = completion
+                .await
+                .lock()
+                .unwrap_or_else(|error| error.into_inner())
+                .take();
+            if let Some(result) = result {
+                on_join(result).await;
```
thepastaclaw: **🟡 Suggestion: Shutdown can still discard a pending join-error callback**

The ordinary observer removes the stored join result with `take()` before awaiting `on_join`. If the ordinary shutdown deadline expires after that removal but before the callback completes, `tasks.shutdown()` aborts the observer. The tracked phase then sees a completed task with no stored result and can return `Complete`, even though the typed panic report or other callback side effect was discarded. The new regression test covers completion after the observer has already been aborted, but not this consume-before-callback-completion interval. Keep the result available until `on_join` completes, or track callback completion as part of the shutdown barrier.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The current head fixes the wallet-teardown ordering and rejected descendant-registration defects. Two shutdown-barrier gaps remain: the MCP network-switch handler performs context publication and outgoing teardown after its tracked dispatch completes, and the join observer can consume a panic result before an abortable callback finishes.

### Review provenance

<details>

```

In `src/mcp/tools/network.rs`:
- [BLOCKING] src/mcp/tools/network.rs:247-250: MCP network switch leaves its final lifecycle work outside the barrier
  `dispatch_task()` tracks only `BackendTask::SwitchNetwork`. Once that task returns the newly created context, the handler separately awaits shutdown of the outgoing backend and then publishes the new context through `swap_context()`. The MCP dispatch uses a deliberately dropped task-result receiver, so the task's earlier `NetworkContextRegistered` result cannot register the new context with GUI shutdown. If shutdown cancels or times out the MCP request during this tail, `TaskManager` can report `Complete` while `finish_wallet_shutdown()` still sees the old context in the MCP `ArcSwap`; the newly initialized backend then misses explicit teardown. Keep outgoing teardown and context publication within tracked work, or publish/register the new context before the tracked dispatch can complete.

In `src/utils/tasks.rs`:
- [SUGGESTION] src/utils/tasks.rs:214-220: Shutdown can still discard a pending join-error callback
  The ordinary observer removes the stored join result with `take()` before awaiting `on_join`. If the ordinary shutdown deadline expires after that removal but before the callback completes, `tasks.shutdown()` aborts the observer. The tracked phase then sees a completed task with no stored result and can return `Complete`, even though the typed panic report or other callback side effect was discarded. The new regression test covers completion after the observer has already been aborted, but not this consume-before-callback-completion interval. Keep the result available until `on_join` completes, or track callback completion as part of the shutdown barrier.
```
</details>

src/utils/tasks.rs
Claudius-Maginificent: Confirmed, narrow but real race. In `register_tracked_completion`'s spawned observer (`tasks.rs:210-220`), `.take()` removes the shared `Option<Result<...>>` before `on_join(result).await` runs. That observer is itself one of the "ordinary" tasks in `shutdown_inner`'s `JoinSet`, so if the ordinary-phase timeout elapses in the window after `.take()` succeeds but before `on_join` finishes, `tasks.shutdown().await` aborts the observer mid-callback. `shutdown_tracked_tasks`'s later `.take()` on the same shared result then sees `None` (already consumed) and reports the task as cleanly `Complete`, silently losing whatever the callback would have reported (e.g. a panic). Agreed this needs either keeping the result available until `on_join` completes, or folding callback completion into the shutdown barrier itself.

<sub>Verified by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

Claudius-Maginificent:
Ran a comment-verification pass on this PR. One note: thepastaclaw's latest review (2026-07-21T03:37:05Z, head `38dbb789`) describes a **blocking** finding in `src/mcp/tools/network.rs:247-250` that doesn't appear to have a corresponding inline review thread, so flagging it here directly:

**MCP network switch leaves its final lifecycle work outside the shutdown barrier.** `dispatch_task()` only tracks `BackendTask::SwitchNetwork`; once that returns `NetworkContextCreated`, the handler (`src/mcp/tools/network.rs:234-250`) separately awaits the outgoing wallet backend's shutdown and then calls `service.swap_context(context)` — both **after** the tracked dispatch has already completed. Confirmed by reading the current code: if an app shutdown lands during that tail (before `swap_context` runs), `TaskManager` can report `Complete` while the MCP service's `ArcSwap` still holds the old context, so `finish_wallet_shutdown()` never sees — and never explicitly tears down — the newly installed backend.

The one existing inline thread (tasks.rs consume-before-callback race) was also verified still open — replied there directly.

src/app.rs
```diff
@@ -1985,6 +2029,210 @@ impl AppState {
...
+    where
+        F: std::future::Future<Output = ShutdownOutcome>,
+    {
+        match task_shutdown_outcome {
+            Some(TaskShutdownOutcome::Complete) => wallet_shutdown.await,
+            Some(TaskShutdownOutcome::BackendTasksTimedOut) => {
+                ShutdownOutcome::BackendTasksTimedOut
+            }
+            None => ShutdownOutcome::TaskManagerFailed,
+        }
```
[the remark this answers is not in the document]
Claudius-Maginificent: Re-verified against current `HEAD` (`26cb9d6d` + local merge of `v1.0-dev` + a CI compile fix, not yet pushed): still valid, not addressed.

`finish_shutdown_after_tasks` (src/app.rs) still routes `forget_all_secrets()` exclusively through `finish_wallet_shutdown`, which only runs on `TaskShutdownOutcome::Complete`. The `BackendTasksTimedOut` and `None` (`TaskManagerFailed`) branches return early with no secret clearing — confirmed by the `backend_task_timeout_skips_wallet_teardown` test itself, which asserts `teardown_calls == 0` for that outcome. The commit that landed since your review (`26cb9d6d`, "close remaining barrier gaps") fixed the join-error-callback and MCP network-switch barrier gaps but didn't touch this one.

Leaving unresolved — the recommended split (unconditional `forget_all_secrets()` on every collected context, `backend.shutdown()` still gated on `Complete`) hasn't been implemented yet. Out of scope for this pass (CI-fix + comment triage only); tracking as a follow-up.

src/mcp/server.rs
```diff
@@ -8,6 +8,27 @@ use rmcp::model::*;
...
+        }
+        Ok(TaskShutdownOutcome::BackendTasksTimedOut) => {
+            tracing::warn!("Managed backend work timed out; skipping standalone wallet teardown")
+        }
+        Err(_) => {
+            tracing::warn!("Managed task shutdown failed; skipping standalone wallet teardown")
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
Claudius-Maginificent: Re-verified against current `HEAD` (`26cb9d6d` + local merge of `v1.0-dev` + a CI compile fix, not yet pushed): still valid, not addressed.

`shutdown_app_context_wallet_backend` (src/mcp/server.rs) still only warns-and-skips on `BackendTasksTimedOut` and `Err(_)` — no `forget_all_secrets()` call on either degraded branch, matching your proposed diff. Same root cause as the sibling finding on `src/app.rs`'s `finish_shutdown_after_tasks`; neither has been fixed by the commits that landed since this review.

Leaving unresolved — out of scope for this pass (CI-fix + comment triage only); tracking as a follow-up alongside the `src/app.rs` counterpart.

Claudius-Maginificent:
Merged `origin/v1.0-dev` in and fixed the CI break:

- **Root cause**: `v1.0-dev` and this branch each touched `AppState` independently with no textual conflict, but the `NetworkContextRegistered` handler referenced `self.secret_prompt_host` — a field that never existed. The `Arc<dyn SecretPrompt>` built in `AppState::new()` was only installed on the initial network contexts, then dropped. This broke both `Test Suite` and `Clippy` (`E0609`).
- **Fix**: store the host as an `AppState` field (`a3a8da6e`) so later-registered contexts can also have it installed.
- Verified: `cargo build`, `cargo clippy --all-features -- -D warnings` (bin), `cargo test --all-features --workspace --no-run` (full compile), and targeted runs of the shutdown/secret-prompt/network-chooser test suites all green locally before pushing.

Also ran a comment-verification pass (fresh fetch, re-checked against current `HEAD`):

- Resolved `thepastaclaw`'s "pending join-error callback" finding (`src/utils/tasks.rs`) — genuinely fixed by `26cb9d6d`'s tracked `callback_completion` + new regression test. Thread resolved.
- Two findings remain **open**, unaddressed by commits since they were raised — replied on both threads:
  - `src/app.rs` `finish_shutdown_after_tasks`: degraded shutdown outcomes (`BackendTasksTimedOut`, `TaskManagerFailed`) still skip `forget_all_secrets()` entirely — only the `Complete` branch clears secrets. `backend_task_timeout_skips_wallet_teardown` itself asserts zero teardown calls for the timeout case.
  - `src/mcp/server.rs` `shutdown_app_context_wallet_backend`: same gap in the standalone MCP shutdown helper.

Both are pre-existing findings, not regressions from this push — flagging for a follow-up rather than blocking on them here.

src/app.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Fixed in `7603be6c`.

`finish_shutdown_after_tasks` now calls `forget_all_secrets()` on every collected wallet backend unconditionally, before branching on the outcome — only the full coordinator `backend.shutdown()` join stays gated on `Complete`. Applies to both `start_async_shutdown` and `run_blocking_shutdown_fallback`. New regression tests (`backend_task_timeout_forgets_secrets_without_wallet_teardown`, `task_manager_failure_forgets_secrets_without_wallet_teardown`) assert secrets are cleared with zero coordinator-shutdown calls on both degraded outcomes.

src/mcp/server.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Fixed in `7603be6c`, following your suggested shape.

`shutdown_app_context_wallet_backend` now calls `forget_all_secrets()` unconditionally before matching on the task-shutdown outcome; the coordinator `backend.shutdown()` join still only runs on `Complete`. New tests (`backend_task_timeout_forgets_standalone_wallet_secrets_without_teardown`, `task_manager_failure_forgets_standalone_wallet_secrets_without_teardown`) cover both degraded branches via a new mockable `StandaloneShutdownWalletBackend` seam.

Claudius-Maginificent:
Implemented the fix for both remaining open findings (degraded-shutdown outcomes skipping `forget_all_secrets()`) in `7603be6c`:

- `finish_shutdown_after_tasks` (`src/app.rs`) now unconditionally forgets secrets on every collected wallet backend before branching on the task-shutdown outcome; the full coordinator `backend.shutdown()` join stays gated on `Complete` only, since backend tasks may still be actively using those coordinators on a degraded outcome. Applies to both the async and blocking-fallback shutdown paths.
- `shutdown_app_context_wallet_backend` (`src/mcp/server.rs`) gets the same treatment via a new mockable `StandaloneShutdownWalletBackend` seam, per CodeRabbit's suggested shape.
- New regression tests in both files assert secrets are cleared with zero coordinator-shutdown calls on both degraded outcomes.

Verified independently (isolated build/target-dir, not reused from the implementing pass): `cargo clippy --all-features --bin dash-evo-tool -- -D warnings` clean, and all 5 new + 5 pre-existing `shutdown_tests`/`mcp::server::tests` pass by name.

Replied on both threads; left them for the original reviewers/human to resolve.

Claudius-Maginificent:
Ran a fresh `/grumpy-review` (security + project-consistency + QA, independent) against the two commits this session added (`177aea3d..7603be6c`) that hadn't been reviewed yet.

**Verdict: no blockers.** 6 non-blocking findings survive dedup (1 MEDIUM, 4 LOW, 1 INFO) — polish and hardening on an otherwise sound fix, not regressions:
- MEDIUM: `mcp/server.rs`'s new Complete-outcome secret-forget branch has zero test coverage (the two new tests only cover the degraded outcomes).
- LOW ×4: undocumented "why unconditional forget" invariant (risk of future regression), a duplicated `ShutdownWalletBackend`-shaped trait between `app.rs`/`mcp/server.rs`, an untested triple-forget-call composition on the real `Complete` path, and the inherent (documented, pre-existing-class, narrower-than-before) residual window where a degraded shutdown doesn't get a second-chance secret clear.
- INFO: clean call-tree walk, no broken callers from the signature changes.

Full report: `/data/artifacts/dash-evo-tool/2026-07-21/pr905-grumpy-review-7603be6c.md`

Treating these as optional follow-up rather than blocking this PR further — happy to implement any of them on request.

PR #906, opened 2026-07-17
fix: re-enable shielded transactions now that mainnet supports protocol v12

Claudius-Maginificent:
## Why this PR exists
- **Problem**: Shielded (Orchard) send/shield/unshield operations have been hardcoded closed in every build since PR #879, pending upstream shipping the shielded state transitions. They've since shipped — the connected network's protocol version is what actually gates them.
- **What breaks without it**: A user on a network that has activated shielded state transitions (confirmed live on testnet at protocol v12 during verification for this PR, and per the reporter already active on mainnet) still sees the Shielded tab's Shield/Send/Unshield controls hidden, with no way to use functionality the network already supports. All the shielded UI/backend code has been built and working since PR #793 — only this one runtime gate was closed.
- **Blocking relationship**: none.

## What was done
- `src/context/feature_gate.rs`: `SHIELDED_ACTIVATION_PROTOCOL_VERSION` now reads `Some(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION)` (upstream `rs-platform-version`'s own constant, protocol v12) instead of the placeholder `None`. `Capability::ShieldedProtocol` is met once the connected network's live protocol version reaches that value.
- Replaced the tripwire test that was designed to fail the moment this constant changed with real coverage of the v12 activation boundary across roles (Developer below/at v12, Everyday at v12).
- Added `FeatureGate::first_unmet_check`, a small generic addition that reports *which* check in a gate's conjunction is failing rather than a single opaque `bool`.
- `src/ui/wallets/shielded_tab.rs`: the "shielded sending unavailable" notice now uses `first_unmet_check` to show one of two accurate messages instead of always blaming the network — distinguishing "your network doesn't support this yet" (capability) from "your interface mode doesn't unlock this yet" (still Power/Developer-only, unchanged gating — this PR only fixes the messaging).
- `CHANGELOG.md` entries for both changes.

## Testing
- `cargo test -p dash-evo-tool -- feature_gate`: 14 tests pass, including the new/changed activation-boundary and `first_unmet_check` coverage (confirmed by name in the test log, not just an aggregate count).
- `cargo test -p dash-evo-tool -- shielded_tab`: passes, including the two renamed/new i18n-label tests.
- `cargo clippy --all-features --all-targets -p dash-evo-tool -- -D warnings`: clean.
- Live GUI verification against the real compiled app (`docs/gui-testing` tier): built and ran the binary, confirmed all three states render correctly —
  - boot state (protocol v0, Developer role): shielded controls correctly hidden with the network-unavailable message
  - live testnet, which is already at protocol v12, Developer role: Shield/Send/Unshield controls appear and are usable
  - live testnet at v12, Everyday role: controls correctly stay hidden (role gate still holds), now showing the role-specific message
  - no panics or crashes in the app log across any of the above

## Breaking changes
None. Shielded operations remain additionally gated behind the existing Power/Developer interface-mode requirement — this only unlocks the network-capability half of the gate, and only on networks that have actually activated the feature.

## Checklist
- [x] Tests added/updated and passing
- [x] Clippy clean
- [x] Verified against the real running app, not just unit tests
- [x] CHANGELOG updated

## Attribution

## Summary by CodeRabbit

* **New Features**
  * Enabled shielded sending, receiving, shielding, and unshielding on supported networks, including mainnet.
  * Added clearer availability notices distinguishing network support from interface permissions.

* **Bug Fixes**
  * Shielded operations now report the specific reason they are unavailable.
  * Prevented unavailable shielded actions from attempting to access wallet functionality.

### DIFF
--- a/src/context/feature_gate.rs
+++ b/src/context/feature_gate.rs
@@ -324,11 +324,16 @@ mod tests {
     }
 
     #[test]
-    fn shielded_operations_are_available_at_activation_for_developer() {
-        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
-        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+    fn shielded_operations_are_available_at_activation_for_unlocked_roles() {
+        for role in [UserRole::Power, UserRole::Developer] {
+            let (_tmp, ctx) = ctx_with_role(role);
+            ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
 
-        assert!(FeatureGate::ShieldedOperations.is_available(&ctx));
+            assert!(
+                FeatureGate::ShieldedOperations.is_available(&ctx),
+                "shielded operations must be available at activation for {role:?}"
+            );
+        }
     }
 
     #[test]
--- a/src/context/feature_gate.rs
+++ b/src/context/feature_gate.rs
@@ -1,21 +1,15 @@
 use crate::context::AppContext;
 use crate::model::user_role::UserRole;
+use dash_sdk::dpp::version::feature_initial_protocol_versions::SHIELDED_POOL_INITIAL_PROTOCOL_VERSION;
 
-/// The platform protocol version that first defines the shielded state
-/// transitions (shield, shielded transfer, unshield, shield from asset lock,
-/// shielded withdrawal) — `None` while they remain unshipped, which is the case
-/// on every released version today, so [`Capability::ShieldedProtocol`] is unmet
-/// on every network.
+/// Shielded state transitions activate at upstream rs-platform-version's
+/// [`SHIELDED_POOL_INITIAL_PROTOCOL_VERSION`] (protocol v12), sourced directly
+/// so DET stays aligned if upstream renumbers the feature.
 ///
-/// TODO: set this to the activation version when upstream ships the shielded state
-/// transitions, then re-check the shielded gates (the tripwire test below fails
-/// until they are). Do NOT infer activation from
-/// `FeatureVersionBounds::max_version > 0` instead: `check_version` is
-/// `v >= min && v <= max`, so `{min: 0, max: 0}` is a legitimately checkable v0
-/// bound — `identity_create_state_transition`, a live feature, ships exactly that
-/// triple — not an "undefined" marker. A shielded transition released at v0 would
-/// read as permanently absent. See the PR #879 review.
-const SHIELDED_ACTIVATION_PROTOCOL_VERSION: Option<u32> = None;
+/// Do not infer activation from `FeatureVersionBounds::max_version > 0`:
+/// `{ min: 0, max: 0 }` is a valid v0 bound, not an undefined marker.
+const SHIELDED_ACTIVATION_PROTOCOL_VERSION: Option<u32> =
+    Some(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
 
 /// A runtime capability of the connected platform, evaluated against the live
 /// context. Independent of the user's role — it answers "does the connected
@@ -38,7 +32,7 @@ impl Capability {
     fn is_met(self, ctx: &AppContext) -> bool {
         match self {
             Capability::ShieldedProtocol => match SHIELDED_ACTIVATION_PROTOCOL_VERSION {
-                // Not shipped anywhere yet, so no network can offer it.
+                // A closed gate makes the capability unavailable on every network.
                 None => false,
                 // The version fetched from the connected network, not the hardcoded
                 // default — the capability is per-network. The boot value (0, "not
@@ -303,30 +297,47 @@ mod tests {
         }
     }
 
-    /// Tripwire. Shielded state transitions have not shipped, so
-    /// [`SHIELDED_ACTIVATION_PROTOCOL_VERSION`] is still `None`: the capability is
-    /// unmet on every protocol version upstream defines, and the "capability met"
-    /// half of the AND cannot be exercised yet.
-    ///
-    /// This test fails the moment that constant names a version upstream actually
-    /// ships. That is the point: whoever activates the capability must then
-    /// re-check the shielded gates and add the missing
-    /// `capability ∧ role ⇒ available` case below.
     #[test]
-    fn no_known_protocol_version_reaches_the_shielded_activation_version() {
+    fn shielded_capability_tracks_the_activation_boundary() {
         let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
         let versions = known_protocol_versions();
         assert!(!versions.is_empty(), "the probe must find some versions");
 
         for version in versions {
             ctx.set_platform_protocol_version(version);
-            assert!(
-                !Capability::ShieldedProtocol.is_met(&ctx),
-                "protocol v{version} now reaches the shielded activation version \
-                 ({SHIELDED_ACTIVATION_PROTOCOL_VERSION:?}) — the shielded gates have a \
-                 reachable capability and need re-checking"
+            assert_eq!(
+                Capability::ShieldedProtocol.is_met(&ctx),
+                version >= SHIELDED_POOL_INITIAL_PROTOCOL_VERSION,
+                "shielded capability on protocol v{version}"
             );
         }
+    }
+
+    #[test]
+    fn shielded_operations_are_available_at_activation_for_developer() {
+        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
+        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+
+        assert!(FeatureGate::ShieldedOperations.is_available(&ctx));
+    }
+
+    #[test]
+    fn shielded_operations_are_unavailable_before_activation_for_developer() {
+        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
+        let version = SHIELDED_POOL_INITIAL_PROTOCOL_VERSION
+            .checked_sub(1)
+            .expect("shielded activation must follow the boot protocol version");
+        ctx.set_platform_protocol_version(version);
+
+        assert!(!FeatureGate::ShieldedOperations.is_available(&ctx));
+    }
+
+    #[test]
+    fn shielded_operations_are_unavailable_at_activation_for_everyday_user() {
+        let (_tmp, ctx) = ctx_with_role(UserRole::Everyday);
+        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+
+        assert!(!FeatureGate::ShieldedOperations.is_available(&ctx));
     }
 
     /// AND semantics. `ShieldedOperations` is the first multi-check gate: it needs
--- a/src/context/feature_gate.rs
+++ b/src/context/feature_gate.rs
@@ -158,6 +158,16 @@ impl FeatureGate {
     pub fn is_available(self, ctx: &AppContext) -> bool {
         self.checks().iter().all(|c| c.is_met(ctx))
     }
+
+    /// The first check that fails to hold for this gate, or `None` if every
+    /// check passes (equivalent to `is_available` returning `true`). Lets a
+    /// caller surface which axis is actually blocking availability — e.g.
+    /// distinguishing "the network doesn't support this yet" (a
+    /// [`Check::Capability`]) from "your interface mode doesn't unlock this" (a
+    /// [`Check::Experimental`]) — instead of a single opaque `false`.
+    pub fn first_unmet_check(self, ctx: &AppContext) -> Option<Check> {
+        self.checks().iter().copied().find(|c| !c.is_met(ctx))
+    }
 }
 
 #[cfg(test)]
@@ -338,6 +348,33 @@ mod tests {
         ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
 
         assert!(!FeatureGate::ShieldedOperations.is_available(&ctx));
+    }
+
+    #[test]
+    fn shielded_operations_reports_the_first_unmet_check() {
+        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
+        let version = SHIELDED_POOL_INITIAL_PROTOCOL_VERSION
+            .checked_sub(1)
+            .expect("shielded activation must follow the boot protocol version");
+        ctx.set_platform_protocol_version(version);
+        assert_eq!(
+            FeatureGate::ShieldedOperations.first_unmet_check(&ctx),
+            Some(Check::Capability(Capability::ShieldedProtocol))
+        );
+
+        let (_tmp, ctx) = ctx_with_role(UserRole::Everyday);
+        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+        assert_eq!(
+            FeatureGate::ShieldedOperations.first_unmet_check(&ctx),
+            Some(Check::Experimental(ExperimentalFeature::Shielded))
+        );
+
+        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
+        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+        assert_eq!(
+            FeatureGate::ShieldedOperations.first_unmet_check(&ctx),
+            None
+        );
     }
 
     /// AND semantics. `ShieldedOperations` is the first multi-check gate: it needs
--- a/src/mcp/resolve.rs
+++ b/src/mcp/resolve.rs
@@ -226,6 +226,10 @@ async fn ensure_legacy_storage_migrated(ctx: &Arc<AppContext>) -> Result<(), Mcp
 /// `WalletStorageNotReady` fast-fail that `run_backend_task` applies while
 /// migration is mid-flight.
 ///
+/// Once synced, an unpopulated Platform protocol cache is refreshed so
+/// headless feature gates evaluate the connected network rather than boot state.
+/// Refresh failures remain best-effort here so Core-only tools can proceed.
+///
 /// ## Why `SpvStatus::Running`, not `OverallConnectionState::Synced`
 ///
 /// `OverallConnectionState::Synced` requires both SPV running **and**
@@ -237,6 +241,26 @@ async fn ensure_legacy_storage_migrated(ctx: &Arc<AppContext>) -> Result<(), Mcp
 /// proof-verifying SDK calls only require a synced chain, not a live DAPI
 /// counter at the `ensure_spv_synced` callsite.
 pub(crate) async fn ensure_spv_synced(ctx: &Arc<AppContext>) -> Result<(), McpToolError> {
+    ensure_spv_ready(ctx, ProtocolRefresh::BestEffortIfUnpopulated).await
+}
+
+/// Require synced SPV and fresh epoch metadata before a shielded fund movement.
+pub(crate) async fn ensure_shielded_operations_ready(
+    ctx: &Arc<AppContext>,
+) -> Result<(), McpToolError> {
+    ensure_spv_ready(ctx, ProtocolRefresh::Required).await
+}
+
+#[derive(Clone, Copy)]
+enum ProtocolRefresh {
+    BestEffortIfUnpopulated,
+    Required,
+}
+
+async fn ensure_spv_ready(
+    ctx: &Arc<AppContext>,
+    protocol_refresh: ProtocolRefresh,
+) -> Result<(), McpToolError> {
     // A throwaway `TaskResult` sender: MCP/CLI has no GUI event loop consuming
     // it, so the receiver is dropped. The `EventBridge` only does non-blocking
     // `try_send`, so a closed channel is harmless. Mirrors `dispatch::dispatch_task`.
@@ -249,11 +273,15 @@ pub(crate) async fn ensure_spv_synced(ctx: &Arc<AppContext>) -> Result<(), McpTo
 
     ensure_legacy_storage_migrated(ctx).await?;
 
-    // Subscribe BEFORE reading the current value so no transition is lost
-    // between the `ensure_wallet_backend_and_start_spv` call above and the
-    // first `borrow_and_update` below. borrow_and_update marks the current
-    // value "seen", so the loop never spins — each iteration always sleeps on
-    // a real change.
+    wait_for_spv_and_refresh_platform_info(ctx, protocol_refresh).await
+}
+
+async fn wait_for_spv_and_refresh_platform_info(
+    ctx: &Arc<AppContext>,
+    protocol_refresh: ProtocolRefresh,
+) -> Result<(), McpToolError> {
+    // Subscribe before reading the current value so no transition is lost.
+    // `borrow_and_update` keeps later iterations asleep until a real change.
     let mut rx = ctx.connection_status().subscribe_spv_status();
 
     let wait = async {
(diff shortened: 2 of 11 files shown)


thepastaclaw:
🕓 Ready for review — 27 ahead in queue (commit 1dd79d40)
_Queue position: 28/29_

src/context/feature_gate.rs
```diff
@@ -303,32 +307,76 @@ mod tests {
...
         }
     }
 
+    #[test]
+    fn shielded_operations_are_available_at_activation_for_developer() {
+        let (_tmp, ctx) = ctx_with_role(UserRole::Developer);
+        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);
+
+        assert!(FeatureGate::ShieldedOperations.is_available(&ctx));
+    }
```
thepastaclaw, line 337: **🟡 Suggestion: Test the Power role at the activation boundary**

The positive activation test covers Developer only, although the actual experimental threshold is `UserRole::Power` and the new UI message promises that Expert view unlocks shielded sending. The Everyday negative test does not verify this lower unlocked boundary, so changing the shielded gate to require Developer would leave all activation-specific tests passing while breaking the advertised Power-role behavior. Exercise both unlocked roles at protocol v12.

```suggestion
#[test]
fn shielded_operations_are_available_at_activation_for_unlocked_roles() {
    for role in [UserRole::Power, UserRole::Developer] {
        let (_tmp, ctx) = ctx_with_role(role);
        ctx.set_platform_protocol_version(SHIELDED_POOL_INITIAL_PROTOCOL_VERSION);

        assert!(
            FeatureGate::ShieldedOperations.is_available(&ctx),
            "shielded operations must be available at activation for {role:?}"
        );
    }
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The protocol-v12 activation boundary and role-specific UI messaging are implemented consistently, but the protocol-version value is only populated by a one-shot epoch-info request. A failed request or an app session spanning activation can therefore leave supported shielded operations disabled until reconnection or a manual refresh. The backend also discards the typed unavailability reason, and the tests do not directly pin the Power-role activation boundary.

### Review provenance

<details>

```

In `src/context/feature_gate.rs`:
- [BLOCKING] src/context/feature_gate.rs:41: Protocol-version cache is never refreshed after initial sync
  This gate treats `platform_protocol_version()` as current network state, but the only production setter runs after a successful `CurrentEpochInfo` request. `ConnectionBanner` dispatches that request only when the connection transitions into `OverallConnectionState::Synced`; it does not retry a failed request while the state remains synced or refresh the value when a later epoch activates protocol v12. Consequently, a transient failure during initial sync or an app session spanning activation leaves both the UI and authoritative backend guards rejecting shielded operations on a supported network until reconnection or a manual Platform Info request. This became an in-scope functional bug when this PR changed the cached value from informational state into the condition enabling shielded fund movements. Retry failed initial requests and refresh epoch information periodically or when the epoch changes.
- [SUGGESTION] src/context/feature_gate.rs:326-332: Test the Power role at the activation boundary
  The positive activation test covers Developer only, although the actual experimental threshold is `UserRole::Power` and the new UI message promises that Expert view unlocks shielded sending. The Everyday negative test does not verify this lower unlocked boundary, so changing the shielded gate to require Developer would leave all activation-specific tests passing while breaking the advertised Power-role behavior. Exercise both unlocked roles at protocol v12.

- [SUGGESTION] src/backend_task/error.rs:1859-1864: Preserve the shielded unavailability reason in the backend error
  The PR introduces `first_unmet_check()` so callers can distinguish an unsupported network from an insufficient interface role, but both authoritative backend guards still return the same fieldless `ShieldedOperationsUnavailable` variant. This is directly observable through the registered MCP shielded tools, which expose the variant's `Display` text: an Everyday user on protocol v12 is told to use a regular payment or wait for a future update even though switching to Expert view enables the operation, while a Power or Developer user below v12 receives application-update advice for a network capability restriction. Carry a typed unavailability reason in `TaskError`, construct it from the failed gate check in both backend guards, and provide an accurate actionable message for each reason.
```
</details>

src/ui/wallets/shielded_tab.rs
```diff
@@ -710,8 +716,12 @@ impl ShieldedTabView {
                 });
             }
         } else {
+            let label = match FeatureGate::ShieldedOperations.first_unmet_check(&self.app_context) {
```
lklimek: 🟠 **`first_unmet_check()` is wired into the UI notice but not into the backend's authoritative refusal error — the PR's own "distinguish network vs role" fix is only half-applied**

This line is the *only* one of four production call sites that uses the new API. `first_unmet_check()`'s own doc comment says it exists to surface "which axis is actually blocking availability... instead of a single opaque `false`" — and the commit title is "distinguish network vs role reason in shielded-unavailable notice." But the other three sites still return the single undifferentiated error:

- `src/backend_task/shielded/mod.rs:77` (`run_shielded_task`)
- `src/backend_task/mod.rs:751` (the pre-check in `run_backend_task`)
- every MCP shielded tool — all five ops route through `dispatch_task` → `run_backend_task` → `run_shielded_task`

All of them surface `TaskError::ShieldedOperationsUnavailable`, whose `#[error(...)]` text (`src/backend_task/error.rs:1861`) is fixed:

```rust
#[error(
    "Shielding, sending, or withdrawing shielded funds is not available right now. Use a regular payment instead, or try again after a future update."
)]
ShieldedOperationsUnavailable,
```

A standalone MCP/CLI context has **no role-switching tool at all** and defaults to `UserRole::Everyday`, which permanently fails the `Experimental(Shielded)` check (needs >= Power) regardless of network readiness. So for that entire integration surface, "try again after a future update" is **false** — no future protocol upgrade fixes a role gate; only exposing a role-switch capability would. That's the exact class of misleading guidance this PR set out to eliminate, left in place at the three sites a headless integrator can actually observe.

**Recommendation**: thread `first_unmet_check()` (or an equivalent typed reason) into `TaskError::ShieldedOperationsUnavailable` so its `Display` differs for `Check::Capability` vs `Check::Experimental`, or add a second `TaskError` variant for the role-blocked case — per the project's own convention of dedicated variants over one generic message, and the rule that action guidance must be accurate.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/ui/wallets/shielded_tab.rs
```diff
@@ -710,8 +716,12 @@ impl ShieldedTabView {
...
+                Some(Check::Experimental(_)) => SHIELDED_OPERATIONS_ROLE_UNAVAILABLE_LABEL,
+                _ => SHIELDED_OPERATIONS_NETWORK_UNAVAILABLE_LABEL,
```
lklimek: 🟡 **Wildcard `_` arm instead of exhaustively matching `Check::Capability` — silently mislabels any future third check as "network"**

`Check` is a 3-variant enum today (`MinRole`, `Capability`, `Experimental`), and `feature_gate.rs`'s own doc comment flags a plausible fourth (`Context`) as a likely future addition, explicitly deferred as YAGNI.

This isn't a live bug — `Capability` is correctly the only other reachable arm right now. But if `ShieldedOperations.checks()` later grows a `Check::MinRole(...)` or `Check::Context(...)` entry, this `_` arm swallows it silently under the "network unavailable" label, with no compiler nudge to revisit. That's precisely the class of mislabeling bug this PR exists to fix, sitting one check-list change away from being reintroduced.

**Recommendation**: match `Some(Check::Capability(_))` explicitly and let the compiler force a decision on any new `Check` variant. (Or keep the wildcard but add a `debug_assert!`/log on the unexpected-variant path.)

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/ui/wallets/shielded_tab.rs
```diff
@@ -710,8 +716,12 @@ impl ShieldedTabView {
...
+            };
```
lklimek: 🟡 **`is_available()` and `first_unmet_check()` both re-evaluate the same check list per frame — redundant, plus a theoretical single-frame TOCTOU window**

Line 637's `if FeatureGate::ShieldedOperations.is_available(...)` and this block's `first_unmet_check(...)` (reached only in the `else` branch) each independently iterate `checks()` against live atomics (`ctx.platform_protocol_version()`, `ctx.user_role()`). Between the two calls — same egui frame — a background task could in principle flip `platform_protocol_version`, making the second read inconsistent with the first.

Low severity and honestly assessed as such: today's `_` wildcard catches the resulting `None` case anyway, so the worst case is one stale frame of wording that self-corrects on the next repaint. Flagging it mainly because the fix also removes the redundancy.

**Recommendation**: call `first_unmet_check()` once and derive both branches from its `Option` — `None` => show controls, `Some(check)` => match on `check` for the label — instead of calling both `is_available()` and `first_unmet_check()`.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

lklimek (commented):
## Grumpy review — consolidated findings

**The core funds-safety logic holds up under adversarial review.** This is the part that matters, since this PR gates real funds-relevant functionality (shielded transactions on mainnet), so it was verified rather than taken on faith — two independent reviewers read the pinned upstream source and ran the suites, and neither found a version-check or bypass defect:

- `SHIELDED_ACTIVATION_PROTOCOL_VERSION` sources upstream's own `SHIELDED_POOL_INITIAL_PROTOCOL_VERSION` (= 12) directly, matching drive-abci's consensus `is_allowed` comparison (activation at v12, rejection at v11). No off-by-one, no inverted logic — the operator is `>=`.
- **Fails closed at boot**: the atomic starts at `0`, so every role is blocked until a live epoch fetch overwrites it.
- **Per-network, not global**: `ctx.platform_protocol_version()` is read fresh from the live `AppContext`, so mainnet stays gated until its *own* connection reports v12, regardless of testnet/devnet state. No cross-network leakage.
- **Conjunctive AND semantics**: capability and role checks are independent — a Developer role cannot buy past a failing capability check. No privilege-escalation path.
- No stray or duplicate hardcoded shielded-version comparison exists anywhere else in the tree — one source of truth, and all four production call sites read the same predicate.
- `shielded_capability_tracks_the_activation_boundary` sweeps *every* upstream-defined protocol version against the exact `>=` predicate — the strongest boundary coverage short of a property test.

CHANGELOG accuracy is also worth crediting: the "Fixed" bullet scopes itself to the `shielded_tab.rs` notice split rather than claiming all messaging now distinguishes both reasons, and the "Changed" bullet's phrasing is conditional ("when the connected network's protocol version supports them") rather than asserting mainnet is live today.

### The one substantive finding

**CALL-001 (medium)** — the PR's own advertised behavior is only half-applied. The commit title is "distinguish network vs role reason," and `first_unmet_check()` was added specifically for that, but only 1 of 4 production call sites uses it. The three backend sites — including every MCP shielded tool — still return the undifferentiated `TaskError::ShieldedOperationsUnavailable`, whose text says "try again after a future update." For MCP/CLI callers (which default to `UserRole::Everyday` and have no role-switching tool at all), that guidance is simply false: no protocol upgrade fixes a role gate. Inline detail below. Worth fixing or explicitly deferring with a tracked follow-up — it's a messaging accuracy issue, not a safety one.

### Two findings that can't be inlined (files outside this diff)

- **CODE-002 (low, pre-existing, not caused by this PR)** — `tests/backend-e2e/framework/shielded_helpers.rs:14-22`: `is_shielded_available` claims to test whether the network supports shielded state transitions, but calls `FeatureGate::Shielded`, whose check list is intentionally empty so balances stay viewable everywhere. It therefore returns `true` before v12. Callers in `tc_074_shielded_lifecycle` and `tc_079_shield_from_balance` don't skip unsupported networks, so TC-079 can fund a Platform address before the (correctly fail-closed) backend rejects the write. This predates the PR — pre-PR the gate was unconditionally closed, so the same wasted setup already happened — and production still fails closed, so it defeats an E2E preflight and wastes funded setup, nothing more. Suggest switching it to `FeatureGate::ShieldedOperations.is_available(app_context)`.
- **PROJ-001 (low)** — `docs/user-stories.md:276-284` (WAL-029) and `368-454` (SND-007/009/010/015/016): these six stories are tagged `[Implemented]` and describe the flows as simply reachable, with no acceptance criterion for the two-axis gating (network protocol v12 **AND** Expert view or higher) that now actually governs them. A reader can't learn from the doc that these were non-functional everywhere until this PR. CLAUDE.md's catalog policy covers adding stories and flipping `[Gap]` tags, not refreshing criteria on already-`[Implemented]` ones — so this is a suggested improvement, not a process violation.

### Dropped as duplicate

One finding (no direct `UserRole::Power` assertion at the activation boundary) was dropped — @thepastaclaw already filed it on `feature_gate.rs:332` with a suggestion patch. Independent agreement on that one; the existing thread stands.

**Verdict**: no funds-safety defect in the activation gate. Resolve CALL-001 or defer it explicitly; the rest are non-blocking follow-ups.

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent — automated grumpy-review (Claude trio-equivalent + Codex Sol)</sub>

src/ui/wallets/shielded_tab.rs
Claudius-Maginificent: The UI now exhaustively matches Check::Capability, Check::MinRole, and Check::Experimental instead of falling through a wildcard arm. Fixed in `ecf98742`.

src/ui/wallets/shielded_tab.rs
Claudius-Maginificent: first_unmet_check() is now called once per frame instead of being re-evaluated redundantly by both is_available() and the notice path. Fixed in `ecf98742`.

src/ui/wallets/shielded_tab.rs
Claudius-Maginificent: first_unmet_check() and the backend's authoritative refusal error now share a typed classifier, so both the UI notice and the backend refusal agree on network-vs-role. Fixed in `ecf98742`.

thepastaclaw (changes requested):

Live PR HEAD and local HEAD both resolve to ecf98742cb5a2c1ddc86b4abf0400ebb3c72f068. The latest delta fixes the Power-role boundary coverage and typed backend unavailability errors, but the prior blocking protocol-cache lifecycle issue remains: GUI refresh is one-shot on a Synced transition, while standalone MCP/CLI never populates the cache. Supported shielded fund-moving operations can therefore remain disabled indefinitely.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 1 blocking

<details>

```

In `src/context/feature_gate.rs`:
- [BLOCKING] src/context/feature_gate.rs:41: Protocol-version cache is never refreshed after initial sync
  The shielded capability now treats `AppContext::platform_protocol_version()` as authoritative network state, but the cache lifecycle does not maintain that invariant. The cache starts at `0`, and the only production setter is the successful `CurrentEpochInfo` path in `src/backend_task/platform_info.rs`. In the GUI, `ConnectionBanner` dispatches that request only when the overall connection state transitions to `Synced`; a transient fetch failure is not retried while the connection remains synced, and a session spanning a later protocol activation retains its previous version. Standalone MCP/CLI initialization and `ensure_spv_synced` never request current epoch information, so those contexts retain `0` and reject every shielded fund-moving tool as network-unavailable even on protocol v12. Because this PR makes the cache the authoritative guard that re-enables these operations, initialize it in every runtime path and refresh or retry it when epoch information changes or fetching fails.
```
</details>

thepastaclaw (changes requested):

The protocol-v12 boundary, Power-role coverage, and typed backend refusal errors are correct at the current head. However, the protocol-version cache remains a one-shot GUI-only update and is never populated by standalone MCP/CLI, so supported shielded writes can remain disabled indefinitely.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 1 blocking

<details>

```

In `src/context/feature_gate.rs`:
- [BLOCKING] src/context/feature_gate.rs:41: Maintain the protocol-version cache as authoritative network state
  Shielded operations now depend on `platform_protocol_version()`, but the cache starts at 0 and its only production setter is the successful `CurrentEpochInfo` path in `src/backend_task/platform_info.rs`. The GUI requests that information only when `ConnectionReconciler` transitions into `Synced`, so a failed request is not retried while the connection remains synced and a session spanning a later protocol activation retains its old value. Standalone MCP/CLI creates the context with version 0, and `ensure_spv_synced()` never fetches epoch information, causing every shielded write to be rejected as network-unavailable even on protocol v12. Populate this state in every runtime path and refresh or retry it when fetching fails or the current epoch changes.
```
</details>

Claudius-Maginificent:
Pushed a fix for the headless shielded-protocol-version gap plus two smaller follow-ups:

- MCP/CLI's `ensure_spv_synced` now fetches/refreshes `CurrentEpochInfo` (with retry) when the protocol version is still unpopulated, so shielded operations actually become available headlessly on a v12 network instead of staying gated forever.
- `tests/backend-e2e/framework/shielded_helpers.rs` now checks the correct `FeatureGate::ShieldedOperations` variant.
- `docs/user-stories.md` SND-007/009/010/015/016 acceptance criteria updated to cover both the protocol-version and role axes.

Commit: `31029372`

src/mcp/resolve.rs
```diff
@@ -274,16 +302,49 @@ pub(crate) async fn ensure_spv_synced(ctx: &Arc<AppContext>) -> Result<(), McpTo
...
             );
-            Err(McpToolError::SpvSyncFailed)
+            return Err(McpToolError::SpvSyncFailed);
         }
     }
+
+    match protocol_refresh {
+        ProtocolRefresh::BestEffortIfUnpopulated if ctx.platform_protocol_version() == 0 => {
+            let _ = refresh_platform_protocol_version(ctx).await;
+            Ok(())
```
thepastaclaw, line 319: **🟡 Suggestion: Bound the best-effort refresh for unrelated Core tools**

`BestEffortIfUnpopulated` awaits the full `CurrentEpochInfo` backend task before returning, even though `ensure_spv_synced` is called by Core-only operations such as `core_address_create`, `core_balances_get`, and `core_funds_send` that do not consume the protocol cache. The SDK is configured with a 10-second request timeout and six retries (`src/sdk_wrapper.rs:14-18`), and the error is discarded only after that wait; if the request fails, the cache remains zero and every later Core call repeats the delay. This is a non-blocking latency regression rather than a correctness blocker because the Core operation is still allowed to proceed after the refresh attempt completes, but the best-effort path should be bounded, single-flight/background, or limited to Platform-dependent callers.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest push correctly refreshes protocol metadata before headless shielded fund movements and fixes the prior Power-role and typed-unavailability gaps. The GUI still treats a transition-only protocol-version cache as live network state, so supported shielded operations can remain hidden or rejected after an initial epoch-query failure or an in-session activation; the new best-effort headless refresh also adds avoidable DAPI latency to Core-only tools.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)

<details>

```

In `src/context/feature_gate.rs`:
- [BLOCKING] src/context/feature_gate.rs:41: Maintain the GUI protocol-version cache as live network state
  This PR activates `ShieldedOperations` by reading `platform_protocol_version()` as authoritative network state, but the GUI still updates that cache only when `ConnectionBanner::update` observes a transition into `OverallConnectionState::Synced` (`src/app/reconcilers.rs:330-337`). The reconciler records `Synced` whether the dispatched `CurrentEpochInfo` request succeeds or fails, so a failed request is not retried while the state stays synced; likewise, a GUI session that cached a pre-v12 epoch never observes a later v12 activation without a connection cycle, a manual epoch query, or a restart. The new refresh in `mcp::resolve` covers MCP/CLI calls only, while GUI controls and both backend shielded guards continue reading the stale atomic value, causing the PR's newly supported operations to remain hidden and be rejected on a supported network.

In `src/mcp/resolve.rs`:
- [SUGGESTION] src/mcp/resolve.rs:316-319: Bound the best-effort refresh for unrelated Core tools
  `BestEffortIfUnpopulated` awaits the full `CurrentEpochInfo` backend task before returning, even though `ensure_spv_synced` is called by Core-only operations such as `core_address_create`, `core_balances_get`, and `core_funds_send` that do not consume the protocol cache. The SDK is configured with a 10-second request timeout and six retries (`src/sdk_wrapper.rs:14-18`), and the error is discarded only after that wait; if the request fails, the cache remains zero and every later Core call repeats the delay. This is a non-blocking latency regression rather than a correctness blocker because the Core operation is still allowed to proceed after the refresh attempt completes, but the best-effort path should be bounded, single-flight/background, or limited to Platform-dependent callers.
```
</details>

src/mcp/resolve.rs
Claudius-Maginificent: Confirmed. `ProtocolRefresh::BestEffortIfUnpopulated` (`src/mcp/resolve.rs:316-319`) still `.await`s the full `refresh_platform_protocol_version` dispatch and only discards the error afterward — it doesn't return early. Since `ensure_spv_synced` (which uses this best-effort path) is the same gate used by Core-only tools (`core_address_create`, `core_balances_get`, `core_funds_send`, etc.), any of them can pay the SDK's full 10s-timeout × 6-retry latency on every call while `platform_protocol_version()` is unpopulated or the query keeps failing. Non-blocking (the Core operation still proceeds after), but agreed it should be bounded, single-flight/background, or skipped entirely for callers that don't consume the protocol cache.

<sub>Verified by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

Claudius-Maginificent:
Ran a comment-verification pass on this PR. One note: thepastaclaw's latest review (2026-07-21T03:30:01Z, head `31029372`) describes a **blocking** finding in `src/context/feature_gate.rs:41` that doesn't appear to have a corresponding inline review thread, so flagging it here directly:

**The GUI treats a one-shot protocol-version cache as live network state.** `FeatureGate::ShieldedOperations` reads `ctx.platform_protocol_version()` as authoritative (`feature_gate.rs:41`), but the GUI only refreshes that value once, on the *first* transition into `OverallConnectionState::Synced` (`src/app/reconcilers.rs:328-337`) — confirmed by reading the code: `self.previous_state` is set to `Synced` regardless of whether the dispatched `CurrentEpochInfo` request succeeds, and there's no re-dispatch on later frames while the connection stays `Synced`. So a failed initial epoch query, or a later in-session protocol activation (e.g. crossing v12), never gets re-observed by the GUI without a full connection cycle or restart — leaving newly-supported shielded operations hidden or wrongly rejected. The new `mcp::resolve` refresh only covers MCP/CLI, not the GUI path or the backend shielded guards.

The one existing inline thread (best-effort refresh latency for Core-only tools) was also verified still open — replied there directly.

PR #914, opened 2026-07-21
fix(wallets): wire "Add Receiving Address" to the same flow as "Receive"

Claudius-Maginificent:
## Why this PR exists
- **Problem**: The "Add Receiving Address" button on the Dash Core account view does not appear to do anything when clicked.
- **What breaks without it**: Select a wallet, open the Dash Core account tab, click "➕ Add Receiving Address". The button issues the `GenerateReceiveAddress` backend task directly, but never opens the Receive dialog. The newly derived address is silently appended to `receive_dialog.core_addresses` — state that's invisible because the dialog was never opened. The user sees no confirmation, no dialog, no new address anywhere: the click looks like a no-op.

## What was done
Route the button through the same flow the "Receive" button's own "New Address" control uses: open the Receive dialog (`open_receive_dialog`) and queue a Core address request through it (`queue_core_address_request`), instead of firing the backend task directly. The new address now lands in a dialog that's actually on screen and gets selected/highlighted when it arrives.

Removed the now-dead `add_receiving_address` helper (its only caller was the button being fixed here).

## Testing
- Added `add_receiving_address_button_opens_receive_dialog` (kittest) — clicks the button and asserts the Receive dialog is visibly open, matching the pattern of the existing `receive_dialog_stays_open_on_triggering_click` test. Strengthened per code review to also assert on the `"Generating a new address…"` status label (only set when `queue_core_address_request` actually fires), not just dialog-open chrome — a regression that reopened the dialog but silently dropped the address request would previously have still passed.
- `cargo test --test kittest --all-features add_receiving_address_button_opens_receive_dialog` — pass.
- `cargo clippy --all-features --bin dash-evo-tool -- -D warnings` — clean.

## Breaking changes
None.

## Checklist
- [x] Tests added/updated
- [x] `cargo fmt --all`
- [ ] User-facing docs (not needed — internal UI wiring fix, no new user-facing surface)

## Attribution

### DIFF
--- a/src/ui/wallets/wallets_screen/mod.rs
+++ b/src/ui/wallets/wallets_screen/mod.rs
@@ -566,33 +566,6 @@ impl WalletsBalancesScreen {
         self.cached_tx_source_len = None;
     }
 
-    /// Request a fresh receive address through the SPV-watched upstream pool.
-    ///
-    /// Routes through the [`GenerateReceiveAddress`](crate::backend_task::wallet::WalletTask::GenerateReceiveAddress)
-    /// backend task (→ `next_receive_address` → upstream `next_unused`) so the
-    /// returned address is always inside the gap-limit window SPV monitors.
-    /// Deriving DET-side here would hand out addresses past the watched window
-    /// and lose deposits sent to them.
-    fn add_receiving_address(&mut self) -> AppAction {
-        let Some(seed_hash) = self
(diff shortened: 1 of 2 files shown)


thepastaclaw:
🕓 Ready for review — 38 ahead in queue (commit a94c7a34)
_Queue position: 39/39_

PR #915, opened 2026-07-21
fix(platform-errors): user-friendly message for duplicate DPNS name

Claudius-Maginificent:
## Why this PR exists
- **Problem**: Registering a DPNS username that's already taken shows the user a raw Rust `Debug`-formatted internal error instead of an actionable message.
- **What breaks without it**: Try to register a DPNS name someone else already holds. The state-transition broadcast fails with a `StateError::DuplicateUniqueIndexError` consensus error (code 40105), but `TaskError`'s classifier has no arm for that variant, so it falls through to the generic `PlatformRejected` catch-all: *"The platform rejected this request. Please check your input and try again."* — no indication the name is taken, no path to self-resolve. In the field this has surfaced as the literal Rust struct dump reaching the banner (`PlatformRejected { source_error: StateTransitionBroadcastError { code: 40105, ... DuplicateUniqueIndexError { ... duplicating_properties: ["normalizedParentDomainName", "normalizedLabel"] } ... } }`).

## What was done
Added a typed classification arm for `ConsensusError::StateError(StateError::DuplicateUniqueIndexError(e))` in `TaskError::from<SdkError>`'s mapper (`src/backend_task/error.rs`):
- `TaskError::DpnsUsernameAlreadyTaken` — message: *"This username is already taken. Please choose a different username and try again."*
- `TaskError::PlatformEntryConflict` — generic fallback for any other duplicate-unique-index case, so those no longer regress to the old vague message either; message: *"This request conflicts with an existing entry. Please use different values and try again."*

Both preserve the original `Box<SdkError>` as a typed `#[source]`, matching the convention used by sibling variants. Detection matches only on the typed `duplicating_properties()` accessor — no string parsing of error text, per this repo's error-handling conventions.

**Post-review follow-up**: the shared `From<SdkError>` boundary is reached by any `DuplicateUniqueIndexError`, not just DPNS registration — `src/backend_task/document.rs` funnels the same generic document/document-action task path (wired to the Contracts & Documents UI, over arbitrary caller-supplied contracts) through this exact conversion. Deciding the DPNS-specific message purely from property-name content at that shared boundary meant a non-DPNS contract that happens to define a unique index over identically-named fields would incorrectly get the DPNS-branded message. Fixed by making the shared classifier always return the generic `PlatformEntryConflict`, and relocating the DPNS-specific rebrand to the one call site that actually knows it's registering a DPNS domain (`src/backend_task/identity/register_dpns_name.rs`) — applied only to the domain-document put, not the preorder put.

## Testing
- `from_sdk_error_duplicate_unique_index_dpns_named_fields_is_generic` / `..._other_document_is_actionable` / `..._boundary_property_counts_are_generic` (`src/backend_task/error.rs`) — the shared classifier stays generic for every `DuplicateUniqueIndexError`, including 0/1-property boundary cases.
- `rebrand_dpns_domain_conflict_maps_platform_entry_conflict` / `..._passes_through_other_errors` (`src/backend_task/identity/register_dpns_name.rs`) — the relocated rebrand function correctly upgrades a DPNS domain conflict and passes every other error through unchanged.
- `cargo test --all-features from_sdk_error_duplicate_unique_index` and `rebrand_dpns_domain_conflict` — all 5 tests pass by name.
- `cargo clippy --all-features --bin dash-evo-tool -- -D warnings` — clean.

## Breaking changes
None.

## Checklist
- [x] Tests added/updated
- [x] `cargo fmt --all`
- [ ] User-facing docs (not needed — error-message copy only)

## Attribution


lklimek:
@coderabbitai review

thepastaclaw:
🕓 Ready for review — 48 ahead in queue (commit 36ecc38d)
_Queue position: 49/49_

PR #916, opened 2026-07-21
fix(ui): clear startup transient error banners once network/migration is ready

Claudius-Maginificent:
**TL;DR:** Keep the "storage still updating" banner visible until storage is genuinely free again, instead of letting it disappear about a frame too early.

## User story
As a user who just finished the app's first-run migration, I want the "storage update is still running" message to stay up until storage is truly available, so a retried action that fails during that window isn't left unexplained.

### Base flow
The app finishes cold-start migration; the user immediately retries a wallet-touching action.

### Actual behavior
Two error banners could persist indefinitely during app cold start ("storage update is still running" and a masternode-list-not-yet-synced error) — even minutes after the underlying condition fully resolved. The initial fix for that made the storage banner reactive, but it could still clear about one frame *before* storage was actually free: a background best-effort node-list refresh kept holding the lock past the point the migration state reported "ready." A retry in that narrow window failed again with the banner already gone, looking like an unexplained glitch.

### Expected behavior
Both banners clear reactively once their condition is genuinely resolved, and the storage banner specifically stays up until storage is truly free — so a failed retry in that window has a visible, honest explanation instead of none.

### What was done
- Added a dedicated `TaskError::MasternodeListNotReady` variant with a clean, actionable message instead of falling through to the generic, opaque `SdkError` text.
- `ConnectionBanner`/`MigrationReconciler` (`app/reconcilers.rs`) now adopt these banner handles and clear them reactively once their readiness condition genuinely holds, instead of leaving them `disable_auto_dismiss()`'d forever.
- **Review fix**: `storage_guard_resolved` was computed from `MigrationState` alone, so it could flip true (and clear the banner) before the `migration_run` mutex the error actually reports on was released — a detached best-effort DAPI-node refresh keeps holding it past `MigrationState` going `Ready`. Gated the clear on both the terminal state AND a non-blocking `migration_run.try_lock()` probe.
- Extracted the repeated adopt/reset/conditional-clear banner-handle pattern (shared by `ConnectionBanner` and `MigrationReconciler`) into a small `TransientBanner` newtype — same behavior, less duplication.
- Hardened the SDK error-string sentinel test to pin the literal upstream string shape (instead of round-tripping the same constant through itself), and cross-referenced the producer/consumer sites of that string so the coupling between `context_provider.rs` and `backend_task/error.rs` is discoverable.
- Addresses `claudius:grumpy-review` findings CODE-001 (blocking), CODE-003 and PROJ-001 (non-blocking). CODE-002 (a pre-existing FIFO-eviction edge case in the "stuck offline forever" scenario) is a known, accepted follow-up — not fixed here.

### Testing
- New/updated reconciler + error-mapping tests (banner-clears-when-ready, plus the new guard/lock-aware clearing) — all green via the narrowest relevant `--lib` scope.
- `cargo clippy --bin dash-evo-tool --all-features -- -D warnings` clean, `cargo fmt --all` clean.
- Independently re-verified by qa-engineer-marvin during grumpy-review consolidation.

### Breaking changes
None.

### Checklist
- [x] Tests added (TDD)
- [x] User-facing strings are plain, complete sentences
- [x] No secrets in diff

### Prior work
- Addresses findings from the combined grumpy-review of PR #916/#917/#918.

### Attribution

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2350,6 +2350,15 @@ impl App for AppState {
                         // TaskError Debug output is shown to users, deliberately.
                         // Ensure inner error types don't expose secrets.
                         handle.with_details(&err);
+                        match &err {
+                            TaskError::WalletStorageNotReady => {
+                                self.migration.track_storage_startup_error(handle);
+                            }
+                            TaskError::MasternodeListNotReady { .. } => {
+                                self.connection_banner.track_quorum_startup_error(handle);
+                            }
+                            _ => {}
+                        }
                         if !is_database_clear {
                             self.visible_screen_mut()
                                 .display_message(&msg, MessageType::Error);
--- a/src/app/reconcilers.rs
+++ b/src/app/reconcilers.rs
@@ -226,6 +226,27 @@ impl SpvBlockReconciler {
     }
 }
 
+#[derive(Default)]
+struct TransientBanner(Option<BannerHandle>);
+
+impl TransientBanner {
+    fn track(&mut self, handle: BannerHandle) {
+        self.0 = Some(handle);
+    }
+
+    fn reset(&mut self) {
+        if let Some(handle) = self.0.take() {
+            handle.clear();
+        }
+    }
+
+    fn clear_if(&mut self, condition: bool) {
+        if condition {
+            self.reset();
+        }
+    }
+}
+
 /// Reconciles the connection-status banner with the overall connection state.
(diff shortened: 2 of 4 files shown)


thepastaclaw:
🕓 Ready for review — 39 ahead in queue (commit 841b70d9)
_Queue position: 40/40_

lklimek (approved):

PR #917, opened 2026-07-21
fix(ui): pointer cursor + tooltips on nav; indicate wallet-less masternodes

Claudius-Maginificent:
## Why this PR exists
- **Problem**: Main navigation items give no visual affordance that they're clickable, and a masternode/evonode whose keys don't belong to any loaded wallet was shown with an arbitrary wallet's name instead of indicating it has none.
- **What breaks without it**: Hovering the "Identities" label next to the nav icon shows no pointer cursor and clicking it does nothing — there's no way to tell it's interactive. Separately, opening a masternode with no owning wallet shows some unrelated wallet's name as if it controlled that masternode's keys, which is actively misleading.

## What was done
- Root cause (traced, not assumed): egui 0.35's `Button` sets no hover cursor by default in this app (no global `interact_cursor` override configured). Applied the existing `ResponseExt::clickable_tooltip` convention (cursor + tooltip) consistently across all nav rows, including the previously-inert text label beside the icon and the no-texture fallback branches.
- Added `WalletAssociation::{InWallet,NotInWallet}` (`model/wallet_association.rs`) — a pure, testable decision function — and a muted "Not in a wallet" pill (distinct styling from a real wallet name) with an explanatory tooltip, replacing the previous silent fallback to `wallets.first()`.

## Testing
- 7 new tests (lib unit + kittest) — all PASS via the narrowest relevant scope, per ledger logs.
- Targeted clippy clean, `cargo fmt --all` clean.

## Breaking changes
None.

## Checklist
- [x] Tests added
- [x] i18n-ready strings, no jargon
- [x] No secrets in diff

## Attribution


thepastaclaw:
🕓 Ready for review — 35 ahead in queue (commit 063b3b71)
_Queue position: 36/37_

PR #922, opened 2026-07-21
test(wallets): make the address-generation assertion genuinely isolate the regression

Claudius-Maginificent:
## Why this PR exists
- **Problem**: PR #920's regression test for the "Add Receiving Address" button claimed to catch a specific regression, but didn't.
- **What breaks without it**: A code review (round 2 of the grumpy-review on the original fix) proved the claim false: `wallet_screen_harness()` never registers its fixture wallet with `WalletBackend`, so `load_core_addresses_for_receive`'s self-heal branch fires the exact same `"Generating a new address…"` status regardless of whether the button handler's own explicit `queue_core_address_request` call exists. Removing that explicit call and re-running the test still passed — the assertion added in #920 to guard against exactly that regression didn't actually guard against it.

## What was done
Added `registered_wallet_screen_harness()`, used only by `add_receiving_address_button_opens_receive_dialog`: registers the fixture wallet with the backend and waits (bounded 5s timeout) for a genuine non-empty monitored-address snapshot before the click, so `load_core_addresses_for_receive` takes its non-self-heal branch and the status can only originate from the button handler's own explicit call. The other three dialog tests are untouched, still using the original in-memory fixture — no risk to their coverage.

Also trims an 8-line tombstone comment in the button handler (narrating removed-code history that already lives in the original PR's commit message) down to present-state rationale only.

## Testing
- **Acceptance criterion, verified directly**: removed the explicit `queue_core_address_request` call from the button handler — test now fails (`FAILED`, panics on the missing status assertion). Restored the call — test passes again.
- `cargo test --test kittest --all-features wallets_screen` — full module, 7/7 pass by name, no regressions in the other three dialog tests.
- `cargo clippy --all-features --bin dash-evo-tool -- -D warnings` and `--test kittest -- -D warnings` — clean.

## Breaking changes
None.

## Checklist
- [x] Tests added/updated
- [x] `cargo fmt --all`
- [ ] User-facing docs (not needed — test-only fix)

## Attribution

## Summary by CodeRabbit

* **Tests**
  * Improved automated coverage for the Wallets screen’s “Add Receiving Address” interaction.
  * Added test setup for registered wallets with monitored receiving addresses, ensuring the receive dialog flow is validated in a realistic state.
* **Documentation**
  * Clarified the inline explanation of how adding a receiving address mirrors the existing Receive and New Address flow.

No user-facing behavior changes were introduced.

### DIFF
--- a/src/ui/wallets/wallets_screen/mod.rs
+++ b/src/ui/wallets/wallets_screen/mod.rs
@@ -799,14 +799,8 @@ impl WalletsBalancesScreen {
                     .button(RichText::new("➕ Add Receiving Address").size(14.0))
                     .clicked()
                 {
-                    // Same flow as the "Receive" button above + its "New Address"
-                    // control: open the Receive dialog and queue a fresh Core
-                    // address request through it, so the newly generated address
-                    // is actually visible to the user (see `queue_core_address_request`
-                    // / `open_receive_dialog`). Previously this issued the backend
-                    // task directly without opening the dialog, so the new address
-                    // landed silently in `receive_dialog.core_addresses` with
-                    // nothing on screen to show for the click.
+                    // Mirrors the "Receive" + "New Address" flow: open the dialog,
+                    // then queue a fresh Core address request so the result is visible.
                     if let Some(wallet) = self.selected_wallet.clone() {
                         action |= self.open_receive_dialog(ui.ctx());
                         self.queue_core_address_request(&wallet);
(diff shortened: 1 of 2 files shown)


thepastaclaw:
🕓 Ready for review — 44 ahead in queue (commit 11a232cd)
_Queue position: 45/72 · 2 reviews active_
_ETA: start ~06:16 UTC · complete ~06:44 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 12h 18m ago · Last checked: 2026-07-22 19:40 UTC_

PR #926, opened 2026-07-22
fix(logging): install SA_ONSTACK + alternate signal stack for fatal handler

Claudius-Maginificent:
## Summary

- Installs a per-thread alternate signal stack (`sigaltstack(2)`, 64 KiB) and adds `SA_ONSTACK` to the fatal-signal handler (`SIGSEGV`/`SIGABRT`/`SIGBUS`/`SIGILL`/`SIGFPE`) so a guard-page **stack-overflow** SIGSEGV can still run the handler and leave a crash marker — today it can't, because there's no stack space left to enter a handler that has no alternate stack.
- `sigaltstack` is per-thread, not process-wide. Covers the main thread directly from `install_fatal_signal_handler()`, and every Tokio worker/blocking thread via `Builder::on_thread_start` in `src/main.rs`.
- `nix` 0.31.3 (pinned) doesn't expose a safe `sigaltstack` wrapper, so this goes through raw `nix::libc::sigaltstack`/`stack_t`, matching this file's existing `nix::libc` FFI convention (`dup`/`dup2`/`write`).

## Context

Found while investigating a SIGSEGV report (core → Platform address funding, proof verification in `grovedb`'s `Op::decode`). `det-stderr.log` was completely empty after the crash — no marker, nothing. Root cause of *that* silence: `install_fatal_signal_handler` was replacing Rust's own default stack-overflow handler (the one that normally prints `"thread '...' has overflowed its stack"`) with a marker-writing handler that had no alternate stack — so on a true stack-overflow SIGSEGV, neither Rust's message nor our own marker could ever be written. This PR closes that gap.

This is a diagnostics fix, not a fix for the SIGSEGV itself — it ensures that *if* a stack-overflow-class crash happens again, `det-stderr.log` will actually say so instead of staying silent. (Separately, production's Tokio runtime has no `.thread_stack_size()` override, so worker threads run on Rust's 2 MiB default while this project's own backend-e2e harness documents a 16 MB floor for the SDK's proof-verification recursion — a real gap, tracked separately.)

## Test plan

- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test --lib logging:: --all-features` — 13/13 passed, including new `install_alt_signal_stack_registers_stack_for_calling_thread` (registers the stack, reads it back via `sigaltstack(NULL, &old)`, asserts `SS_DISABLE` unset and size matches)
- [x] `cargo clippy --bin dash-evo-tool --all-features -- -D warnings` — clean
- [ ] CI (full suite + clippy) on this PR

Co-Authored-By: Codex Sol <noreply@openai.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fatal-signal handling by using a dedicated alternate signal stack.
  * Applied the safeguard consistently across runtime worker and blocking threads.
  * Preserved existing crash-reporting behavior while improving reliability during stack-related failures.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -219,6 +219,16 @@ fn redirect_stderr_to(_path: &Path) {
     );
 }
 
+/// Size of the alternate signal stack used by [`install_alt_signal_stack`].
+///
+/// The fatal-signal handler itself does a single `write(2)` call with
+/// nothing beneath it, but the kernel/libc signal-delivery machinery needs
+/// some stack space of its own to build the signal frame before the handler
+/// runs. 64 KiB is a generous multiple of `libc::SIGSTKSZ` (8 KiB on this
+/// target) for what is a one-time, process-lifetime allocation per thread.
+#[cfg(unix)]
+const ALT_SIGNAL_STACK_SIZE: usize = 64 * 1024;
+
 /// Installs a handler for the synchronous fatal signals (`SIGSEGV`, `SIGABRT`,
 /// `SIGBUS`, `SIGILL`, `SIGFPE`) that writes a one-line marker to stderr before
 /// the process dies, then lets the default handler take over.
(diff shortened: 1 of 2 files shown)


thepastaclaw:
🕓 Ready for review — 55 ahead in queue (commit 14c5ecb6)
_Queue position: 56/72 · 2 reviews active_
_ETA: start ~08:53 UTC · complete ~09:21 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 4h 43m ago · Last checked: 2026-07-22 19:40 UTC_

Claudius-Maginificent:
Fixed in 14c5ecb6 — switched `install_alt_signal_stack_impl` from `Box::leak` to thread-local storage (`thread_local! { static ALT_SIGNAL_STACK_BUF: RefCell<Option<Box<[u8]>>> }`), freed by the TLS destructor at thread exit instead of living for the process. Good catch on the blocking-pool churn.

PR #934, opened 2026-07-24
fix: app review batch — identity actions, native crash, asset-lock errors

Claudius-Maginificent:
**TL;DR:** Three fixes from an ongoing app-review session: duplicate/confusing Identity Home action buttons collapsed into one clear row, a native crash during deposit verification fixed, and deposits that can never be reused now say so plainly instead of telling the user to retry. A follow-up review round then caught (and fixed) a real bug in round 1's own withdrawal-eligibility fix, plus a rejection-message regression and a test-coverage gap.

## User story
As a Dash Evo Tool user funding an identity, I want the action buttons on my Identity Home screen to each do something different, I want the app to not crash while my deposit is being verified, and I want a deposit that's already been used to tell me that plainly — not ask me to retry something that can never succeed.

### Base flow
A user manages an identity from the Identity Home screen, funds it via a deposit (asset lock), and occasionally hits a network-side rejection during that process.

### Actual behavior
1. Identity Home showed six action buttons where two pairs opened the exact same screen (`Send`/`Send to another identity` both → Transfer; `Receive`/`Add funds` both → Top Up), rendered at inconsistent visual weights.
2. The app could crash (`SIGSEGV`) during deposit broadcast/verification — root-caused to an undersized default Tokio worker-thread stack overflowing during proof/quorum-signature verification.
3. A deposit already consumed by Platform (e.g. because the app crashed before recording that success locally) produced a generic "Please try again in a moment" banner — misleading, since that failure is permanent, not transient.
4. (Found reviewing round 1's own fix) A Developer-role user whose identity has an on-chain key but no locally-held TRANSFER/OWNER key got an *enabled* `Send to wallet` button that led to a blank withdrawal screen or a signing error — the button lied about what the screen could actually do.

### Expected behavior
1. One row of four clearly distinct, correctly labeled buttons: Add funds, Send to wallet, Send to another identity, Add contact.
2. The app no longer crashes during deposit verification.
3. A permanently-unusable deposit says so directly: *"This deposit has already been used and cannot be used again. Choose a different deposit, or start a new deposit."*
4. `Send to wallet` is enabled only when a real, locally-signable key exists — for every role — matching what the withdrawal backend actually accepts.

### What was done

**1. Identity Home action-row redesign** (`398ee3f4`, `2df2d829`, plus the design note `b94aeec3`/`bc59de0d`)
- Removed the duplicate `Send`/`Receive` `HomeButton` variants; one row of four buttons (`Add funds` primary, `Send to wallet` / `Send to another identity` / `Add contact` secondary), rendered via `StyledButton`/`ComponentStyles`.
- `Send to wallet` gated on `available_withdrawal_keys()`, matching the existing `Add contact` gating pattern.
- Replaced the stale regression test (which pinned the duplicate routing as intended) with an injectivity check over the action row's destinations.
- Renamed "Top Up Identity" → "Add funds" consistently across the funding wizard's breadcrumb, heading, sub-heading, and all CTA buttons; rewrote several jargon/ASCII-decorated status strings per the project's error-message/i18n rules.
- Fixed a duplicate ID in `docs/user-stories.md` and added `[Gap]` entries for the unimplemented payment-sheet Send/Receive affordances that originally caused the duplication.
- Full UX rationale: `docs/ai-design/2026-07-24-identity-home-actions/README.md`.

**2. Native crash fix** (`659e32e4`, `9df18fea`)
- Root-caused via log analysis: the app's Tokio runtime used the default worker-thread stack size, which could overflow during deposit proof/quorum-signature verification, producing a silent `SIGSEGV` (confirmed via the app's own crash-stderr capture mechanism) — the process died mid-verify with no panic, no error, just a restart 7–16 seconds later. This pattern recurred repeatedly during testing.
- Fix: explicit 4 MiB worker-thread stack size (`SDK_THREAD_STACK_SIZE`), applied consistently across the main GUI runtime and all `det_cli` binary entry points, replacing the previous default (undersized) stack.
- Verified live: post-fix, the same deposit-verification path that previously crashed now completes without incident.

**3. Non-retryable error message for already-consumed deposits** (`38f4f97d`)
- Added `TaskError::AssetLockOutPointAlreadyConsumed`, routed via the existing SDK-error classifier when a deposit's asset lock has already been consumed by Platform (`IdentityAssetLockTransactionOutPointAlreadyConsumedError`), for both identity registration and top-up.
- Wired into the actual call path (`map_identity_register_error`/`map_identity_top_up_error`), which previously bucketed every SDK rejection into a single generic, retry-implying message.
- **Note on scope**: the underlying reason a deposit can get stuck in this state (its local "already used" status is only ever recorded on the *success* path of the funding flow, so a crash or interruption anywhere in between leaves it permanently offered as usable again, and no client-side resync can detect Platform-side consumption after the fact) is an upstream `platform-wallet` gap, not fixable in this repo — filed as [dashpay/platform#4227](https://github.com/dashpay/platform/issues/4227). This PR only fixes the *message*; the stuck state itself will persist until the upstream issue is addressed.
- A related third call site (`map_platform_address_fund_error`, platform-address funding from an existing deposit) had the identical gap, found during live testing after this PR's initial commits landed — fixed in `65d28226`, reusing the same `AssetLockOutPointAlreadyConsumed` variant/message rather than introducing a duplicate.

**4. Reviewer findings addressed — round 1** (`978bb632`, `05647971`)
- CodeRabbit: clarified an ambiguous CHANGELOG line ("it opens" read as referring to the wrong button).
- Automated review (Medium): Identity Home's `Send to wallet` gate (`available_withdrawal_keys().is_empty()`) was stricter than `WithdrawalScreen`'s own role-aware check, wrongly disabling the button (with a factually-wrong tooltip) for a Developer-role user whose identity has an on-chain key but no locally-held TRANSFER/OWNER key — a real regression versus pre-PR behavior, and previously untested. Fixed by hoisting a single `QualifiedIdentity::can_attempt_withdrawal(role)` predicate that Home, `WithdrawalScreen`, and the Identities-list "Withdraw" popup (a third, previously balance-only gate, now reconciled too) all call, so the three surfaces can no longer disagree.

**5. Reviewer findings addressed — round 2** (`ea455bf1`, `6fbfb3c2`, `18110b0b`)

A fresh review pass on round 1's own commits caught a real bug in the fix itself, plus two smaller gaps:

- **Blocking (thepastaclaw + CodeRabbit, independently):** round 1's `can_attempt_withdrawal(role)` predicate had a Developer-only carve-out treating *any* on-chain public key as withdrawal-eligible. Nothing downstream honors that — the withdrawal backend task (`resolve_withdrawal_signing_key`) unconditionally requires a **locally-held** TRANSFER/OWNER key, for every role; there is no signing override wired into the withdrawal path. So a Developer-role identity with only an authentication key got an *enabled* button leading to a blank `WithdrawalScreen` (`selected_key` stayed `None`) or a signing error — never a working withdrawal. Fixed by making the predicate uniform across every role and removing the now-dead Developer-only key-selection fallback in `WithdrawalScreen::new`.
- **Blocking (thepastaclaw):** the same round's SDK-error fix in `wallet_backend` (`38f4f97d`/`65d28226`) routed *every* `PlatformWalletError::Sdk` variant through the generic classifier to catch the already-consumed case, but that also caught an ordinary *unmapped* broadcast rejection, losing its funding-specific recovery guidance (retained funding lock, affected identity, existing-lock instructions) in favor of a generic "platform rejected this" message. Fixed with a small helper (`classify_sdk_error_or`) that only falls back to the generic path when the richer SDK classifier truly has nothing better.
- **Test coverage (thepastaclaw + CodeRabbit):** the Identity Home kittest suite queried only `Send to wallet`'s disabled/enabled state and discarded `home::render`'s `(AppAction, HomeOutcome)` result — a button rendered twice, omitted, or wired to the wrong destination would have passed every test. Added coverage asserting the full four-button row renders exactly once, and that each enabled click routes to its intended screen/outcome.
- **Changelog (thepastaclaw):** the Unreleased section covered the Identity Home redesign but omitted the native-crash stack-size fix and the consumed-deposit guidance change (fix #2 and #3 above). Added both entries.
- All six review threads from this round verified against current code and resolved.

### Branch sync
Merged `v1.0-dev` (`18221698`) to pick up the upstream `platform`/`platform-wallet` pin bump to PR #3968 tip (seedless rehydration, new SQLite persistence backend — unrelated to this PR's scope). One textual conflict: both branches added new test functions at the same anchor point in `wallet_backend`'s test module (pure add/add, no logic overlap) — resolved by keeping both. Full build, `cargo test --lib`/`--test kittest` (all-features), `cargo fmt --check`, and `cargo clippy --all-features --all-targets -- -D warnings` all verified clean post-merge.

### Testing
- `cargo test --lib --all-features` — 2135/2135 passing (post-merge; includes all changes from both rounds and the upstream sync).
- `cargo test --test kittest --all-features` — 286/286 passing, including `identity_home::` (8 tests: disabled/enabled states across every role, full four-button-row assertion, and click-routing for all four buttons) and `withdraw_screen::` (6 tests, no regression from the predicate/fallback cleanup).
- `wallet_backend::tests::map_{identity_register,identity_top_up,platform_address_fund}_error_falls_back_for_unmapped_broadcast_rejection` — new round-2 regression tests, all passing.
- `model::qualified_identity::withdrawal_key_tests::can_attempt_withdrawal_{ignores_on_chain_only_keys,uses_local_withdrawal_keys}_for_every_role` — new round-2 unit tests covering every `UserRole`, all passing.
- `cargo clippy --all-features --all-targets -- -D warnings` — clean (matches CI's exact gate).
- Crash fix verified live against the real app on testnet: the previously-crashing deposit-verification path now completes without a `SIGSEGV`.

### Breaking changes
None — UI wording/layout, thread-runtime configuration, and error-message classification only; no persisted data, public API, or consensus-affecting change.

### Checklist
- [x] `cargo fmt --all`
- [x] Relevant tests updated and passing
- [x] `CHANGELOG.md` updated

### Prior work
- [dashpay/platform#4227](https://github.com/dashpay/platform/issues/4227) — tracks the upstream root cause behind fix #3 above (why a deposit's local state doesn't self-correct).
- [dashpay/platform#4011](https://github.com/dashpay/platform/pull/4011) — prior, partial fix for an adjacent case of the same symptom family; referenced in #4227.

This PR aggregates everything found so far in an ongoing app-review session; further unrelated fixes from the same review may still be added to this branch.

### Attribution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   identity`, `Receive`/`Add funds`). It's now one row of four: **Add funds**,
   **Send to wallet**, **Send to another identity**, and **Add contact**.
   **Send to wallet** is disabled with an explanation when the identity has no
-  withdrawal key loaded. The funding screen it opens, and its wording
+  withdrawal key loaded. The **Add funds** screen, and its wording
   throughout, now consistently says "Add funds" instead of "Top Up Identity",
   and its step-by-step deposit messages no longer use technical wording.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   throughout, now consistently says "Add funds" instead of "Top Up Identity",
   and its step-by-step deposit messages no longer use technical wording.
 
+- **Fewer native crashes while verifying a deposit**: background worker
+  threads now get a larger stack, fixing a crash that could occur during
+  deposit verification.
+
+- **Clearer guidance for an already-used deposit**: registering an identity,
+  topping one up, or funding a platform address with a deposit that was
+  already consumed by another operation now tells you directly to choose a
+  different deposit or start a new one, instead of the generic rejection
+  message that suggested retrying the same one.
+
 ### Changed
 
 - **Shielded transactions are available on supported networks**: sending,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   does not support shielded sending from when the current interface mode does
   not unlock it.
 
+- **Identity Home actions simplified**: the action row previously had six
+  buttons — several of which opened the same screen (`Send`/`Send to another
+  identity`, `Receive`/`Add funds`). It's now one row of four: **Add funds**,
+  **Send to wallet**, **Send to another identity**, and **Add contact**.
+  **Send to wallet** is disabled with an explanation when the identity has no
+  withdrawal key loaded. The funding screen it opens, and its wording
+  throughout, now consistently says "Add funds" instead of "Top Up Identity",
+  and its step-by-step deposit messages no longer use technical wording.
+
 ### Changed
 
 - **Shielded transactions are available on supported networks**: sending,
--- a/src/model/qualified_identity/mod.rs
+++ b/src/model/qualified_identity/mod.rs
@@ -9,6 +9,7 @@ pub mod qualified_identity_public_key;
 use crate::backend_task::error::TaskError;
 use crate::model::qualified_identity::encrypted_key_storage::{KeyStorage, ResolvedPrivateKey};
 use crate::model::qualified_identity::qualified_identity_public_key::QualifiedIdentityPublicKey;
+use crate::model::user_role::UserRole;
 use crate::model::wallet::{Wallet, WalletSeedHash};
 use bincode::{Decode, Encode};
 use dash_sdk::dashcore_rpc::dashcore::{PubkeyHash, signer};
@@ -886,6 +887,15 @@ impl QualifiedIdentity {
         keys
     }
 
+    /// Whether the active role can attempt a withdrawal with this identity.
+    pub fn can_attempt_withdrawal(&self, role: UserRole) -> bool {
+        if role.at_least(UserRole::Developer) {
+            !self.identity.public_keys().is_empty()
+        } else {
+            !self.available_withdrawal_keys().is_empty()
+        }
+    }
+
     /// Returns the key to pre-select for signing a withdrawal.
     ///
     /// Only keys whose private material is held locally are considered (via
@@ -1343,6 +1353,28 @@ mod withdrawal_key_tests {
             .map(|qk| qk.identity_public_key.id())
             .collect();
         assert_eq!(ids, vec![1]);
+    }
+
+    #[test]
+    fn can_attempt_withdrawal_developer_uses_on_chain_keys() {
+        let without_keys = build_identity(IdentityType::User, vec![], vec![]);
+        assert!(!without_keys.can_attempt_withdrawal(UserRole::Developer));
+
+        let authentication = key(1, Purpose::AUTHENTICATION);
+        let on_chain_only = build_identity(IdentityType::User, vec![authentication], vec![]);
+        assert!(on_chain_only.can_attempt_withdrawal(UserRole::Developer));
+    }
+
+    #[test]
+    fn can_attempt_withdrawal_non_developer_uses_local_withdrawal_keys() {
+        let authentication = key(1, Purpose::AUTHENTICATION);
+        let on_chain_only = build_identity(IdentityType::User, vec![authentication], vec![]);
+        assert!(!on_chain_only.can_attempt_withdrawal(UserRole::Everyday));
+
+        let transfer = key(2, Purpose::TRANSFER);
+        let locally_signable =
+            build_identity(IdentityType::User, vec![transfer.clone()], vec![transfer]);
+        assert!(locally_signable.can_attempt_withdrawal(UserRole::Power));
     }
 
     /// A wholly disabled key set leaves no signable withdrawal key.
--- a/src/model/qualified_identity/mod.rs
+++ b/src/model/qualified_identity/mod.rs
@@ -888,12 +888,20 @@ impl QualifiedIdentity {
     }
 
     /// Whether the active role can attempt a withdrawal with this identity.
-    pub fn can_attempt_withdrawal(&self, role: UserRole) -> bool {
-        if role.at_least(UserRole::Developer) {
-            !self.identity.public_keys().is_empty()
-        } else {
-            !self.available_withdrawal_keys().is_empty()
-        }
+    ///
+    /// Uniform across every role: a withdrawal is only ever executable with a
+    /// locally-held TRANSFER or OWNER key (see
+    /// [`resolve_withdrawal_signing_key`](Self::resolve_withdrawal_signing_key),
+    /// the backend's unconditional enforcement layer — no role currently
+    /// relaxes it). A Developer-only "any on-chain key" carve-out previously
+    /// existed here and in [`default_withdrawal_key`](Self::default_withdrawal_key)'s
+    /// caller, but nothing downstream ever accepted an on-chain-only key for
+    /// signing, so that carve-out only produced an enabled button that led to
+    /// a blank or failing withdrawal screen. Gate on the same invariant the
+    /// backend enforces, so this predicate never lies about what the screen
+    /// can actually do.
+    pub fn can_attempt_withdrawal(&self, _role: UserRole) -> bool {
+        !self.available_withdrawal_keys().is_empty()
     }
 
     /// Returns the key to pre-select for signing a withdrawal.
@@ -1355,26 +1363,34 @@ mod withdrawal_key_tests {
         assert_eq!(ids, vec![1]);
     }
 
+    /// An on-chain-only key (no local private material) can never be used to
+    /// sign a withdrawal (see `resolve_withdrawal_signing_key`), for any
+    /// role — including Developer. Enabling the gate for it previously led
+    /// `WithdrawalScreen` to a blank or failing screen.
     #[test]
-    fn can_attempt_withdrawal_developer_uses_on_chain_keys() {
+    fn can_attempt_withdrawal_ignores_on_chain_only_keys_for_every_role() {
         let without_keys = build_identity(IdentityType::User, vec![], vec![]);
-        assert!(!without_keys.can_attempt_withdrawal(UserRole::Developer));
-
         let authentication = key(1, Purpose::AUTHENTICATION);
         let on_chain_only = build_identity(IdentityType::User, vec![authentication], vec![]);
-        assert!(on_chain_only.can_attempt_withdrawal(UserRole::Developer));
+
+        for role in [UserRole::Everyday, UserRole::Power, UserRole::Developer] {
+            assert!(!without_keys.can_attempt_withdrawal(role));
+            assert!(
+                !on_chain_only.can_attempt_withdrawal(role),
+                "on-chain-only AUTHENTICATION key must not enable withdrawal for {role:?}"
+            );
+        }
     }
 
     #[test]
-    fn can_attempt_withdrawal_non_developer_uses_local_withdrawal_keys() {
-        let authentication = key(1, Purpose::AUTHENTICATION);
-        let on_chain_only = build_identity(IdentityType::User, vec![authentication], vec![]);
-        assert!(!on_chain_only.can_attempt_withdrawal(UserRole::Everyday));
-
+    fn can_attempt_withdrawal_uses_local_withdrawal_keys_for_every_role() {
         let transfer = key(2, Purpose::TRANSFER);
         let locally_signable =
             build_identity(IdentityType::User, vec![transfer.clone()], vec![transfer]);
-        assert!(locally_signable.can_attempt_withdrawal(UserRole::Power));
+
+        for role in [UserRole::Everyday, UserRole::Power, UserRole::Developer] {
+            assert!(locally_signable.can_attempt_withdrawal(role));
+        }
     }
 
     /// A wholly disabled key set leaves no signable withdrawal key.
--- a/src/ui/identity/home.rs
+++ b/src/ui/identity/home.rs
@@ -7,18 +7,18 @@
 //!    identity-type badge, network pill. Two variants: social profile set
 //!    (`IdentityHeroCard` with display name) and no social profile (type-glyph
 //!    monogram + inline `Set up your social profile` card below the hero).
-//! 2. Quick-actions row: **Send**, **Receive**, **Add contact**. `Add contact`
-//!    is gated behind a social profile (see §B.3).
-//! 3. Secondary actions row (ghost buttons): `Add funds`, `Send to wallet`,
-//!    `Send to another identity`. All three visible for all personas (§B.2).
-//! 4. Onboarding checklist strip (until all three steps are complete or the
+//! 2. Actions row: **Add funds**, **Send to wallet**, **Send to another
+//!    identity**, **Add contact**. `Add contact` is gated behind a social
+//!    profile (see §B.3).
+//! 3. Onboarding checklist strip (until all three steps are complete or the
 //!    user dismisses it).
-//! 5. Recent activity preview (up to 5 rows), currently an empty-state
+//! 4. Recent activity preview (up to 5 rows), currently an empty-state
 //!    preview; the `See all activity` link routes to the Activity tab.
 //!    Richer content is parked until the activity aggregator lands.
-//! 6. Advanced details expander (raw Identity ID, revision, last updated).
+//! 5. Advanced details expander (raw Identity ID, revision, last updated).
 //!
-//! Strings are taken verbatim from §B.2 / §B.3 and the wording audit in §C.
+//! Strings follow the Identity Home action-set redesign and the wording audit
+//! in §C.
 //!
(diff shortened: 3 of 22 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 18221698)

src/wallet_backend/mod.rs
```diff
@@ -2813,9 +2813,16 @@ fn map_shielded_op_error(e: platform_wallet::error::PlatformWalletError) -> Task
...
 /// finality failures become `AssetLockFinalityTimeout`; everything else
-/// falls through to the generic `WalletBackend` wrapper. Structural match
+/// falls through to the generic `WalletBackend` wrapper. SDK errors use the
+/// richer `TaskError` classifier before this coarse mapping. Structural match
 /// — never parses error strings.
 fn map_identity_register_error(e: platform_wallet::error::PlatformWalletError) -> TaskError {
+    let e = match e {
+        platform_wallet::error::PlatformWalletError::Sdk(sdk_error) => {
+            return TaskError::from(sdk_error);
+        }
```
thepastaclaw, line 2932: **🔴 Blocking: Preserve funding-flow guidance for unclassified SDK rejections**

The unconditional early return sends every `PlatformWalletError::Sdk` through the global `From<SdkError>` classifier. Specialized classifications such as `AssetLockOutPointAlreadyConsumed` and insufficient balance are appropriate, but an ordinary `StateTransitionBroadcastError` with an unmapped consensus cause now becomes `PlatformRejected` instead of the operation-specific fallback. Registration consequently loses the instruction to start a new identity using the retained funding lock; top-up loses the affected identity context; and the identical early return at lines 2902-2905 makes platform-address funding lose its explicit existing-lock recovery instructions. Upstream retains the tracked lock whenever submission returns an error, and the existing `IdentityCreateRejected` and `PlatformAddressFundRejected` messages and tests encode that recovery invariant. Preserve dedicated global classifications when available, but fall back to the corresponding funding-operation envelope for generic or unclassified SDK rejections. Add regression tests using an unmapped SDK broadcast rejection for all three mappers.

<sub>source: ['codex']</sub>

CHANGELOG.md
```diff
@@ -57,6 +57,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
   not unlock it.
 
+- **Identity Home actions simplified**: the action row previously had six
+  buttons — several of which opened the same screen (`Send`/`Send to another
+  identity`, `Receive`/`Add funds`). It's now one row of four: **Add funds**,
+  **Send to wallet**, **Send to another identity**, and **Add contact**.
+  **Send to wallet** is disabled with an explanation when the identity has no
+  withdrawal key loaded. The funding screen it opens, and its wording
+  throughout, now consistently says "Add funds" instead of "Top Up Identity",
+  and its step-by-step deposit messages no longer use technical wording.
```
thepastaclaw, line 67: **🟡 Suggestion: Document the crash and consumed-deposit fixes**

The PR explicitly aggregates three user-visible fixes, but the changelog records only the Identity Home redesign. It omits both the larger Tokio worker stack that prevents a native crash during deposit verification and the new non-retryable guidance for already-consumed deposits. Add separate `Fixed` entries for those two reliability improvements so the changelog matches the PR scope and its completed changelog checklist item.

<sub>source: ['codex']</sub>

src/ui/identity/home.rs
```diff
@@ -903,30 +825,25 @@ mod tests {
...
+            HomeButtonKind::Outcome(outcome) => {
+                panic!("action-row button {button:?} produced hub outcome {outcome:?}")
+            }
+        });
+
+        for (index, destination) in destinations.iter().enumerate() {
+            for other in &destinations[index + 1..] {
+                assert_ne!(destination, other);
+            }
+        }
```
thepastaclaw, line 847: **🟡 Suggestion: Test the rendered Identity Home action row**

The replacement regression test exercises only the pure dispatcher. It would still pass if `render` omitted an action, rendered a duplicate label, wired the wrong `HomeButton` at a click site, or failed to disable `Send to wallet` when `available_withdrawal_keys()` is empty. This PR's primary UI change is the rendered four-button row and its new withdrawal-key gate, and the project expects egui behavior to be covered in `tests/kittest/`. Add a populated Identity Home kittest that verifies the four labels and routing, plus identities with and without a usable withdrawal key to verify the enabled and disabled `Send to wallet` states.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The Identity Home redesign, runtime stack configuration, and consumed-deposit classification are otherwise coherent. However, routing every funding-flow SDK error directly through the global classifier regresses operation-specific recovery guidance for ordinary broadcast rejections, so the PR needs a fallback that preserves funding context. The changelog and rendered UI coverage should also be completed for the other user-visible fixes and the new action-row behavior.

Source: reviewers gpt-5.6-sol; final verifier gpt-5.6-sol

### Review provenance

<details>

```

- [BLOCKING] src/wallet_backend/mod.rs:2820-2823: Preserve funding-flow guidance for unclassified SDK rejections
  The unconditional early return sends every `PlatformWalletError::Sdk` through the global `From<SdkError>` classifier. Specialized classifications such as `AssetLockOutPointAlreadyConsumed` and insufficient balance are appropriate, but an ordinary `StateTransitionBroadcastError` with an unmapped consensus cause now becomes `PlatformRejected` instead of the operation-specific fallback. Registration consequently loses the instruction to start a new identity using the retained funding lock; top-up loses the affected identity context; and the identical early return at lines 2902-2905 makes platform-address funding lose its explicit existing-lock recovery instructions. Upstream retains the tracked lock whenever submission returns an error, and the existing `IdentityCreateRejected` and `PlatformAddressFundRejected` messages and tests encode that recovery invariant. Preserve dedicated global classifications when available, but fall back to the corresponding funding-operation envelope for generic or unclassified SDK rejections. Add regression tests using an unmapped SDK broadcast rejection for all three mappers.

In `CHANGELOG.md`:
- [SUGGESTION] CHANGELOG.md:60-67: Document the crash and consumed-deposit fixes
  The PR explicitly aggregates three user-visible fixes, but the changelog records only the Identity Home redesign. It omits both the larger Tokio worker stack that prevents a native crash during deposit verification and the new non-retryable guidance for already-consumed deposits. Add separate `Fixed` entries for those two reliability improvements so the changelog matches the PR scope and its completed changelog checklist item.

In `src/ui/identity/home.rs`:
- [SUGGESTION] src/ui/identity/home.rs:829-846: Test the rendered Identity Home action row
  The replacement regression test exercises only the pure dispatcher. It would still pass if `render` omitted an action, rendered a duplicate label, wired the wrong `HomeButton` at a click site, or failed to disable `Send to wallet` when `available_withdrawal_keys()` is empty. This PR's primary UI change is the rendered four-button row and its new withdrawal-key gate, and the project expects egui behavior to be covered in `tests/kittest/`. Add a populated Identity Home kittest that verifies the four labels and routing, plus identities with and without a usable withdrawal key to verify the enabled and disabled `Send to wallet` states.
```
</details>

src/model/qualified_identity/mod.rs
```diff
@@ -886,6 +887,15 @@ impl QualifiedIdentity {
         keys
     }
 
+    /// Whether the active role can attempt a withdrawal with this identity.
+    pub fn can_attempt_withdrawal(&self, role: UserRole) -> bool {
+        if role.at_least(UserRole::Developer) {
+            !self.identity.public_keys().is_empty()
+        } else {
+            !self.available_withdrawal_keys().is_empty()
+        }
```
thepastaclaw: **🔴 Blocking: Do not enable withdrawals for keys the execution path cannot use**

The Developer branch returns true for any on-chain public key, including the authentication-only, private-key-free identity used by the new kittest. Home and the identities popup consequently enable Withdrawal, but `WithdrawalScreen::new` only falls back to an on-chain `TRANSFER` key; for the authentication-only case `selected_key` remains `None`, the shared predicate suppresses the no-key state, and rendering exits at `withdraw_screen.rs:560-562` without showing a form or guidance. An on-chain-only `TRANSFER` key is also unusable because `resolve_withdrawal_signing_key` accepts only active, locally held `TRANSFER` or `OWNER` keys, and the backend withdrawal passes `None` for transition options before invoking a signer that requires private material. Base this predicate on the same eligibility invariant as the backend, or update constructor selection, the chooser, and backend signing coherently if a Developer override is intended. Add a click-through test proving every enabled launcher reaches a usable form and submission path.

<sub>source: ['codex']</sub>

tests/kittest/identity_home.rs
```diff
@@ -0,0 +1,141 @@
...
+    app_context.set_selected_identity(Some(identity_id));
+}
+
+fn mount_home(app_context: Arc<AppContext>) -> Harness<'static, (HomeState, ProfileCache)> {
+    let mut harness = Harness::builder()
+        .with_size(egui::vec2(1280.0, 900.0))
+        .build_ui_state(
+            move |ui, state: &mut (HomeState, ProfileCache)| {
+                let _ = home::render(ui, &app_context, &state.0, &mut state.1);
+            },
```
thepastaclaw, line 106: **🟡 Suggestion: Test the rendered Identity Home action row**

The new kittests mount the real renderer and cover the enabled and disabled states of `Send to wallet`, which addresses only the withdrawal-gating part of the prior finding. `mount_home` still discards the `(AppAction, HomeOutcome)` returned by `home::render`, and every test queries only `Send to wallet`; none verifies that `Add funds`, `Send to another identity`, and `Add contact` are each rendered once or that enabled clicks produce their intended destinations. The tests would therefore pass if those buttons were omitted, duplicated, or wired to the wrong `HomeButton`. Capture the render result and extend the kittest to assert the complete four-button row and its enabled click routing.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest delta introduces a new blocking mismatch: Developer mode enables withdrawal whenever any on-chain key exists, even though the screen and backend require a locally held, active TRANSFER or OWNER key. Three carried-forward findings remain valid: the funding mappers still discard operation-specific recovery guidance for unclassified SDK rejections, the changelog still omits two of the PR's three user-visible fixes, and the rendered action-row tests still cover only Send to wallet rather than the complete row and its routing. With the new withdrawal blocker plus the carried-forward funding blocker, changes are required before merge.

Source: reviewers gpt-5.6-sol; final verifier gpt-5.6-sol

### Review provenance

🔴 2 blocking | 🟡 2 suggestion(s)

<details>

```

- [BLOCKING] src/wallet_backend/mod.rs:2820-2823: Preserve funding-flow guidance for unclassified SDK rejections
  This unconditional return still sends every `PlatformWalletError::Sdk` through the global `From<SdkError>` classifier. Dedicated classifications such as `AssetLockOutPointAlreadyConsumed` and insufficient balance are appropriate, but a structured `StateTransitionBroadcastError` with an unmapped consensus cause becomes `PlatformRejected` at `src/backend_task/error.rs:3082-3084` instead of the operation-specific rejection envelope used before this PR. Registration therefore loses the instruction to reuse the retained funding lock, top-up loses the affected identity context, and the identical return at lines 2902-2905 removes the platform-address flow's existing-lock recovery instructions. Preserve applicable specialized SDK classifications, but fall back to `IdentityCreateRejected`, `IdentityTopUpRejected`, or `PlatformAddressFundRejected` for unclassified broadcast rejections, with regression tests for all three mappers.

In `src/model/qualified_identity/mod.rs`:
- [BLOCKING] src/model/qualified_identity/mod.rs:891-896: Do not enable withdrawals for keys the execution path cannot use
  The Developer branch returns true for any on-chain public key, including the authentication-only, private-key-free identity used by the new kittest. Home and the identities popup consequently enable Withdrawal, but `WithdrawalScreen::new` only falls back to an on-chain `TRANSFER` key; for the authentication-only case `selected_key` remains `None`, the shared predicate suppresses the no-key state, and rendering exits at `withdraw_screen.rs:560-562` without showing a form or guidance. An on-chain-only `TRANSFER` key is also unusable because `resolve_withdrawal_signing_key` accepts only active, locally held `TRANSFER` or `OWNER` keys, and the backend withdrawal passes `None` for transition options before invoking a signer that requires private material. Base this predicate on the same eligibility invariant as the backend, or update constructor selection, the chooser, and backend signing coherently if a Developer override is intended. Add a click-through test proving every enabled launcher reaches a usable form and submission path.

In `CHANGELOG.md`:
- [SUGGESTION] CHANGELOG.md:60-67: Document the crash and consumed-deposit fixes
  The PR description explicitly presents three user-visible fixes and marks the changelog checklist complete, but the Unreleased section records only the Identity Home redesign. It still omits the larger Tokio worker stack that prevents the deposit-verification crash and the permanent-consumption guidance for already-used deposits. Add separate `Fixed` entries for those two changes so the release notes cover the complete stated PR scope.

In `tests/kittest/identity_home.rs`:
- [SUGGESTION] tests/kittest/identity_home.rs:82-85: Test the rendered Identity Home action row
  The new kittests mount the real renderer and cover the enabled and disabled states of `Send to wallet`, which addresses only the withdrawal-gating part of the prior finding. `mount_home` still discards the `(AppAction, HomeOutcome)` returned by `home::render`, and every test queries only `Send to wallet`; none verifies that `Add funds`, `Send to another identity`, and `Add contact` are each rendered once or that enabled clicks produce their intended destinations. The tests would therefore pass if those buttons were omitted, duplicated, or wired to the wrong `HomeButton`. Capture the render result and extend the kittest to assert the complete four-button row and its enabled click routing.
```
</details>

Issue #909, opened 2026-07-20
Send fails with InsufficientFunds when change would be zero — reported `available` exceeds `required`, and `Max` reliably produces the failing amount


DashBot-0001:
**Version:** `v1.0.0-weekly.20260717+3b32bed26` (Linux x86_64, prebuilt zip)

## Summary

Sending an amount that leaves **zero change** fails at transaction-build time, and the
error reports **more available than required**:

```
WalletPaymentBuildFailed {
  source: CoinSelection(InsufficientFunds { available: 10000000, required: 9999780 })
}
```

`10000000 > 9999780`. Two further problems compound it:

- `required` (9999780) is the send amount **without** the fee, so the number shown is not
  the quantity actually being tested.
- The user-facing banner reads *"The payment could not be prepared. Please review the
  amount and recipient, then try again"*, which directs the user to check the address and
  amount — neither is the cause.

**The `Max` button reliably produces exactly the failing amount**, so the one control meant
to make this easy is the one that cannot succeed.

## Steps to reproduce

1. Core wallet holding a single UTXO of 0.1 DASH.
2. Wallets → Send → click **Max** → amount `0.0999978`, quoted fee `0.0000022` (sums
   exactly to the UTXO).
3. Send → fails with the above. Reproduces every time.
4. Change the amount to `0.09` (leaving change well above dust) → succeeds immediately.

Same UTXO, same recipient, same session — the only variable is whether change is non-zero.

## Expected

- When change would be zero or below dust, fold the remainder into the fee and build the
  transaction (standard wallet behaviour) instead of failing.
- `Max` should yield a spendable amount.
- The error should name the actual constraint, and `available` / `required` should be
  comparable quantities.

## Possibly related: UTXO height is never backfilled

In the same wallet, after the funding transaction confirmed:

```
core_transactions:  height = 2507739   finalized = 1
core_utxos:         height = 0         spent = 0
```

The transaction is final at a known height while its UTXO row still records `height = 0`.
If coin selection treats height 0 as unconfirmed, this may share a root cause with the
above. Observed to persist across app restarts and across a `Refresh mode: Core + Platform`
refresh. A second UTXO (the change output from the successful send) shows the same
`height = 0` while its transaction is likewise `finalized = 1`.

## Relationship to existing issues

- #540 asks for the send dialog's *pre-flight* balance check to account for fees. This
  report is about the **build stage failing after** validation passes, plus the
  contradictory `available > required` figures — related area, different failure.
[...]

PR #937, opened 2026-07-27
fix(wallets): use selector ceiling for asset-lock Max

### LINKED ISSUE #929: `Max` on Shield reads the DISPLAY-ONLY snapshot balance, not the coin selector's set — over-shoots and the asset lock always fails
### Summary

In the Shield flow (and every `SourceSelection::CoreWallet` flow that builds an asset lock), the **Max** button and the amount validation both read the **snapshot** balance, while the transaction is actually built by the coin selector. When the wallet holds any UTXO that coin selection will not spend, Max produces an amount that passes validation and is then rejected by the builder — so Max can never succeed until the wallet happens to contain no such UTXO.

Observed on `v1.0.0-weekly.20260721+81fddd0a1` (mainnet, GUI, standard BIP44 wallet).

### Symptom

Clicking **Max** on Shield, then Shield DASH:

```
WalletBackend { source: AssetLockTransaction("Asset lock builder failed:
  Transaction builder error: Coin selection error: Insufficient funds:
  available <A>, required <R>") }
```

with `R > A`. The distinguishing signature is that **`available` stays exactly
constant across retries no matter how the amount is adjusted** — it is the coin
selector's own view, unaffected by the requested amount. Lowering the amount only
helps once it drops below `available`; the gap between the displayed balance and
`available` is invisible in the UI.

### Why it happens

The codebase already states the invariant that is being violated:

```rust
// src/wallet_backend/mod.rs:2387
// The snapshot is DISPLAY-ONLY — coin selection / tx construction MUST go
// through `send_payment` / `create_asset_lock_proof` (A04 gate).
```

But the Max computation reads the snapshot:

```rust
// src/ui/wallets/send_screen.rs:2334
Some(SourceSelection::CoreWallet) => {
    let mut max = ... self.app_context
        .snapshot_balance(&wallet.seed_hash())
        .spendable() * CREDITS_PER_DUFF
    ...
    Some(AddressKind::Shielded) => {
        // subtract estimated shield fees
    }
```

and so does the validation that gates the send:

```rust
// src/ui/wallets/send_screen.rs:1541 (send_core_to_shielded)
let balance = self.get_core_balance();     // -> snapshot_balance(..).spendable()  (:599)
if amount_duffs > balance { return Err("Insufficient balance...") }
```

Because Max and the validation share the same (snapshot) source, the validation
cannot catch what Max over-shot — it agrees with it. The disagreement only
surfaces inside `create_asset_lock_proof`, after the user has committed.

The comment right above the Max computation shows this failure mode was already
recognised once:

```rust
// Reserve against the spendable set (confirmed + unconfirmed), not `total` —
// `total` counts immature and locked funds coin selection can't spend, so a
// total-based Max over-shoots and the send fails.
```

The fix went from `total` to `spendable`, but `spendable` is still not the coin
selector's accepted set. Any UTXO the selector declines still inflates Max.

### Two independent ways I hit this

1. **Unconfirmed, not InstantSend-locked.** A Platform → Core withdrawal lands as an
   output with `confirmations = 0` and `txlock = false`. `spendable()` deliberately
   includes unconfirmed funds (per the comment at `send_screen.rs:594`, to avoid
   understating IS-locked funds that are not yet flagged locally) — but a Platform
   withdrawal output carries no IS lock, so the selector correctly refuses it while
   Max counts it. Waiting for confirmation resolves this instance.

2. **A confirmed, IS-locked, unspent UTXO that the selector still declines.** In the
   same wallet another output was excluded even though `confirmations` was in the
   thousands, `txlock = true`, `spentTxId = null`, script type `pubkeyhash`, and its
   address was present in `core_address_pool` under the same wallet / `standard_bip44`
   / account 0. I could not determine from the outside why the selector skips it, so I
   am not claiming a cause for this one — I mention it only because it shows the gap is
   not solely an unconfirmed-funds issue, and it persisted across confirmations.

Case 1 gives a reliable reproduction; case 2 is what makes Max unusable rather than
merely occasionally wrong.

### Reproduction

1. Withdraw from a Platform identity to a Core address in the same wallet.
2. Immediately open **Wallets → Shield**, source = Core wallet, and press **Max**.
3. Press **Shield DASH**.

The amount passes validation and the asset lock builder rejects it. `available` in the
error equals the confirmed/selectable subset; the displayed balance includes the
withdrawal output.

### Suggested direction

Have Max (and the amount validation) obtain the spendable ceiling from the same path
that builds the transaction — e.g. a dry-run of `create_asset_lock_proof` /
`send_payment` returning the selector's accepted total — rather than reading the
snapshot. That would satisfy the A04 invariant the code already documents, and would
make Max correct regardless of *why* a particular UTXO is unselectable.

A cheaper mitigation, if a dry-run is too heavy for the frame loop: surface the
selector's `available` in the error as a user-facing "maximum you can send right now"
so the amount can be corrected without reading logs.

### Related

- #909 — the Send (Core → Core) Max path fails differently: `available > required` with
  a small delta, from change-output sizing in coin selection. Same button, different
  root cause; that one is tracked upstream in dashpay/rust-dashcore#911.
- #4178 (dashpay/platform) — `core_utxos.height` never backfilled. Observed here too
  (every row `height = 0`), but it does not appear to drive this behaviour.

---

*Investigated and drafted with assistance from Claude (Anthropic). Source excerpts are
quoted verbatim from `81fddd0a1`; the error strings and field values are from actual
mainnet runs on the version stated. Amounts and addresses are omitted deliberately.*

Claudius-Maginificent:
## TL;DR

Pressing **Max** when shielding DASH, funding a Platform address, creating an identity, or topping up an identity from your Core wallet balance — whether from your wallet balance or a received deposit — could suggest an amount larger than the wallet could actually send. The transaction was then rejected, no matter how you adjusted the amount, because the number the UI showed wasn't the number the wallet's real balance-selection logic was actually working from. Max and the amount check now both ask the wallet directly what it can send, so the two stay in agreement across every one of those flows.

## User story

As a **DET wallet user**, I want the **Max** button on Shield / Fund Platform Address / Create Identity / Top Up Identity to give me an amount that actually sends, so that I don't hit a confusing "insufficient funds" error after already confirming the send.

## Scenario

**Actual behavior (before this PR):** After receiving funds that hadn't fully settled yet (for example, right after withdrawing from a Platform identity), or with a wallet holding an unusually large number of small UTXOs, pressing Max on Shield, Fund Platform Address, or funding an identity from the Core wallet filled in an amount that looked correct and passed the on-screen check — then failed when actually sent, with a technical error the app never explained. Lowering the amount by trial and error was the only way through.

**Expected behavior (after this PR):** Max always reflects what the wallet can actually send right now, across all four of those flows, and stays in agreement with dispatch-time validation for every one of them (including the deposit-address-scoped ReceiveDeposit method). If the real amount is temporarily lower than the wallet's headline balance (e.g. funds still settling), or the wallet's UTXO composition changes between the last check and the moment you send, Max and the send both reflect that instead of over-promising. If the real amount is capped by how many inputs a single transaction can hold, Max reflects that too, and the send succeeds on the first try. While the wallet is being asked, the amount field shows "Checking the available amount…"; if that check fails, "The available amount could not be checked." appears with a "Retry available amount check" button, and you can still switch to a different funding method at any point.

## Detailed discussion

Fixes #929.

**Root cause:** `send_screen.rs`'s Max computation and its pre-send validation for `SourceSelection::CoreWallet` asset-lock flows (Shield-from-Core, Core→Identity top-up) both read `AppContext::snapshot_balance(..).spendable()` — a display-only push snapshot (`confirmed + unconfirmed`) that the codebase's own A04 fund-safety invariant says must never feed coin selection. Because Max and validation shared that same wrong source, validation could never catch what Max over-shot; both agreed with each other and disagreed with the real coin selector, which is only consulted later, inside `create_asset_lock_proof`, after the user had already committed to the send. The same pattern was present in the identity "fund from wallet balance" screens (Create Identity, Top Up Identity).

**Fix:** added `WalletBackend::asset_lock_max_amount()` — a read-only counterpart to `create_asset_lock_proof` that queries the *live* `TransactionBuilder` for the largest credit output it will actually accept, cleaning up its own dry-run UTXO reservations afterward. Max and validation for Shield, Core→Identity top-up, Create Identity from wallet balance, and Top Up Identity from wallet balance now consume this builder-derived ceiling (via an async per-wallet UI cache, `AssetLockBalanceCache`) instead of the display-only snapshot. `DetWalletBalance::spendable()`'s doc comment was corrected to state plainly that it is display-only. A new `TaskError::AssetLockBalanceQueryFailed` variant surfaces probe failures with a plain, actionable message ("Wait a moment and try again") rather than raw internals.

**Follow-up hardening (post-review):** an independent architecture audit of the initial implementation found it was a hand-rebuilt parallel path rather than routing through the real chokepoint, and two of its duplicated parameters had already drifted:
- The probe read `synced_height()` while the real `create_asset_lock_proof` path reads `last_processed_height()` — a different, independently-advanced watermark. Diverging by enough blocks could sweep a live in-flight UTXO reservation via the coinbase-maturity/TTL logic — a real side effect from a supposedly read-only probe. Fixed: the probe now uses the same accessor the real path uses, with a regression test proving it (`asset_lock_max_uses_last_processed_height_when_sync_watermarks_diverge`).
- The cached Max ceiling (`AssetLockBalanceCache`) only invalidated on screen refresh, not on new blocks or balance changes, so validation could run against a stale ceiling. Fixed: `WalletSnapshot` now carries a per-wallet publish generation, and the cache rejects stale-generation results and re-queries.
- The probe's fee-rate constant (`ASSET_LOCK_FEE_PER_KB`) is a hand-copied duplicate of the real path's `DEFAULT_FEE_PER_KB` — verified against the pinned `rs-platform-wallet` source that this constant is `pub(super)` with no public accessor, so DET genuinely cannot import or derive it today. Left in place with an explicit `TODO(upstream)` rather than faked closed; tracked as a follow-up for upstream to export the constant or expose a shared ceiling-quote primitive.

**Simplification (post-review, round 2):** the initial ceiling query binary-searched `[0, MAX_MONEY]` against the real builder — correct, but up to ~34-51 real `TransactionBuilder` builds per Max query, each holding the wallet's global async write lock. A first attempt at trimming this added a hand-derived "68-byte" padding correction to a single-call drain probe, tuned by adding placeholder outputs until tests passed — rejected on review: independently checked against the vendored coin-selector and could only verify a *single* 34-byte change-output assumption in its fallback path, not two, with no local ground truth to confirm the larger number generalized (DET's own code never touches `TransactionBuilder` directly for the real send — it's wrapped inside `platform-wallet`'s `AssetLockManager`). Replaced instead with a seed-then-bisect approach that assumes nothing about the real strategy's internal fee formula: one `SelectionStrategy::All` drain call gives a fast, structurally-verified upper bound (the builder already drops the change output for a drain), then a short exponential-search-then-bisect against the *real* default strategy (`BranchAndBound`, matching `create_asset_lock_proof`) finds the exact boundary — typically ~15 calls instead of ~34-51, all against real code, none against a guessed formula. A wallet with more spendable UTXOs than fit in one transaction (where the drain seed itself is rejected) falls back to the original full-range search unchanged, preserving prior behavior for that edge case. All three regression tests pass with their assertions unchanged, proving the seeded result is exact.

**Closing the gaps found by automated CI review (round 3):** an automated code review of this PR found 14 issues; 9 confirmed real and independently re-verified (diff-read, not just trusted) before fixing, 5 deferred with reasons noted below:
- **The "too many UTXOs" edge case collapsed Max to a hard 0** instead of finding the largest amount achievable from an in-cap subset — present since round 1, missed by every prior review pass in this PR. The drain-ceiling seed and the full-range fallback now both route through an input-cap-aware search: a `LargestFirst` probe finds a real, builder-verified achievable seed (nudged by one well-known P2PKH input's fee as a starting point only — never trusted, always re-verified against the real default strategy via real builder calls). New test with 517 UTXOs proves a non-zero max that builds within the input cap, and one duff more does not.
- The drain-ceiling probe's placeholder credit-output script was 0 bytes where the real send path produces a ~25-byte P2PKH script, skewing the fee estimate — and therefore the quoted ceiling — in the unsafe (too-generous) direction. Now sized to match.
- **`AssetLockBalanceCache` discarded its last loaded value on every wallet-snapshot generation bump** and had no in-flight-request guard, so during active sync the funding screens could get stuck on "Checking the available amount…" indefinitely. Now stale-while-revalidate: the last loaded value stays displayed while a refresh runs in the background, and a second probe never dispatches while one is already in flight.
- **Fund Platform Address (Core wallet → Platform address) was never migrated** off the display-only snapshot — the exact #929 bug, in a fourth call site none of the prior rounds scoped. Now uses the same builder ceiling as Shield and Identity destinations.
- The **ReceiveDeposit** funding method on both Create Identity and Top Up Identity was still reading a separate per-address snapshot even though it dispatches through the identical wallet-level asset-lock builder as `UseWalletBalance` — now shares the same ceiling cache.
- Pre-selection affordability gates (deciding whether to offer/pre-select the wallet-balance funding option) still read the old, larger snapshot figure while the banner they're meant to match already read the new ceiling — now aligned; an unloaded ceiling doesn't block the option.
- `SnapshotStore::publish`'s own generation-tracking fix (round 1, above) had turned a cheap `Arc` refcount clone into a full deep-clone of the wallet's transaction/UTXO history on every wallet event — a real hot-path regression introduced by this PR itself. Generation is now assigned via a dedicated counter before entering the publish closure, which only clones the `Arc`.
- A failed background asset-lock-max probe raised a duplicate global error banner and could reset in-flight wizard step state — confirmed concretely: without suppression it invoked `step_after_task_failure` on the identity screens and reset `send_status` on the send screen, mid-registration. All three screens now suppress the global banner for this specific background task; the existing inline "could not be checked" / Retry UI is the right surface.
- This PR's own CHANGELOG entry overclaimed ("Max always produces an amount that goes through"), omitted the Platform-address flow, and reused "spendable" — the term the code comments elsewhere demote in favor of the builder-verified ceiling. Corrected to name all four fixed flows without the absolute claim.

**Closing the gaps found by a holistic full-diff review (round 4):** with rounds 1-3 landed, a full-diff review (not per-commit — specifically looking for cross-commit interaction bugs the incremental rounds couldn't see) found 4 more blocking issues, each independently reproduced against current code before fixing:
- **ReceiveDeposit's dispatch-time validation checked the wrong ceiling.** It validated against the wallet-wide asset-lock balance instead of `receive_deposit_ceiling_duffs()` — the deposit-address-scoped bound the Max button itself already used — contradicting this PR's own "the two stay in agreement" guarantee for that one funding method. Fixed: Create Identity and Top Up Identity now share one `available_ceiling_duffs()` helper for both display and dispatch validation, structurally preventing the two paths from diverging again.
- **The round-3 debounce cache (`AssetLockBalanceCache::ensure_requested`) had two flaws of its own.** (a) Its debounce signal (`spendable = confirmed + unconfirmed`) can't distinguish an unconfirmed UTXO from the same UTXO once confirmed or InstantLocked — but the real builder only accepts confirmed-or-instantlocked inputs, so a same-signal balance change (a confirmation landing) could be silently ignored. (b) A same-generation dedup check could mask a genuine signal change reached via a real two-read race: `snapshot_generation()` and `snapshot_balance()` are two independent reads that can straddle a background publish — reproduced with a standalone test (`ensure_requested` at a given generation, `store`, then a second `ensure_requested` at the *same* generation with a *changed* signal silently returned `None` instead of re-dispatching) before fixing. Fixed: the debounce key is now `(generation, final_funds_duffs)`, where `final_funds_duffs` is the confirmed-or-instantlocked total computed once at snapshot-publish time, and both fields are read from one snapshot load — eliminating the race at its source.
- **Both identity screens kept their asset-lock ceiling cached across a network switch.** `WalletSeedHash` is network-independent, so a mainnet-probed ceiling could survive into a Testnet session. `WalletSendScreen` already handled this; `AddNewIdentityScreen` and `TopUpIdentityScreen` did not. Fixed: both now invalidate the cache on `change_context`/`refresh`/`refresh_on_arrival`, matching the send screen.
- **The safe-Max probe now runs automatically on every funding-screen render, not just on explicit Send.** This converts an already-accepted upstream algorithmic-complexity risk (`dashpay/rust-dashcore#918` — BranchAndBound has no feasibility prune for near-total-balance targets) from rare/user-initiated into routine, automatic, and remotely re-triggerable by anyone who knows a receive address and sends dust to add a UTXO. Rust cannot forcibly kill an in-flight `spawn_blocking` search, so this added cooperative cancellation instead: every search-loop function checked a `CancellationToken` between iterations, cancelled by a 5-second timer started alongside the probe. On timeout the caller got a distinguishable error with a plain, actionable message instead of an unbounded wait. This was explicitly a bounding workaround, not a fix for the underlying complexity — see round 5 below for what replaced it.

Deferred from round 4 (reviewed, not fixed here): cancelling the probe when the user navigates away, and deduplicating concurrent probes for the same wallet — both real, both able to reuse the round-4 `CancellationToken` mechanism, neither required to close the round-4 findings. (Cancel-on-navigate-away and probe dedup remain open; the `CancellationToken` mechanism itself was removed in round 5 below, so any future work here needs a fresh mechanism.)

**Closing the gaps found by a bot review of the reconciled whole (round 5):** after round 4 landed, `thepastaclaw`'s review of the current tip found three further issues. All three were independently re-verified by reading the actual pinned dependency source (not taken on trust) before fixing:
- **The round-4 timeout couldn't actually bound the case it existed to bound.** The `CancellationToken` was only checked *between* search-loop iterations — never *during* a single call into the real `TransactionBuilder`. The one call most likely to hit the exponential `BranchAndBound` case is exactly the one the timeout couldn't interrupt, so the "bound worst-case wait" claim didn't hold for a genuinely pathological wallet. Rather than build a bespoke fallback search algorithm to patch around this, the actual upstream fix was confirmed to already exist: `dashpay/rust-dashcore#919` adds a suffix-sum feasibility/undershoot prune to `BranchAndBound` itself, closing `#918` at the algorithmic level (verified directly against the vendored source — the exact `BnbSearch`/`suffix`-array/`BNB_NODE_BUDGET` code, with a test explicitly asserting "what actually fixes #918 is the feasibility bound, not the node budget"). With that confirmed present on the `dashpay/platform` branch this repo now pins to, the round-4 timeout workaround — `ASSET_LOCK_PROBE_TIMEOUT`, the `CancellationToken` threaded through every search function, and the distinguishable timeout error and its UI handling — was removed entirely as unnecessary complexity now that the thing it was working around is fixed at the source.
- **A stale cached ceiling could still be used to authorize a real send.** `AssetLockBalanceCache::ensure_requested` deduped re-queries on the wallet's aggregate spendable subtotal only, so a UTXO composition change that didn't move that subtotal left pre-send validation trusting a ceiling computed before the change — exactly the class of bug this PR exists to fix, reintroduced by the cache's own debounce logic. Fixed: `WalletSnapshot` now tracks a canonical signature of eligible asset-lock inputs and a revision counter that advances only when that exact set changes; a new `get_current()` requires a quote matching the current UTXO revision for dispatch-time validation, while display keeps the existing stale-while-revalidate behavior unchanged (verified: all of round 3's anti-thrash/debounce tests still pass unmodified).
- **The dry-run probe could silently drop a real send's live UTXO reservation.** Confirmed by reading the pinned `key-wallet` source directly: `ReservationSet` is a bare `Arc<Mutex<HashMap<OutPoint,u32>>>` with no owner/session tag, and `release()` matches by outpoint alone. Real sends correctly hold the wallet-manager write lock across their whole select→reserve sequence; the probe previously took only a brief read lock to clone the account, then ran its whole dry-run search — potentially many iterations, each capable of reserving and releasing outpoints in the *same shared* reservation set — with no lock held at all. A concurrent real send and a probe iteration could both reserve the same UTXO, and the probe's own cleanup could then release the real send's still-live reservation out from under it. Fixed: the probe now acquires an owned wallet-manager write guard before cloning the account and holds it for the entire dry-run search, matching the real-send path's own locking discipline (new test `asset_lock_max_probe_waits_for_wallet_manager_write_lock` proves genuine lock contention, not a superficial check).

The `dashpay/platform` pin bump needed to make the first finding's fix real (`288a6cae4f...` → `a18bd1586858ef680124e150caad6a7dc21d0b64`, the `feat/platform-wallet-storage-rehydration` branch tip) initially failed to compile — a signature drift between `key-wallet`'s `build_asset_lock_with_signer` and `platform-wallet`'s own call site, introduced by the same branch's `rust-dashcore` bump — and had to wait for that to be fixed upstream (in `dashpay/platform`, not this repo) before landing here.

`github-actions[bot]`'s earlier review on this PR was filed against the very first pre-remediation commit and is now stale — every item it raised maps to a fix already landed in rounds 1-4.

**Closing the gaps found by a fresh CodeRabbit review of round 5's own diff (round 6):** CodeRabbit's review of the round-5 commits found 5 issues, all embedded in the review body ("outside diff range" — no resolvable GitHub thread), each independently verified against the code before deciding what to do with it. Four were real and fixed:
- **A stale probe reply could get mislabeled as the current request's result.** `AssetLockBalanceCache::store()`/`mark_loading_failed()` matched an incoming reply against the cache's *current* in-flight request using only `snapshot_generation` — because `ensure_requested` can legitimately re-arm at the same generation when UTXO composition changes (existing, tested behavior since round 4), a superseded probe's stale reply, arriving late, could be silently applied to the *newer* request's identity: reopening the exact stale-ceiling-authorizes-a-send risk round 5 closed, via a different mechanism. Fixed: every probe now carries a monotonic per-cache `request_id`, threaded through `WalletTask::GetAssetLockMaxAmount`, `BackendTaskContext`, and `BackendTaskSuccessResult`, so a reply can only ever be applied to the exact request it belongs to.
- **Dispatch-time messaging couldn't tell "still loading" from "permanently failed".** All three funding screens showed "still being checked" even when the probe had already given up and was waiting on an explicit Retry. Fixed consistently across Send, Create Identity, and Top Up Identity — not just the one screen CodeRabbit happened to flag.
- **Test coverage gap**: added the stale-UTXO-composition dispatch-rejection test to both identity screens, mirroring `send_screen.rs`'s existing coverage.
- **`snapshot.rs` cloned a full `Vec<(OutPoint, u64)>` UTXO signature on every wallet event.** Replaced with a constant-size, order-independent `u64` fingerprint of the same set; the round-6 diff's own fingerprint test only proved order-independence, so a distinctness assertion (two different UTXO sets must not collide) was added before committing.

One finding (a claimed unit mismatch between duffs and credits in `top_up_identity_screen.rs`'s Max computation) was verified to be a **false positive** — `max_amount_after_fee_reserve(spendable_duffs, fee_credits)` deliberately takes a duffs first argument and a credits second argument by its own signature and doc comment; the flagged call site is correct as written, and the suggested "fix" would have introduced a real ~1000x under-reservation bug. Left untouched, with the reasoning recorded on the PR.

**Closing the gaps found by a holistic 3-agent review of the full diff (round 7):** with rounds 1-6 landed, a fresh security/project-consistency/QA review of the complete diff (25 commits at the time) found 4 more blocking issues — all second-order effects of round 5's own lock-widening commit, plus a pre-existing gap in the Send screen that the two identity screens had already been hardened against. Each was independently re-verified against the actual pinned dependency source before fixing, not taken on trust:
- **The freshness signal round 5 added never accounted for reservations.** `asset_lock_final_inputs`'s eligibility check didn't consult the shared `ReservationSet`, so a real send's live reservation didn't move the tracked UTXO composition — meaning `get_current()` could still authorize dispatch against a ceiling the real builder would now reject, the exact failure the round-5 revision counter exists to prevent, reopened via a different gap in the same mechanism. Fixed at the root rather than patched: the probe now clones the live `ManagedCoreFundsAccount` (its `ReservationSet` is a shared `Arc<Mutex<...>>`, so the clone sees the same live reservations) and drives it through the actual `TransactionBuilder::set_funding`/`build_unsigned` path — the same code a real send runs — instead of a parallel reimplementation of the eligibility rule. Independently confirmed by reading `set_funding`'s reservation-filtering logic directly against the pinned `key-wallet` checkout.
- **Round 5's own fix for the reservation-release race went further than the funds-safety justification required, with two side effects.** Holding the wallet-manager write lock for the *entire* dry-run search (previously up to ~15 real builder calls) also blocks real sends from acquiring that same lock while a Max probe runs, and freezes `SnapshotStore::recompute` (which needs the same lock) for the same duration — stalling the very freshness recompute the lock-widening was meant to help. Fixed: a deadline is now checked between each individually-bounded builder call (an in-flight call itself still can't be interrupted — no change there from what's achievable), bounding the worst-case lock hold to ~5 seconds; on expiry the probe returns a already-proven-safe lower bound and marks the quote partial rather than blocking indefinitely or fabricating a number.
- **The Send screen could silently drop a same-frame probe dispatch or navigation action.** `AppAction::BitOrAssign` is a pure overwrite, not a merge — if a Max-probe dispatch and a confirmation-dialog action landed in the same UI frame, one could silently clobber the other. The two identity screens (Create Identity, Top Up Identity) already had a concurrent-task-preserving helper for exactly this; the Send screen never got it, since the pattern predates this PR touching that screen. Fixed: the helper is now shared (relocated out of `ui::identities` into `ui::mod`) and applied at the Send screen's probe-dispatch and confirmation-dialog call sites too.
- **A dropped probe reply left the UI stuck on "Checking the available amount…" forever, with no Retry offered** — contradicting this PR's own stated UX contract that a failed check always offers Retry. Fixed: in-flight requests now carry a 15-second deadline; on expiry the cache automatically redispatches with a fresh request id and shows Retry even while still nominally "loading," so a lost reply can no longer wedge a funding screen.

Bundled in the same pass: the round-6 `DefaultHasher`-based UTXO fingerprint (deterministic, zero-keyed, not the right primitive for a funds-safety equality check even as defense-in-depth) was replaced with exact input-set comparison; `send_core_to_shielded`'s over-maximum error wording was aligned with the other three dispatch-validation sites; and this PR's own CHANGELOG/user-stories entries were corrected to scope the "Max" fix to the Simple builder-driven funding form — the Advanced manual-input Platform-address path remains governed by the Core inputs the user selects, not this ceiling.

**Closing the gaps found by two more holistic full-diff reviews (rounds 8-11):**

*Round 8* — a fresh 3-agent (security/project-consistency/QA) review of the complete diff (28 commits at the time) found 4 more blocking issues, each either a new regression round 7's own fix introduced or an incomplete fix for a round-6 finding:
- The Send screen could still silently drop a confirmed send action on a same-frame navigation collision — a new instance of the exact defect class round 7 had just fixed elsewhere on the same screen.
- `ProbeDeadline` didn't bound the phase that costs the most: the observation/clone work ran first, unbounded, before the deadline clock even started.
- An asset-lock observation failure on the per-block recompute hot path silently discarded the *entire* snapshot recompute — balance and UTXOs included, not just the asset-lock composition.
- Max and validation still had different freshness guarantees (`get()` vs `get_current()`), leaving a narrow window where Max could offer an amount validation would immediately reject; the CHANGELOG/user-stories entries overclaimed an absolute "always agree" guarantee the code didn't provide.

All four fixed in round 9 (which also resynced the branch with `v1.0-dev`'s then-current tip — one `CHANGELOG.md` conflict, resolved by keeping both sides' entries), each independently re-verified against source before being accepted rather than taken on the fixing agent's own report.

*Round 10* — a second independent whole-diff review, this time of round 9's own fixes, explicitly briefed to re-derive every safety claim from source rather than trust the prior verification pass. It confirmed round 9's four fixes were themselves correct, but found 2 new blocking issues in the machinery round 9 built to make them work:
- The observation function classified an ordinary "wallet's eligible balance is at or below the network dust threshold" selection outcome as fatal, permanently disabling the asset-lock Max check for any near-drained wallet.
- The balance cache had no backoff for a probe reply that could never satisfy its own freshness key, so a persistent mismatch caused an unbounded per-frame re-dispatch — each iteration seizing a process-wide observation lock for up to 5 seconds and starving real sends and sync for as long as the mismatch lasted.

Both fixed in round 11: the dust-balance outcome is now treated as an ordinary result — mirroring a classification already used elsewhere in the same file — instead of an error; the cache now caps automatic re-dispatch at one transient retry before requiring an explicit Retry, with the suppression keyed to the wallet's actual UTXO composition (not sync-driven churn) so it can't be defeated by ordinary sync activity. Both fixes were independently re-verified from source, including tracing every consumer of the affected data to confirm the one flagged edge case (a since-reserved outpoint transiently entering the observed composition) can only ever cause a stale mismatch — never an inflated Max amount or a send targeting an already-reserved outpoint.

Both rounds' remaining non-blocking findings — documentation gaps, code duplication the fixes moved rather than removed, message-wording inconsistencies, and a handful of pre-existing residuals already accepted as follow-up debt in round 8 (a transitively mutable dependency pin, an undocumented process-global mutex, uneven adoption of a shared helper) — were reviewed and intentionally left out of this PR's scope: none are required to close #929 or any explicit acceptance criterion.

Deferred (reviewed, not fixed here):
- A reservation clone+release "redundancy" question — already independently verified earlier in this PR's own review that `ReservationSet` shares state via `Arc<Mutex<...>>`, so the release is load-bearing, not dead code. Not a bug.
- The backend regression tests build their own `TransactionBuilder` rather than the real entry point, since the real path has no unit-test-reachable seam (wrapped several layers inside `platform-wallet`'s `AssetLockManager`) — a documented limitation, not fixable without a backend-e2e-level test.
- A DRY extraction of the request/status/validate UI blocks duplicated across the three funding screens — real, but a larger refactor; the three screens are more consistent with each other now as a side effect of the round-3 fixes above.

(The wallet-manager write-lock hold during the ceiling search — previously listed here as an accepted, deferred follow-up — is no longer deferred: round 5 implements it, as the fix for the reservation-release race above.)

Out of scope, left as-is:
- `src/ui/dashpay/send_payment.rs` still reads `.spendable()` for ordinary (non-asset-lock) sends — that path doesn't hit the final-input rule the asset-lock builder does, needs different plumbing, and is a separate follow-up.
- Issue #909 (Send Core→Core Max) is a distinct, already-triaged upstream defect in `dashpay/rust-dashcore`'s coin selection (tracked at rust-dashcore#911) — untouched here.
- Cancel-on-navigate-away and single-flight-per-wallet probe dedup (deferred since round 4) remain open — real, but not required to close any currently-open finding.

### Testing

- `wallet_backend::payments::tests::asset_lock_max_excludes_unconfirmed_funds_counted_by_snapshot`: constructs a wallet with a confirmed and an unconfirmed UTXO and asserts the builder-derived ceiling excludes the unconfirmed value — a direct reproduction of the reported scenario.
- `wallet_backend::payments::tests::asset_lock_max_uses_last_processed_height_when_sync_watermarks_diverge`: proves the probe uses the real path's height watermark, not a diverging one.
- `wallet_backend::payments::tests::asset_lock_max_uses_an_in_cap_subset_when_the_wallet_has_too_many_utxos`: 517-UTXO wallet (over the per-tx input cap) proves Max is a real, non-zero, builder-verified amount instead of collapsing to 0, and one duff more is rejected.
- `wallet_backend::payments::tests::asset_lock_max_probe_waits_for_wallet_manager_write_lock` (round 5): proves the probe genuinely blocks on a concurrent real-send write-lock holder and only proceeds once it's released, closing the reservation-release race.
- `model::asset_lock::tests::builder_ceiling_validation_reserves_operation_fee` for the stateless ceiling/reserve arithmetic.
- `ui::wallets::send_screen::tests::core_asset_lock_max_and_validation_use_builder_quote` covering Shield, Core→Identity, and Core→Platform-address Max, exact-ceiling acceptance, and one-duff-over rejection.
- `ui::wallets::send_screen::tests::core_asset_lock_validation_rejects_quote_for_stale_utxo_composition` (round 5): drives the real send-dispatch path and confirms it rejects a quote whose UTXO revision is stale rather than authorizing the send.
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_requeries_when_unconfirmed_funds_become_final` and `..._requeries_same_generation_after_loaded_signal_changes`: reproduce round 4's two debounce-cache flaws before the fix, then prove them closed.
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_blocks_stale_validation_after_utxo_composition_change` (round 5): shows `get()` still returns a stale display value (correct, stale-while-revalidate) while `get_current()` correctly refuses to authorize a send against it.
- `wallet_backend::snapshot::tests::asset_lock_probe_snapshot_reads_generation_and_final_funds_together` and `..._final_funds_tracks_confirmation_and_instant_lock`: prove generation and the confirmed-or-instantlocked total are read from one atomic snapshot load.
- `ui::identities::add_new_identity_screen::tests::receive_deposit_dispatch_rejects_amount_above_deposit_address_balance` and `..._network_switch_and_refresh_invalidate_asset_lock_balance`: round 4's ReceiveDeposit and network-switch fixes, the latter driving the real `Screen::change_context` entry point.
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_rejects_superseded_reply_at_same_generation` (round 6): reproduces the stale-probe-reply race directly — dispatches a superseded request, then a superseding one at the same generation, and proves the superseded reply's late arrival cannot taint the superseding request's state.
- `backend_task::tests::backend_task_context_preserves_asset_lock_request_identity`, `ui::wallets::send_screen::tests::core_asset_lock_dispatch_distinguishes_failed_probe_from_loading`, `ui::identities::add_new_identity_screen::tests::register_identity_dispatch_distinguishes_failed_probe_from_loading` and `..._rejects_quote_for_stale_utxo_composition`, `ui::identities::top_up_identity_screen::tests::top_up_dispatch_distinguishes_failed_probe_from_loading` and `..._rejects_quote_for_stale_utxo_composition` (round 6): the request-id correlation, failed-vs-loading messaging, and stale-composition dispatch-rejection coverage added to both identity screens.
- `wallet_backend::snapshot::tests::asset_lock_input_fingerprint_is_independent_of_utxo_order` (round 6, extended after independent review to also assert two genuinely different UTXO sets don't collide): proves the new fingerprint is order-independent without being degenerate.
- `wallet_backend::payments::tests::asset_lock_max_probe_deadline_bounds_the_wallet_manager_write_lock` (round 7): with a zero-duration deadline, proves the write lock is released within 1 second and the returned quote is marked partial — reproduces round 5's unbounded-lock-hold regression before the fix.
- `wallet_backend::snapshot::tests::asset_lock_input_revision_changes_when_an_eligible_utxo_is_reserved` (round 7): reserves an eligible UTXO via the shared `ReservationSet` and proves the tracked input composition changes — reproduces the reservation-blind freshness gap directly.
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_validates_the_observed_not_dispatched_composition` (round 7): stores a reply whose observed composition differs from what was dispatched and proves `get_current()` authorizes only against the observed value.
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_redispatches_and_offers_retry_after_reply_deadline` (round 7): lets an in-flight request's deadline expire and proves it redispatches with a new request id and offers Retry while still loading.
- `ui::wallets::send_screen::tests::same_frame_probe_and_confirmation_dispatch_are_both_preserved` (round 7): dispatches a Max probe and a confirmation action in the same frame and proves neither is silently dropped by `AppAction::BitOrAssign`.
- `ui::wallets::send_screen::tests::confirmed_send_action_survives_same_frame_navigation` (round 9): proves a same-frame Max-probe dispatch and a confirmed send no longer clobber each other.
- `wallet_backend::payments::tests::asset_lock_observation_reports_incomplete_when_the_deadline_has_expired` and `..._asset_lock_probe_deadline_bounds_the_observation_phase` (round 9): prove the observation phase itself is now deadline-bounded, not just the phases after it.
- `wallet_backend::snapshot::tests::contended_asset_lock_observation_keeps_the_fresh_balance_and_utxos` (round 9): holds the real global observation lock during a live recompute and proves an observation failure no longer discards the balance/UTXO fields alongside it.
- `ui::wallets::send_screen::tests::max_button_offers_nothing_while_the_builder_quote_is_revalidating` (round 9): proves Max and validation now read the same accessor rather than two accessors with different freshness guarantees.
- `wallet_backend::snapshot::tests::asset_lock_observation_treats_a_sub_dust_balance_as_observed_not_failed` (round 11): a single 300-duff UTXO now yields a real, current zero quote instead of a permanent "could not be checked".
- `ui::state::asset_lock_balance::tests::asset_lock_balance_cache_stops_redispatching_after_persistent_composition_mismatch` and `..._retry_rearms_a_mismatch_suppressed_wallet` (round 11): prove a persistent composition mismatch stops automatic re-dispatch after one transient retry, and that Retry still re-arms it.
- Rounds 8-11 (this update): `cargo fmt --all -- --check` clean and scoped `cargo clippy`/`cargo test` for every touched module passing at each round, each independently re-run and re-verified from source rather than trusted from the fixing agent's own report; full-workspace sweep left to CI per this repo's own local-vs-CI policy.
- Full joint sweep across the reconciled final state, after rounds 5-7 and the platform pin bump: `cargo fmt --all -- --check` clean; `cargo test --all-features --workspace` — 2487 tests, 0 failed (2177 unit + 310 kittest/e2e/legacy/mcp-auth/backend-e2e, +7 from round 7's own new tests, no other unexpected count swings); `cargo test --doc --all-features --workspace` — 7 passed (5 + 2 compile-fail), unchanged from round 5; `cargo clippy --all-features --all-targets -- -D warnings` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated **“Max”** for shielding and identity/wallet-balance top-ups to use the backend’s true buildable maximum (including reserved fee room), avoiding over-limit submissions.
  * Improved asset-lock ceiling/available-amount checks for send and top-up flows, with received-deposit amounts capped to what actually arrived/address-balances allow.

* **User Experience**
  * Added clearer **“Checking…”** and **“could not be checked”** states for amount ceilings, with a **Retry** action and automatic re-checks after refresh/network changes.
  * Improved over-maximum validation messages to reflect the correct fee-reserve-aware limits.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 8ea190d9)
_Canonical validated blockers: 4_

src/wallet_backend/payments.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Not fixing this one: independently re-verified earlier in this PR's review that `ReservationSet` shares state via `Arc<Mutex<HashMap>>`, so the release call IS load-bearing (a plain clone would otherwise leak reserved inputs) — the pairing is intentional, not defensive duplication. Also confirmed `reservations.reserve()` only runs as the last statement before a successful `Ok`, so rejected probes never reserve and have nothing to release. Leaving as-is; a confirming regression test would still be a reasonable follow-up.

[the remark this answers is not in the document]
Claudius-Maginificent: Correct that these tests build their own `TransactionBuilder` rather than the real entry point — the real send path has no unit-test-reachable seam (it's wrapped several layers inside `platform-wallet`'s `AssetLockManager`, confirmed while investigating this PR). Not fixable here without a backend-e2e-level test against a live wallet. Leaving as a documented limitation rather than forcing a fix that doesn't reach the real code.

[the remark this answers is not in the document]
Claudius-Maginificent: Real, but this is the same lock-hold-cost item already raised and accepted as a non-blocking follow-up earlier in this PR's own review. Deferring for the same reason (responsiveness/stall risk, not correctness) — will fold your specific narrow-the-guard-scope suggestion into that follow-up.

src/ui/identities/add_new_identity_screen/by_using_unused_balance.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Agreed this is duplicated across the three funding screens, and your diagnosis that it's directly why the three screens drifted in behavior is exactly right. Deferring the full extraction as a larger refactor rather than doing it under this PR's scope, but the round of fixes just pushed did bring the three screens' patterns back into alignment with each other, which reduces (doesn't eliminate) the risk you're flagging.

src/ui/state/asset_lock_balance.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+        let request_key = (snapshot_generation, final_funds_duffs);
+        if state.in_flight == Some(request_key)
+            || state.loaded.is_some_and(|(generation, final_funds, _)| {
+                (generation, final_funds) == request_key
+            })
+            || state.failed == Some(request_key)
+        {
+            return None;
+        }
```
thepastaclaw: **🔴 Blocking: Wallet-input changes can leave a stale ceiling authoritative**

When `final_funds_duffs` changes, `ensure_requested` starts a replacement query but deliberately retains `state.loaded`. `get()` therefore keeps returning the previous ceiling, and every affected flow uses that value not only for display but also for Max and dispatch-time validation. If final funds fall from 1 DASH to 0.5 DASH, a cached 0.9-DASH ceiling can still authorize a transaction while the refresh runs. The debounce key also uses only the aggregate final-funds sum, so a newer snapshot with different UTXO count, value distribution, outpoints, or reservations but the same subtotal is ignored indefinitely even though those properties affect BranchAndBound, fees, and the 500-input cap. Keep stale data display-only, expose a separate current authoritative value for validation, and key refreshes on selector-relevant input state rather than only the subtotal.

<sub>source: ['codex']</sub>

src/wallet_backend/payments.rs
```diff
@@ -10,11 +10,543 @@
...
+
+        let probe_token = cancellation_token.clone();
+        let result = tokio::task::spawn_blocking(move || {
+            asset_lock_max_amount_from_account(
+                &managed_account,
+                &account,
+                current_height,
+                &probe_token,
+            )
+        })
```
thepastaclaw, line 612: **🔴 Blocking: Dry-run probes can clear reservations owned by real sends**

`ManagedCoreFundsAccount::clone` copies the UTXO map but shares the live `ReservationSet` through `Arc<Mutex<_>>`. The probe drops the wallet-manager guard before running `set_funding` through `build_unsigned`, even though the pinned key-wallet implementation explicitly requires that sequence to run under one uninterrupted wallet-lock hold. A probe can observe an input as free, a real builder can then select and reserve the same input under the wallet lock, and the probe can subsequently reserve and unconditionally release that outpoint. Because `ReservationSet` is an ownership-free `HashMap`, the release deletes the real transaction's reservation too, allowing a later build to reuse an in-flight input. Run probes in the same synchronization domain as real builders or use an upstream quote path with a detached, non-reserving reservation set.

<sub>source: ['codex']</sub>

src/wallet_backend/payments.rs
```diff
@@ -10,11 +10,543 @@
...
+        .await?;
+        if cancellation_token.is_cancelled() {
+            return Err(TaskError::AssetLockMaxAmountTimedOut {
+                seed_hash: *seed_hash,
+            });
+        }
```
thepastaclaw: **🔴 Blocking: The five-second timer does not bound an active selector call**

The timer only cancels a cooperative token, while `asset_lock_max_amount` still awaits the `spawn_blocking` handle before checking that token. Cancellation is checked between dry-run builds, but the known exponential work occurs inside a single synchronous `build_unsigned()` call, and the pinned BranchAndBound recursion has no cancellation check. One selector invocation can therefore run far beyond five seconds, leaving the UI query in flight and occupying a blocking-pool thread until it finishes. An async timeout around the join handle would at least restore the UI promptly, but the blocking work would continue in the background; fully enforcing the bound requires a cancellable or bounded upstream selector, or avoiding unbounded BranchAndBound on this automatic path.

<sub>source: ['codex']</sub>

CHANGELOG.md
```diff
@@ -63,6 +63,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+  two stay in agreement, and both reserve room for the fee. Funding from a
+  received deposit is also now capped by what actually arrived at that
+  deposit address, never by unrelated funds elsewhere in the wallet. While
+  the check is running, the amount field shows "Checking the available
+  amount…"; if it fails, "The available amount could not be checked."
+  appears with a "Retry available amount check" button, and you can still
+  switch to a different funding method at any point. For a very large
+  wallet, this check is capped at a few seconds; if it can't finish in
+  time, the amount field explains that and suggests sending a smaller
+  amount or trying again shortly.
```
thepastaclaw: **🔴 Blocking: The changelog overpromises absolute sendability**

The entry says Max matches what the wallet can actually send, Max and validation stay in agreement, and the large-wallet check is capped at a few seconds. The cache currently validates against stale ceilings after wallet-input changes, and the timeout still waits for an active BranchAndBound call to finish. A builder quote is also inherently a point-in-time result whose reservation state can change before dispatch. Rephrase the changelog and SND-017 as a builder-derived point-in-time ceiling that reduces over-limit attempts, and claim a bounded wait only after the backend actually returns at the deadline.

<sub>source: ['codex']</sub>

src/backend_task/error.rs
```diff
@@ -234,6 +235,23 @@ pub enum TaskError {
         >,
     },
 
+    /// The non-broadcasting asset-lock builder probe exceeded its bounded wait.
+    #[error(
+        "We couldn't work out the maximum amount you can send from this wallet in time. Try sending a smaller amount, or wait a few minutes and try again."
+    )]
+    AssetLockMaxAmountTimedOut { seed_hash: WalletSeedHash },
```
thepastaclaw: **🟡 Suggestion: Timeout guidance recommends an action the UI still blocks**

After this error, each screen marks the quote request as failed, and all affected dispatch paths still require `asset_lock_balance.get(...)` to return a loaded ceiling. Entering a smaller amount cannot bypass that prerequisite, so the suggested action cannot send anything. Direct users to the visible Retry control or to wait and retry the check.

```suggestion
    /// The non-broadcasting asset-lock builder probe exceeded its bounded wait.
    #[error(
        "We couldn't check how much this wallet can send in time. Select \"Retry available amount check\", or wait a few minutes and try again."
    )]
    AssetLockMaxAmountTimedOut { seed_hash: WalletSeedHash },
```

<sub>source: ['codex']</sub>

src/ui/identities/funding_common.rs
```diff
@@ -158,6 +193,11 @@ pub fn max_amount_after_fee_reserve(spendable_duffs: u64, fee_credits: u64) -> u
         .saturating_sub(fee_credits)
 }
```
thepastaclaw, line 159: **💬 Nitpick: Funding helpers still describe builder ceilings as spendable balances**

`max_amount_after_fee_reserve` and `spendable_covers_minimum` still name their inputs `spendable_duffs` and document them as wallet subtotals, while current callers also pass builder-derived ceilings and deposit-address bounds. The statement that the value never exceeds what the wallet can actually send is also too strong for a cached point-in-time quote. Rename these inputs around an available ceiling and describe the helper as domain-neutral arithmetic so future callers are not encouraged to restore `DetWalletBalance::spendable()` as an authoritative funding gate.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The builder-derived ceiling now covers the intended funding flows and most prior findings are fixed, but this head still has three implementation blockers: stale ceilings remain authoritative after wallet-input changes, dry-run probes race real builders through the shared reservation set, and the five-second timer cannot stop or bound an active selector call. The changelog also promises guarantees contradicted by those paths; two smaller message and terminology issues remain.

### Review provenance

🔴 4 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `src/ui/state/asset_lock_balance.rs`:
- [BLOCKING] src/ui/state/asset_lock_balance.rs:43-64: Wallet-input changes can leave a stale ceiling authoritative
  When `final_funds_duffs` changes, `ensure_requested` starts a replacement query but deliberately retains `state.loaded`. `get()` therefore keeps returning the previous ceiling, and every affected flow uses that value not only for display but also for Max and dispatch-time validation. If final funds fall from 1 DASH to 0.5 DASH, a cached 0.9-DASH ceiling can still authorize a transaction while the refresh runs. The debounce key also uses only the aggregate final-funds sum, so a newer snapshot with different UTXO count, value distribution, outpoints, or reservations but the same subtotal is ignored indefinitely even though those properties affect BranchAndBound, fees, and the 500-input cap. Keep stale data display-only, expose a separate current authoritative value for validation, and key refreshes on selector-relevant input state rather than only the subtotal.

In `src/wallet_backend/payments.rs`:
- [BLOCKING] src/wallet_backend/payments.rs:502-537: Dry-run probes can clear reservations owned by real sends
  `ManagedCoreFundsAccount::clone` copies the UTXO map but shares the live `ReservationSet` through `Arc<Mutex<_>>`. The probe drops the wallet-manager guard before running `set_funding` through `build_unsigned`, even though the pinned key-wallet implementation explicitly requires that sequence to run under one uninterrupted wallet-lock hold. A probe can observe an input as free, a real builder can then select and reserve the same input under the wallet lock, and the probe can subsequently reserve and unconditionally release that outpoint. Because `ReservationSet` is an ownership-free `HashMap`, the release deletes the real transaction's reservation too, allowing a later build to reuse an in-flight input. Run probes in the same synchronization domain as real builders or use an upstream quote path with a detached, non-reserving reservation set.
- [BLOCKING] src/wallet_backend/payments.rs:522-543: The five-second timer does not bound an active selector call
  The timer only cancels a cooperative token, while `asset_lock_max_amount` still awaits the `spawn_blocking` handle before checking that token. Cancellation is checked between dry-run builds, but the known exponential work occurs inside a single synchronous `build_unsigned()` call, and the pinned BranchAndBound recursion has no cancellation check. One selector invocation can therefore run far beyond five seconds, leaving the UI query in flight and occupying a blocking-pool thread until it finishes. An async timeout around the join handle would at least restore the UI promptly, but the blocking work would continue in the background; fully enforcing the bound requires a cancellable or bounded upstream selector, or avoiding unbounded BranchAndBound on this automatic path.

In `CHANGELOG.md`:
- [BLOCKING] CHANGELOG.md:66-82: The changelog overpromises absolute sendability
  The entry says Max matches what the wallet can actually send, Max and validation stay in agreement, and the large-wallet check is capped at a few seconds. The cache currently validates against stale ceilings after wallet-input changes, and the timeout still waits for an active BranchAndBound call to finish. A builder quote is also inherently a point-in-time result whose reservation state can change before dispatch. Rephrase the changelog and SND-017 as a builder-derived point-in-time ceiling that reduces over-limit attempts, and claim a bounded wait only after the backend actually returns at the deadline.

- [SUGGESTION] src/backend_task/error.rs:238-242: Timeout guidance recommends an action the UI still blocks
  After this error, each screen marks the quote request as failed, and all affected dispatch paths still require `asset_lock_balance.get(...)` to return a loaded ceiling. Entering a smaller amount cannot bypass that prerequisite, so the suggested action cannot send anything. Direct users to the visible Retry control or to wait and retry the check.
```
</details>

Claudius-Maginificent:
Reviewed the 5 "outside diff range" findings from the [latest review](https://github.com/dashpay/dash-evo-tool/pull/937#pullrequestreview-4818160049) (round 5's diff) individually against the code before acting on any of them:

**Fixed in `8f5c1406`:**
- `src/ui/state/asset_lock_balance.rs:80-141` — confirmed real: a superseded probe's stale reply could get mislabeled as the *current* in-flight request's result whenever `ensure_requested` re-arms at the same `snapshot_generation` (which it legitimately does on a UTXO composition change). Fixed by threading a monotonic per-cache `request_id` through the whole dispatch/result round trip so a reply can only ever be applied to the exact request it belongs to. New regression test: `asset_lock_balance_cache_rejects_superseded_reply_at_same_generation`.
- `src/ui/identities/add_new_identity_screen/mod.rs:1135-1167` (and the same gap in `send_screen.rs`/`top_up_identity_screen.rs`) — dispatch-time messaging now distinguishes a permanently failed probe from one still loading, across all three screens, not just the one flagged.
- `src/ui/identities/add_new_identity_screen/mod.rs:1897-2033` — added `register_identity_dispatch_rejects_quote_for_stale_utxo_composition`, and the same test for Top Up Identity, mirroring `send_screen.rs`'s existing coverage.
- `src/wallet_backend/snapshot.rs:97-99` — replaced the per-event `Vec<(OutPoint, u64)>` UTXO signature with a constant-size, order-independent `u64` fingerprint.

**Not fixed — verified false positive:** `src/ui/identities/top_up_identity_screen/mod.rs:584-622` ("unit mismatch" claim). `max_amount_after_fee_reserve(spendable_duffs: u64, fee_credits: u64)` deliberately takes a duffs first argument and a **credits** second argument (its own signature and doc comment say so) — the flagged call site passing the fee estimate unconverted is correct per that contract. The suggested fix would introduce a real ~1000x under-reservation bug by subtracting a duffs value from a credits total. Leaving as-is.

All four fixes independently re-verified (diff read against actual code, not the fix's own claims) before committing; full targeted test suites re-run for real, no test silently removed or weakened.

src/wallet_backend/payments.rs
```diff
@@ -10,11 +10,592 @@
...
+                .get(&DEFAULT_BIP44_ACCOUNT)
+                .ok_or(TaskError::WalletStateInconsistent)?
+                .clone();
+            (managed_account, account, current_height)
+        };
+
+        tokio::task::spawn_blocking(move || {
+            let observed_inputs =
+                asset_lock_final_input_state(&managed_account, &account, current_height)?;
+            let amount_duffs = asset_lock_max_amount_from_account_until(
```
thepastaclaw, line 599: **🔴 Blocking: The observation mutex can extend the wallet write lock past its deadline**

The probe acquires the wallet-manager write guard and starts its five-second deadline before calling `asset_lock_final_input_state()`. That call waits indefinitely on the process-wide `ASSET_LOCK_INPUT_OBSERVATION_LOCK`, and the deadline is not checked until the subsequent amount search. A snapshot observation for another fragmented wallet can hold this mutex while running transaction builders over every 500-input batch, so the probe can retain its own wallet's exclusive manager lock for longer than five seconds and block real sends or state updates. Do not wait for the global observation mutex while holding a wallet-manager write guard; use per-wallet serialization or return a retryable partial result when the observation lock is unavailable.

<sub>source: ['codex']</sub>

src/ui/state/asset_lock_balance.rs
```diff
@@ -0,0 +1,732 @@
...
+    failed: Option<FailedRequest>,
+    retry_available: bool,
+}
+
+/// Async fetch state for asset-lock maximum amounts, keyed by wallet.
+#[derive(Default)]
+pub struct AssetLockBalanceCache {
+    states: BTreeMap<WalletSeedHash, FetchState>,
+    next_request_id: u64,
+}
```
thepastaclaw, line 72: **🔴 Blocking: Per-screen request IDs can attach an old reply to a new screen**

Every newly constructed `AssetLockBalanceCache` starts its request counter at zero, while backend results and errors are routed to whichever screen is visible when they arrive. If screen A dispatches request 1, the user navigates away, and a newly constructed funding screen B dispatches request 1 for the same wallet and snapshot generation, A's late failure matches B's in-flight request and marks B failed. B's actual later result is then ignored because the in-flight entry was cleared. A late partial success can similarly replace and discard B's fresher result. Include an application-global unique request token or a per-cache instance nonce in the correlation key.

<sub>source: ['codex']</sub>

src/wallet_backend/snapshot.rs
```diff
@@ -160,12 +206,111 @@ impl TransactionHistoryStatus {
...
+                }));
+                dry_run_account.release_reservation(&transaction);
+            }
+            Err(BuilderError::CoinSelection(SelectionError::NoUtxosAvailable)) => {}
+            Err(source) => return Err(source),
+        }
+    }
+
+    Ok(AssetLockInputState::from_inputs(observed))
+}
```
thepastaclaw, line 312: **🟡 Suggestion: Asset-lock input discovery runs transaction builders on every wallet event**

`observe_asset_lock_inputs_locked()` clones all eligible UTXOs, constructs a value map, partitions the UTXOs into 500-input batches, clones the managed account for every batch, and runs `TransactionBuilder::build_unsigned()`. `SnapshotStore::recompute()` performs this synchronously while holding the non-blocking wallet state guard, and `EventBridge::on_wallet_event()` invokes `recompute()` for every wallet event, including `SyncHeightAdvanced` and `ChainLockProcessed`. Catch-up synchronization therefore repeatedly performs work proportional to the wallet's UTXO count inside a callback whose documented contract is to remain non-blocking, while also delaying wallet-manager writers. Move this discovery out of the event callback, cache it behind a meaningful input/reservation revision, or obtain reservation-aware inputs through an upstream constant-time snapshot.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The builder-derived quote now covers the intended asset-lock flows, and the previous reservation-release race, unbounded selector path, and unusable timeout guidance are fixed. This exact head still permits stale reservation state to authorize dispatch, can retain a wallet's exclusive lock beyond the advertised deadline, and can mis-correlate requests after navigation; the release documentation also overstates a potentially partial point-in-time quote. Changes are required before final review.

### Review provenance

🔴 4 blocking | 🟡 1 suggestion(s)

<details>

```

In `src/wallet_backend/snapshot.rs`:
- [BLOCKING] src/wallet_backend/snapshot.rs:802-857: Wallet-input changes can leave a stale ceiling authoritative
  Dispatch validation compares the cached quote only with `asset_lock_inputs` from the last published wallet snapshot. A real transaction build can reserve inputs after that snapshot without publishing a new snapshot or advancing its input revision, so another funding flow continues to accept the old quote even though the real builder now excludes those inputs. The contention fallback preserves the same problem: when either `try_state()` or `try_asset_lock_final_input_state()` is unavailable, `recompute()` republishes the prior input composition with a newer generation, while `AssetLockBalanceCache::ensure_requested()` intentionally ignores generation-only changes. Reservation changes must advance a directly observed revision, unavailable observations must invalidate validation, or quote validation and transaction construction must share one live wallet-lock operation.
- [SUGGESTION] src/wallet_backend/snapshot.rs:236-290: Asset-lock input discovery runs transaction builders on every wallet event
  `observe_asset_lock_inputs_locked()` clones all eligible UTXOs, constructs a value map, partitions the UTXOs into 500-input batches, clones the managed account for every batch, and runs `TransactionBuilder::build_unsigned()`. `SnapshotStore::recompute()` performs this synchronously while holding the non-blocking wallet state guard, and `EventBridge::on_wallet_event()` invokes `recompute()` for every wallet event, including `SyncHeightAdvanced` and `ChainLockProcessed`. Catch-up synchronization therefore repeatedly performs work proportional to the wallet's UTXO count inside a callback whose documented contract is to remain non-blocking, while also delaying wallet-manager writers. Move this discovery out of the event callback, cache it behind a meaningful input/reservation revision, or obtain reservation-aware inputs through an upstream constant-time snapshot.

In `src/wallet_backend/payments.rs`:
- [BLOCKING] src/wallet_backend/payments.rs:551-579: The observation mutex can extend the wallet write lock past its deadline
  The probe acquires the wallet-manager write guard and starts its five-second deadline before calling `asset_lock_final_input_state()`. That call waits indefinitely on the process-wide `ASSET_LOCK_INPUT_OBSERVATION_LOCK`, and the deadline is not checked until the subsequent amount search. A snapshot observation for another fragmented wallet can hold this mutex while running transaction builders over every 500-input batch, so the probe can retain its own wallet's exclusive manager lock for longer than five seconds and block real sends or state updates. Do not wait for the global observation mutex while holding a wallet-manager write guard; use per-wallet serialization or return a retryable partial result when the observation lock is unavailable.

In `src/ui/state/asset_lock_balance.rs`:
- [BLOCKING] src/ui/state/asset_lock_balance.rs:58-63: Per-screen request IDs can attach an old reply to a new screen
  Every newly constructed `AssetLockBalanceCache` starts its request counter at zero, while backend results and errors are routed to whichever screen is visible when they arrive. If screen A dispatches request 1, the user navigates away, and a newly constructed funding screen B dispatches request 1 for the same wallet and snapshot generation, A's late failure matches B's in-flight request and marks B failed. B's actual later result is then ignored because the in-flight entry was cleared. A late partial success can similarly replace and discard B's fresher result. Include an application-global unique request token or a per-cache instance nonce in the correlation key.
```
</details>

src/wallet_backend/payments.rs
thepastaclaw: Resolved in `33622bb` — _The five-second timer does not bound an active selector call_ no longer present.

src/backend_task/error.rs
thepastaclaw: Resolved in `33622bb` — _Timeout guidance recommends an action the UI still blocks_ no longer present.

src/wallet_backend/payments.rs
thepastaclaw: Resolved in `33622bb` — _Dry-run probes can clear reservations owned by real sends_ no longer present.

src/ui/state/asset_lock_balance.rs
thepastaclaw: Resolved in `33622bb` — _Wallet-input changes can leave a stale ceiling authoritative_ no longer present.

src/wallet_backend/snapshot.rs
```diff
@@ -160,12 +210,139 @@ impl TransactionHistoryStatus {
...
+            | Err(BuilderError::InsufficientFunds { .. }) => {
+                // A sub-dust batch is an ordinary near-drained state, not a
+                // failure; its eligible candidates still shape the real
+                // builder's cross-batch choices, so they stay in the key.
+                // Unlike the success branch above, this arm cannot filter out
+                // an already-reserved outpoint, so one may transiently enter
+                // `observed`. That is safe: this key only ever drives quote
+                // *match* detection (see the doc comment below), never the
+                // amount `TransactionBuilder` reports or actually spends.
+                observed.extend(batch.iter().map(|utxo| (utxo.outpoint, utxo.value())));
```
thepastaclaw, line 305: **🔴 Blocking: Wallet-input changes can leave a stale ceiling authoritative**

The insufficient-funds arm records every candidate in the batch, including inputs that `TransactionBuilder::set_funding` excluded as reserved. A batch can therefore produce the same composition key across a real reservation change: for example, two free 250-duff inputs plus reserved inputs remain sub-dust both before and after one of the 250-duff inputs is reserved, so this arm records the entire batch both times. A larger free input in another batch lets the full-wallet builder use those small inputs and makes the earlier quote higher, but after the reservation the real builder excludes one of them. Because the published composition remains unchanged, `AssetLockBalanceCache::get_current` still accepts the obsolete higher quote and dispatch validation can authorize an amount the builder now rejects. Filter the insufficient batch through the live reservation set, include a reservation revision in the key, or validate and construct the transaction under one authoritative wallet operation.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The exact head improves sub-dust handling and stops persistent composition mismatches from causing per-frame probe dispatches, but four in-scope blockers remain. Reservation changes can still leave an obsolete quote authoritative, the global observation mutex can defeat the wallet-lock deadline, screen-local request IDs can misroute late replies, and the release notes promise stronger freshness and maximality guarantees than the implementation provides. The event-path builder workload and funding-helper terminology also remain valid non-blocking findings.

### Review provenance

🔴 4 blocking

<details>

```

In `src/wallet_backend/snapshot.rs`:
- [BLOCKING] src/wallet_backend/snapshot.rs:295-305: Wallet-input changes can leave a stale ceiling authoritative
  The insufficient-funds arm records every candidate in the batch, including inputs that `TransactionBuilder::set_funding` excluded as reserved. A batch can therefore produce the same composition key across a real reservation change: for example, two free 250-duff inputs plus reserved inputs remain sub-dust both before and after one of the 250-duff inputs is reserved, so this arm records the entire batch both times. A larger free input in another batch lets the full-wallet builder use those small inputs and makes the earlier quote higher, but after the reservation the real builder excludes one of them. Because the published composition remains unchanged, `AssetLockBalanceCache::get_current` still accepts the obsolete higher quote and dispatch validation can authorize an amount the builder now rejects. Filter the insufficient batch through the live reservation set, include a reservation revision in the key, or validate and construct the transaction under one authoritative wallet operation.
```
</details>

src/ui/state/asset_lock_balance.rs
thepastaclaw: Resolved in `9a5105d` — _Per-screen request IDs can attach an old reply to a new screen_ no longer present.

thepastaclaw (changes requested):

At exact head 8ea190d9883430736139f5d9bbccfbaa8a35095a, all six indexed findings remain valid; the latest merge does not change the relevant implementations. Four blockers remain around stale quote authorization, deadline-defeating lock contention, cross-screen reply correlation, and inaccurate release-note guarantees. The targeted asset-lock cache tests pass, but they do not exercise cross-screen request-ID reuse or the sub-dust reservation-composition case.

### Review provenance

🔴 4 blocking

src/wallet_backend/snapshot.rs
thepastaclaw: Resolved in this update — _Wallet-input changes can leave a stale ceiling authoritative_ no longer present.

PR #941, opened 2026-07-28
feat(identity): recover keys stranded by legacy migration (#889)

### LINKED ISSUE #889: Recovery flow for legacy-only identity keys stranded by a partial pre-migration load
## Background

PR #885 imports legacy v0.9.3 identities and their keys into the modern per-network k/v store during cold-start migration (`src/backend_task/migration/finish_unwire.rs`).

The importer uses a **skip-if-present** rule: an identity already in the modern store is skipped wholesale and never re-persisted. This is deliberately conservative — see the design doc §7 "Known limitation" (`docs/ai-design/2026-07-13-legacy-identity-migration/design.md`).

## The gap

If a v0.9.3 identity was already loaded into the modern store *before* migration but only **partially**, migration skips it and its legacy-only keys stay in the legacy `data.db`, inaccessible through the current UI. The canonical cases:

- A **masternode** brought in from just its ProTxHash persists a *bare* record (no private keys); the legacy blob still holds its owner/voting/payout keys.
- The masternode load form accepts voting/owner/payout keys independently, so a modern record may hold one key while the legacy blob holds the others.

No data is destroyed — the legacy `data.db` is preserved verbatim (migration never deletes source rows), so the keys remain recoverable. They are just not surfaced by any current flow:

- The normal load form rejects the stored ProTxHash (`RejectIfExists`).
- The per-key screen requires the user to re-enter the missing WIF and cannot read it from `data.db`.
- Adding a missing key directly to a **protected** identity fails — it introduces plaintext without the identity password (`IdentityKeyProtectionDowngrade`).

## Why not a migration-time heuristic

Three heuristic attempts (unconditional merge, empty-key-map gate, bare-only reconcile) were each shown unsound during review: field *absence* cannot be distinguished from a deliberate user removal ("Remove private key from DET" leaves no tombstone; a cleared alias persists as `None`) without provenance the model does not carry, and merging a plaintext legacy key into a protected identity trips the vault-first downgrade guard. Keeping the sentinel pending instead would re-fire an unactionable warning on every launch. The agreed resolution was to revert to safe skip-if-present and track proper recovery here.

## User Story

**As** a Dash Evo Tool user upgrading from v0.9.3 who had already partially loaded one of my identities (e.g. a masternode added by ProTxHash before the upgrade),
**I want** an explicit, opt-in recovery action that reads my preserved legacy `data.db` and restores the private keys and role associations that are still stranded there,
**so that** I regain full control of that identity without having to re-enter WIFs I no longer have on hand.

### Acceptance criteria

- A discoverable, opt-in recovery entry point (e.g. on the Key Info / identity screen) for an identity detected as present-but-partial with recoverable legacy material.
- Recovery reads the preserved legacy blob from `data.db` and merges only **genuinely-missing** key material and role associations into the modern record — never overwriting anything the user currently holds.
- For a **protected** identity, recovery is gated behind the identity password so no plaintext key is written that would trip `IdentityKeyProtectionDowngrade`.
- The user makes an explicit decision per identity; nothing is reconciled silently at migration time.
- No legacy source data is deleted; recovery is idempotent and safely repeatable.

## References

- PR #885 (`feat/legacy-identity-migration`) — introduces the migration and the skip-if-present rule.
- Design doc: `docs/ai-design/2026-07-13-legacy-identity-migration/design.md` §7 (edge cases + "Known limitation").
- Reverted reconcile attempt: commit `08242ecd`.

---

Claudius-Maginificent:
**TL;DR:** Adds a discoverable, opt-in "restore stranded keys" flow so identities partially migrated from the previous app version can recover keys and masternode role links that were left behind, without re-entering any secrets.

## User story
As an **Everyday User** who upgraded from the previous app version, I want a simple way to restore keys and role associations that got stranded during migration, to achieve a fully working identity or masternode without hunting for backups or re-typing private keys.

### Base flow
A user who ran the previous app version (v0.9.3) had identity and masternode key data stored in the old local database. Upgrading to the current version migrates what it safely can into the new storage, but some installs are left with a "bare" or partially-populated identity — a key, or a voting/owner/payout role association, that exists in the old database but never made it into the new one.

### Actual behavior
Today there is no way to recover that stranded data through the UI. The old database still holds it (nothing is ever deleted), but the only paths back in are re-entering a WIF key by hand or re-running the full migration, neither of which is discoverable from where the user actually notices the problem (a masternode showing a missing voting key, or a Key Info screen missing a key the user knows they had).

### Expected behavior
On the masternode detail screen and the identity Key Info screen, a self-appearing/self-disappearing section offers to restore exactly the keys and role links that are genuinely missing — never touching anything already present. For a password-protected identity, restoring requires the identity's password, same as any other key-sealing operation. The offer disappears on its own once there is nothing left to recover, and running it again after a full recovery is a safe no-op.

### What was done

Closes #889.

Ground-up design (not derived from PR #925 — see "Prior work" below) covering:
- `model/legacy_recovery.rs` — pure, side-effect-free candidate computation (`compute_recovery_plan`) and additive-only merge (`apply_recovery_plan`). A candidate is only ever something the current identity record lacks; nothing held today can be overwritten or removed by construction. Candidates also require a correspondence check (the restored key must still resolve to a live, non-disabled key on the identity, and a wallet-derivation reference must name a wallet this install holds) so a stale or rotated-away legacy key can't silently satisfy a role without actually fixing anything. For a key or role link held on a separate voting/operator identity, that correspondence can only be verified against this identity's *own* current records — the previous version's saved copy of that other identity, or of the link itself, is never trusted to vouch for itself, which also closes a two-step case where an unverified link, once restored, would otherwise become the very evidence a later pass uses to justify restoring more from the same source. An unlinked voting/operator key or link is listed with an explanation rather than silently restored — see "Known limitation" below for what this means in practice for a brand-new masternode's voting key, and #942 for the planned follow-up.
- `database/legacy_import.rs` — single-row read of the frozen legacy database, `SQLITE_OPEN_READ_ONLY`. No write path to the legacy store exists anywhere in this change.
- `backend_task/identity/recover_legacy_keys.rs` — the enforcement layer: recomputes candidates server-side (never trusts the UI's preview), gates on the identity's password when the identity is protected, and performs exactly one write to persist the merge. Holds the identity-load claim for the duration but keeps the storage-migration guard scoped to synchronous work only, so an open password prompt can't block unrelated operations.
- UI: a shared, render-only "restore stranded keys" component surfaced from both the masternode detail screen and the identity Key Info screen, reusing this app's existing key-role labelling so there is one source of truth for how a key's role is displayed.
- No new database tables/columns, no new persisted key-value state — detection is a pure recomputation on every screen visit, which is also what makes recovery safely idempotent and repeatable.

### Testing
- Unit tests for the merge algorithm (model layer), including the correspondence rules and the voter/operator-link exclusion described above.
- Backend integration tests against an offline `AppContext`: end-to-end restore for both protection tiers, wrong/cancelled/unavailable password prompts, double-run idempotency, deleted-identity handling, unreadable/absent legacy rows, concurrent-load exclusion, wallet-link preservation, and a mid-flight protection-tier-change guard.
- UI tests: the section's appear/disappear behavior, cross-identity result attribution (a restore completing for one identity never touches another identity's screen state), and off-screen restore survival (returning to a screen after a restore landed elsewhere no longer risks overwriting it).
- `cargo test --all-features --lib --test kittest`: 2198 lib + 293 kittest tests passing, 0 failures, 0 regressions across the workspace (independently re-verified from a clean recompile, not just self-reported).
- `cargo clippy --all-features --all-targets -- -D warnings` and `cargo fmt --all` clean.
- Round-3 hardening (see Review status below) added 8 further tests: a key_id/embedded-id mismatch rejection, a malformed-key-derivation rejection, a same-material-different-slot collision rejection, a live concurrent-writer-vs-recovery-guard race across all five whole-record writers, and a stale-password re-protect-during-recovery regression.

### Known limitation
A masternode loaded from just its ProTxHash (no locally-held keys yet) can have its owner and payout keys restored automatically by this flow, but **not** its voting key when that key lives on a separate voter identity the current record has no independent link to — which is the common case for this scenario. This isn't an oversight: it's the direct, necessary consequence of the correspondence check above refusing to let old, unverifiable data vouch for itself. The voting key still has a path back in — entering it by hand, same as before this PR — it's just not a one-click restore. Closing that gap safely requires checking a restored/entered voting key against the masternode's actual current on-chain voting address, which neither this flow nor the existing manual-entry path currently does; tracked as a separate follow-up in #942 rather than expanding this PR's scope beyond what #889 asked for (migrating stranded data, not adding new on-chain verification).

### Review status
This PR went through two independent review passes after opening, deliberately not shown prior review history to avoid confirmation bias:
1. **Adversarial review** (three reviewers, security/project-consistency/QA lenses): 23 findings, none CRITICAL/HIGH. 4 assessed as blocking, fixed — a circular-validation gap in the voter/operator key check, a fresh-install detection gate that didn't actually engage in the shipped binary, a residual stale-data window on the identity Key Info screen, and a cross-identity misattribution bug in the restore-completion handler. The remaining 19 are minor (security hygiene, doc accuracy, code-quality nits), deliberately left for a follow-up.
2. **Goal-alignment audit** (does the current code, not just the diff, actually satisfy #889's literal acceptance criteria): found that the voter/operator correspondence fix above narrowed voting-key auto-restore more than initially understood (see "Known limitation"), plus left some now-dead UI code and doc/CHANGELOG text describing the pre-fix behavior. Both fixed; the scope question was resolved as above (filed as #942, not expanded here). This pass also caught the same self-vouching shape applying to the operator-link case, closed in this PR, and flagged a structurally similar (currently unreachable, closed only by luck rather than design) case on the voter-link side — noted on #942 rather than fixed here, since a full fix would remove an existing auto-approval feature and is a larger, separate change.
3. **Automated review** (`thepastaclaw`): 8 findings, 4 flagged blocking — all verified as real defects against current code (not stale reviewer snapshots) and fixed. A recovered key is now checked against the exact live key slot it's meant to replace (matching `key_id`, not just type+bytes scanned across every key on the identity), closing both a malformed-key-derivation bypass and a same-material-under-a-different-slot collision. A new per-identity record lock serializes recovery's read-merge-write against every other whole-record writer — refresh, top-up, key edit, rename, protect/unprotect, delete — not only the migration/load guards it previously relied on. The verified password is now re-checked against the identity's *current* protected state immediately before sealing, closing a window where a mid-flight re-protect-with-a-new-password could seal recovered keys under the stale one. An unrelated backend-task error can no longer reset an in-progress restore back to "offered" on the Key Info or Masternodes screens. One suggestion (preserving SQLite error causes in a recovery-read failure) and two documentation nitpicks were left unaddressed, out of this round's scope.

### Breaking changes
None. No schema change, no new dependency, no change to any existing flow's behavior — the feature is entirely dormant on any install without a legacy database.

### Checklist
- [x] Tests added/updated
- [x] `cargo fmt --all`
- [x] `docs/user-stories.md` updated (new story, IDN-020)
- [x] `CHANGELOG.md` updated
- [x] Design record added (`docs/ai-design/2026-07-28-legacy-identity-recovery/`)

### Prior work
- #925 attempted to address this issue but grew, over 57 commits, into a much broader "masternode/evonode identity lifecycle + unload" feature whose scope diverged substantially from this issue's literal request, and surfaced a data-loss-risk finding in its own key-write path during review. This PR is an independent, ground-up redesign scoped strictly to what #889 actually asks for, informed by (but not built on top of) what was learned reviewing #925.
- #942 tracks the on-chain voting-key verification follow-up described above — out of scope for #889 itself.

### Attribution

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in legacy recovery flow to restore keys and role associations left behind after an upgrade.
  * Recovery offers appear per identity on identity details, Key Info, and masternode/evonode detail screens when eligible.
  * Users can preview what will be restored and apply by role (owner/voting/payout) without replacing or removing existing keys.
  * Protected identities prompt for the identity password; cancel/wrong password leaves state unchanged and restore remains repeatable.
  * Unavailable/unreadable or non-matching keys are shown with explanations (not silently restored).
* **Bug Fixes**
  * Improved refresh/retirement of recovery offers and success messaging after returning from Key Info.
* **Tests**
  * Added UI tests covering restoring, exclusions, and in-flight behavior.


thepastaclaw:
🕓 Ready for review — 7 ahead in queue (commit 61ae1541)
_Queue position: 8/10 · 3 reviews active_
_ETA: start ~13:16 UTC · complete ~13:42 UTC (median 25m across 30 recent reviews; 3 slots)_
_Queued 3h 7m ago · Last checked: 2026-07-30 12:20 UTC_

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,659 @@
...
+on-chain revision), `dpns_names`, `status`, wallet link
+(`wallet_hash`/`wallet_index` — preserved from the modern record by the
+`update_local_qualified_identity` writer's existing contract), and all runtime
+wiring (`associated_wallets`, `secret_access`, `top_ups`).
+
+### 3.3 Application — `apply_recovery_plan(modern, legacy, approved) -> AppliedRecovery`
+
+Pure, same module. Starting from a clone of the **fresh modern record**:
+
+```
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed still open: the pseudocode fences at lines 250 and 360 have no language identifier (still triggers markdownlint MD040). Not fixed in this pass — left for the PR author to apply (`text` is a reasonable fence language for both).

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,659 @@
...
+
+New `TaskError` variants (no `String` payloads; `Display` written for the
+Everyday User — what happened + what to do, details via `with_details`):
+
+| Variant | When | User-facing direction |
+|---|---|---|
+| `LegacyIdentityUnreadable { identity_id }` | legacy row present but blob/columns fail decode | "The saved copy of this identity from the previous version could not be read. You can still add the key by entering it on the identity's key screen." |
+| `LegacyRecoveryNothingApproved` | empty allowlist reaches the backend (UI bug / MCP misuse) | re-run the check and select items |
+| existing `IdentityNotFoundLocally`, `IdentityLoadInProgress`, `SecretPromptUnavailable`, prompt-cancel, `IdentityKeyProtectionIncomplete` / `IdentityKeyProtectionLegacyFormat` | as today | existing copy |
+
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed against current code: `recover_legacy_identity_data` returns `LegacyRecoveryNothingApproved` exactly when `approved.is_empty()` (pre-execution), and separately returns success `LegacyRecoveryCompleted { nothing_to_recover }` whenever `applied.applied.is_empty()` post-execution (including a non-empty-but-entirely-stale allowlist). The table row for `LegacyRecoveryNothingApproved` already says "empty allowlist," so it's accurate — but lines 370-372 don't call out that a non-empty, all-stale approval set is a *different, successful* outcome. Valid doc-clarity gap, not fixed in this pass.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,722 @@
...
+- `ui/components/legacy_recovery_section.rs` — new render-only component
+  (component pattern: builder config, `show(ui) -> ComponentResponse`); takes
+  the plan summary, emits the approved-items dispatch intent. It renders egui,
+  so it is a component, not `ui/state/`.
+- `ui/masternodes/detail_screen.rs` + `ui/identities/keys/key_info_screen.rs`
+  — hold `Option<plan>` screen state, dispatch `CheckLegacyRecovery` on
+  arrival (behind the legacy-rows gate of §2.1), render the section, handle both
+  results in `display_task_result`, refresh on completion. Progress/success
+  banners via the standard `BannerHandle` lifecycle; failure banners carry
+  details via `with_details`.
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed: §4.5 still describes `CheckLegacyRecovery` as dispatched "on arrival," while §10.1 explicitly documents the shipped behavior as a render-latch dispatch instead. §10.1 does self-disclose the discrepancy ("§2.1 step 2 and §4.5 put the dispatch on refresh_on_arrival... Shipped: a dispatch-once latch..."), so a reader of the full doc isn't misled, but §4.5 read alone is stale. Valid, not fixed in this pass — left for the PR author.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,722 @@
...
+predicate that already decides whether the link is worth offering — so the two
+cannot disagree about which candidate the link belongs to.
+
+### 10.11 The fresh-install gate is on rows, not on the file
+
+§2.1 step 1 gated detection on `data.db` existing. Every install passes that
+gate: `AppState::boot_inputs` opens an existing file read-only *or* creates one
+and runs the fresh-install schema ladder, so `FetchState::Unavailable` was
+unreachable in the shipped binary and every identity or node screen dispatched
+a detection task against a table the ladder had just created empty. The kittest
```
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed still present at (current) line 733: "...a table the ladder had just created empty." Reads as missing an object; "created an empty database" is clearer. Trivial, not fixed in this pass.

Claudius-Maginificent:
Triage on the two `src/ui/masternodes/detail_screen.rs` findings from CodeRabbit's outside-diff-range comments (not postable as inline threads, so noting disposition here instead):

- `refresh_from_store()` reloading all masternode identities synchronously on the UI thread (312-316) — confirmed real, out of scope for this PR, not fixed here.
- Stale `contest_summary`/`open_contests` after a recovery adds a voter association (317-320) — confirmed real, out of scope for this PR, not fixed here.

The test-coverage nitpick at 1124-1160 needs no action — the suggested branch is already exercised by the adjacent `manage_keys_labels_fall_back_to_key_id_on_residual_collision` test.

src/model/legacy_recovery.rs
```diff
@@ -0,0 +1,1332 @@
...
+        return Some(ExclusionReason::KeyNoLongerOnIdentity);
+    }
+    let Some(identity) = reference_identity(modern, target) else {
+        return Some(ExclusionReason::LinkedIdentityUnverified);
+    };
+    let live = identity.public_keys().values().any(|on_identity| {
+        on_identity.key_type() == public_key.key_type()
+            && on_identity.data() == public_key.data()
+            && !on_identity.is_disabled()
+    });
```
thepastaclaw: **🔴 Blocking: Validate recovered keys against the exact live key slot**

The correspondence check accepts two classes of unusable keys. First, any error from `public_key_data_from_private_key_data` skips private/public validation, so malformed bytes such as an all-zero ECDSA secret can be restored when their saved public metadata matches the identity. Second, the live-key search compares only type and public bytes across every key, ignoring the approved map key's `key_id`; a legacy entry under one ID can therefore be accepted because the same material is live under another ID. The entry is still stored under the legacy `(target, key_id)`, while signing resolves private material by the requested on-chain key ID, so role-presence checks can report a key that cannot sign. Pass `key_id` into this check, require the saved map key and embedded public-key ID to agree, compare with the active current key at exactly that ID, and reject derivation failures unless a specific supported key type has an independent correspondence validator.

<sub>source: ['codex']</sub>

src/backend_task/identity/recover_legacy_keys.rs
```diff
@@ -0,0 +1,1679 @@
...
+        if self
+            .protected_identity_verify_scope(&applied.merged)?
+            .is_some()
+        {
+            let password = password.ok_or(TaskError::LegacyRecoveryIdentityChanged)?;
+            reject_resident_identity_plaintext(&modern.private_keys)?;
+            self.seal_merged_plaintext_keys(&mut applied.merged, password)?;
+        }
+
+        self.update_local_qualified_identity(&applied.merged)?;
```
thepastaclaw: **🔴 Blocking: Serialize recovery with all writers of the identity record**

Neither guard held by this read/merge/write section excludes ordinary same-identity writers. `migration_run` coordinates migration and removal, while the identity-load claim is observed only by load and recovery paths; refreshes, key edits, top-ups, transfers, DPNS registration, and other callers of `update_local_qualified_identity` do not take either guard. Because these paths persist whole `QualifiedIdentity` snapshots, another writer can update the record after the read at lines 204-206 and have its change overwritten at line 242, or write an older snapshot immediately after recovery and erase the restored keys. The existing prompt-race test only covers an update that lands before this final reread. Use a per-identity mutation lock shared by every whole-record writer, or a versioned compare-and-swap with retry, so the final read-modify-write is atomic relative to all identity updates.

<sub>source: ['codex']</sub>

src/backend_task/identity/recover_legacy_keys.rs
```diff
@@ -0,0 +1,1679 @@
...
+        // downgrade guard evaluates, so its trigger is false either way: Tier-2
+        // seals every merged key before the write, and Tier-1 has no protected
+        // key for the guard to protect.
+        if self
+            .protected_identity_verify_scope(&applied.merged)?
+            .is_some()
+        {
+            let password = password.ok_or(TaskError::LegacyRecoveryIdentityChanged)?;
+            reject_resident_identity_plaintext(&modern.private_keys)?;
+            self.seal_merged_plaintext_keys(&mut applied.merged, password)?;
```
thepastaclaw, line 262: **🔴 Blocking: Do not seal with a password verified against an obsolete vault state**

`VerifiedIdentityPassword` only proves that its password opened one protected key when the prompt completed; it is not tied to the current vault generation. Protection and unprotection tasks take neither the recovery claim nor `migration_run`. If the identity is unprotected and then protected under a new password while recovery is waiting, the second scope check still sees a protected identity, but `seal_merged_plaintext_keys` seals the recovered keys with the previously verified password. Existing and recovered keys then require different passwords, violating the one-password invariant. Serialize protection changes with recovery through the final persist, or revalidate the password against the current protected scope while holding a shared same-identity mutation guard. Add a regression test for concurrent unprotect-and-reprotect with a different password.

<sub>source: ['codex']</sub>

src/ui/identities/keys/key_info_screen.rs
```diff
@@ -209,6 +261,12 @@ impl ScreenLike for KeyInfoScreen {
             self.protection_in_flight = false;
             self.protection_status = None;
         }
+        // End whatever recovery operation was in flight. A failed restore
+        // returns to its offer, so a mistyped identity password can be
+        // corrected and Restore pressed again.
+        if matches!(message_type, MessageType::Error) {
+            self.recovery.failed();
+        }
```
thepastaclaw, line 276: **🔴 Blocking: Keep unrelated task errors from resetting an active recovery**

`display_message` is invoked for generic errors routed to the currently visible screen, so this calls `recovery.failed()` for every failing backend task, not only this identity's recovery operation. An unrelated failure while a restore is awaiting its password or running changes `Restoring` back to `Offered` and re-enables Restore even though the original task still owns the identity-load claim; pressing it again produces a misleading `IdentityLoadInProgress` error. `MasternodesScreen::display_task_error` has the same unconditional behavior through `detail.recovery_failed()`. Give recovery checks and restores distinct `BackendTaskContext` values carrying the identity and dispatch, then transition recovery state only when the failing context matches the in-flight operation.

<sub>source: ['codex']</sub>

src/backend_task/identity/recover_legacy_keys.rs
```diff
@@ -0,0 +1,1679 @@
...
+                Err(TaskError::LegacyIdentityUnreadable { identity_id })
+            }
+            Err(error) => {
+                tracing::warn!(
+                    target = LOG_TARGET,
+                    identity = %identity_id,
+                    error = ?error,
+                    "Could not read the previous version's saved copy of an identity",
+                );
+                Err(TaskError::LegacyIdentityUnreadable { identity_id })
```
thepastaclaw, line 330: **🟡 Suggestion: Preserve SQLite causes in legacy recovery errors**

Failures to open or query the legacy database are logged and then collapsed into `LegacyIdentityUnreadable`, whose `Debug` output contains only the identity ID and has no `#[source]`. Consequently, `BannerHandle::with_details`, callers, and tests cannot inspect whether recovery failed because of permissions, I/O, schema, or another SQLite error once the process log is unavailable. Add a `TaskError` variant with the `rusqlite::Error` as a source for open/query failures, while keeping row-decoding corruption as a separate source-free outcome if no underlying error object is retained.

<sub>source: ['codex']</sub>

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,750 @@
...
+              Some(scope) => { reject_resident_identity_plaintext(&modern.private_keys)?;
+                               Some(verify_identity_object_password(&scope).await?) }
+              None => None }
+applied = apply_recovery_plan(&modern, legacy.qi, approved ∩ plan.items)
+if applied.applied.is_empty() ⇒ Ok(LegacyRecoveryCompleted { nothing_to_recover })
+if let Some(pw) = &password { seal_merged_plaintext_keys(&mut applied.merged, pw)? }  // Tier-2: InVault before persist
+update_local_qualified_identity(&applied.merged)?           // ONE write; preserves wallet link
+claim.loaded()
+Ok(LegacyRecoveryCompleted { identity_id, applied, skipped_stale, excluded })
+```
```
thepastaclaw, line 379: **💬 Nitpick: Add language identifiers to the pseudocode fences**

The pseudocode blocks beginning at lines 253 and 363 use bare fenced code blocks, triggering markdownlint MD040. Mark both opening fences as `text`.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
```diff
@@ -0,0 +1,750 @@
...
+different question (a link with no key behind it reports a node as able to vote
+when it cannot).
+
+### 10.11 The fresh-install gate is on rows, not on the file
+
+§2.1 step 1 gated detection on `data.db` existing. Every install passes that
+gate: `AppState::boot_inputs` opens an existing file read-only *or* creates one
+and runs the fresh-install schema ladder, so `FetchState::Unavailable` was
+unreachable in the shipped binary and every identity or node screen dispatched
+a detection task against a table the ladder had just created empty. The kittest
```
thepastaclaw, line 740: **💬 Nitpick: Complete the fresh-database sentence**

The phrase “the ladder had just created empty” is grammatically incomplete. Change it to “the ladder had just created an empty database.”

thepastaclaw (changes requested):

The recovery flow has four blocking correctness issues: it can accept unusable key material, lose concurrent identity updates, create mixed-password vault entries, and reset an active restore after an unrelated task error. The implementation also drops typed SQLite causes, and the design document has two minor lint/wording issues.

### Review provenance

🔴 4 blocking | 🟡 1 suggestion(s) | 💬 2 nitpick(s)

<details>

```

In `src/model/legacy_recovery.rs`:
- [BLOCKING] src/model/legacy_recovery.rs:466-481: Validate recovered keys against the exact live key slot
  The correspondence check accepts two classes of unusable keys. First, any error from `public_key_data_from_private_key_data` skips private/public validation, so malformed bytes such as an all-zero ECDSA secret can be restored when their saved public metadata matches the identity. Second, the live-key search compares only type and public bytes across every key, ignoring the approved map key's `key_id`; a legacy entry under one ID can therefore be accepted because the same material is live under another ID. The entry is still stored under the legacy `(target, key_id)`, while signing resolves private material by the requested on-chain key ID, so role-presence checks can report a key that cannot sign. Pass `key_id` into this check, require the saved map key and embedded public-key ID to agree, compare with the active current key at exactly that ID, and reject derivation failures unless a specific supported key type has an independent correspondence validator.

In `src/backend_task/identity/recover_legacy_keys.rs`:
- [BLOCKING] src/backend_task/identity/recover_legacy_keys.rs:196-242: Serialize recovery with all writers of the identity record
  Neither guard held by this read/merge/write section excludes ordinary same-identity writers. `migration_run` coordinates migration and removal, while the identity-load claim is observed only by load and recovery paths; refreshes, key edits, top-ups, transfers, DPNS registration, and other callers of `update_local_qualified_identity` do not take either guard. Because these paths persist whole `QualifiedIdentity` snapshots, another writer can update the record after the read at lines 204-206 and have its change overwritten at line 242, or write an older snapshot immediately after recovery and erase the restored keys. The existing prompt-race test only covers an update that lands before this final reread. Use a per-identity mutation lock shared by every whole-record writer, or a versioned compare-and-swap with retry, so the final read-modify-write is atomic relative to all identity updates.
- [BLOCKING] src/backend_task/identity/recover_legacy_keys.rs:229-239: Do not seal with a password verified against an obsolete vault state
  `VerifiedIdentityPassword` only proves that its password opened one protected key when the prompt completed; it is not tied to the current vault generation. Protection and unprotection tasks take neither the recovery claim nor `migration_run`. If the identity is unprotected and then protected under a new password while recovery is waiting, the second scope check still sees a protected identity, but `seal_merged_plaintext_keys` seals the recovered keys with the previously verified password. Existing and recovered keys then require different passwords, violating the one-password invariant. Serialize protection changes with recovery through the final persist, or revalidate the password against the current protected scope while holding a shared same-identity mutation guard. Add a regression test for concurrent unprotect-and-reprotect with a different password.
- [SUGGESTION] src/backend_task/identity/recover_legacy_keys.rs:284-307: Preserve SQLite causes in legacy recovery errors
  Failures to open or query the legacy database are logged and then collapsed into `LegacyIdentityUnreadable`, whose `Debug` output contains only the identity ID and has no `#[source]`. Consequently, `BannerHandle::with_details`, callers, and tests cannot inspect whether recovery failed because of permissions, I/O, schema, or another SQLite error once the process log is unavailable. Add a `TaskError` variant with the `rusqlite::Error` as a source for open/query failures, while keeping row-decoding corruption as a separate source-free outcome if no underlying error object is retained.

In `src/ui/identities/keys/key_info_screen.rs`:
- [BLOCKING] src/ui/identities/keys/key_info_screen.rs:256-269: Keep unrelated task errors from resetting an active recovery
  `display_message` is invoked for generic errors routed to the currently visible screen, so this calls `recovery.failed()` for every failing backend task, not only this identity's recovery operation. An unrelated failure while a restore is awaiting its password or running changes `Restoring` back to `Offered` and re-enables Restore even though the original task still owns the identity-load claim; pressing it again produces a misleading `IdentityLoadInProgress` error. `MasternodesScreen::display_task_error` has the same unconditional behavior through `detail.recovery_failed()`. Give recovery checks and restores distinct `BackendTaskContext` values carrying the identity and dispatch, then transition recovery state only when the failing context matches the in-flight operation.
```
</details>

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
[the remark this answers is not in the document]
thepastaclaw: The exact-head document already makes this distinction: the error table limits `LegacyRecoveryNothingApproved` to an empty allowlist, and the execution sketch returns successful completion whenever the recomputed application is empty.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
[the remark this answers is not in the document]
thepastaclaw: Sections 1-9 intentionally preserve the original proposal, while the document header makes section 10 authoritative for shipped deviations. Section 10.1 already documents the render-latch behavior and the narrower role of `refresh_on_arrival`.

src/backend_task/identity/recover_legacy_keys.rs
Claudius-Maginificent: Confirmed still valid against current code — `legacy_identity_record` (`src/backend_task/identity/recover_legacy_keys.rs:307-331`) still collapses both the open-connection failure and the query/decode failure into `TaskError::LegacyIdentityUnreadable { identity_id }`, which carries no `#[source]`.

Deferred, not fixed in this round — tracked in PR #941's description under "Review status" (round 3) as intentionally out of scope alongside the two doc nitpicks, so it isn't lost. Will be picked up as a follow-up.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
Claudius-Maginificent: Confirmed still valid — both pseudocode fences in `docs/ai-design/2026-07-28-legacy-identity-recovery/design.md` are still bare triple-backtick blocks (currently around lines 253 and 363).

Deferred, not fixed in this round — tracked in PR #941's description under "Review status" (round 3) as intentionally out of scope this pass, alongside the SQLite-source-preservation suggestion. Will be picked up as a follow-up.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
Claudius-Maginificent: Confirmed still valid — the incomplete sentence ("...a table the ladder had just created empty.") is still present (currently around line 740, shifted by later doc additions).

Deferred, not fixed in this round — tracked in PR #941's description under "Review status" (round 3) as intentionally out of scope this pass, alongside the other doc nitpick and the SQLite-source suggestion. Will be picked up as a follow-up.

src/backend_task/identity/recover_legacy_keys.rs
Claudius-Maginificent: Closing as won't-fix: the existing `tracing::warn!` at the call site already logs the underlying `rusqlite::Error` for diagnostics, and this is a narrow, cold error path (previous-version DB unreadable) with no user-facing recovery action that would benefit from the distinction. Not worth a new `TaskError` variant for this PR's scope.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
Claudius-Maginificent: Closing as won't-fix: cosmetic markdownlint (MD040) nit on a design-doc pseudocode block, no functional or reader-comprehension impact.

docs/ai-design/2026-07-28-legacy-identity-recovery/design.md
Claudius-Maginificent: Closing as won't-fix: cosmetic doc wording nit, meaning is unambiguous in context.

PR #947, opened 2026-07-30
fix(identity): delete a removed key's stored secret and verify it before use (#889)

Claudius-Maginificent:
**TL;DR:** Removing a saved key now actually erases it from this device's secure storage, and signing with (or displaying) a saved key now checks the stored secret really is the key being asked about before using it.

## User story
As a **masternode operator or identity owner using Dash Evo Tool**, I want removing a key from my device to genuinely erase it, and any action that signs with or reveals a saved key to only ever use the exact key I asked about, so that a key I've removed can't resurface from this device and I never sign or export under the wrong key.

### Base flow
A user manages their masternode/identity keys in Dash Evo Tool: they add keys, occasionally remove a key's private copy from the device (e.g. after moving it elsewhere), and use saved keys to sign messages or view a key's WIF for backup.

### Actual behavior
Removing a key only forgot it from the on-screen list — the secret itself stayed in the app's secure storage with nothing left pointing at it. It could not be used or recovered through the app, but it also could not be cleared, even by deleting the whole identity afterward. Separately, when the app signed with or displayed a saved key, it trusted whatever internal placement it was given without checking that placement still held the key it was supposed to; a stale or inconsistent record could have silently signed with, or shown, the wrong key.

### Expected behavior
Removing a key's private copy now deletes it from secure storage as part of the same action — if that fails, nothing is removed and the removal can simply be retried. Signing with or displaying a saved key now confirms the stored secret matches the key being asked about; if an inconsistency is ever found, the action stops with a clear message instead of proceeding silently.

## Detailed discussion

Stacked on #946 (built directly on that PR's key-placement resolution work) — closes the two non-blocking, disclosed-but-deferred items from `docs/ai-design/2026-07-30-key-placement-resolution/design.md` §8, both raised as review findings on #946 (CALL-001, SEC-001) and left open there pending this follow-up.

### What was done

**Commit `15f27292` — a removed key takes its stored secret with it**
- `AppContext::delete_identity_key_secrets` (`src/context/identity_db.rs`), the per-key counterpart of the existing whole-identity vault sweep `clear_identity_vault_keys`.
- `key_info_screen.rs::remove_held_private_key` now deletes the resolved placements' vault secrets *before* evicting them from the in-memory key map — the map is what makes a vault label enumerable by the whole-identity sweep, so doing it in the old order stranded the secret permanently, unreachable even by deleting the identity. A vault-delete failure aborts before the map or the persisted identity blob are touched, so the user can just retry.

**Commit `fd787136` — check the key a placement names before signing with it**
- `AppContext::with_identity_secret_key` — the chokepoint behind both `SignMessageWithIdentityKey` and `DeriveIdentityKeyForDisplay` — now reads the public key the stored identity records at the exact requested placement and refuses to hand off vault bytes that don't derive it, before the caller's closure runs.
- New fieldless `TaskError::IdentityKeyMismatch`, distinct from `IdentityKeyMalformed` (bytes present but unreadable) and `IdentityKeyMissing` (nothing there).
- A key type this build cannot derive a public half for skips the check rather than failing it (mirrors the existing `key_exclusion` precedent in `legacy_recovery.rs`) — deliberately unverifiable rather than wrongly rejected.
- `WalletTask`'s `(target, key_id)` fields are unchanged — they're now checked rather than believed. Removing them entirely (resolving purely by public key, like `resolve_private_key_bytes` does) is a larger follow-up, out of scope here.

Also updates `CHANGELOG.md` and closes the two corresponding bullets in `design.md` §8.

### Testing
- `cargo fmt --all -- --check`
- `cargo clippy --bin dash-evo-tool --all-features --all-targets -- -D warnings` — clean
- Targeted test scope (`backend_task::wallet`, `context::identity_db`, `key_info_screen`, `backend_task::error`, `wallet_backend::identity_key_store`): 176 passed, 0 failed, including 5 new tests written RED-first pinning both defects and their happy-path regression locks.
- Full workspace suite left to CI per this repo's local-testing convention.

### Breaking changes
None — no public API/field changes; `WalletTask` variant shapes are unchanged.

### Checklist
- [x] `CHANGELOG.md` updated
- [x] `design.md` design record updated
- [x] Tests added (TDD, RED-first)
- [x] `cargo fmt` / `cargo clippy -D warnings` clean

### Prior work
- #946 — the key-placement resolution PR this one is stacked on; both fixes here close findings raised in its review (CALL-001, SEC-001) as disclosed, non-blocking follow-ups.
- #889 — parent tracking issue for the identity-key consistency work this and #945/#946 are part of.

### Attribution


Claudius-Maginificent:
Superseded — both commits (vault-secret deletion on remove, vault-to-record verification before sign/show) merged into fix/889-key-resolution-triage-followup at commit dca69163, combined with that branch's own persist-rollback logic. See PR #948.

PR #948, opened 2026-07-30
fix(identity): close two rounds of key-placement-resolution review findings

Claudius-Maginificent:
## TL;DR

Follow-up to #946 (merged): closes 14 findings from that PR's first code review, 9 more from a second review of this branch itself, 2 fixes absorbed from #947 (now closed, merged in here instead of staying a separate PR) for a removed key's vault secret being left behind and an unverified vault-to-record match on Show/Sign, and — as of this update — the final 7 blocking findings from a combined review of the whole branch. No new user-facing feature beyond one small addition: a new section in the Keys screen listing held keys not published on any on-chain key list, so users have somewhere to find and manage a key that would otherwise be invisible. Otherwise this fixes edge cases in the key-placement-resolution redesign: better error messages when a key can't be saved or removed, a fix so cancelling a password prompt doesn't wrongly block a key that didn't need one, a removed key's saved copy actually being deleted everywhere, key paste/removal now a properly locked read-modify-write instead of a snapshot-and-rollback that could lose concurrent writes, and several places where the app now finds a key's data correctly instead of missing it under certain conditions.

## User story

As a **user managing identity keys** (pasting a private key, removing one, or viewing/signing with one), I want the app to correctly find, save, and report on my keys in every configuration they might legitimately be stored in, to have "remove" actually remove everything, to get an error message that actually tells me what to do when something goes wrong, and to be able to find a key I hold even if it isn't published anywhere on-chain yet.

### Actual behavior (before this branch, on top of #946)
Several review-found edge cases: a persist failure could leave the screen claiming a key was saved when it wasn't; a concurrent write to the identity record while the Key Info screen was open could be silently lost; a cancelled password prompt could block a sibling key placement that needed no password at all, or block Show/Sign from reaching a live copy of a key when the placement the app checked first was a dead vault placeholder; an "already occupied" error told the user to do something that wasn't actually possible; the identities list's Keys popup could miss a key filed under an older layout convention; removing a key's saved copy left its secret behind in the vault forever, unreachable even by deleting the identity; nothing verified a vault-stored secret actually matched the key it was being shown or signed as; a key held locally but not published on any on-chain key list had nowhere in the UI to be found.

### Expected behavior (after this branch)
Each of the above is fixed with a pinned regression test. See `docs/ai-design/2026-07-30-key-placement-resolution/design.md` on this branch for the full design and its "what is not covered" section for what's still knowingly deferred.

## Detailed discussion

**Stacked on:** #946 (merged into `v1.0-dev`). This branch carries #946's full history plus `v1.0-dev`'s current tip, merged in twice (first the pre-squash branch tip, then the actual squash) to avoid resurrecting an already-fixed bug from a stale squash — both independently verified.

**Absorbed, not stacked:** #947 (`fix/889-key-vault-secret-lifecycle`) was a sibling branch fixing 2 issues this branch's own review also flagged. Rather than merge two overlapping PRs into `v1.0-dev` separately (both modified the same function), #947's 2 commits were merged directly into this branch, the one real conflict resolved by composing both invariants (vault-delete-first ordering from #947, persist-failure rollback from this branch) rather than picking one — verified via the full test suite that both hold together. #947 is now closed.

**Review status:** this branch has been through 4 review passes, all closed:
1. A review of #946 itself (16 findings) — closed by this branch's first 14 commits.
2. A review of those 14 commits (22 findings, 11 triaged by the user as fix/defer/accept-risk) — closed by this branch's next 9 commits.
3. A combined review of the whole thing (#946 + both follow-up rounds + both merges) as one diff (18 findings). Absorbing #947 closed 2 *non-blocking* findings from that review as a bonus (the vault-secret-orphan and vault-verification gaps it independently also flagged). **7 blocking findings** remained.
4. All 7 blocking findings from pass 3 are now fixed and independently verified (10 more commits): the paste/removal race is closed via a new locked read-modify-write helper (`AppContext::edit_local_qualified_identity`), Show/Sign now fall through a dead vault placeholder to a live sibling instead of failing, the resolver serves prompt-free resident keys before ever prompting, the placement error messages are reworded to fit both the read and write path, the Keys popup no longer clones key material just to decide a tint, the design doc's claimed public-surface narrowing is corrected to what the code actually enforces, and the new "held but unpublished" Keys-screen section (feature addition, called out separately from the bug fixes above) gives both callers of the reworded "occupied slot" error a real place to point users at.

Of pass 3's remaining 10 non-blocking / out-of-scope-follow-up findings, none were required for this PR; they're tracked in the review report for a future pass if wanted, not filed as issues automatically.

Every commit is independently verified (build, full test suite, `cargo fmt --all -- --check`, `cargo clippy --all-features --all-targets -- -D warnings`) — not just trusted from the implementing agent's own claim.

### Testing

`cargo test --all-features --workspace` and `cargo test --test kittest --all-features` both green at HEAD (2257 lib tests, 316 kittest tests, 0 failures). `cargo fmt --all -- --check` and `cargo clippy --all-features --all-targets -- -D warnings` both clean.

## Summary by CodeRabbit

* **New Features**
  * Locally saved keys missing from published identity lists now appear in a separate section and can be opened or removed.
  * Key details now support safer placement, wallet association, and private-key validation.

* **Bug Fixes**
  * Improved key lookup, password prompting, cancellation handling, and duplicate-key protection.
  * Key removal now cleans up securely and updates the interface only after successful persistence.
  * Prevented mismatched or invalid keys from being used for signing.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit fcdde0fb)
_Canonical validated blockers: 3_

src/model/qualified_identity/encrypted_key_storage.rs
```diff
@@ -500,6 +515,56 @@ impl KeyStorage {
...
+    /// different stores.
+    pub fn first_live_candidate(
+        &self,
+        key: &IdentityPublicKey,
+    ) -> Option<(PrivateKeyTarget, KeyID)> {
+        // Two lazy passes rather than one collected Vec: this runs per key per
+        // frame, and the fallback re-probe costs at most three map reads.
+        self.candidates(key)
+            .find(|placement| !self.is_in_vault(placement))
+            .or_else(|| self.candidates(key).next())
```
thepastaclaw, line 537: **🔴 Blocking: Synchronous resolution can prefer a prompting or unusable placement**

`first_live_candidate` treats every non-vault placement as resident, including `AtWalletDerivationPath`, which may require unlocking a wallet, and legacy `Encrypted`, which this code cannot resolve. If either variant is filed under an earlier target and the same key has a `Clear` or `AlwaysClear` copy under a later target, `held_private_key_data` makes Key Info prompt for the wallet or report the key unavailable. In contrast, `resolve_private_key_bytes` continues to the prompt-free plaintext copy. This breaks the PR's resident-first invariant and makes the synchronous UI disagree with the authoritative resolver.

```suggestion
self.candidates(key)
    .find(|placement| {
        matches!(
            self.private_keys.get(placement),
            Some((
                _,
                PrivateKeyData::Clear(_) | PrivateKeyData::AlwaysClear(_)
            ))
        )
    })
    .or_else(|| self.candidates(key).next())
```

<sub>source: ['codex']</sub>

src/model/qualified_identity/encrypted_key_storage.rs
```diff
@@ -500,6 +515,56 @@ impl KeyStorage {
...
+    ///
+    /// A typed probe, not a liveness one: it answers "is this a wallet-derived
+    /// key, and from which wallet", so a key filed under two placements where
+    /// the first is not wallet-derived still finds its wallet.
+    pub fn wallet_derived_at(&self, key: &IdentityPublicKey) -> Option<&WalletDerivationPath> {
+        self.candidates(key)
+            .find_map(|placement| match self.private_keys.get(&placement) {
+                Some((_, PrivateKeyData::AtWalletDerivationPath(path))) => Some(path),
+                _ => None,
+            })
```
thepastaclaw, line 565: **🔴 Blocking: Do not require a wallet when a resident copy is usable**

`wallet_derived_at` returns a path from any duplicate placement even when another placement contains immediately usable plaintext. Callers treat a returned path as proof that the operation needs that wallet: `WithdrawalScreen`, for example, returns before rendering the form while the wallet is locked. The actual key resolver uses the resident copy without touching the wallet, so a later wallet-derived duplicate can unnecessarily block an operation. The new `a_wallet_is_found_under_a_later_placement_too` test pins exactly this clear-first, wallet-later behavior and should instead verify that later wallet lookup is used only when no resident copy can answer.

```suggestion
if self.candidates(key).any(|placement| {
    matches!(
        self.private_keys.get(&placement),
        Some((
            _,
            PrivateKeyData::Clear(_) | PrivateKeyData::AlwaysClear(_)
        ))
    )
}) {
    return None;
}
self.candidates(key)
    .find_map(|placement| match self.private_keys.get(&placement) {
        Some((_, PrivateKeyData::AtWalletDerivationPath(path))) => Some(path),
        _ => None,
    })
```

<sub>source: ['codex']</sub>

src/context/identity_db.rs
```diff
@@ -1129,6 +1162,26 @@ impl AppContext {
...
+    ///
+    /// Idempotent — a placement whose vault label is already absent is not an
+    /// error, so a caller need not know whether the key was vault-backed.
+    pub fn delete_identity_key_secrets(
+        &self,
+        identity_id: &Identifier,
+        placements: impl IntoIterator<Item = (PrivateKeyTarget, KeyID)>,
+    ) -> std::result::Result<(), TaskError> {
+        crate::wallet_backend::IdentityKeyView::new(&self.secret_store, identity_id.to_buffer())
+            .delete_all(placements)
```
thepastaclaw, line 1182: **🔴 Blocking: Multi-placement vault deletion is not failure-atomic**

The new per-key deletion delegates to `IdentityKeyView::delete_all`, which performs one vault write per label, continues after an error, and returns only the first error afterward. For a key filed under multiple placements, one deletion can therefore succeed while another fails. `remove_held_private_key` then aborts before removing or persisting the map entries, leaving the record and UI claiming the key is fully held even though one or more vault copies are gone. This contradicts the documented and changelog guarantee that a vault failure leaves the key exactly unchanged. The labels must be deleted through a failure-atomic batch operation or restored on failure; merely stopping at the first error would still permit partial deletion when a later write fails.

<sub>source: ['codex']</sub>

src/backend_task/wallet/mod.rs
```diff
@@ -129,6 +187,18 @@ impl AppContext {
...
+                    .key_type()
+                    .public_key_data_from_private_key_data(key, network)
+                    && derived.as_slice() != recorded.data().as_slice()
+                {
+                    tracing::warn!(
+                        identity = %identity_id,
+                        key_id,
+                        "Vault key at the requested placement is not the key recorded there",
+                    );
+                    return Err(TaskError::IdentityKeyMismatch);
```
thepastaclaw, line 200: **🟡 Suggestion: Reject vault keys whose correspondence cannot be verified**

The `if let Ok(derived)` check treats every derivation failure as permission to use the vault bytes unchecked. With the current DPP implementation, `BIP13_SCRIPT_HASH` always returns `NotSupported`, so `DeriveIdentityKeyForDisplay` can present any valid secp256k1 scalar stored at that label as the private key for the script-hash record even though no correspondence was established. The signing task rejects this key type, but the display/export path does not. Fail closed with a typed unverifiable-key error when derivation is unsupported instead of handing unverified bytes to the closure.

<sub>source: ['codex']</sub>

src/ui/identities/keys/keys_screen.rs
```diff
@@ -154,14 +161,39 @@ impl KeysScreen {
...
+        let mut keys: Vec<(PrivateKeyTarget, IdentityPublicKey)> = Vec::new();
+        for ((target, _), (stored, _)) in self.identity.private_keys.iter() {
+            let key = &stored.identity_public_key;
+            if !matches!(self.identity.placement_of(key), KeyPlacement::Unknown) {
+                continue;
+            }
+            if keys.iter().any(|(_, listed)| same_key(listed, key)) {
+                continue;
+            }
+            keys.push((target.clone(), key.clone()));
```
thepastaclaw, line 178: **🟡 Suggestion: Use the same role fallback for unpublished rows and Key Info**

An unpublished row is named from the historical storage target, while `KeyInfoScreen::published_on` defaults an unknown placement to `PrivateKeyOnMainIdentity`. An unpublished authentication key stored under `PrivateKeyOnVoterIdentity` is consequently called “Voting key” in the list and “Authentication key” after it is opened. Because no on-chain list publishes this key, its storage address is not a published role. Use the same main-identity fallback as Key Info so the shared naming rule remains consistent.

```suggestion
for (_, (stored, _)) in self.identity.private_keys.iter() {
    let key = &stored.identity_public_key;
    if !matches!(self.identity.placement_of(key), KeyPlacement::Unknown) {
        continue;
    }
    if keys.iter().any(|(_, listed)| same_key(listed, key)) {
        continue;
    }
    keys.push((
        PrivateKeyTarget::PrivateKeyOnMainIdentity,
        key.clone(),
    ));
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR substantially improves placement-aware key resolution, persistence, vault verification, and recovery, but three in-scope correctness issues remain. Synchronous and wallet selection can still bypass a usable resident copy, while multi-placement vault deletion can partially remove a key despite reporting failure. There are also smaller issues in unsupported-key validation, unpublished-key naming, and the release notes.

### Review provenance

🔴 3 blocking | 🟡 3 suggestion(s)

<details>

```

In `src/model/qualified_identity/encrypted_key_storage.rs`:
- [BLOCKING] src/model/qualified_identity/encrypted_key_storage.rs:535-537: Synchronous resolution can prefer a prompting or unusable placement
  `first_live_candidate` treats every non-vault placement as resident, including `AtWalletDerivationPath`, which may require unlocking a wallet, and legacy `Encrypted`, which this code cannot resolve. If either variant is filed under an earlier target and the same key has a `Clear` or `AlwaysClear` copy under a later target, `held_private_key_data` makes Key Info prompt for the wallet or report the key unavailable. In contrast, `resolve_private_key_bytes` continues to the prompt-free plaintext copy. This breaks the PR's resident-first invariant and makes the synchronous UI disagree with the authoritative resolver.
- [BLOCKING] src/model/qualified_identity/encrypted_key_storage.rs:560-565: Do not require a wallet when a resident copy is usable
  `wallet_derived_at` returns a path from any duplicate placement even when another placement contains immediately usable plaintext. Callers treat a returned path as proof that the operation needs that wallet: `WithdrawalScreen`, for example, returns before rendering the form while the wallet is locked. The actual key resolver uses the resident copy without touching the wallet, so a later wallet-derived duplicate can unnecessarily block an operation. The new `a_wallet_is_found_under_a_later_placement_too` test pins exactly this clear-first, wallet-later behavior and should instead verify that later wallet lookup is used only when no resident copy can answer.

- [BLOCKING] src/context/identity_db.rs:1181-1182: Multi-placement vault deletion is not failure-atomic
  The new per-key deletion delegates to `IdentityKeyView::delete_all`, which performs one vault write per label, continues after an error, and returns only the first error afterward. For a key filed under multiple placements, one deletion can therefore succeed while another fails. `remove_held_private_key` then aborts before removing or persisting the map entries, leaving the record and UI claiming the key is fully held even though one or more vault copies are gone. This contradicts the documented and changelog guarantee that a vault failure leaves the key exactly unchanged. The labels must be deleted through a failure-atomic batch operation or restored on failure; merely stopping at the first error would still permit partial deletion when a later write fails.

In `src/backend_task/wallet/mod.rs`:
- [SUGGESTION] src/backend_task/wallet/mod.rs:190-200: Reject vault keys whose correspondence cannot be verified
  The `if let Ok(derived)` check treats every derivation failure as permission to use the vault bytes unchecked. With the current DPP implementation, `BIP13_SCRIPT_HASH` always returns `NotSupported`, so `DeriveIdentityKeyForDisplay` can present any valid secp256k1 scalar stored at that label as the private key for the script-hash record even though no correspondence was established. The signing task rejects this key type, but the display/export path does not. Fail closed with a typed unverifiable-key error when derivation is unsupported instead of handing unverified bytes to the closure.

In `src/ui/identities/keys/keys_screen.rs`:
- [SUGGESTION] src/ui/identities/keys/keys_screen.rs:170-178: Use the same role fallback for unpublished rows and Key Info
  An unpublished row is named from the historical storage target, while `KeyInfoScreen::published_on` defaults an unknown placement to `PrivateKeyOnMainIdentity`. An unpublished authentication key stored under `PrivateKeyOnVoterIdentity` is consequently called “Voting key” in the list and “Authentication key” after it is opened. Because no on-chain list publishes this key, its storage address is not a published role. Use the same main-identity fallback as Key Info so the shared naming rule remains consistent.

In `CHANGELOG.md`:
- [SUGGESTION] CHANGELOG.md:225-233: Remove the stale key-placement limitation
  This paragraph still warns that saving or removing a same-numbered voting key can affect the linked identity's key, may leave the requested key in place, and will be fixed by an in-progress key-placement change. Those are the behaviors the preceding entries and this PR claim to resolve. Remove the paragraph or replace it only with limitations that remain after this change so the release notes do not describe the current fix as unfinished.
```
</details>

PR #950, opened 2026-07-31
fix(platform): stop epoch-fetch retry storm from exhausting the DAPI rate limit

Claudius-Maginificent:
**TL;DR:** Adding funds to an identity (and other network actions) could fail with "All Dash network servers are temporarily unreachable" because an unrelated background check was silently burning through the app's shared connection budget.

## User story
As a **Dash Evo Tool user**, I want adding funds to my identity to succeed reliably, so that a temporary unrelated background hiccup doesn't block me from using the app.

### Base flow
A user opens DET, SPV sync connects, and they try to add funds to an identity (or do any other Platform action).

### Actual behavior
Every time the app's connection state settled from "syncing" back to "synced" — which can happen repeatedly in a short window — it silently made a network request that was guaranteed to fail on every server. Each failed attempt used up a slice of a budget shared across all of the app's network requests. When that budget ran out, unrelated actions like adding funds to an identity failed with a generic "All Dash network servers are temporarily unreachable" error, even though the network itself was healthy.

### Expected behavior
The app no longer makes that guaranteed-to-fail request. The fee-rate figure it was trying (and failing) to refresh now shows as a fixed, accurate value instead, and the Platform Info screen says so plainly instead of implying something is broken.

### What was done
Root-caused via live-log triage: `ExtendedEpochInfo::fetch_current` sends a descending-epoch-without-explicit-start query that the platform proof verifier at the current pin (`a18bd158`) unconditionally rejects (dashpay/platform#4231, open/unmerged). It fails identically on every DAPI node. `CurrentEpochInfo` runs automatically on every SPV `Syncing`→`Synced` transition (`src/app/reconcilers.rs`), so each transition cycled the SDK's entire address pool and drained the shared per-client DAPI rate-limit budget — confirmed as the direct cause of `DapiAllAddressesExhausted` in unrelated flows such as identity top-up.

- `src/backend_task/platform_info.rs`: the broken `fetch_current()` call is commented out behind a `TODO(platform#4231)` explaining the failure mode and citing the exact revert condition (PR merges, `Cargo.toml`/`Cargo.lock` rev advances past `a18bd158`). `fee_multiplier_permille` is now explicitly hardcoded to `1000` (1.0x) — not a placeholder default, but the value the pinned platform crate's own fee schedules (`rs-platform-version` `v1.rs:13` / `v2.rs:14`, `uses_version_fee_multiplier_permille: Some(1000)`) actually write on-chain for every protocol version this pin knows. The DPNS-fetch protocol-version ratchet added in #936 is untouched — it works and is unrelated to this bug. `format_unavailable_current_epoch_info` → `format_hardcoded_current_epoch_info`, wording now states the multiplier is fixed rather than "unavailable".
- `src/mcp/resolve.rs`: three tests sourced their mocked protocol version from the now-disabled `fetch_current` branch and would have gone red; repointed at the working DPNS-fetch mock. Assertions still prove the ratchet mechanism itself works (a failed-ratchet case still asserts the version stays unconfirmed), not just "doesn't panic".
- `CHANGELOG.md`: user-facing entry, follows the PR #936 entry it builds on.

No `FetchUnproved` variant exists for `ExtendedEpochInfo` (checked `rs-sdk/src/platform/fetch_unproved.rs`), so there's no way to dodge the broken proof verifier client-side — this is a stopgap until dashpay/platform#4231 merges and the pin advances.

### Testing
- `cargo test --lib --all-features -- platform_info::tests resolve::tests` — 9 passed, 0 failed (touched tests only, per repo convention).
- `cargo clippy --all-features --all-targets -- -D warnings` — clean.

### Breaking changes
None — internal error-handling and hardcoded-value change only; no persisted data, public API, or consensus-affecting change.

### Checklist
- [x] `cargo fmt --all`
- [x] Relevant tests added/updated and passing
- [ ] `docs/user-stories.md` — not applicable; restores existing behavior, no new story
- [x] `CHANGELOG.md` updated

### Prior work
Follows up on #936 (`fix(platform): work around dash-sdk epoch-proof regression via version ratchet`), which fixed protocol-version detection via the DPNS ratchet but left the broken `fetch_current()` call itself running, which this PR now disables.

### Attribution

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced connection failures during unrelated actions by pausing epoch-detail requests.
  - Platform Info now reports protocol-version confirmation status, the fixed send-fee multiplier, and unavailable epoch details.
  - Improved protocol refresh handling with retry support.

- **Documentation**
  - Added changelog details covering the updated connection behavior and fixed fee reporting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -308,6 +308,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   stay unavailable and the app keeps retrying, instead of assuming the version
   the app was built with.
 
+- **Fewer connection failures during unrelated actions**: the app kept asking
+  the network for epoch details through a request every server currently
+  refuses. Each attempt consumed part of the app's shared request allowance, so
+  other actions — adding funds to an identity, for example — could fail with a
+  connection error. That request is now paused until the upstream fix is
+  released. The send-fee rate it was meant to refresh is the standard rate every
+  network charges today, so fees are unchanged, and the Platform Info screen now
+  says plainly that the rate shown is fixed rather than read from the network.
+
 ### Changed
 
 - **Upstream wallet backend updated (`platform-wallet` / `platform-wallet-storage`)**:
--- a/src/backend_task/platform_info.rs
+++ b/src/backend_task/platform_info.rs
@@ -1,6 +1,7 @@
 use crate::backend_task::BackendTaskSuccessResult;
 use crate::backend_task::error::TaskError;
 use crate::context::AppContext;
+use crate::model::fee_estimation::PlatformFeeEstimator;
 use dash_sdk::Error as SdkError;
 use dash_sdk::Sdk;
 use dash_sdk::dashcore_rpc::RpcApi;
@@ -187,6 +188,9 @@ pub enum WithdrawalParseError {
 }
 
 // Helper functions for formatting platform data
+/// Kept for the restore path of the disabled live epoch fetch; see the
+/// `TODO(platform#4231)` in the `CurrentEpochInfo` arm.
+#[allow(dead_code)]
 fn format_extended_epoch_info(
     epoch_info: ExtendedEpochInfo,
     network: Network,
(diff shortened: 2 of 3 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 943822d5)

PR #951, opened 2026-07-31
fix(wallet): persist the identity funding account DET provisions

Claudius-Maginificent:
## TL;DR

Setting up an identity, or topping one up, first creates an account on this device for the payment to travel through. That account was never written to disk, so closing the app between sending the payment and the identity appearing left nothing able to work out where the money had gone — the payment was stranded and the funds with it. This writes the account down as it is created, retries temporary save failures a few times, and stops the payment with a clear message if saving still fails.

## User story

As a **wallet user**, I want an identity payment that is interrupted by closing the app to still be recoverable when I reopen it, so that funds I have already sent are not lost to a restart.

## Scenario

**Given** I fund a new identity, or top up an existing one
**When** the app closes after the payment is sent but before the identity appears
**Then** reopening the app finds the account the payment travelled through and resumes.

### Actual behavior

Reopening the app fails to resume the payment, reporting that the funding account cannot be found. The money is stranded. This is also a route by which a wallet could appear unusable after updating the app.

### Expected behavior

The account survives the restart, so the payment is picked up where it left off.

### Root cause

`load()` rebuilds `Wallet.accounts` from the `account_registrations` manifest alone. `provision_identity_funding_account` created the account in both upstream in-memory collections and persisted nothing — so upstream's own creator, the only writer of that row, hit its `contains_*` presence guards, concluded both sides already held the account, and took its early return, skipping the registration store its own docs call load-bearing for crash recovery. The account was live in memory and absent from disk on every launch. On the next start `resume_asset_lock` cannot re-derive the credit output and fails with *"Funding account IdentityTopUp not found for re-derivation"*.

### The write path

Provisioning writes the `AccountRegistrationEntry` through the wallet persister so a cold boot rebuilds the account from the manifest.

A single `store` is the whole write. DET's persister runs in `FlushMode::Immediate`, where a successful `store` has already committed, and the code now *asks* rather than assumes — `persister.store_commits_inline()`, added upstream at the pinned revision, decides whether a follow-up `flush` is needed at all. An unconditional flush after a committed store would be actively harmful: the persister buffer is shared per wallet, so the flush adopts whatever changeset another writer parked there and reports that stranger's terminal failure as this registration's, evicting an account whose row is already durable.

### Failure handling

Persist failures are classified, not lumped together. A transient failure is retried with bounded exponential backoff, every attempt re-supplying the entries (safe: the changeset merges by `extend` and applies through an idempotent `ON CONFLICT … DO UPDATE`). If the retry budget runs out while the write is still retryable, the account stays in memory and a pending marker records the intent so the next attempt rewrites it. A terminal failure on the first attempt evicts the account this call created. A terminal retry after an earlier transient failure retains both the account and its pending rewrite: another writer may already have committed the shared buffer during backoff. The next user attempt rewrites the registration idempotently before funding proceeds. Failures surface as the dedicated `TaskError::IdentityFundingAccountPersistFailed`. Funding provisioning is serialised per wallet, and a pending registration cannot outlive the wallet it belongs to.

### Scope

Duplicate identity index handling belongs to upstream storage and is outside this PR. The branch includes `v1.0-dev` through `13d8309e3` and retains its Platform pin `67d4ef3f` and Rust 1.98.

### Testing

- `a_provisioned_identity_topup_account_survives_a_restart` — provisions a top-up account, tears the backend down, cold-boots over a copy of the on-disk state, asserts the `account_registrations` row is present, and then asserts the *reloaded* account matches the provisioned one by type, index, xpub, and managed-collection membership. Confirmed RED before the fix (0 rows, expected 1). A sibling test covers the index-less top-up account.
- `a_committed_registration_write_is_not_followed_by_a_second_write` — asserts exactly one `store` and zero `flush` calls. Negative-controlled: reintroducing the chained flush turns it red on the flush count.
- `an_inline_committing_store_is_the_whole_write` / `a_buffering_store_is_completed_by_a_flush` — pin both branches of the `store_commits_inline()` decision. The buffering case was confirmed RED against the pre-change body.
- A fault-injection harness over the real persister covers transient retry, exhausted retry budget, terminal and constraint rollback, transient-escalating-to-terminal, inherited pending registrations, foreign buffer drains, and concurrent provisioning — asserting against real persisted rows rather than a fake's bookkeeping.
- Two terminal-retry regressions cover both a foreign successful commit and a discarded shared buffer, asserting retained accounts, pending rewrites, and exactly one persisted row after retry. Both failed before the fix and passed afterward.
- Full `wallet_lifecycle` plus `wallet_backend::identity_ops` suite: 116 passed, 0 failed. `cargo clippy --all-features --lib --tests -- -D warnings` clean, `cargo fmt --all` clean.

### Known residual

The address-pool snapshot upstream writes alongside the registration is not reachable from DET (`account_address_pool_entries` is crate-private), so pool depth for that account re-warms on the next sync instead of being restored. The account itself — the part re-derivation needs — is restored.

## Summary by CodeRabbit

* **Bug Fixes**
  * Identity-funding accounts are now reliably saved and remain available after wallet restarts.
  * Failed account saves now roll back cleanly, preventing inconsistent wallet state.
  * Transient persistence failures are retried automatically, while terminal failures preserve consistent wallet state.
  * Provisioning is serialized per wallet to prevent overlapping operations.
  * Wallet removal now clears pending account registrations and stale state.
  * Persistence failures provide actionable guidance, including when additional disk space may be needed.
* **Tests**
  * Added coverage for restarts, persistence failures, retries, rollback behavior, cleanup, and concurrent provisioning.


thepastaclaw:
🕓 Queued for automated review — 66th in line, estimated start in ~108 h (commit ec87d7f6)
_Estimated review time once started: ~3 h (two-phase automated review; median of recent runs)._

src/wallet_backend/identity_ops.rs
```diff
@@ -387,12 +403,69 @@ impl WalletBackend {
...
+            account_registrations: vec![AccountRegistrationEntry {
+                account_type,
+                account_xpub,
+            }],
+            ..Default::default()
+        };
+        self.inner
+            .wallet_persister
+            .store(*wallet_id, changeset)
+            .and_then(|()| self.inner.wallet_persister.flush(*wallet_id))
```
thepastaclaw, line 614: **🟡 Suggestion: Retry the retained changeset before rolling back memory**

DET uses the default immediate-flush `SqlitePersister`. When its inline flush fails transiently, `store` restores the account-registration changeset to the per-wallet buffer and returns a transient `PersistenceError`, as required by `PlatformWalletPersistence`. The `and_then` here skips the subsequent bare `flush`, and the caller's `inspect_err` removes both in-memory account entries even though the registration remains queued. A later unrelated store can therefore commit that retained registration while the running wallet has no corresponding account in either in-memory collection. Handle `source.is_transient()` by retrying the existing buffer through `flush` without re-submitting the changeset; roll back the in-memory inserts only after a terminal failure for which the backend has dropped the buffered delta. Add an injected transient-failure test covering the two account maps, buffered state, durable row, and retry behavior.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The focused duplicate-index and funding-account restart tests pass, but the production identity-registration seams still bypass the new slot guard and can spend funds before recreating the collision the PR is intended to prevent. The persistence rollback also mishandles transient store failures by discarding in-memory state while the changeset remains buffered; identity removal itself is intentionally DET-local and is not a defect in this PR. Source: reviewers `gpt-5.6-sol` (general, rust-quality); verifier `gpt-5.6-sol`.

### Review provenance

<details>

```

In `src/wallet_backend/identity_ops.rs`:
- [BLOCKING] src/wallet_backend/identity_ops.rs:69-76: Enforce identity-index occupancy before every paid registration
  The new uniqueness check protects only `ensure_identity_managed`, while wallet-funded creation calls `register_identity_with_funding` directly after provisioning. At the pinned upstream revision, `IdentityManager::add_identity` checks only for a duplicate identity ID and then uses `BTreeMap::insert(identity_index, managed_identity)`, which replaces an existing occupant; the orchestrator performs that insertion only after Platform has accepted and consumed the registration funding. This path is reachable after a cold boot because DET's `Wallet.identities` map starts empty and is not hydrated from stored qualified identities, so the UI's used-index list can miss an existing upstream slot. Address-funded creation at `src/backend_task/identity/register_identity.rs:334-379` likewise submits payment before checking the slot, then saves an identity that reconciliation cannot manage alongside the occupant; add-existing at `src/backend_task/identity/load_identity_from_wallet.rs:250-267` also admits duplicate local ownership. Centralize an atomic occupancy reservation and use it before broadcasting, submitting address-funded creation, or saving an existing identity. A check that is released before the asynchronous registration is insufficient because concurrent registrations could both pass it.
- [SUGGESTION] src/wallet_backend/identity_ops.rs:462-465: Retry the retained changeset before rolling back memory
  DET uses the default immediate-flush `SqlitePersister`. When its inline flush fails transiently, `store` restores the account-registration changeset to the per-wallet buffer and returns a transient `PersistenceError`, as required by `PlatformWalletPersistence`. The `and_then` here skips the subsequent bare `flush`, and the caller's `inspect_err` removes both in-memory account entries even though the registration remains queued. A later unrelated store can therefore commit that retained registration while the running wallet has no corresponding account in either in-memory collection. Handle `source.is_transient()` by retrying the existing buffer through `flush` without re-submitting the changeset; roll back the in-memory inserts only after a terminal failure for which the backend has dropped the buffered delta. Add an injected transient-failure test covering the two account maps, buffered state, durable row, and retry behavior.
```
</details>

src/wallet_backend/identity_ops.rs
```diff
@@ -55,6 +85,15 @@ impl WalletBackend {
...
         settings: Option<dash_sdk::platform::transition::put_settings::PutSettings>,
     ) -> Result<dash_sdk::platform::Identity, TaskError> {
+        // Refuse before the asset lock is built, so a collision costs nothing:
+        // upstream registers the identity at `identity_index` without an
+        // occupancy check and deliberately swallows that add's failure to
+        // guarantee the spent lock is consumed, so a collision there is
+        // persisted rather than reported. A pre-flight, not a mutex — DET
+        // cannot hold the manager lock across the network round-trip.
+        self.reject_taken_identity_index(seed_hash, identity_index)
+            .await?;
```
thepastaclaw, line 99: **🔴 Blocking: Reserve the identity index across every registration path**

The new preflight covers only wallet-funded registration and releases the manager read lock before account provisioning, asset-lock creation, and the Platform round trip. Backend tasks and unlock-time reconciliation can run concurrently, so two registrations—or registration racing `reconcile_managed_identities`—can observe a free slot before either records its identity. At the pinned upstream revision, `IdentityManager::add_identity` checks only for a duplicate identity ID and then uses `BTreeMap::insert(identity_index, ...)`, replacing an existing slot occupant after Platform has already accepted payment. The same invariant is bypassed entirely by `register_identity_from_platform_addresses`, which spends address credits at `src/backend_task/identity/register_identity.rs:334-356` before saving the wallet/index association, and by `load_user_identity_from_wallet`, which persists that association at `src/backend_task/identity/load_identity_from_wallet.rs:250-267` without consulting the guard. Use a shared per-`(wallet_id, identity_index)` reservation for wallet-funded registration, address-funded registration, reconciliation, and existing-identity attachment, and retain it until the operation reaches its terminal bookkeeping outcome.

<sub>source: ['codex']</sub>

src/backend_task/identity/remove_identity.rs
```diff
@@ -16,11 +16,13 @@ impl AppContext {
                     .map(|(voter_identity, _)| voter_identity.id())
             });
 
+        self.release_identity_index(&identity_id).await;
         self.delete_local_qualified_identity(&identity_id)?;
```
thepastaclaw, line 20: **🔴 Blocking: Keep the local identity until index release is durable**

`release_identity_index` is best-effort, but the local identity is deleted unconditionally immediately afterward. If backend initialization was deferred, `wallet_backend()` fails and release becomes a no-op. Even with an initialized backend, the pinned upstream `IdentityManager::remove_identity` mutates memory, calls `persister.store`, logs and swallows any persistence error, and still returns success. A transient failure leaves the tombstone only in the persister buffer, while a terminal failure drops it; exiting before a retained write commits lets the old occupant reappear on restart after DET has deleted the identity and its only cleanup retry anchor. Reusing the apparently free slot can then recreate duplicate persisted identity rows or leave a paid replacement unmanaged. Require an observably durable tombstone, or save a durable cleanup-retry record, before deleting the local identity and reporting removal success.

<sub>source: ['codex']</sub>

src/context/wallet_lifecycle/tests.rs
```diff
@@ -4105,6 +4168,171 @@ fn basic_test_identity() -> dash_sdk::dpp::identity::Identity {
...
+    .await
+    .expect("removing an identity must succeed");
+
+    backend
+        .ensure_identity_managed(&seed_hash, &replacement, 0)
+        .await
+        .expect(
+            "the index of a removed identity must be free again; a phantom occupant \
+             means the removal did not reach the wallet backend",
+        );
```
thepastaclaw, line 4216: **💬 Nitpick: Assert that removal admits a distinct replacement identity**

The test ignores the boolean returned by the second `ensure_identity_managed` call, so `Ok(false)` is accepted even though it means no replacement was newly registered. The fixtures currently generate random IDs, making accidental equality extremely unlikely, but the test should state both assumptions explicitly: the identities are distinct, and the replacement was newly admitted after removal.

```suggestion
let removed = test_identity_with_key();
let replacement = test_identity_with_key();
assert_ne!(removed.id(), replacement.id());
assert!(
    backend
        .ensure_identity_managed(&seed_hash, &removed, 0)
        .await
        .expect("first identity registers at a free index"),
);

ctx.run_backend_task(
    BackendTask::IdentityTask(IdentityTask::RemoveIdentity {
        identity_id: removed.id(),
    }),
    sender,
)
.await
.expect("removing an identity must succeed");

assert!(
    backend
        .ensure_identity_managed(&seed_hash, &replacement, 0)
        .await
        .expect(
            "the index of a removed identity must be free again; a phantom occupant \
             means the removal did not reach the wallet backend",
        ),
);
```

thepastaclaw (changes requested):

The focused duplicate-index, removal, and funding-account restart tests pass, and the new preflight closes the ordinary sequential wallet-funded collision path. The identity-index invariant is still bypassed by address-funded and local-load paths and remains non-atomic under concurrent registration or reconciliation; identity removal can also discard its only retry anchor before the upstream tombstone is durable. The transient account-registration persistence issue from the prior review is unchanged.

### Review provenance

🔴 2 blocking | 💬 1 nitpick(s)

<details>

```

In `src/wallet_backend/identity_ops.rs`:
- [BLOCKING] src/wallet_backend/identity_ops.rs:88-95: Reserve the identity index across every registration path
  The new preflight covers only wallet-funded registration and releases the manager read lock before account provisioning, asset-lock creation, and the Platform round trip. Backend tasks and unlock-time reconciliation can run concurrently, so two registrations—or registration racing `reconcile_managed_identities`—can observe a free slot before either records its identity. At the pinned upstream revision, `IdentityManager::add_identity` checks only for a duplicate identity ID and then uses `BTreeMap::insert(identity_index, ...)`, replacing an existing slot occupant after Platform has already accepted payment. The same invariant is bypassed entirely by `register_identity_from_platform_addresses`, which spends address credits at `src/backend_task/identity/register_identity.rs:334-356` before saving the wallet/index association, and by `load_user_identity_from_wallet`, which persists that association at `src/backend_task/identity/load_identity_from_wallet.rs:250-267` without consulting the guard. Use a shared per-`(wallet_id, identity_index)` reservation for wallet-funded registration, address-funded registration, reconciliation, and existing-identity attachment, and retain it until the operation reaches its terminal bookkeeping outcome.

In `src/backend_task/identity/remove_identity.rs`:
- [BLOCKING] src/backend_task/identity/remove_identity.rs:19-20: Keep the local identity until index release is durable
  `release_identity_index` is best-effort, but the local identity is deleted unconditionally immediately afterward. If backend initialization was deferred, `wallet_backend()` fails and release becomes a no-op. Even with an initialized backend, the pinned upstream `IdentityManager::remove_identity` mutates memory, calls `persister.store`, logs and swallows any persistence error, and still returns success. A transient failure leaves the tombstone only in the persister buffer, while a terminal failure drops it; exiting before a retained write commits lets the old occupant reappear on restart after DET has deleted the identity and its only cleanup retry anchor. Reusing the apparently free slot can then recreate duplicate persisted identity rows or leave a paid replacement unmanaged. Require an observably durable tombstone, or save a durable cleanup-retry record, before deleting the local identity and reporting removal success.
```
</details>

src/backend_task/identity/remove_identity.rs
```diff
@@ -39,4 +41,27 @@ impl AppContext {
...
+            return;
+        };
+        if let Err(error) = backend.forget_identity(identity_id).await {
+            tracing::warn!(
+                %identity_id,
+                %error,
+                "Identity index could not be released in the wallet backend; \
+                 adding an identity at that index may be refused"
+            );
+        }
```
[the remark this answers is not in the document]
thepastaclaw: The finding is valid, but propagating `forget_identity` errors alone is insufficient. At the pinned upstream revision, `IdentityManager::remove_identity` swallows `persister.store` failures and returns success after mutating memory, so DET must establish a durable tombstone itself or persist a durable cleanup-retry record before deleting the local identity.

Claudius-Maginificent:
Follow-up from a fresh repro against the reporting user's actual DB, done just now on the current `v1.0-dev` pin (which already includes #953):

**The status-update note above is stale.** Loading the user's real `platform-wallet.sqlite` on the current pin does *not* hit `UnrecognizedScript` — it hits `OrphanedIdentityEntry`, the exact duplicate-`(wallet_id, identity_index)` mechanism this PR hardens against:

```
WalletLocalDataLoadFailed { source: PersisterLoad(Backend { kind: Fatal,
    source: OrphanedIdentityEntry { owner: 63c48e0f...977303 } }) }
```

Querying the DB directly confirms it: two identity rows share `(wallet_id c59da1...67e0b0, identity_index 1)` — `63c48e0f...977303` and `e7fc8f3e...14fb5f5`, both `tombstoned = 0`. So either #953 masked an earlier failure and this duplicate-index corruption was already sitting underneath it, or the original investigation moved on before re-testing against #953. Either way, this PR's own bug report *is* what's currently blocking that wallet.

**However, merging this PR as-is will not unblock it.** All three fixes here (`ensure_identity_managed`, `register_identity`, `AppContext::remove_identity`) are DET-side guards against *creating* a new collision. The fatal error happens earlier and lower — inside upstream `rs-platform-wallet-storage`'s `load_from_persistor()` → `merge_contacts_and_keys()`, before any DET code (including this PR's guards, or the boot-reconciler leniency the "residual, by design" note refers to) ever runs. Pre-existing on-disk duplicates — e.g. carried in by a migration, as here — still hard-fail wallet load with no recovery path.

Logged the upstream side of this as a `platform` TODO (memcan `4fbe235c`): `rs-platform-wallet-storage` needs either a write-time uniqueness constraint on `(wallet_id, identity_index)`, or `merge_contacts_and_keys` needs to treat an orphaned entry as skip-and-warn (like it already does for tombstoned identities) instead of rejecting the whole wallet's saved state. This DET-side PR is still correct and worth keeping — it stops new collisions — but the upstream fix is the one that actually recovers a wallet already in this state.

src/context/wallet_lifecycle/tests.rs
```diff
@@ -4105,6 +4168,171 @@ fn basic_test_identity() -> dash_sdk::dpp::identity::Identity {
...
+            backend
+                .ensure_identity_managed(&seed_hash, &resident, 0)
+                .await
+                .expect("the first identity at a free index must register"),
+        );
+
+        let error = backend
+            .ensure_identity_managed(&seed_hash, &colliding, 0)
+            .await
+            .expect_err("a second identity at a taken index must be refused");
```
thepastaclaw, line 4272: **🟡 Suggestion: Exercise the paid registration guard rather than only the reconciler helper**

The collision regression calls `ensure_identity_managed` directly for both identities, so it never executes the new `WalletBackend::register_identity` preflight. Removing the paid-path check at lines 94-95 leaves this test green. Contrary to the commit message, this path is testable offline: seed an occupied manager slot and call `register_identity`; the occupancy error is returned before the secret session, funding-account provisioning, asset-lock work, or any network request. Add that direct regression, plus barrier-controlled concurrency coverage and equivalent tests for the address-funded and existing-identity paths when the shared reservation is introduced.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The dependency-only delta from the previously reviewed head updates empty-script UTXO handling; the identity lifecycle files are unchanged, and the relevant upstream identity-manager and persister code is identical across the old and current platform pins. Two blockers remain: identity indices are not reserved across concurrent and alternative entry paths, and local deletion can outrun a durable upstream tombstone; transient funding-account persistence handling and regression-test gaps also remain open.

Source: reviewers `gpt-5.6-sol` (general, rust-quality); verifier `gpt-5.6-sol`.

### Review provenance

_4 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `src/context/wallet_lifecycle/tests.rs`:
- [SUGGESTION] src/context/wallet_lifecycle/tests.rs:4294-4304: Exercise the paid registration guard rather than only the reconciler helper
  The collision regression calls `ensure_identity_managed` directly for both identities, so it never executes the new `WalletBackend::register_identity` preflight. Removing the paid-path check at lines 94-95 leaves this test green. Contrary to the commit message, this path is testable offline: seed an occupied manager slot and call `register_identity`; the occupancy error is returned before the secret session, funding-account provisioning, asset-lock work, or any network request. Add that direct regression, plus barrier-controlled concurrency coverage and equivalent tests for the address-funded and existing-identity paths when the shared reservation is introduced.
```
</details>

src/wallet_backend/identity_ops.rs
thepastaclaw: Resolved in this update — _Reserve the identity index across every registration path_ no longer present.

thepastaclaw: Correction: the preceding auto-resolve reply was incorrect. This finding remains open at `4c1013a1672e7e8617f87731801e7eb28042aea7` and is one of the two blockers retained by the preliminary review: https://github.com/dashpay/dash-evo-tool/pull/951#pullrequestreview-4832133827

No thread resolution occurred; please treat the earlier auto-resolve message as superseded by this correction.

Claudius-Maginificent:
Merged \`v1.0-dev\` to resolve the mergeable-conflict state — it was pure adjacency (both this PR and #954 inserting a method at the same anchor line), now resolved with both additions kept.

Two conflicts auto-merged **without markers** and needed a closer look:
- #954 added a \`Funding::TopUpNotBound\` variant that made a \`match funding\` in this PR's rollback closure non-exhaustive — added the missing arm.
- This PR's new restart-survival test hardcoded the wallet DB path #954 relocated — repointed at the shared path helper.

Both fixed, verified: full \`wallet_lifecycle\` suite green (75/75), clippy clean, all four PRs' new tests confirmed passing by name.

**Flagging for review**: the added \`TopUpNotBound\` rollback arm is untested — it only fires if \`persist_account_registration\` fails, and no existing test exercises that failure path. Worth a look before merge.

thepastaclaw (changes requested):

The merged delta adds the index-less foreign-top-up funding branch, but it does not change the non-atomic identity-index preflight, best-effort removal ordering, or transient persistence chain. Two blocking data-integrity issues remain, and the paid-registration, unbound-account restart, and removal regressions still leave important behavior unproven.

Source: reviewers `gpt-5.6-sol` (general, rust-quality); verifier `gpt-5.6-sol`.

### Review provenance

<details>

```

In `src/context/wallet_lifecycle/tests.rs`:
- [SUGGESTION] src/context/wallet_lifecycle/tests.rs:4574-4581: Prove the unbound top-up account survives a restart
  The merged branch routes `Funding::TopUpNotBound` through this PR's account-registration persistence path, but this test calls `ensure_unbound_topup_funding_account` twice in the same process. That proves only that the two in-memory account collections were populated; it does not assert the `identity_topup_unbound` manifest row or verify that a cold boot reconstructs the account needed to resume a foreign-identity asset lock. The existing restart regression covers only the indexed `IdentityTopUp { registration_index }` variant. Add a cold-boot regression that provisions the unbound account, verifies its account-registration row, reloads the wallet, and confirms the account remains available for unbound top-up recovery.
```
</details>

src/context/wallet_lifecycle/tests.rs
```diff
@@ -3786,6 +3786,82 @@ async fn ensure_identity_funding_accounts_succeeds_on_cold_booted_watch_only_wal
...
+
+    let (ctx2, sender2) = offline_testnet_context_at(cold_dir.path());
+    ctx2.ensure_wallet_backend(sender2)
+        .await
+        .expect("cold boot must load the persisted wallet");
+    let backend2 = ctx2.wallet_backend().expect("backend");
+    assert!(
+        backend2.is_wallet_registered(&seed_hash),
+        "the wallet must still come back registered with the extra account row",
+    );
```
thepastaclaw, line 3885: **🟡 Suggestion: Assert that the indexed top-up account was reloaded**

`is_wallet_registered` only verifies that the reloaded backend's `id_map` contains the wallet seed hash. The preceding SQL assertion proves that the registration row exists, but neither assertion proves that cold-boot reconstruction placed `IdentityTopUp { registration_index }` and its expected xpub into the wallet account collections. Since the test and user story specifically require the account to remain usable after restart, add a test-only account accessor and assert the reloaded account type, index, and xpub.

thepastaclaw (commented):
## Final validation — Codex/Sol only (Phase 2 disabled)

The normal path correctly persists the provisioned account registration, and the manifest assertion covers the core restart regression. A transient immediate-flush error can still leave buffered persistence out of sync with rolled-back in-memory state, while restart coverage does not directly inspect the reloaded indexed account or cover the unbound variant.
Source: reviewer backend model `gpt-5.6-sol` (general and rust-quality); final verifier backend model `claude-opus-4-6`; orchestration-only, not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

Validated zero-blocker Codex/Sol precheck evidence was promoted to final because Phase 2 (Sonnet/Opus) is temporarily disabled. This is Codex/Sol-only final validation, not Codex + Sonnet/Opus coverage.

### Review provenance
- Sonnet/Opus: **not run (Phase 2 disabled — temporary Codex/Sol-only final)**
- Secondary pass: disabled (`temporary_phase2_sonnet_disable`)

🟡 2 suggestion(s)

<details>

```

In `src/context/wallet_lifecycle/tests.rs`:
- [SUGGESTION] src/context/wallet_lifecycle/tests.rs:4424-4431: Prove the unbound top-up account survives a restart
  The new persistence path also handles `Funding::TopUpNotBound`, whose manifest encoding is distinct (`identity_topup_unbound`). This test calls `ensure_unbound_topup_funding_account` twice in one process, proving only that both in-memory collections were populated. It does not verify that the unbound registration row was written or that a cold boot reconstructs the account needed to resume a foreign-identity asset lock. Add a cold-boot regression that provisions the unbound account, checks its manifest row, reloads the wallet, and confirms the account and expected xpub are present after restart.
- [SUGGESTION] src/context/wallet_lifecycle/tests.rs:3857-3860: Assert that the indexed top-up account was reloaded
  `is_wallet_registered` only verifies that the reloaded backend's `id_map` contains the wallet seed hash. The preceding SQL assertion proves that the registration row exists, but neither assertion proves that cold-boot reconstruction placed `IdentityTopUp { registration_index }` and its expected xpub into the wallet account collections. Since the test and user story specifically require the account to remain usable after restart, add a test-only account accessor and assert the reloaded account type, index, and xpub.

In `src/wallet_backend/identity_ops.rs`:
- [SUGGESTION] src/wallet_backend/identity_ops.rs:494-497: Retry the retained changeset before rolling back memory
  (existing thread: https://github.com/dashpay/dash-evo-tool/pull/951#discussion_r3691598908)
  The default `SqlitePersister` uses immediate flushing. At the pinned platform revision, a transient failure during that inline flush restores the submitted changeset to the per-wallet buffer, and the persistence contract requires retrying it through a bare `flush` without submitting the changeset again. Here, `and_then` skips that retry when `store` returns the transient error, after which the caller removes the account from both in-memory collections even though its registration remains buffered. A later unrelated persistence operation can therefore commit the row while the live wallet lacks the account. Retry transient failures with a bounded, backoff-based bare `flush`; roll back memory only for a terminal failure that discarded the staged delta. If the retry budget is exhausted while the error remains transient, preserve memory and ensure a subsequent provisioning attempt flushes the pending buffer instead of taking the current idempotent early return.
```
</details>

src/context/wallet_lifecycle/tests.rs
thepastaclaw: Resolved in this update — _Exercise the paid registration guard rather than only the reconciler helper_ no longer present.

src/context/wallet_lifecycle/tests.rs
thepastaclaw: Resolved in this update — _Assert that removal admits a distinct replacement identity_ no longer present.

src/backend_task/identity/remove_identity.rs
thepastaclaw: Resolved in this update — _Keep the local identity until index release is durable_ no longer present.

src/context/wallet_lifecycle/tests.rs
```diff
@@ -3786,6 +3786,82 @@ async fn ensure_identity_funding_accounts_succeeds_on_cold_booted_watch_only_wal
...
+/// the upstream creator that would otherwise write that row skips it once both
+/// in-memory collections already hold the account — which DET's own
+/// provisioning puts there first. A memory-only account leaves a restart
+/// between an asset-lock broadcast and its consumption unable to re-derive the
+/// credit-output path, stranding the lock and the funds in it.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn a_provisioned_identity_topup_account_survives_a_restart() {
+    let _guard = backend_reopen_lock().await;
+    let source_dir = tempfile::tempdir().expect("source tempdir");
+    let seed = [0xF3u8; 64];
```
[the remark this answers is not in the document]
thepastaclaw: This is a deterministic Testnet-only fixture, not a user seed or live credential. The suite uses fixed seed arrays extensively to make wallet derivation tests reproducible; replacing this disposable fixture with runtime randomness would not remove an exposed secret.

src/wallet_backend/identity_ops.rs
Claudius-Maginificent: Confirmed still open: `persist_account_registration` (identity_ops.rs) rolls back the in-memory account entries on any `store`/`flush` error, including a transient one where the upstream persister retains the changeset in its per-wallet buffer per `PlatformWalletPersistence`. A later unrelated `store` call can flush that retained buffer and durably commit the registration while the in-memory account maps no longer have it.

Not fixed in this revision — filing for follow-up: distinguish `source.is_transient()` and retry via a bare `flush` before rolling back memory; only roll back on a terminal failure that actually discarded the buffered delta.

src/context/wallet_lifecycle/tests.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Confirmed still open: `a_provisioned_identity_topup_account_survives_a_restart` only asserts the `account_registrations` row count and `is_wallet_registered(&seed_hash)` after cold boot. Neither proves the reloaded `IdentityTopUp { registration_index }` account (and its xpub) actually landed in the wallet's managed account collections — a loader that drops the account would still pass.

Not fixed in this revision — needs a test-only accessor to assert the reloaded account type/index/xpub. (Same gap also applies to the unbound top-up variant, per the latest coderabbitai/thepastaclaw review — not covered by an inline thread since that line is outside the diff.)

Claudius-Maginificent: Confirmed still open: `a_provisioned_identity_topup_account_survives_a_restart` only asserts the `account_registrations` row count and `is_wallet_registered(&seed_hash)` after cold boot. Neither proves the reloaded `IdentityTopUp { registration_index }` account (and its xpub) actually landed in the wallet's managed account collections — a loader that drops the account would still pass.

Not fixed in this revision — needs a test-only accessor to assert the reloaded account type/index/xpub. (Same gap also applies to the unbound top-up variant, per the latest coderabbitai/thepastaclaw review — not covered by an inline thread since that line is outside the diff.)

src/backend_task/identity/remove_identity.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Closing as out of scope for this PR. This finding was raised against `release_identity_index`/durable-tombstone handling in `remove_identity.rs`, but that work isn't part of this PR anymore — the current diff only touches `identity_ops.rs`, `tests.rs`, and `error.rs`, and `remove_identity.rs` on this branch is identical to `v1.0-dev` (42 lines, no `forget_identity`/`release_identity_index` call at all). Looks like this PR's scope was narrowed to just the account-persistence fix at some point after the earlier reviews. Not tracked in another open PR right now — leaving that as a known gap rather than filing a new issue, per current instructions.

src/wallet_backend/identity_ops.rs
```diff
@@ -55,6 +85,15 @@ impl WalletBackend {
...
+
```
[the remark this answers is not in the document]
Claudius-Maginificent: Closing as out of scope for this PR. This targets a `reject_taken_identity_index`-style preflight around lines 88-95 that no longer exists — `grep -rn reject_taken_identity_index src/` finds nothing, and the current `identity_ops.rs` diff for this PR only adds `persist_account_registration` (account-persistence fix), not an index-reservation guard. Same scope-narrowing as the sibling `remove_identity.rs` findings.

src/wallet_backend/identity_ops.rs
Claudius-Maginificent: Closing as out of scope for this PR. Confirmed the correction below is right that this was a real blocker at the time — but the shared per-`(wallet_id, identity_index)` reservation it describes isn't present anywhere in the repo today (`grep -rn "reject_taken_identity_index\|identity_index.*reserv" src/` finds nothing), and this PR's current diff doesn't touch the registration-preflight code path at all. This PR's scope has been narrowed since the July reviews to just the funding-account persistence fix. Not tracked in another open PR right now — leaving that as a known gap rather than filing a new issue, per current instructions.

src/backend_task/identity/remove_identity.rs
Claudius-Maginificent: Closing as out of scope for this PR, same as the sibling finding above — `remove_identity.rs` isn't in this PR's current diff and has no `forget_identity`/`release_identity_index` call to make durable. This blocker applied to an earlier, larger revision of this branch; the current scope is limited to the identity-funding-account persistence fix.

src/context/wallet_lifecycle/tests.rs
Claudius-Maginificent: Closing as out of scope for this PR. The test this nitpick targets (asserting removal admits a distinct replacement identity via `RemoveIdentity`) doesn't exist in the current `tests.rs` — the removal-regression coverage this comment refers to isn't part of this PR's diff. Same scope-narrowing as the other stale threads on this PR.

src/context/wallet_lifecycle/tests.rs
Claudius-Maginificent: Closing as out of scope for this PR. This targets a "new preflight" in `register_identity` (lines 94-95 in the prior revision) that isn't present in the current code — there's no occupancy/reservation check to exercise. This PR's current diff is limited to the identity-funding-account persistence fix; the registration-preflight work this test was meant to cover isn't part of it.

src/wallet_backend/identity_ops.rs
```diff
@@ -34,6 +53,216 @@ enum Funding {
...
+    entries: &[AccountRegistrationEntry],
+) -> Result<(), RegistrationPersistFailure> {
+    retry_transient(|| {
+        let changeset = PlatformWalletChangeSet {
+            account_registrations: entries.to_vec(),
+            ..Default::default()
+        };
+        persister
+            .store(wallet_id, changeset)
+            .and_then(|()| persister.flush(wallet_id))
```
thepastaclaw: **🔴 Blocking: Do not roll back an account after its immediate store committed**

`WalletBackend::new` constructs the persister with `SqlitePersisterConfig::new`, whose default `FlushMode::Immediate` makes a successful `store` durable before it returns. The following `flush` therefore cannot improve the durability of this registration; it operates on the shared per-wallet buffer. A concurrent writer can place an unrelated changeset in that buffer between the two calls, and this flush can take that changeset and return its terminal error. The caller then classifies the registration as discarded and evicts the newly created account at lines 738-748 even though its manifest row is already durable, producing a false funding failure and an in-memory/on-disk disagreement. Treat the successful immediate-mode `store` as the durability acknowledgement and add coverage for a successful registration store followed by a terminal failure from the redundant flush.

```suggestion
persister.store(wallet_id, changeset)
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The current head resolves all three prior findings: transient failures are retried without losing live account state, and both indexed and unbound top-up accounts are verified after cold boot. One blocking concurrency defect remains: the redundant explicit flush can fail on another writer's changeset and roll back an account whose registration was already committed.
Source: reviewers `gpt-5.6-sol` (general, rust-quality); final verifier `claude-opus-4-6`; orchestration-only `openclaw-agent/cliproxy/gpt-5.6-sol` is not reviewer evidence.

Validated blockers were found in the Codex precheck. Opus is deferred until a fresh Codex revalidation clears the blocker gate.

### Review provenance

🔴 1 blocking

<details>

```

In `src/wallet_backend/identity_ops.rs`:
- [BLOCKING] src/wallet_backend/identity_ops.rs:259-261: Do not roll back an account after its immediate store committed
  `WalletBackend::new` constructs the persister with `SqlitePersisterConfig::new`, whose default `FlushMode::Immediate` makes a successful `store` durable before it returns. The following `flush` therefore cannot improve the durability of this registration; it operates on the shared per-wallet buffer. A concurrent writer can place an unrelated changeset in that buffer between the two calls, and this flush can take that changeset and return its terminal error. The caller then classifies the registration as discarded and evicts the newly created account at lines 738-748 even though its manifest row is already durable, producing a false funding failure and an in-memory/on-disk disagreement. Treat the successful immediate-mode `store` as the durability acknowledgement and add coverage for a successful registration store followed by a terminal failure from the redundant flush.
```
</details>

src/context/wallet_lifecycle/tests.rs
thepastaclaw: Resolved in `1330c1b` — _Assert that the indexed top-up account was reloaded_ no longer present.

src/wallet_backend/identity_ops.rs
thepastaclaw: Resolved in this update — _Retry the retained changeset before rolling back memory_ no longer present.

src/context/wallet_lifecycle/tests.rs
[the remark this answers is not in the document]
thepastaclaw: This was subsequently fixed at the exact head. The test now retrieves the cold-booted `IdentityTopUp` account through `identity_funding_account` and compares its complete view with the account captured before restart.

src/wallet_backend/identity_ops.rs
lklimek: Verified and fixed in `39695b103` (plus `ad7997929`).

Every premise checked out against the pinned upstream (`762c66c`): `WalletBackend::new` builds the persister via `SqlitePersisterConfig::new` with no `.with_flush_mode`, leaving the default `FlushMode::Immediate`, under which `store` calls `flush_inner` internally — so a successful `store` has already committed. The trailing `flush` could add no durability, and because the buffer is shared per wallet it could adopt a changeset another writer parked there after its own transient failure and hand us that stranger's terminal error, classifying an already-durable registration as `Discarded` and evicting its account from memory.

`persist_account_registrations` now issues a single `store`, with the flush-mode assumption and its establishing site recorded on the function.

On the requested coverage: a test for "successful store followed by a terminal flush" is now permanently unwritable, because that sequence no longer exists. The substitute is `a_committed_registration_write_is_not_followed_by_a_second_write`, which asserts exactly 1 `store` and 0 `flush` calls (plus account live and `managed`, nothing staged, one durable row after shutdown). It was negative-controlled: reintroducing `.and_then(flush)` turns it red on `left: 1, right: 0`.

Two adjacent defects surfaced during verification and are fixed in the same change: the fault-injection harness documented a per-`flush` fault capability that was unreachable at its only call site — the dead capability is removed, the three misleading doc comments corrected, and `store`/`flush` call counters added in its place — and a stale comment describing a bare-flush retry shape that no longer exists was rewritten.

One correction to the finding itself: the 🔴 Blocking label is inflated. No funds move before provisioning succeeds, nothing deletes `account_registrations` rows, and the next attempt re-UPSERTs the identical deterministic xpub, so the divergence is in-memory and self-healing. Real defect, worth fixing, but LOW and non-blocking.

thepastaclaw (commented):
## Final validation — Codex/Sol only (Phase 2 disabled)

The exact head fixes the prior redundant-flush blocker by treating an inline-committing store as the durability boundary and retaining an explicit flush only for buffering persisters. One non-blocking durability ambiguity remains when a transient failure allows another writer to commit the shared buffer before this call encounters a terminal retry failure.
Source: reviewers `gpt-5.6-sol` (general, rust-quality); final verifier `claude-opus-4-6`. Orchestration-only `openclaw-agent/cliproxy/gpt-5.6-sol` is not reviewer evidence.

Validated zero-blocker Codex/Sol precheck evidence was promoted to final because Phase 2 (Sonnet/Opus) is temporarily disabled. This is Codex/Sol-only final validation, not Codex + Sonnet/Opus coverage.

### Review provenance
- Sonnet/Opus: **not run (Phase 2 disabled — temporary Codex/Sol-only final)**
- Secondary pass: disabled (`temporary_phase2_sonnet_disable`)

🟡 1 suggestion(s)

<details>

```

In `src/wallet_backend/identity_ops.rs`:
- [SUGGESTION] src/wallet_backend/identity_ops.rs:215-220: Preserve the account when a prior transient makes durability ambiguous
  A terminal error is always classified as `Discarded`, even after an earlier transient attempt exposed this registration through the shared per-wallet buffer. During the retry backoff, an unrelated writer for the same wallet can successfully drain and commit that buffer, including this registration; the provisioning mutex does not cover those writers. If the re-supplied retry then fails terminally, it discards only the new attempt and does not undo the row already committed by the other writer. The current classification consequently evicts the account at lines 972-982 even though its manifest row is durable. Track whether this retry loop has observed a transient failure. A later terminal result is durability-unknown, so retain the account, record the pending rewrite marker, and return the typed error; only a terminal failure before any transient attempt proves that this call's registration never escaped. Add a regression covering a transient failure, a successful foreign buffer commit, and a terminal retry.
```
</details>

src/wallet_backend/identity_ops.rs
```diff
@@ -41,6 +58,229 @@ enum Funding {
...
+{
+    let mut backoff = PERSIST_RETRY_INITIAL_BACKOFF;
+    let mut attempt: u32 = 1;
+    loop {
+        match op() {
+            Ok(()) => return Ok(()),
+            Err(e) if !e.is_transient() => return Err(RegistrationPersistFailure::Discarded(e)),
+            Err(e) if attempt >= PERSIST_RETRY_MAX_ATTEMPTS => {
+                return Err(RegistrationPersistFailure::Staged(e));
+            }
```
thepastaclaw, line 218: **🟡 Suggestion: Preserve the account when a prior transient makes durability ambiguous**

A terminal error is always classified as `Discarded`, even after an earlier transient attempt exposed this registration through the shared per-wallet buffer. During the retry backoff, an unrelated writer for the same wallet can successfully drain and commit that buffer, including this registration; the provisioning mutex does not cover those writers. If the re-supplied retry then fails terminally, it discards only the new attempt and does not undo the row already committed by the other writer. The current classification consequently evicts the account at lines 972-982 even though its manifest row is durable. Track whether this retry loop has observed a transient failure. A later terminal result is durability-unknown, so retain the account, record the pending rewrite marker, and return the typed error; only a terminal failure before any transient attempt proves that this call's registration never escaped. Add a regression covering a transient failure, a successful foreign buffer commit, and a terminal retry.

<sub>source: ['codex']</sub>

src/context/wallet_lifecycle/tests.rs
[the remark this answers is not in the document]
thepastaclaw: This was open when the reply was written, but it is fixed at the exact head. The test now retrieves the cold-booted `IdentityTopUp` account through `identity_funding_account` and compares its complete view with the account captured before restart.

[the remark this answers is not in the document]
thepastaclaw: This automated mute comment contains no review finding to validate. The underlying restart-assertion finding is fixed at the exact head.

PR #955, opened 2026-08-01
fix(identity): register masternode/evonode identities without requiring a wallet

Claudius-Maginificent:
**TL;DR:** Dash Evo Tool can now track masternode/evonode identities that have no owning wallet, so they survive restarts and wallet removal instead of silently vanishing.

## User story
As a **masternode operator running Dash Evo Tool without a funded wallet**, I want an identity I've registered or loaded to persist across restarts and survive removing any wallet, so I don't lose track of nodes I'm monitoring.

### Base flow
An operator loads or registers an identity for a masternode/evonode that isn't tied to any wallet in the app.

### Actual behavior
DET's local record for that identity had no durable, wallet-independent backing. Because the underlying storage keyed identity data to a wallet, the identity's associated data could be silently dropped — e.g. removing an unrelated wallet could cascade-delete metadata that happened to be co-located with it.

### Expected behavior
The identity is written through a dedicated "unowned" scope (no wallet association, a NULL/sentinel wallet reference upstream) so it survives removing every wallet in the app and reconciles correctly at startup — including on an install with zero wallets at all, the pure masternode-operator case.

### What was done
- `ensure_identity_unowned` / `remove_unowned_identity` (originally `register_`/`forget_unowned_identity`, renamed for consistency with the module's other idempotent-registration methods) added to `WalletBackend` (`src/wallet_backend/identity_ops.rs`).
- Write-through wired at `insert_local_qualified_identity` for every identity stored without a wallet association, mirrored on delete at `delete_local_qualified_identity` (`src/context/identity_db.rs`) — otherwise a node the user removes would linger upstream forever.
- `reconcile_unowned_identities` runs at boot, ahead of and independent of the per-wallet bootstrap passes (`src/context/wallet_lifecycle/bootstrap.rs`).
- `CHANGELOG.md` and `docs/kv-keys.md` updated with the scope and why it's load-bearing.

This branch also carries the base-branch merge from `v1.0-dev` (including the platform pin bump to PR #3968 head, `4784de03`, and its fixed-closed load-failure policy — unrelated to this feature, picked up by the merge) and a lockfile-drift fix from that merge.

### Review round
Independently reviewed for architectural direction and code quality (two parallel passes, cross-checked against a prior kv-storage investigation). Verdict: **good direction, merge-worthy as designed** — the NULL-scope decision is sound and well-evidenced. Every finding raised was then acted on:

- **Docs**: three stale/misleading claims corrected — `delete_local_qualified_identity`'s premise that DET never writes to the upstream `identities` table (it now tombstones a wallet-less identity's row there; the *conclusion*, that the delete cascade never fires for this path, still holds, for the right reason), four places implying "wallet-less" means only masternode/evonode, and `docs/kv-keys.md`'s section contradicting its own opener and omitting the removal half. Both new public methods gained `# Errors` sections and now disclose the mirror's real limitations (no public keys, write-once) — [tracked upstream](https://github.com/dashpay/platform/issues/4443), where two more findings were filed as follow-up context, plus a separate memcan TODO for an upstream swallowed-error case.
- **API**: `ensure_identity_unowned` now returns `Result<bool, TaskError>` (matching `ensure_identity_managed`'s contract) instead of `Result<()>`, so the boot reconcile logs real registrations instead of every attempt.
- **Two-way reconcile**: the boot reconcile was add-only — a tombstone lost to a crash or a not-yet-wired backend at delete time left upstream permanently advertising a node the device no longer has. It now withdraws every upstream registration whose sidecar record is gone too, closing that drift; a new test (`reconcile_withdraws_a_stale_registration_the_sidecar_no_longer_has`) proves it.
- **Boot cost**: the reconcile used to fully hydrate — decode + a secret-vault key migration — every wallet-less identity on every boot just to diff ids. It now diffs two cheap, pre-decode id sets (sidecar vs. upstream) and hydrates only an id it's actually about to register; a steady-state boot touches neither the blob decoder nor the vault.
- **Test coverage**: five new tests close the gaps found — the dangerous path the existing wallet-owned test didn't reach (`&None` against an identity already wallet-owned upstream, pinning upstream's own upsert guard rather than trusting it silently), a wallet-less `User` identity (not just nodes) getting mirrored too, the update-path-then-reconcile path, direct idempotency at the `WalletBackend` layer, and the removal half above. One existing test's fixture was also fixed — it simulated the pre-upgrade state via a tombstone (revive-a-tombstone path) rather than a genuinely absent row (the real upgrade path); a new test-only helper writes the true shape.

The one item left as "hygiene, not a performance defect" per the original review's own framing: `ensure_identity_unowned`/`remove_unowned_identity` each still rebuild their own upstream manager per call rather than sharing one across a reconcile batch. Small tables; not pursued further here.

### Testing
- 10 new tests total (5 original + 5 from the review round) covering write/read/reconcile/removal/idempotency/update-path/wallet-owned-guard, including proof the hazard this closes is real: a wallet-scoped version of the same scenario still fails (removing every wallet loses the identity) while the unowned-scope version survives.
- `wallet_lifecycle::` suite (82 tests, includes all of the above plus the merged-in pin bump's tests): 82 passed, 0 failed.
- `cargo clippy --all-features --all-targets -- -D warnings`: clean.
- `cargo fmt --all` / `cargo doc --no-deps --all-features`: clean.

### Breaking changes
None.

### Checklist
- [x] Tests added
- [x] `cargo fmt` / `cargo clippy` clean
- [x] CHANGELOG updated

### Attribution

## Summary by CodeRabbit

- **New Features**
  - ProTxHash-loaded masternodes and evonodes are stored independently of wallets.
  - Existing wallet-independent nodes and identities are imported automatically when the app starts.
  - Wallet-independent identities remain available after wallet removal.

- **Bug Fixes**
  - Added reconciliation to restore missing registrations and remove stale ones.
  - Improved handling of identity ownership changes, deletion cleanup, and temporary storage failures.
  - Added verification and warnings when identity changes cannot be confirmed.
  - Wallet-linked identities are preserved when ownership cannot be conclusively determined.


thepastaclaw:
⛔ Blockers found — Opus deferred (commit df503af2)

src/context/identity_db.rs
```diff
@@ -598,10 +598,69 @@ impl AppContext {
             wallet_index,
         };
         index_add_identity(&kv, &id)?;
+        kv.put(DetScope::Identity(&id), IDENTITY_KEY, &stored)
+            .map_err(identity_err)?;
+        if wallet_hash.is_none() {
+            self.register_unowned_identity_upstream(qualified_identity);
```
thepastaclaw: **🔴 Blocking: A wallet-owned row cannot be converted to the durable unowned scope**

These lines persist `wallet_hash: None` in the sidecar before attempting the best-effort unowned registration. At the pinned platform revision, `add_out_of_wallet_identity` swallows `persister.store` failures, and the SQLite upsert silently skips a sentinel-scoped write when the existing identity row belongs to a real wallet. Consequently, the mutation can report success while the sidecar says wallet-less, the upstream row remains attached to the old wallet, and no unowned registration exists. The new `none_insert_of_a_wallet_owned_identity_does_not_disturb_the_upstream_row` test confirms this contradictory state. Removing the old wallet then cascades through its identity row and deletes the `meta_identity` sidecar, defeating this PR's durability guarantee. Verify the persisted postcondition after unowned mutations and either preserve the existing wallet association when the mirror cannot be established or explicitly re-scope the row safely before committing the wallet-less sidecar. The regression test should remove the old wallet and assert that the local identity survives; removal mutations should likewise verify that the tombstone actually persisted before reporting success.

<sub>source: ['codex']</sub>

src/context/wallet_lifecycle/bootstrap.rs
```diff
@@ -275,6 +275,97 @@ impl AppContext {
...
+        if added > 0 {
+            tracing::info!(
+                added,
+                "Registered wallet-less identities with the wallet store"
+            );
+        }
+
+        let mut removed = 0usize;
+        for id in registered.difference(&wallet_less) {
+            match backend.remove_unowned_identity(id) {
```
thepastaclaw, line 421: **🟡 Suggestion: Reconciliation can withdraw a concurrently inserted identity**

The reconcile snapshots the sidecar before the upstream scope. An insertion that completes between those reads is absent from `wallet_less` but present in `registered`, so the removal loop immediately tombstones the registration that the successful insertion just created. This is reachable because `ensure_wallet_backend` publishes the backend before awaiting `bootstrap_loaded_wallets`; another task can take the populated-backend fast path and insert an identity while reconciliation is running. Reverse the snapshot order or revalidate that each removal candidate is still absent from the current wallet-less sidecar set immediately before withdrawing it. A mere `has_local_qualified_identity` check is too broad because a wallet-owned sidecar should not preserve a stale unowned registration.

<sub>source: ['codex', 'coderabbit']</sub>

src/wallet_backend/identity_ops.rs
```diff
@@ -158,6 +165,151 @@ impl WalletBackend {
...
+    ///
+    /// `AppContext::reconcile_unowned_identities` diffs this set against its
+    /// own cheap sidecar id scan on both sides — registering what's missing
+    /// here and withdrawing what's missing there — so this is also the read
+    /// half of that reconcile's two-way contract, not only a test accessor.
+    pub(crate) fn unowned_identity_ids(
+        &self,
+    ) -> Result<std::collections::BTreeSet<dash_sdk::platform::Identifier>, TaskError> {
+        Ok(self.load_unowned_identities()?.into_keys().collect())
+    }
```
thepastaclaw, line 189: **🟡 Suggestion: The upstream side of the ID diff still hydrates complete identities**

`unowned_identity_ids` calls `load_unowned_identities`, which decodes every persisted `IdentityEntry`, loads keys, contacts, and tombstones, and constructs a `ManagedIdentity` map before `into_keys()` discards all values. Therefore the steady-state boot scan is not a pre-decode ID-only read, contrary to the rustdoc in `reconcile_unowned_identities` and the PR description. Add or use an ID-only storage accessor for this scan. If full hydration is intentionally acceptable for the expected table size, correct the performance documentation instead of claiming that neither side decodes identity rows.

<sub>source: ['codex']</sub>

src/context/identity_db.rs
```diff
@@ -1051,14 +1149,22 @@ impl AppContext {
...
+    /// blob, the top-up history, and every scheduled vote queued for this
+    /// identity — and removes the Global index entries that the trigger
+    /// would not touch.
+    ///
+    /// For a wallet-less identity this also *tombstones* (never row-deletes)
+    /// its mirrored row in the upstream unowned scope, via
+    /// [`WalletBackend::remove_unowned_identity`](crate::wallet_backend::WalletBackend::remove_unowned_identity)
+    /// below — so upstream stops advertising a node this device no longer
+    /// has. Best-effort and, unlike registration, not retried on failure: a
+    /// tombstone lost here stays lost until the identity is re-inserted.
```
thepastaclaw: **🟡 Suggestion: Removal documentation contradicts the two-way reconcile**

This rustdoc says failed removals are never retried, but `reconcile_unowned_identities` computes `registered.difference(&wallet_less)` and retries stale withdrawals at boot. The same obsolete add-only description remains in `src/wallet_backend/identity_ops.rs:252-257` and `docs/kv-keys.md:94`. Update all three sites to state that both missing registrations and stale registrations are retried by the next boot reconciliation.

<sub>source: ['codex', 'coderabbit']</sub>

src/context/identity_db.rs
```diff
@@ -772,12 +831,51 @@ impl AppContext {
...
+
     /// The masternode/evonode identities for the active network — the
     /// Masternodes-page card list and the page-scoped masternode pill source.
     /// The complement of [`Self::load_local_user_identities`] over the FR-6 type
-    /// boundary. Filters the hydrated full load, so each card's top-up history
-    /// is available (unlike the pre-decode [`Self::load_local_voting_identities`],
-    /// which is un-hydrated and named for the DPNS voting flows).
+    /// boundary (`identity_type`) — a different partition from
+    /// [`Self::load_local_wallet_less_identities`] (`wallet_hash`); the two
+    /// sets need not match. Filters the hydrated full load, so each card's
```
thepastaclaw, line 1032: **💬 Nitpick: Rustdoc still links to the removed wallet-less loader**

The link names `load_local_wallet_less_identities`, but that method does not exist after the ID-only refactor; the replacement is `local_wallet_less_identity_ids`. This leaves a newly introduced unresolved intra-doc link.

```suggestion
    /// [`Self::local_wallet_less_identity_ids`] (`wallet_hash`); the two
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The dedicated unowned scope and two-way reconciliation are sound in principle, but the wallet-owned-to-wallet-less path can still leave the sidecar attached to a wallet-owned upstream row and lose the identity when that wallet is removed. The reconciliation race and mutation postcondition gaps also weaken immediate consistency, while several performance and lifecycle claims in the documentation do not match the implementation.
Source: Claude reviewer (exact backend model ID was not supplied), Codex general and Rust-quality reviewers (exact backend model IDs were not supplied), and Claude Agent SDK final verifier (exact backend model ID was not exposed). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and is not reviewer evidence.

Validated blockers were found in the Codex precheck. Opus is deferred until a fresh Codex revalidation clears the blocker gate.

### Review provenance

🔴 1 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

<details>

```

- [BLOCKING] src/context/identity_db.rs:601-604: A wallet-owned row cannot be converted to the durable unowned scope
  These lines persist `wallet_hash: None` in the sidecar before attempting the best-effort unowned registration. At the pinned platform revision, `add_out_of_wallet_identity` swallows `persister.store` failures, and the SQLite upsert silently skips a sentinel-scoped write when the existing identity row belongs to a real wallet. Consequently, the mutation can report success while the sidecar says wallet-less, the upstream row remains attached to the old wallet, and no unowned registration exists. The new `none_insert_of_a_wallet_owned_identity_does_not_disturb_the_upstream_row` test confirms this contradictory state. Removing the old wallet then cascades through its identity row and deletes the `meta_identity` sidecar, defeating this PR's durability guarantee. Verify the persisted postcondition after unowned mutations and either preserve the existing wallet association when the mirror cannot be established or explicitly re-scope the row safely before committing the wallet-less sidecar. The regression test should remove the old wallet and assert that the local identity survives; removal mutations should likewise verify that the tombstone actually persisted before reporting success.
- [SUGGESTION] src/context/identity_db.rs:1162-1167: Removal documentation contradicts the two-way reconcile
  This rustdoc says failed removals are never retried, but `reconcile_unowned_identities` computes `registered.difference(&wallet_less)` and retries stale withdrawals at boot. The same obsolete add-only description remains in `src/wallet_backend/identity_ops.rs:252-257` and `docs/kv-keys.md:94`. Update all three sites to state that both missing registrations and stale registrations are retried by the next boot reconciliation.
- [NITPICK] src/context/identity_db.rs:870-875: Rustdoc still links to the removed wallet-less loader
  The link names `load_local_wallet_less_identities`, but that method does not exist after the ID-only refactor; the replacement is `local_wallet_less_identity_ids`. This leaves a newly introduced unresolved intra-doc link.

In `src/context/wallet_lifecycle/bootstrap.rs`:
- [SUGGESTION] src/context/wallet_lifecycle/bootstrap.rs:300-352: Reconciliation can withdraw a concurrently inserted identity
  The reconcile snapshots the sidecar before the upstream scope. An insertion that completes between those reads is absent from `wallet_less` but present in `registered`, so the removal loop immediately tombstones the registration that the successful insertion just created. This is reachable because `ensure_wallet_backend` publishes the backend before awaiting `bootstrap_loaded_wallets`; another task can take the populated-backend fast path and insert an identity while reconciliation is running. Reverse the snapshot order or revalidate that each removal candidate is still absent from the current wallet-less sidecar set immediately before withdrawing it. A mere `has_local_qualified_identity` check is too broad because a wallet-owned sidecar should not preserve a stale unowned registration.

In `src/wallet_backend/identity_ops.rs`:
- [SUGGESTION] src/wallet_backend/identity_ops.rs:178-182: The upstream side of the ID diff still hydrates complete identities
  `unowned_identity_ids` calls `load_unowned_identities`, which decodes every persisted `IdentityEntry`, loads keys, contacts, and tombstones, and constructs a `ManagedIdentity` map before `into_keys()` discards all values. Therefore the steady-state boot scan is not a pre-decode ID-only read, contrary to the rustdoc in `reconcile_unowned_identities` and the PR description. Add or use an ID-only storage accessor for this scan. If full hydration is intentionally acceptable for the expected table size, correct the performance documentation instead of claiming that neither side decodes identity rows.
```
</details>

CHANGELOG.md
```diff
@@ -101,6 +101,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 ### Fixed
 
+- **Masternodes and evonodes are now recorded in the wallet store too**: a node
+  loaded from its ProTxHash was only ever written to this app's own records, so
+  the wallet store the app shares with the rest of the wallet stack had no entry
+  for it at all. Such a node is now recorded there as belonging to no wallet,
+  which is what it is — nodes already on this device are added the next time the
+  app starts, with nothing to press. Recording it under one of your wallets was
+  not an option: removing that wallet would then have deleted the node along
+  with it, keys and alias included.
```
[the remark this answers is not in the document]
thepastaclaw: The changelog accurately describes the PR's stated user-facing scope. It does not claim that only masternodes and evonodes use the generalized wallet-less storage path, so omitting the unexpected wallet-less User case is not contradictory.

src/context/wallet_lifecycle/bootstrap.rs
```diff
@@ -275,6 +275,97 @@ impl AppContext {
...
+                Ok(()) => removed += 1,
+                Err(error) => tracing::debug!(
+                    identity = %id,
+                    %error,
+                    "Unowned-identity removal deferred; will retry at next boot"
+                ),
+            }
+        }
```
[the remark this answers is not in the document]
thepastaclaw: The race is valid, but `has_local_qualified_identity` is too broad: it would also preserve an unowned registration when the current sidecar exists but is wallet-owned. Reverse the snapshot order or re-check whether the current sidecar specifically has `wallet_hash: None` before withdrawing.

src/context/identity_db.rs
Claudius-Maginificent: Fixed in 2f808fe19/72250ce73: `insert_local_qualified_identity` now readbacks the mirror write and, on `UnownedIdentityMirrorMissing`, keeps the existing wallet link instead of flipping the sidecar to unowned. Regression test `none_insert_of_a_wallet_owned_identity_does_not_disturb_the_upstream_row` now asserts the sidecar link survives.

src/context/wallet_lifecycle/bootstrap.rs
Claudius-Maginificent: Fixed in 2f808fe19: the withdrawal loop now re-checks `stored_identity_is_wallet_less` under the same `identity_record_lock` the insert path holds, before withdrawing — wallet-less-ness specifically, not mere existence (per your note that a generic existence guard would be too broad). New test: `reconcile_does_not_withdraw_an_identity_inserted_between_its_two_snapshots`.

src/wallet_backend/identity_ops.rs
Claudius-Maginificent: Accepted as a documented limitation rather than reworked — upstream exposes no id-only read door at this pin, so `unowned_identity_ids` still fully hydrates via `load_unowned_identities`. Rustdoc on both `unowned_identity_ids` and `reconcile_unowned_identities` no longer claims a pre-decode/ID-only read on the upstream side; it now says so explicitly.

src/context/identity_db.rs
Claudius-Maginificent: Fixed in 72250ce73: all 3 stale "not retried on failure" claims (identity_db.rs, docs/kv-keys.md, wallet_backend/identity_ops.rs) rewritten to describe the actual two-way reconcile that both registers and withdraws every boot.

src/context/identity_db.rs
Claudius-Maginificent: Fixed: the dead rustdoc link to the removed `load_local_wallet_less_identities` now points to `local_wallet_less_identity_ids`, which exists.

thepastaclaw (changes requested):

The dedicated unowned scope and lock-protected two-way reconciliation are sound, and the prior race and documentation defects are fixed. However, two fall-through paths still persist a wallet-less sidecar without a verified durable unowned mirror, allowing wallet removal to delete a record DET reports as wallet-less; this remains blocking.
Source: Codex reviewers gpt-5.6-sol (general and rust-quality); final verifier backend model grok-4.5 via Claude Agent SDK. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

Validated blockers were found in the Codex precheck. Opus is deferred until a fresh Codex revalidation clears the blocker gate.

### Review provenance

🔴 1 blocking

<details>

```

- [BLOCKING] src/context/identity_db.rs:653-668: A wallet-owned row cannot be converted to the durable unowned scope
  (existing thread: https://github.com/dashpay/dash-evo-tool/pull/955#discussion_r3830585869)
  These fall-through branches still commit `wallet_hash: None` without verifying that the upstream unowned mirror exists. There are two destructive cases: a transient `load_unowned_identities` failure discards an existing sidecar wallet link even when the upstream row remains wallet-owned, and `UnownedIdentityMirrorMissing` writes a wallet-less sidecar when no previous sidecar link exists even though the upstream row may already belong to a wallet. On the next reconcile, the guarded upstream upsert still cannot re-scope that row; removing its wallet then deletes the upstream identity and triggers deletion of the sidecar DET reports as wallet-less. The read-failure test misses this because it never registers the identity with the upstream wallet manager. Abort the mutation whenever the durable mirror cannot be verified, or recover and persist the actual upstream wallet owner; add regressions for both a wallet-owned upstream row with a failed unowned read and a wallet-owned upstream row with no existing sidecar, including wallet removal.
```
</details>

src/wallet_backend/identity_ops.rs
```diff
@@ -158,6 +165,195 @@ impl WalletBackend {
...
+        let mut manager = self.unowned_identity_manager()?;
+        if manager.identity(identity_id).is_none() {
+            return Ok(());
+        }
+        manager
+            .remove_identity(identity_id, &self.unowned_scope_persister())
+            .map(|_| ())
+            .map_err(|e| TaskError::WalletBackend {
+                source: Arc::new(e),
+            })
```
thepastaclaw: **🟡 Suggestion: Verify removal because upstream swallows tombstone persistence failures**

At the pinned platform revision, `IdentityManager::remove_identity` calls `persist_removal`, which logs `persister.store(...)` failures and still returns success. This wrapper can consequently return `Ok(())`, increment the reconciliation removal count, and report a successful deletion even though the unowned row remains on disk. The next boot can retry, so this is not data loss, but the new API's immediate result and logs are incorrect. Read the unowned scope back after removal and return a dedicated typed `TaskError` when the identity remains present; add a swallowed-removal test seam analogous to `swallow_next_unowned_write`.

<sub>source: ['codex']</sub>

docs/kv-keys.md
```diff
@@ -89,7 +89,13 @@ The identity blob and top-up history are **identity-scoped** (`DetScope::Identit
...
+
+`wallet_hash` is decided by that mirror as well as by the caller: `insert_local_qualified_identity` writes the caller's `None` unless the record **already had** a wallet link **and** the unowned row was read back absent, in which case the existing link is kept and the user is warned. A wallet-store read that merely failed is not proof of absence and leaves the caller's `None` standing, for the boot reconcile to mirror later. A proven-absent mirror on an identity with **no** prior link still writes `None` — that is the pre-existing mismatch this design does not repair, and the boot reconcile reports it at `warn` on every boot.
```
thepastaclaw: **🟡 Suggestion: Document the current wallet-scope resolution behavior**

This paragraph still describes the behavior before the latest fail-closed changes. The implementation now keeps an existing wallet link on any mirror error, not only a proven-absent readback; it refuses an unestablished insert on either a read failure or a missing mirror and persists nothing; and only an already-wallet-less record is allowed to proceed without verification. Update the paragraph to match the three `StoredWalletScope` branches and the corresponding regression tests.

thepastaclaw (changes requested):

The new verification logic fixes the previously identified linked-record and no-record fall-throughs, but the already-wallet-less branch still permits a sidecar to remain attached to a wallet-owned upstream row, leaving the PR's wallet-removal durability guarantee incomplete. Removal also needs a persistence postcondition check, and the key-storage documentation still describes superseded behavior.
Source: Codex reviewers gpt-5.6-sol (general and rust-quality); final verifier: Claude via Anthropic Claude Agent SDK (exact backend model ID was not exposed). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

Validated blockers were found in the Codex precheck. Opus is deferred until a fresh Codex revalidation clears the blocker gate.

### Review provenance

<details>

```

In `src/wallet_backend/identity_ops.rs`:
- [SUGGESTION] src/wallet_backend/identity_ops.rs:305-310: Verify removal because upstream swallows tombstone persistence failures
  At the pinned platform revision, `IdentityManager::remove_identity` calls `persist_removal`, which logs `persister.store(...)` failures and still returns success. This wrapper can consequently return `Ok(())`, increment the reconciliation removal count, and report a successful deletion even though the unowned row remains on disk. The next boot can retry, so this is not data loss, but the new API's immediate result and logs are incorrect. Read the unowned scope back after removal and return a dedicated typed `TaskError` when the identity remains present; add a swallowed-removal test seam analogous to `swallow_next_unowned_write`.

In `docs/kv-keys.md`:
- [SUGGESTION] docs/kv-keys.md:96: Document the current wallet-scope resolution behavior
  This paragraph still describes the behavior before the latest fail-closed changes. The implementation now keeps an existing wallet link on any mirror error, not only a proven-absent readback; it refuses an unestablished insert on either a read failure or a missing mirror and persists nothing; and only an already-wallet-less record is allowed to proceed without verification. Update the paragraph to match the three `StoredWalletScope` branches and the corresponding regression tests.

- [BLOCKING] src/context/identity_db.rs:673-681: A wallet-owned row cannot be converted to the durable unowned scope
  (existing thread: https://github.com/dashpay/dash-evo-tool/pull/955#discussion_r3830585869)
  The linked-record and no-record cases now fail safely, but `StoredWalletScope::WalletLess` still accepts the write after `mirror_identity_unowned` proves that no durable unowned row can be established. A pre-#955 wallet-less sidecar can coexist with the same identity stored upstream under a real wallet. In that state the sentinel upsert is silently skipped by upstream's guarded `WHERE`, readback returns `UnownedIdentityMirrorMissing`, and this branch nevertheless rewrites the sidecar with `wallet_hash: None`. Boot reconciliation reaches the same unresolved state and only logs a warning. Deleting the owning wallet then cascades through the upstream identity row and `cascade_meta_on_identity_delete`, deleting the sidecar DET reports as wallet-less. The existing read-failure test is safe because it begins with a successfully persisted unowned mirror and therefore does not exercise this shape. Do not treat the sidecar's existing wallet-less bit as proof of upstream durability: durably re-scope the row or recover and persist its actual wallet owner. Add a regression starting with a sidecar-only wallet-less record plus a wallet-owned upstream row, then remove that wallet and verify the identity either survives unowned or remains correctly linked.
```
</details>

src/context/identity_db.rs
Claudius-Maginificent: Re-verified: the remedy this finding asks for (durably re-scope the row, or recover and persist the actual wallet owner) is not implementable at pin 4784de03. Traced the full read surface this session: `SqlitePersister`'s only identity read is `load_unowned_identities()` (unowned scope only); the whole-store `PlatformWalletPersistence::load()` would answer 'which wallet owns this' but only as a `WalletId`, and DET's only `WalletSeedHash -> WalletId` map (`id_map`) is in-memory, populated on wallet registration, not persisted — there is no reverse map to turn an unloaded wallet's `WalletId` back into the `(WalletSeedHash, u32)` the sidecar stores. Re-scoping the upstream row itself has no primitive either: `add_out_of_wallet_identity`'s upsert is guarded `WHERE wallet_id IS NULL OR wallet_id IS excluded.wallet_id` and silently no-ops against an already-scoped row by design.

This PR's own docs already state this residual as an accepted, unfixable-at-this-pin limitation (a mislabel, not a data loss — the record persists, just under the wrong scope) for the exact same class of pre-existing divergence, confirmed independently by two rounds of adversarial QA. Building the reverse map to make recovery possible is a real feature addition, out of scope for this bugfix PR.

Doing the one thing that IS implementable: the WalletLess restate-arm's silent `debug!` becomes a loud `warn!` + sticky banner, matching the visibility the Linked-record arm already gets, so a hit is surfaced rather than invisible — even though it can't be repaired. Leaving this thread open / the review as CHANGES_REQUESTED; proceeding with the two independent fixes from the same review round (removal-postcondition check, doc staleness).

src/backend_task/error.rs
```diff
@@ -349,6 +349,36 @@ pub enum TaskError {
...
+    /// not survive. Names only that, because the two causes are
+    /// indistinguishable at this pin: upstream's out-of-wallet upsert skips a
+    /// row already filed under a wallet, and it also swallows a persist
+    /// failure and returns `Ok(())`. Both leave the same evidence and take
+    /// the same remedies. Carries the identity id (data, not a message).
+    #[error(
+        "Identity {identity_id} could not be saved as belonging to no wallet, because this \
+         device's wallet data still links it to a wallet. Open that wallet and load the identity \
+         from there, or try again in a moment."
+    )]
```
thepastaclaw, line 363: **🟡 Suggestion: Mirror error states an indistinguishable cause as certain**

`UnownedIdentityMirrorMissing` deliberately represents two indistinguishable outcomes: the guarded upsert may reject an identity already owned by a wallet, or upstream may swallow a persistence failure and leave no row at all. Its user-facing `Display` nevertheless states that wallet data still links the identity to a wallet. A persistent write failure therefore receives a false diagnosis and a wallet-loading remedy that may not apply. Keep the message neutral about why confirmation failed while preserving actionable remedies for both cases.

```suggestion
#[error(
    "Identity {identity_id} could not be saved as belonging to no wallet because this device could not confirm the wallet data update. Check available disk space and try again. If a wallet holds this identity, open that wallet and load it from there."
)]
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The dedicated unowned scope, mutation readbacks, and lock-protected two-way reconciliation are sound for normal paths, and the prior removal-verification and storage-documentation findings are fixed. However, an already-wallet-less sidecar can still collide with a wallet-owned upstream row and be deleted with that wallet, leaving the PR's core durability guarantee incomplete; the error wording and user-story catalog also need smaller corrections.
Source: Codex reviewers gpt-5.6-sol (general and Rust-quality); final verifier grok-4.5 via the Anthropic Claude Agent SDK. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

Validated blockers were found in the Codex precheck. Opus is deferred until a fresh Codex revalidation clears the blocker gate.

### Review provenance

<details>

```

- [SUGGESTION] src/backend_task/error.rs:359-363: Mirror error states an indistinguishable cause as certain
  `UnownedIdentityMirrorMissing` deliberately represents two indistinguishable outcomes: the guarded upsert may reject an identity already owned by a wallet, or upstream may swallow a persistence failure and leave no row at all. Its user-facing `Display` nevertheless states that wallet data still links the identity to a wallet. A persistent write failure therefore receives a false diagnosis and a wallet-loading remedy that may not apply. Keep the message neutral about why confirmation failed while preserving actionable remedies for both cases.

In `docs/user-stories.md`:
- [SUGGESTION] docs/user-stories.md:1543-1553: Catalog the new wallet-independent identity persistence story
  The Masternodes catalog covers loading, viewing, voting, removal, and key management, but no existing story requires a loaded node to persist across restarts or survive removal of unrelated or all wallets. This PR significantly changes that user-facing behavior and its description supplies a distinct masternode-operator story. The repository policy therefore requires adding an `[Implemented]` story with acceptance criteria covering zero-wallet startup reconciliation, restart persistence, and unrelated or all-wallet removal.

- [BLOCKING] src/context/identity_db.rs:685-693: A wallet-owned row cannot be converted to the durable unowned scope
  (existing thread: https://github.com/dashpay/dash-evo-tool/pull/955#discussion_r3830585869)
  The `StoredWalletScope::WalletLess` branch still accepts the rewrite after `mirror_identity_unowned` establishes that no durable unowned row can be confirmed. A pre-#955 wallet-less sidecar can coexist with the same identity stored upstream under a real wallet. In that state, the sentinel upsert is skipped by upstream's guarded `WHERE identities.wallet_id IS NULL OR identities.wallet_id IS excluded.wallet_id`, this branch preserves `wallet_hash: None`, and startup reconciliation repeatedly reaches the same missing-mirror result without repairing the ownership conflict. Removing the actual upstream wallet then deletes its identity row and fires `cascade_meta_on_identity_delete`, which removes the sidecar DET reports as wallet-independent. The warning and proposed restart do not repair this permanent wallet-owned-row case. The implementation must either durably re-scope the upstream row, recover and persist its actual owner, or store the DET record somewhere the wallet deletion cascade cannot reach. Add a regression beginning with a sidecar-only wallet-less record plus a wallet-owned upstream row, then remove that wallet and verify the identity either survives under a verified unowned scope or remains attached to its actual owner.
```
</details>

src/wallet_backend/identity_ops.rs
thepastaclaw: Resolved in `b0bd0c1` — _Verify removal because upstream swallows tombstone persistence failures_ no longer present.

docs/kv-keys.md
thepastaclaw: Resolved in `b0bd0c1` — _Document the current wallet-scope resolution behavior_ no longer present.

PR #958, opened 2026-08-21
chore(deps): bump platform pin to PR #3968 latest (4784de03)

Claudius-Maginificent:
## TL;DR

Updates the app's connection to `dashpay/platform` (the shared library that talks to Dash Core and the Dash network) to its latest version. This brings in a fix for a bug where certain wallets, damaged by an older release, would refuse to open at all — those wallets now open again automatically, with funds intact.

## User story

As an **Everyday User** who updated from an older Dash Evo Tool release, I want a wallet damaged by that older release to open normally again, so that I don't have to restore from a recovery phrase to get back to my funds.

## Scenario

**Actual behavior (before this PR):** A wallet database written by certain earlier app versions could contain one leftover, unreadable record. On any later version, that single bad record stopped the *entire* wallet file from opening — every wallet inside it, funded ones included — reporting "Saved wallet data appears damaged and cannot be loaded."

**Expected behavior (after this PR):** The bad record is cleared automatically, once, the first time the updated app opens the file. The affected wallets open normally afterward, with their balances and history intact.

Separately, this PR also changes how the app reacts to *other*, unrelated forms of data corruption it detects while loading a wallet: instead of silently skipping the damaged piece and continuing, it now stops and reports the problem plainly, rather than risking a partial or misleading picture of the wallet's state. This matches the app's general policy: prevent bad data from being written in the first place, and refuse to guess when it's found anyway.

## Detailed discussion

Moves all four `dashpay/platform` git dependencies (`dash-sdk`, `rs-sdk-trusted-context-provider`, `platform-wallet`, `platform-wallet-storage`) from `762c66cf` to `4784de03`, the current head of upstream PR #3968. This is a merge of `v4.2-dev` into that branch, so unlike the previous bump it carries a release cycle of unrelated upstream development (`4.1.0` → `4.2.0-dev.1`, key-wallet's `rust-dashcore` reference `branch=` → `rev=173ffac0`, grovedb `tag=` → `rev=a2791bbd`) and required adaptation across ~44 call sites, landed as three commits on this branch.

### What moved upstream, and how DET adapted (`chore(deps): bump platform pin to PR #3968 head`)

- **Document-query pagination + crate move** — `DocumentQuery` now lives in `dash-platform-queries` and gained a required `offset: Option<u32>`. All 14 literals set `None`, matching upstream's own constructor default: a set offset is served only by the protocol-v14 ranked surface and rejected everywhere else, so `None` keeps the wire bytes identical.
- **Transaction-builder redesign** — `set_funding` → additive `add_funding`; `build_unsigned` replaced by `build_unsigned_reserved`, which returns the `ReservationToken` from `dashpay/platform#4185`. DET's asset-lock dry-run probes — which reserve real wallet outpoints through an `Arc`-shared `ReservationSet` — now release with the owner-guarded `release_reservation_if_owner`.
- **New wallet-error and asset-lock variants** — 14 new `PlatformWalletError` variants bucketed explicitly in both no-wildcard matches; `AssetLockStatus::RecoveredFromChain` given its own honest label; `AssetLockFunding::DrainAccountBalance` fails closed through a new typed `TaskError`; `DocumentPropertyType::IdentifierWithReference` folded into the existing `Identifier` handling.
- **Restored `#[must_use]` returns** — the shutdown report and coordinator drain flags are inspected again, retiring three stale `TODO(platform-pr3954)`s.

### Load-failure policy (`test(wallet): assert the fail-closed load contract on a corrupt viewing-key row`)

Upstream converted `platform-wallet-storage`'s per-row corruption skips into a `LoadPolicy` gate defaulting to `Strict`: a row that will not decode aborts the load rather than being silently skipped. That is the intended, documented upstream design, and DET adopts it. The cold-boot test that asserted the old silent-skip contract is rewritten and renamed (`cold_boot_fails_closed_on_a_corrupt_fvk_row`), asserting the full typed error chain down to the column that failed, plus that a failed load leaves no half-wired backend behind.

A `Strict` → `Recovery` fallback (degraded, read-only wallet access instead of no access) is deliberately deferred — it's its own feature: `LoadPolicy` is fixed at persister-open time, `Recovery` makes the *entire* persister read-only (including the per-network KV backing settings, scheduled votes, DashPay overlays and the token cache), and it needs a new persistent read-only UI surface.

### Legacy-row migration (`chore(deps): bump platform pin to PR #3968 latest`)

The `Strict` policy exposed one previously-silent gap: upstream's used-address readers have no `LoadCtx` tolerance at all, so a leftover placeholder row an older DET build could write for a spend of an output it never recorded would fail the load of every wallet in the file. This is fixed upstream by `fix(platform-wallet-storage): purge legacy empty-script spent UTXO rows`, landing migration `V012`, which deletes exactly those rows (`spent = 1 AND script = ''`) once per database — balance-neutral, address-reuse-guard-neutral. A DET-side stopgap for this same gap was considered and dropped in favor of this upstream fix, per the project's "fix producers, migrate existing data upstream" policy. No DET-side mitigation exists or is needed after this bump.

### Verification

- `cargo build --all-features` — clean, no source changes required beyond the earlier 44-call-site adaptation.
- `cargo fmt --all` — clean.
- `cargo clippy --all-features --all-targets -- -D warnings` — exit 0, zero warnings.
- Scoped `cargo test --all-features --lib` runs green for the rewritten cold-boot test and its neighbours.
- CI owns the full workspace sweep.

## Summary by CodeRabbit

* **New Features**
  * Added support for editing and creating documents with identifier fields that include references.
  * Added a clear message when an unsupported funding method is selected for identity top-ups.

* **Bug Fixes**
  * Funding transactions rediscovered from the network now accurately show that network confirmation is known, while spent status remains unknown.
  * Wallet loading now fails safely when local viewing-key data is corrupted, avoiding partial setup.
  * Improved transaction reservation handling to prevent releasing funds reserved by another operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -347,6 +347,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- **A funding transaction found again on the network is now labelled honestly**:
+  when the app rediscovers a saved funding transaction from the chain rather
+  than tracking it from the start, it can tell that the network confirmed it but
+  not whether it was already spent on an identity. The funding list says exactly
+  that instead of claiming it is ready to use or already used. Selecting it still
+  works — the network has the final say and refuses one that was already spent.
+
 - **Upstream wallet backend updated (`platform-wallet` / `platform-wallet-storage`)**:
   the `dashpay/platform` dependency is bumped to the PR #3968 tip
   (`d18020f` → `288a6ca`), which lands an embeddable SQLite persistence backend with
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1876,8 +1876,8 @@ dependencies = [
 
 [[package]]
 name = "dapi-grpc"
-version = "4.1.0"
-source = "git+https://github.com/dashpay/platform?rev=762c66cf7716b9b3264868a49e8d29991c643c7e#762c66cf7716b9b3264868a49e8d29991c643c7e"
+version = "4.2.0-dev.1"
+source = "git+https://github.com/dashpay/platform?rev=993584a6b53831026f2f5fe30f21a8f4d14a06e8#993584a6b53831026f2f5fe30f21a8f4d14a06e8"
 dependencies = [
  "dash-platform-macros",
  "futures-core",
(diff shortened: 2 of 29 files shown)


thepastaclaw:
🕓 Ready for review — 1 ahead in queue (commit 37aaa65b)
_Queue position: 2/2 · 1 review active_
_ETA: start ~11:00 UTC · complete ~11:11 UTC (median 11m across 30 recent reviews; 3 slots)_
_Queued 8m ago · Last checked: 2026-08-21 11:00 UTC_

PR #959, opened 2026-08-25
ci: unblock the weekly Flatpak build and auto-publish scheduled pre-releases

lklimek:
## TL;DR

The weekly pre-release has silently failed to publish for the last two weeks. One of the two Flatpak builds runs out of time, and because the step that saves its build cache never gets reached, it starts from scratch again the following week and runs out of time again. This gives that build enough time to finish, makes the cache survive a failure so the next run is fast, and publishes the weekly automatically instead of leaving it waiting for someone to click a button.

## User story

As a **person who downloads Dash Evo Tool weekly builds**, I want the Tuesday pre-release to actually appear, so that I can install the latest version without waiting for someone to notice the release never went out.

As a **maintainer**, I want a scheduled release that has passed every check to publish itself, so that shipping does not depend on me remembering to un-draft it by hand.

### Actual behavior

The last two Tuesday releases never appeared. Both runs got most of the way through — every application build succeeded, all installers were produced — and then a single packaging step ran past its time limit. When any part fails, the release is rolled back and its tag deleted, so nothing was published at all. The newest release available is still the one from 21 July.

The failure also repeats itself by design: the run gets cut off before it can save the work it has done, so the next week starts from nothing and runs out of time in exactly the same place. It cannot recover on its own.

### Expected behavior

The Tuesday release is produced and published automatically, and a slow run makes the following run faster rather than repeating the same failure.

### Root cause

The Flatpak Cargo cache is keyed on `Cargo.lock`. The platform pin bump in #958 changed that hash, so the exact-key restore missed. The `restore-keys` prefix fallback still recovers the registry, but every upstream crate whose source moved must recompile — roughly **16 minutes** of `cargo build --release` on x86_64, leaving under four minutes for `flatpak-builder` export and `flatpak build-bundle`. The job's `timeout-minutes: 20` expired mid-bundle: 20m02s on the Aug 24 run, 20m15s on Aug 25.

The self-perpetuating part is step ordering. `Save Cargo cache` sits after `Build Flatpak` → `Verify` → `Create Flatpak bundle` and carried no `if: always()`, so a cancelled job saved nothing. x86_64 therefore re-enters every subsequent run just as cold, times out again, and saves nothing again.

`aarch64` escaped purely on timing: it finished its first cold run in 15m37s — inside the limit by about four minutes — and wrote its cache. On Aug 25 it restored that and finished in 9m8s. Same workflow, same commit; the only difference is which arch got its cache written.

Nothing was wrong with the build itself. On the Aug 25 run `cargo` finished cleanly at 04:06:52, `appstreamcli compose` reported success, and the `ldd` sandbox dependency check passed.

### Blast radius

`cleanup` fires whenever `finalize` does not succeed and deletes the tag and draft, restoring the "a weekly tag exists ⟺ a complete release exists" invariant. Correct behaviour, but it means one arch's timeout discards the entire release. Confirmed via the API: the newest release object in the repository is `v1.0.0-weekly.20260721`.

### Changes

**`.github/workflows/flatpak.yml`**

- `timeout-minutes: 20` → `45`, sized for a cold build rather than a cache hit. Sibling platform builds in the same run legitimately take 21–35 minutes. A warm Flatpak run finishes in well under half the new budget; the headroom exists so a cold run can reach the cache-save step.
- `if: always()` on `Clean Cargo cache for storage` and `Save Cargo cache`. Partial compilation output is a valid cache — the next run restores it and compiles only what remains. Without this, the next `Cargo.lock` bump re-arms the identical trap.
- `continue-on-error: true` on the save. Re-running a job whose exact key already exists makes the save a no-op, and that must never turn a green build red.

**`.github/workflows/weekly-build.yml`**

- New `Publish the weekly pre-release` step at the end of `finalize`, gated on `github.event_name == 'schedule'`. `finalize` already carries `if: success()`, so reaching this step means every build leg is green and every artifact is attached — the draft state was protecting nothing.
- `workflow_dispatch` cuts deliberately stay drafts: a manual run is usually an inspection or a re-cut, and the person who asked for it decides when it ships.
- The publish re-asserts `--prerelease` rather than assuming it. Publishing a release not flagged prerelease would make it the repository's "Latest" and hand weekly builds to everyone tracking stable. The flag is set at creation; this keeps a change there from silently promoting a weekly.
- Header comment updated to describe the draft/published split.

Ordering is deliberate: bundles are attached before the release is published, so a published release always has its assets. If the publish step fails, `finalize` fails and `cleanup` removes the release — consistent with the existing invariant.

### Note for reviewers

The workflow's display name is still `Weekly Draft Pre-release`, which is now only accurate for manual runs. Renaming it would break run-history continuity and anything keyed on the name, so it is left alone deliberately — say the word if you would rather it changed.

### Testing

Both workflow files parse cleanly and the resulting job graph was verified: `timeout-minutes: 45`, `always()` on both cache steps with `continue-on-error` on the save, and the publish step correctly gated to `schedule` inside a `finalize` job that is itself gated on `success()`. `actionlint` is not available on this machine, so the change has not been linted by it.

The real proof is the next scheduled Tuesday run, which should complete the cold x86_64 build, save its cache, and publish. The run after that should drop back to roughly nine minutes.

## Summary by CodeRabbit

* **Chores**
  * Improved Flatpak build reliability by allowing longer build times and preserving caches after interrupted or failed runs.
  * Updated weekly release workflow documentation to clarify scheduled and manually triggered release behavior.
  * Scheduled weekly releases can now be finalized automatically as prereleases, while manually created releases remain drafts.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -37,7 +37,14 @@ jobs:
             runs-on: ubuntu-24.04-arm
 
     runs-on: ${{ matrix.runs-on }}
-    timeout-minutes: 20
+    # Sized for a COLD build, not a cache hit. The Cargo cache key below is
+    # keyed on `Cargo.lock`, so every dependency bump starts both arches from
+    # a prefix-restore that still recompiles the whole upstream tree: ~16
+    # minutes of `cargo build --release` on x86_64 before flatpak-builder even
+    # starts exporting. A warm run finishes in well under half this budget;
+    # the headroom exists so the cold run can reach "Save Cargo cache" and
+    # make the next one warm.
+    timeout-minutes: 45
 
     steps:
       - name: Check out code
@@ -181,14 +188,27 @@ jobs:
         run: |
           flatpak build-bundle flatpak-repo dash-evo-tool-linux-${{ matrix.arch }}.flatpak org.dash.DashEvoTool
 
+      # `always()` on both steps below is what keeps a slow build from
+      # starving itself. These run after the build, so without it a job that
+      # times out or fails saves nothing, comes back just as cold next week,
+      # times out again, and never writes the cache that would have made it
+      # fast — a deadlock only a human raising the timeout can break. Partial
+      # compilation output is still a valid cache: the next run restores it
+      # and compiles only what is left.
       - name: Clean Cargo cache for storage
+        if: always()
         run: |
           rm -rf cargo-target/release/incremental
           rm -rf cargo-cache/registry/src
           echo "=== Cargo cache sizes ==="
           du -sh cargo-cache cargo-target || true
 
+      # `continue-on-error` because saving is best-effort: re-running a job
+      # whose exact key already exists makes the save a no-op, and that must
+      # never turn a green build red.
       - name: Save Cargo cache
+        if: always()
+        continue-on-error: true
         uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684  # v4.2.3
         with:
           path: |
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -139,7 +139,13 @@ jobs:
             cargo-cache
             cargo-target
           key: cargo-flatpak-${{ matrix.arch }}-${{ hashFiles('Cargo.lock') }}
+          # Ordered most to least specific. The middle entry is what makes a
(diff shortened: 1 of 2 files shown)


thepastaclaw:
ℹ️ Review skipped (commit e5f9683e)
_Last checked: 2026-08-25 09:31 UTC_

PR #965, opened 2026-08-29
fix(build): support native Windows builds with the GNU toolchain

lklimek:
## Problem

Windows release binaries are cross-compiled from Linux with the mingw toolchain (`x86_64-pc-windows-gnu`, see `.github/workflows/release.yml`), so building natively on Windows was never exercised and did not work.

An earlier revision of this PR made the native build work against the **MSVC** target. That approach is dropped here in favour of building natively against `x86_64-pc-windows-gnu` — the same target that ships. Developers then exercise the configuration that is actually released, and the MSVC-specific workarounds disappear rather than becoming something to maintain.

## Why GNU rather than MSVC

- **`rs-x11-hash` compiles out of the box.** The crate compiles its C sources with clang (hardcoded in its own build script) and its `sph_types.h` includes the POSIX `<unistd.h>` unconditionally, which MSVC does not ship. mingw does, so the empty stand-in header the MSVC route needed is gone.
- **No divergence from the release artifacts.** A native build and a release build now differ only in where they run.
- **`build.rs` needs no target-env special-casing.** GNU targets use `windres`/`ar`, which is what it already did.

## Changes

**1. `.cargo/config.toml`.** Removed `[target.x86_64-pc-windows-msvc]`. The main-thread stack reservation it carried is a genuine fix, so it moves onto the GNU target as `-Wl,--stack,8388608`: `RUST_MIN_STACK` covers spawned threads only, the main thread's stack is reserved at link time, and mingw's default of 2 MiB is not enough for the deep call chains noted in that file. This reserve is address space, not committed memory, and it now also applies to the released binaries. Verified in the PE header: `SizeOfStackReserve 0x800000`, with `SizeOfStackCommit` still `0x1000`.

**2. `build.rs`.** Reverted to the base revision — the `CARGO_CFG_TARGET_ENV != "msvc"` gate only existed to let MSVC skip the mingw tools.

**3. `build-support/windows-msvc/unistd.h`.** Deleted.

**4. Dead-code warnings.** Two statics/methods warned as dead on any Windows target, MSVC included. `ORIGINAL_STDERR_FD` is read only from `#[cfg(unix)]` code and `SpvBlockReconciler::armed` has a single caller gated on the `testing` feature, so each is now gated on the configuration that uses it rather than silenced with an `allow`. They appear and disappear together with their callers, and CI's `--all-features` clippy sees both present.

**5. Documentation.** The *Building natively on Windows* section in `CONTRIBUTING.md` is split into install, alias and build steps.

## The one manual step

`.cargo/config.toml` names the Debian cross-compilers the release workflow uses, and is deliberately left alone — CI is untouched by this PR. MSYS2 uses different names for the same tools: the `-posix` suffix is Debian's `update-alternatives` naming for the posix/win32 threading choice, which MSYS2 does not have (its `gcc` already uses posix threads), and it ships no prefixed `ar`. No package provides those names, so the setup creates three symlinks:

```
x86_64-w64-mingw32-gcc-posix.exe -> gcc.exe
x86_64-w64-mingw32-g++-posix.exe -> g++.exe
x86_64-w64-mingw32-ar.exe        -> ar.exe
```

Symlinks rather than copies or hard links: `pacman` upgrades by extracting and renaming over the old file, so either would keep pointing at the previous binary and silently build with a stale compiler after the next `pacman -Syu`.

Without them the build fails in `cc-rs` with `failed to find tool "x86_64-w64-mingw32-gcc-posix": program not found`. `CONTRIBUTING.md` documents the alternative for anyone who cannot create symlinks — overriding `CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER`, `CC_`, `CXX_` and `AR_x86_64_pc_windows_gnu` per session, which works because cargo's `[env]` entries do not force.

## Testing

Windows 11, Rust 1.92, MSYS2 gcc 16.2.0, binutils 2.47, cmake 4.4.2.

- `cargo build --target x86_64-pc-windows-gnu` succeeds with **no environment overrides at all** — only the symlinks and mingw on `PATH`. Run against a forced rebuild (`cargo clean -p rs-x11-hash -p libsqlite3-sys -p aws-lc-sys -p dash-evo-tool`) so that `cc-rs` genuinely re-invoked the compiler and archiver symlinks and rustc re-invoked the linker symlink, rather than reporting a no-op from cached artifacts. Finished in 4m50s, 0 errors, 0 warnings.
- The dependencies that most plausibly break under this target all compile: `rs-x11-hash` (clang locates the mingw sysroot unaided), `aws-lc-sys` (cmake), `libsqlite3-sys`.
- `cargo fmt --all -- --check` passes; `--features testing` compiles.
- No effect on other targets. `build.rs` returns to its base behaviour, `.github/workflows/release.yml` is untouched, and the only change reaching CI is the added `-Wl,--stack` link argument on the GNU target.

## Note on the branch name

The branch is still `fix/windows-msvc-native-build`, which no longer describes the change. Renaming it would close and recreate this PR, so it is left as is.


thepastaclaw:
🕓 Ready for review — 12 ahead in queue (commit 0822cf09)
_Queue position: 13/15 · 2 reviews active_
_ETA: start ~14:14 UTC · complete ~14:54 UTC (median 39m across 30 recent reviews; 2 slots)_
_Queued 12m ago · Last checked: 2026-08-31 10:10 UTC_

PR #966, opened 2026-08-29
fix(secret): stop the mlock guard from aborting the app on drop

lklimek:
## Problem

Closing the "Add existing identity" screen killed the app in debug builds:

```
Panic occurred: unlocking region: Err(SystemCall(Os { code: 158, kind: Uncategorized,
  message: "Segment jest juz odblokowany." }))
  10: region::lock::impl$1::drop            (region-3.0.2/src/lock.rs:90)
  13: drop_in_place<dash_evo_tool::model::secret::Secret>
  14: drop_in_place<...::password_input::PasswordInput>
  15: drop_in_place<...::add_existing_identity_screen::AddExistingIdentityScreen>
  18: dash_evo_tool::app::impl$7::ui        (src/app.rs - AppAction::PopScreen)
```

The panic escaped the winit event handler and eframe re-raised it, taking the process down.

## Cause

`Secret` held a `region::LockGuard`, whose `Drop` is `debug_assert!(result.is_ok(), "unlocking region: {result:?}")`.

Memory locks operate on whole pages, and a 4096-byte unaligned `String` buffer always spans two of them — so buffers of neighbouring `Secret`s share a page. Windows' `VirtualUnlock` is **not** reference-counted: the first `Secret` to drop releases the shared page, and the next one gets `ERROR_NOT_LOCKED` (158) and aborts.

`relock_if_moved()` had the same problem in a second form: it took the new lock *before* dropping the stale guard, so the old guard's unlock could immediately undo the lock just taken.

Only debug builds abort (`debug_assert!`); release builds silently ignored the failed unlock.

## Fix

Replace the guard with a local `MemLock` that owns the unlock itself and logs failures at debug level instead of asserting — locking is documented as best-effort, so a benign "already unlocked" must not be fatal. `relock_if_moved()` now releases the stale lock first.

## Testing

Two regression tests added; both abort the test process without the fix (`STATUS_STACK_BUFFER_OVERRUN`):

```
test model::secret::tests::test_dropping_page_sharing_secrets_does_not_panic ... ok
test model::secret::tests::test_reallocation_relock_does_not_panic ... ok
test result: ok. 19 passed; 0 failed; 1 ignored
```

`cargo fmt --all` run; `cargo clippy --lib --all-features -- -D warnings` clean for this change.

## Not addressed

Page sharing still means one `Secret`'s drop unlocks pages other live `Secret`s sit on, weakening the anti-swap guarantee. Fixing that needs page-aligned allocations, which `String` cannot provide — a much larger change than this crash fix, and the type already documents locking as best-effort.

## Summary by CodeRabbit

- **Security Enhancements**
  - Sensitive values now use centralized secure storage and constant-time comparisons.
  - Secret schemas and serialization align with secure storage requirements.

- **Bug Fixes**
  - Passwords and passphrases over 4,080 UTF-8 bytes are rejected before storage.
  - Clear length-limit messages replace generic storage errors across wallet, key, and protected-object operations.

- **Documentation**
  - Updated guidance to describe password limits and validation behavior.

- **Testing**
  - Expanded coverage for boundary lengths, multi-byte characters, and refusal handling.


thepastaclaw:
🕓 Ready for review — 42 ahead in queue (commit 04a08fd3)
_Queue position: 43/50 · 2 reviews active_
_ETA: start ~03:45 UTC · complete ~04:40 UTC (median 54m across 30 recent reviews; 2 slots)_
_Queued 17h 23m ago · Last checked: 2026-09-04 08:30 UTC_

src/model/secret.rs
```diff
@@ -39,6 +39,50 @@ impl CompareOnlyPtr {
...
+/// A best-effort `mlock`/`VirtualLock` on a memory region, released on drop.
+///
+/// This exists instead of [`region::LockGuard`] because that guard's `Drop`
+/// carries a `debug_assert!` on the unlock result, which aborts a debug build
+/// whenever the unlock reports the region was not locked. That is a routine
+/// occurrence here, not a bug: locks operate on whole pages, so two `Secret`
+/// buffers that share a page share a lock. Windows' `VirtualUnlock` is not
+/// reference-counted — the first `Secret` to drop unlocks the shared page, and
+/// the second then sees `ERROR_NOT_LOCKED` (158), which took the whole process
+/// down. Locking is documented as best-effort, so a failed unlock is logged and
```
lklimek: this is a wrong approach; we should rather ensure that secrets don't share pages with other secrets - just use whole page.

src/model/secret.rs
```diff
@@ -39,6 +39,50 @@ impl CompareOnlyPtr {
     }
 }
 
+/// A best-effort `mlock`/`VirtualLock` on a memory region, released on drop.
```
lklimek: I think we have secret implementation on platform-wallet-storage, why don't we use it? Is it affected by the issue fixed here?

lklimek (commented):
LGTM

PR #969, opened 2026-08-31
fix(ui): start window maximized on first launch

Claudius-Maginificent:
**TL;DR:** The app now opens as a maximized window the very first time you run it, instead of a small window that cuts off the onboarding page.

## User story
As a **first-time user**, I want the app window to open large enough to see the whole onboarding page, to achieve a usable setup experience without having to resize the window myself.

### Base flow
A user installs Dash Evo Tool and launches it for the very first time.

### Actual behavior
The window opens at a small default size that doesn't fit the onboarding page, forcing the user to resize or maximize it manually before they can comfortably use it.

### Expected behavior
On first launch, the window opens maximized. On every later launch, the window remembers whatever size/position the user left it at, exactly as before.

### What was done
Added `.with_maximized(true)` to the `egui::ViewportBuilder` passed to `eframe::NativeOptions` in `src/main.rs`.

`persist_window: true` (with `persistence_path` pointing at `app.ron`) means eframe unconditionally overrides `inner_size`/`maximized`/`fullscreen` from the persisted `WindowSettings` once a valid entry exists in `app.ron` — which happens after the very first run. So `.with_maximized(true)` only ever takes effect on a genuine first launch (no persisted window state yet); it is silently overridden by the user's actual window state on every subsequent run and does not "stick" as forced-maximized forever.

Also adds the corresponding `CHANGELOG.md` entry under `[Unreleased] / Fixed`.

### Testing
- `cargo build --bin dash-evo-tool` — clean.
- `cargo clippy --bin dash-evo-tool --all-targets -- -D warnings` — clean.

### Breaking changes
None.

### Checklist
- [x] `cargo fmt --all`
- [x] `cargo clippy` (scoped to the touched binary)
- [x] `cargo build`
- [ ] Manual verification of first-launch maximize on a real display (no persisted `app.ron`)

### Attribution

## Summary by CodeRabbit

* **New Features**
  * The app now opens maximized on first launch.
  * Your preferred window size and position are remembered for subsequent launches.

* **Documentation**
  * Updated the changelog to document the window behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **The app now opens maximized the very first time it runs**: on a fresh
+  install, the window used to open at a small default size that didn't fit
+  the onboarding page. It now opens maximized on first launch; any size or
+  position you set afterward is remembered and used on every later start, as
+  before.
+
 - **A sent payment whose confirmation couldn't be verified no longer tells you
   to retry it**: when the network doesn't confirm a payment quickly enough,
   the app used to show a generic "please retry" message — but the payment may
--- a/src/main.rs
+++ b/src/main.rs
@@ -74,7 +74,13 @@ async fn start(app_data_dir: &std::path::Path) -> Result<(), eframe::Error> {
         persistence_path: Some(app_data_dir.join("app.ron")),
         viewport: egui::ViewportBuilder::default()
             .with_icon(icon_data)
-            .with_app_id("org.dash.DashEvoTool"),
+            .with_app_id("org.dash.DashEvoTool")
+            // Only takes effect on a genuine first launch: once `persist_window`
+            // has written window geometry to `app.ron`, the persisted state
+            // (size/maximized/fullscreen) unconditionally overrides this on
+            // every later run. Without it, winit falls back to its own small
+            // platform-default size, which doesn't fit the onboarding page.
+            .with_maximized(true),
         // Use wgpu instead of glow (OpenGL) to avoid platform-specific rendering
         // issues, e.g. NSOpenGLContext idle/sleep crashes on macOS (#629)
         renderer: eframe::Renderer::Wgpu,


thepastaclaw:
🕓 Ready for review — 24 ahead in queue (commit ed5d4356)
_Queue position: 25/25 · 2 reviews active_
_ETA: start ~22:45 UTC · complete ~23:30 UTC (median 45m across 30 recent reviews; 2 slots)_
_Queued 5m ago · Last checked: 2026-08-31 13:30 UTC_

Claudius-Maginificent:
This weekend's `release-review` QA campaign (baseline `v1.0.0-weekly.20260825` vs this branch's HEAD `1bcb20ae2`) live-A/B tested the unticked checklist item — **"Manual verification of first-launch maximize on a real display."**

Result: **it does not verify.** Both binaries were launched fresh (`mktemp -d` data dir, no persisted `app.ron`) on a private `Xvfb :99` display and open at the same **800×600** window. `with_maximized(true)` (`src/main.rs:83`) has no observable effect on this platform/build.

Not classified as a regression (baseline behaves identically — neither build maximizes), so it's non-blocking for this release, but the CHANGELOG's "Fixed" claim and this PR's own intent aren't actually verified working. Screenshot evidence (`R5-window-sizing-01-first-launch-*`) and full writeup: `GUI-004` in this campaign's consolidated report, `/data/artifacts/dash-evo-tool/2026-08-31/release-review-campaign/report.md`.

Recommend re-opening as a follow-up rather than leaving the checklist box as-is — worth checking whether `eframe`'s `centered: true` branch is overriding the maximize request before persistence ever kicks in.

<sub>Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

thepastaclaw:
Thanks for the A/B evidence. I traced the exact startup path in the pinned `eframe 0.35.0` / `winit 0.30.13` sources. On fresh state, `with_maximized(true)` survives `eframe::native::epi_integration::viewport_builder` and becomes `WindowAttributes::maximized`. On X11, winit then maps the window and sends `_NET_WM_STATE_MAXIMIZED_HORZ` / `_NET_WM_STATE_MAXIMIZED_VERT` to the root window.

That request requires an EWMH-capable window manager to act on it; Xvfb itself is only an X server. A bare Xvfb display will leave the mapped window at winit's 800×600 default even though the application correctly requested maximization. Also, `NativeOptions::centered` only adds a position in eframe 0.35.0; it does not clear the maximized flag, so it does not explain this first-launch result.

Was a window manager (for example Openbox) running on `:99` during GUI-004? If not, this result is an environment limitation rather than evidence that #969 fails on a user desktop. Please rerun this specific check with a WM/compositor, or on a normal desktop session, before opening a source-code follow-up. If a WM was running, please share which one and its `_NET_SUPPORTED` state so we can narrow the real platform path.

PR #971, opened 2026-09-01
chore(ui): remove the superseded standalone identities screen

Claudius-Maginificent:
**TL;DR:** Removes the old, no-longer-visible "Identities" list screen and points its few remaining entry points at the current Identity Hub instead.

## User story
As an **everyday user**, I want every path that used to lead to identity management to land on the Identity Hub, so I never end up stuck on an old screen with no way back.

### Base flow
A user creates or imports a wallet, then chooses to create or load a Dash Platform identity from the screen that follows.

### Actual behavior
That choice, and a few others (registering a DPNS name from a newly-created identity, its breadcrumb), lands the user on a standalone "Identities" list screen. That screen has had no entry in the left-hand navigation for a while, so nothing is highlighted and there's no way to navigate back to it on purpose — and because the app remembers your last screen, a returning user can get parked there again on every future launch.

### Expected behavior
All of those actions land on the Identity Hub, the app's current, fully-navigable home for identity management. The old list screen is gone; everything it offered (adding/loading identities, top-up, withdraw, transfer, key management, DPNS registration) is already available through the Hub, which reuses the same underlying screens.

### What was done
Investigated whether `src/ui/identities/` (the legacy screen set) was genuinely dead code, per a report that it had become inaccessible. It turned out to be a mix: most of the directory is live shared infrastructure the Identity Hub itself depends on (`TransferScreen`, `TopUpIdentityScreen`, `WithdrawalScreen`, key management, DPNS registration, etc.) — only the standalone root list screen (`identities_screen.rs`) was actually superseded, and it was still reachable through four live navigation targets, not fully dead as assumed.

This PR:
- Deletes `src/ui/identities/identities_screen.rs` and its dedicated test file.
- Removes the `RootScreenType::RootScreenIdentities` route and its `ScreenType`/`Screen` plumbing (`settings.rs`, `ui/mod.rs`, `app.rs`), leaving the usual numeric-encoding gap comment.
- Reroutes the four live navigation targets that pointed at the retired screen to `RootScreenIdentityHub`: post-wallet-creation "Create Platform Identity", post-import "Create Identity"/"Load Existing Identity", the add-identity success screen's "Register DPNS Name", and the Register-DPNS breadcrumb.
- Fixes seven `left_panel` nav-highlight arguments that referenced the retired route and highlighted nothing; they now name the hub, matching the pattern already used by `keys_screen.rs`.
- Fixes a knock-on regression found in self-review: retiring the numeric encoding for the old screen made a returning user's persisted setting fall through to the *settings default* (`RootScreenDashpay`, itself absent from the nav) before the app's own fallback logic got a say. The default is now `RootScreenIdentityHub`, matching `app::FALLBACK_ROOT_SCREEN`, under a new regression test.
- Removes `OnboardingAction::ImportIdentity`, which was already dead (absent from `ONBOARDING_CARDS`, constructed nowhere).
- Corrects a stale doc comment in `left_panel.rs` claiming the retired routes stayed reachable via "deep links, MCP tools, direct screen construction" — none of that exists in the codebase.
- Updates `docs/user-stories.md` (DPN-010) to point at the surviving UI locations for the pending-DPNS-registration hint.
- Moves three still-relevant app-shell tests out of the deleted test file into `startup.rs`; drops a fourth that duplicated an existing `startup.rs` test; retargets ~25 test mount/assert sites, judged individually (e.g. a test specifically covering a *non-hub* everyday root was moved to `RootScreenDocumentQuery`, not the hub, to keep covering what it actually tests).

### Testing
- `cargo clippy --all-features --all-targets -- -D warnings` — exit 0 (wide scope deliberate: this is a project-wide enum-variant removal).
- `cargo test --test kittest --all-features` — 313 passed, 0 failed.
- `cargo test --lib --all-features` — 2432 passed, 0 failed, 2 ignored.
- `cargo fmt --all` applied.
- New regression test (`the_retired_identities_encoding_falls_back_to_the_hub`) confirmed failing before the settings-default fix, passing after.

### Breaking changes
None for end users functionally — the fresh-install default root screen changes from DashPay (a screen with no nav entry) to the Identity Hub; this is treated as a bugfix, not a regression, since DashPay was equally unreachable via nav.

### Checklist
- [x] `cargo fmt --all`
- [x] Targeted tests green (see Testing)

### Attribution

## Summary by CodeRabbit

* **New Features**
  * Consolidated identity management into the Identity Hub as the primary identities destination.
  * Updated identity-related flows—including creation, loading, transfers, withdrawals, key management, and funding—to return to the Identity Hub.
  * Pending username registrations now display a “Pending” pill in Identity Settings while remaining visible from Identity Home.

* **Bug Fixes**
  * Retired legacy identities navigation and updated saved navigation settings to fall back to the Identity Hub.


thepastaclaw:
🕓 Ready for review — 41 ahead in queue (commit a1037d0b)
_Queue position: 42/42_
_ETA: start ~05:19 UTC · complete ~06:09 UTC (median 49m across 30 recent reviews; 2 slots)_
_Queued 19m ago · Last checked: 2026-09-01 12:50 UTC_

lklimek (approved):

PR #956, opened 2026-08-01
test(wallet): add cross-wallet top-up isolation e2e coverage

Claudius-Maginificent:
**TL;DR:** Adds a live-testnet end-to-end test proving cross-wallet identity top-up no longer corrupts the paying wallet — regression coverage for a fix (#954) that merged without its test.

## User story
As a **developer maintaining Dash Evo Tool's wallet top-up flow**, I want an end-to-end regression test for cross-wallet top-up isolation, so a future change can't silently reintroduce the corruption #954 fixed.

### Base flow
One wallet (the payer) funds another wallet's identity top-up using an asset lock.

### Actual behavior
#954 fixed cross-wallet top-up corrupting the payer wallet's own state, and was developed with a matching end-to-end test — but the PR landed as a squash merge that didn't include the test file, so the fix shipped with no regression coverage on `v1.0-dev`.

### Expected behavior
A backend-e2e test (`cross_wallet_topup_e2e`) exercises the real flow against Dash testnet and asserts: the payer wallet's own identity/UTXO state stays intact, both wallets reload cleanly afterward, resuming a spent unbound asset-lock is correctly refused, and the existing owned-identity top-up path still works.

### What was done
Recovered `tests/backend-e2e/cross_wallet_topup.rs` and its `tests/backend-e2e/main.rs` module registration from the branch that originally accompanied #954's fix, rebased cleanly onto current `v1.0-dev`, and cleaned up references that had gone stale in the rebase (a commit hash squashed away by the merge, an outdated file path, internal tracking IDs replaced with plain-language rationale).

### Testing
Run live against Dash testnet earlier in development (~104s, all 4 assertions passed). This PR's rebased version compiles clean (`cargo test --test backend-e2e --all-features --no-run`, exit 0). The test is `#[ignore]`-gated and network-dependent per this repo's backend-e2e conventions, so it's not part of CI — see `tests/backend-e2e/README.md` for how to run it live before merge if a reviewer wants to re-confirm.

### Breaking changes
None (test-only addition).

### Checklist
- [x] Compiles (`--no-run` check)
- [x] Follows `tests/backend-e2e/README.md` conventions
- [x] `cargo fmt` clean

### Prior work
Fixed by #954 — this PR only adds the test coverage that PR's squash merge dropped.

### Attribution

<sub>🤖 Co-authored by [Claudius the Magnificent](https://github.com/lklimek/claudius) AI Agent</sub>

## Summary by CodeRabbit

- **Tests**
[...]

Issue #799, opened 2026-03-26
refactor(wallet): unify all asset lock paths and add transaction recovery


lklimek:
## Summary

All asset lock creation flows should use a single shared code path with deterministic HD-derived keys and built-in recovery for broken/rejected transactions. Currently there are multiple independent implementations with inconsistent error handling and key management.

## Background

PR #790 (`fix/deterministic-asset-lock-keys`) introduced deterministic HD key derivation for generic asset lock transactions, replacing `OsRng`-generated keys that caused permanent fund loss when Platform rejected a state transition after Core had already broadcast the asset lock. It also introduced `broadcast_and_commit_asset_lock()` as a shared helper for the broadcast → track → UTXO cleanup sequence.

However, several asset lock flows still have their own inline implementations, and recovery from broken transactions is incomplete.

### 1. Multiple divergent asset lock paths

The codebase has at least 4 separate asset lock creation flows:

| Flow | File | Key derivation | Broadcast path | Recovery |
|---|---|---|---|---|
| Identity registration | `create_asset_lock.rs` | `registration_asset_lock_transaction()` — HD-derived ✅ | `broadcast_and_commit_asset_lock()` ✅ | Partial (Search for Asset Locks) |
| Identity top-up | `create_asset_lock.rs` | `top_up_asset_lock_transaction()` — HD-derived ✅ | `broadcast_and_commit_asset_lock()` ✅ | Partial |
| Fund platform address | `fund_platform_address_from_wallet_utxos.rs` | `generic_asset_lock_transaction()` — HD-derived ✅ (PR #790) | Inline broadcast + manual UTXO cleanup ❌ | None |
| Shield from asset lock | `shielded/bundle.rs` (`shield_from_asset_lock`) | `generic_asset_lock_transaction()` — HD-derived ✅ (PR #790) | Inline broadcast + manual UTXO cleanup ❌ | None |

### 2. Inconsistent broadcast + commit sequence

`broadcast_and_commit_asset_lock()` (introduced in PR #790 for identity flows) handles:
- Register tx in `transactions_waiting_for_finality`
- Broadcast raw transaction
- Remove used UTXOs from wallet
- Persist UTXO removal to DB
- Recalculate affected address balances
- Clean up finality tracking on failure

The platform address funding and shielded flows do this sequence **inline** with subtle differences:
- `fund_platform_address_from_wallet_utxos.rs` does UTXO cleanup but doesn't register in `transactions_waiting_for_finality` before broadcast
[...]

PR #975, opened 2026-09-08
test(backend-e2e): stop init-panic cascade, fix 5 stale assertions, fix TC-MN-021

Claudius-Maginificent:
## Summary

Follow-up to #956's backend-e2e QA run: fixes the issues found there that are
safe to fix in test-only code, root-causes and documents the ones that aren't.

## Item A — root cause of the `AlreadyOpen` cascade (not what it looked like)

`--test-threads=1` was never the cause. `tokio::sync::OnceCell` serializes
correctly; the real trigger in the original failing run was a bad
`E2E_WALLET_MNEMONIC`, and every retry afterward panicked with `AlreadyOpen`
instead of the real cause because `AppContext` is immortal in-process (a
strong `Arc` cycle through `SpvProvider::bind_app_context` — see the doc
comment on `CTX` in `harness.rs` for the full trail), so a leaked
`app_kv`/`secret_store` handle poisons the workdir slot for the rest of the
process.

Fixed in `harness.rs` (test-only):
- `E2E_WALLET_MNEMONIC` is read/parsed before any store opens, so the most
  common init failure can't poison a slot.
- `open_available_workdir` also opens the app k/v and secret vault as part of
  slot selection — a slot still held open in-process is now treated as taken,
  same as a slot locked by another process.
- Init attempts capped at 3, then a clear message pointing at the first
  failure instead of silently burning all 10 workdir slots.

**Not changed, deliberately:** `SpvProvider::app_context` staying a strong
`Arc` is the actual production root cause, but fixing it (strong → `Weak`)
touches the SDK proof-verification/quorum path where a silent "no app
context" is funds-adjacent. Needs its own reviewed PR — tracked separately,
not in this diff.

`--test-threads=1` is still required, for unrelated reasons (shared framework
wallet UTXO set, `SHARED_IDENTITY` fixture, a cleanup sweep that removes every
non-framework wallet) — now documented on `CTX`.

## Item B — 5 stale result-variant assertions

`core_tasks::tc004/tc005/tc012`, `wallet_tasks::tc_018`,
`identity_tasks::tc_028` all asserted a `BackendTaskSuccessResult` variant the
task under test cannot construct anymore (`Message` where the backend has
returned `AssetLockBroadcast{txid}` for a while; `identity_tasks::tc_028`
never handled `IdentitiesLoaded{count}`). Verified against `src/`, not just
the QA log. Fixed the assertions; did not touch backend behavior.

Heads-up: `tc_018` now runs on into the documented `TODO(#799)` funding gap
and fails there on its 360s confirmation timeout instead of failing in 1s on
the wrong assertion — correct but slower.

## Item C

- **C1 (`identity_masternode_withdraw::test_mn021`)** — stale test, fixed.
  `MasternodeNotFound` is a deliberate, unit-tested variant
  (`src/backend_task/error.rs`) for non-`User` identity types; the test
  predates it. Now asserts the variant and that it echoes the resolved
  ProTxHash.
- **C2 (`shielded_tasks::tc_074_shielded_lifecycle`)** — investigated, NOT
  fixed. Attempted a bounded-poll fix for the balance-read race the original
  QA run observed, but two live verification runs against testnet both failed
  *earlier*, at the shield broadcast itself
  (`TransactionConfirmationUnknown` / `TransactionBroadcastUnconfirmed`),
  reproducibly. That's a different, bigger problem than what the fix
  addressed, so it was reverted rather than shipped unverified. `tc_079`
  (shield from an existing balance, no asset lock) passes — the asset-lock
  shield path is the suspect. Recommend a follow-up there, not on this test.
- **C3 (`wallet_tasks::tc_014_wallet_platform_lifecycle`)** — investigated,
  not a stale test. `step_withdraw` already polls and confirms the funds are
  present, yet Platform's ST processor rejects the same address as
  under-funded — a real sync-proof vs ST-processor disagreement (already
  TODO'd in `wallet_tasks.rs`). No test-side fix exists that isn't a
  papered-over retry; leaving as-is.

## Scope

Test-only: `tests/backend-e2e/{core_tasks,framework/harness,framework/task_runner,identity_masternode_withdraw,identity_tasks,wallet_tasks}.rs`. No `src/` changes. `cross_wallet_topup.rs` and this directory's `README.md` untouched (already fixed on #956).

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy --test backend-e2e --all-features -- -D warnings` exit 0
- `cargo test --test backend-e2e --all-features -- --list` — 70 tests enumerated (link-checks the full binary)
- 2 live testnet runs against `tc_074` (both failed at shield broadcast, leading to the C2 revert above)

No full-suite live run performed (network/funds cost) — the mechanical fixes (A, B, C1) are verified by log evidence + compile/link checks; C2 was reverted specifically because it could not be verified live.

## Summary by CodeRabbit

- **Tests**
  - Improved backend end-to-end test reliability with deterministic temporary workspaces and bounded initialization retries.
  - Added stricter validation for asset-lock broadcast results, identity lookup errors, and wallet identity search outcomes.
  - Expanded coverage and clarified test expectations across identity, wallet, and core task workflows.

- **Documentation**
  - Updated backend end-to-end testing documentation with current workspace handling, cleanup guidance, and the expanded test module registry.

### DIFF
--- a/tests/backend-e2e/README.md
+++ b/tests/backend-e2e/README.md
@@ -92,13 +92,15 @@ ctx().await  -->  OnceCell::get_or_init(BackendTestContext::init)
   testnet with SPV running
 - **`framework_wallet_hash`** -- the `WalletSeedHash` of the "bank" wallet used
   to fund per-test wallets
-- **`_workdir`** -- path to a persistent temp directory keyed by git revision
-  (e.g., `/tmp/dash-evo-e2e-testnet-abc1234`)
+- **`_workdir`** -- path to a persistent temp directory (e.g.,
+  `/tmp/dash-evo-e2e-testnet`, or a numbered fallback slot; see "Persistent
+  workdir" below -- not git-rev-keyed)
 
 ### Initialization sequence
 
 1. Initialize tracing subscriber for structured log output.
-2. Create a persistent workdir under `/tmp/` keyed by `git rev-parse --short HEAD`.
+2. Create a persistent workdir under `/tmp/` (deterministic path, with numbered
+   fallback slots guarded by a `.lock` file -- see "Persistent workdir" below).
 3. Copy `.env.example` into the workdir via `ensure_env_file()`.
 4. Create a SQLite database and `AppContext` for `Network::Testnet`, passing the workdir as `data_dir`.
 5. Start SPV in light-client mode and wait for peer connections (60s timeout).
@@ -110,10 +112,14 @@ ctx().await  -->  OnceCell::get_or_init(BackendTestContext::init)
 
 ### Persistent workdir
 
-The workdir survives across test runs for the same git revision. This means:
+The workdir is deterministic, not git-rev-keyed: it's always
+`<tmp>/dash-evo-e2e-testnet` (slot 0), with numbered fallback slots
+(`dash-evo-e2e-testnet-1`, `-2`, ...) that `pick_available_workdir` cycles
+through via a per-slot `.lock` file when an earlier slot is still locked by
+another process (see `framework/harness.rs`). This means:
 
-- The SQLite database is reused, so wallets registered in prior runs are already
-  present.
+- The SQLite database is reused across runs, so wallets registered previously
+  are already present, regardless of which commit produced them.
 - The framework wallet registration handles the "already imported" case
   gracefully.
 - SPV sync is faster on repeat runs because prior state may be cached.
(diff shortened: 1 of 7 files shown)


thepastaclaw:
🕓 Queued for automated review — 2nd in line, estimated start in ~15 min (commit 068af9e8)
_Estimated review time once started: ~30 min (two-phase automated review; median of recent runs)._

